PR #14, opened 2018-01-16
new hashsample processor

quipo:
This new processor provides a deterministic data sampling pipeline. Internally, it hashes the message part(s) configured, and checks if the hash is within the [min, max) range (where min and max are floats between 0 and 100).  This way it's possible to send a portion of the input messages to one output channel, and the other portion (complementary to the first) to another output channel.


Jeffail:
Dope, just added underscores to the proc type.

PR #36, opened 2018-07-27
Add Duplicate, QoS, Retained, Topic, and MessageID from mqtt.Message …

chrisriley:
This adds the results of Duplicate(), Qos(), Retained(), Topic(), and MessageID() from [mqtt.Message](https://github.com/eclipse/paho.mqtt.golang/blob/master/message.go#L24) to the metadata of the MQTT input. I have another pull request coming for the AMQP output that then makes use of this metadata.

### DIFF
--- a/lib/input/reader/mqtt.go
+++ b/lib/input/reader/mqtt.go
@@ -21,6 +21,7 @@
 package reader
 
 import (
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


Jeffail:
Awesome, thanks @chrisriley! I like the idea of namespacing the metadata keys so I did the same to the Kafka ones. Hope you don't mind but I changed `mqtt_messageid` to `mqtt_message_id`, just a style preference on my part.

chrisriley:
You're welcome @Jeffail, glad I could help. I don't mind the change at all, consistency is always better :-)

PR #37, opened 2018-07-27
Add interpolation for AMQP routing key and mapping of metadata to AMQ…

chrisriley:
This adds interpolation to the AMQP routing key. It also adds any defined metadata as AMQP message headers.


Jeffail:
Awesome, thanks @chrisriley!

PR #39, opened 2018-07-29
Expanded the metadata fields supported in the AMQP input

chrisriley:
This PR expands AMQP input metadata functionality to include all fields in the [amqp.Delivery](https://github.com/streadway/amqp/blob/master/delivery.go#L28) struct with the exception of Acknowledger and MessageCount. It also adds support for all types supported by [amqp.Table](https://godoc.org/github.com/streadway/amqp#Table) and includes support for nested amqp.Table structs.

I created a [gist](https://gist.github.com/chrisriley/9bf7f79331f0ecd90d52fecabdb5dda7) that demonstrates these additions.

-Chris

### DIFF
--- a/docs/inputs/README.md
+++ b/docs/inputs/README.md
@@ -128,13 +128,27 @@ Exchange type options are: direct|fanout|topic|x-custom
 This input adds the following metadata fields to each message:
 
 ```
+- amqp_content_type
+- amqp_content_encoding
+- amqp_delivery_mode
+- amqp_priority
+- amqp_correlation_id
+- amqp_reply_to
+- amqp_expiration
+- amqp_message_id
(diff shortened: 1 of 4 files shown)


Jeffail:
Hey @chrisriley, looks great. I'm just playing with the changes locally. I think I might take away the prefix for headers, the intention there is to make a best attempt at passing the unchanged preexisting metadata to sinks.

A general rule I'm leaning towards right now is any specific headers/metadata fields that are taken from a message we try to copy as-is (obviously here you do a best attempt by walking recursive header tables, which I like), but things we extract that aren't explicitly metadata or headers (such as the key, timestamps, etc) we prefix so that they can be easily distinguished.

Issue #49, opened 2018-08-20
Extending jmespath to add metrics from JSON fields


bigkraig:
What are your thoughts on extending (or replicating) the jmespath condition to fetch fields and expose them as metrics? Adding counters/gauges would allow you to scrape the stream from Prometheus and generate alerts and dashboards. We'd be most interested at doing this off of a KSQL table, but I imagine it could be useful from all sorts of inputs.

Jeffail:
Hey @bigkraig, would the goal here be to extract the value of a field from a document and change a metric by that amount? So, for example, for a document `{"foo":{"bar":4}}` we might increment the counter `stats.foo_bars` by 4?

bigkraig:
I guess it could go a few ways.

I may want to count the amount of time I see a field (like a 4xx error) or I may want to expose the value of them, like a transaction total. If I think even further if someone is reading in their logs with response times, they may want to do a histogram.

Obviously it would make sense to just instrument the application, but we have quite a lot of legacy products where that would be difficult to do.

Jeffail:
That sounds interesting, we could add something like a `metric` processor that allows you to mutate either a `counter`, `gauge` or `timing` metric, and the value you specify in configuration could use function interpolation (https://github.com/Jeffail/benthos/blob/master/docs/config_interpolation.md#functions) so that either you specify a static value, or extract an integer value either from metadata, a json path, etc.

If I add that as a processor you could then use a https://github.com/Jeffail/benthos/tree/master/docs/processors#conditional processor to make it conditional based on the type/contents of the message.

It could look something like this:

``` yaml
type: conditional
conditional:
  condition:
    type: jmespath
    jmespath:
      query: 'doctor.error == "people died"'
  processors:
  - type: metric
    metric:
      type: counter
      path: people.that.have.died
      operator: increment
      value: ${!json_field:doctor.people.killed}
```

Or, the shorthand version might just be:

``` yaml
type: metric
metric:
  type: counter
  path: people.that.have.died
  operator: increment
  value: ${!json_field:doctor.people.killed}
```

In which case if `doctor.people.killed` doesn't exist, isn't a number, or returns 0 it's a no-op.

bigkraig:
@Jeffail I think the labels should get included at this point, its going to be kind of limiting without them as I try to apply this to some use cases.

[...]

PR #50, opened 2018-08-20
Adding `metric` processor type

bigkraig:
Addressing #49. I'd like to get some feedback before I go any further. As it is, it appears to be working with my test sources but I have yet to add any tests.

While writing this I did think it might be nice to be able to define some labels, but I didn't see a way of doing that with the metrics interface.

For example if I had this stream of data:
```
{"name": "kraig"} 
{"name": "ashley"} 
{"name": "kraig"} 
{"name": "kraig"} 
```

And a configuration like this:
```
  processors:
    - type: metric
      metric:
        type: counter
        path: names
        labels:
          name: ${!json_field:name}
        value: ${!json_field:name}
```

I'd get the metrics:
```
benthos.names 1 name=ashley
benthos.names 3 name=kraig
```

### DIFF
--- a/lib/processor/constructor.go
+++ b/lib/processor/constructor.go
@@ -75,6 +75,7 @@ const (
 	TypeJSON         = "json"
 	TypeMergeJSON    = "merge_json"
 	TypeMetadata     = "metadata"
+	TypeMetric       = "metric"
 	TypeNoop         = "noop"
 	TypeProcessField = "process_field"
 	TypeProcessMap   = "process_map"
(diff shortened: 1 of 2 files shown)


Jeffail:
Yeah unfortunately right now the metrics package doesn't have any form of labels/tags. We can always add that in later. Looks good so far, although it might be nice to allow an empty value when the type is `counter`, which would be interpretted as 1.

Jeffail:
I would do the check anyway and if false try and parse the value as an int, if that fails return an error.

Jeffail:
Hey @bigkraig, is this ready to merge? If so I'll do it this week.

bigkraig:
@Jeffail I think its good to go 👍

Jeffail:
Done, thanks @bigkraig!

I noticed in unit tests that when interpolating json fields we actually get `null` instead of an empty string when the field doesn't exist, which I'd completely forgotten about. I therefore simplified the behaviour to just always count by 1, if you need this to be conditional on the existence of a field then you can still do that with the `conditional` processor. I'm pretty sure this shouldn't change the way that you were using it, but let me know if that breaks your use case and I'll take a another look.

PR #66, opened 2018-08-28
wip: refactors kinesis output to utilize batch write

cludden:


Jeffail:
Hey @cludden, this looks great. Is it ready to merge?

lib/output/writer/kinesis.go
```diff
@@ -87,16 +113,82 @@ func NewKinesis(
...
+func (a *Kinesis) toRecords(msg types.Message) ([]*kinesis.PutRecordsRequestEntry, error) {
+	entries := make([]*kinesis.PutRecordsRequestEntry, msg.Len())
+
+	err := msg.Iter(func(i int, p types.Part) error {
+		m := message.New(nil)
+		m.Append(p)
+
+		entry := kinesis.PutRecordsRequestEntry{
+			Data:         p.Get(),
+			PartitionKey: aws.String(a.partitionKey.Get(m)),
```
cludden: I believe this is a breaking change from the previous implementation which applied the same partition key to all Kinesis records, would you prefer I change the logic here to be backwards compatible?

cludden:
@Jeffail almost, need to add handling for chunking messages with over 500 parts and at least one integration test.

lib/output/writer/kinesis.go
Jeffail: Nah this is fine, it's a new target and I think this behaviour makes much more sense. In fact, I might change the Kafka output to do the same. As long as it's documented in the output description it should be fine.

cludden:
@Jeffail I think this is at a good point, but let me know if you see anything I missed

Jeffail:
Awesome, thanks @cludden! I merged it basically unchanged but I moved the `kinesis.backoff.max_retries` field to `kinesis.retries` since that's how other similar components look. I really like the backoff settings here so I might replicate them elsewhere (like the `http` processor). I also set some defaults to help new users understand the time string formats.

Jeffail:
Released: https://github.com/Jeffail/benthos/releases/tag/v0.26.3

Jeffail:
@cludden FYI I've done a small refactor of the backoff configuration fields so they can be reused in other components, the impact is that the field `retries` has been renamed (back) to `max_retries` and also the defaults have changed for the interval fields.

PR #68, opened 2018-08-29
Add zip archive/unarchive

Oscil8:
Also adds metadata for the filename of the extracted file, to allow filtering or other behavior based on filename.


Jeffail:
Awesome, thanks @Oscil8!

Jeffail:
Released: https://github.com/Jeffail/benthos/releases/tag/v0.26.3

PR #76, opened 2018-08-31
wip: adds mysql input type

cludden:
@Jeffail what are your thoughts on adding this to the benthos core?


Jeffail:
Hey @cludden, I've just rolled out some experimental plugin options to most of the core Benthos types. It means we could potentially add this in as a plugin, which wouldn't show up as one of the "main" inputs but would still be available in the core project with the following config:

``` yaml
type: mysql
plugin:
  batch_size: 0
  buffer_timeout: ""
  cache: ""
  ... etc ...
```

And would be documented in some other `plugin` section of the Benthos docs. I'm still fleshing out the API and the way that plugins would work functionally by dog fooding with some internal projects we have. I think this is probably a good candidate for a project plugin as it's potentially useful for many people but more of a niche use case.

Would you be willing to try this out? I'd like to take over the development side and when it's ready you could try running it and maybe give some feedback before we merge into master.

cludden:
@Jeffail I like the sound of that. When you say plugin, are you referring to go plugins, benthos plugins, or a combination of both? I'm definitely willing to give it a shot. Keep me posted on progress if possible. Thanks!

Jeffail:
just a pure Benthos plugin, it'll be using this call: https://godoc.org/github.com/Jeffail/benthos/lib/input#RegisterPlugin

Jeffail:
@cludden Got it here: https://github.com/Jeffail/benthos/blob/feature/plugin-example/lib/input/plugins/mysql.go#L48
Might be good to get your eyes on it, does the API look reasonable to you? Going to take another look myself tomorrow and if you are okay with it I'll get it merged into master.

cludden:
While the API is intuitive and I think it could work, my initial thoughts are summarized below.

After thinking it over, my biggest concern is: what is benthos gaining by introducing a second tier of inputs, outputs, processors, and conditions? I think that most users' initial reaction would be that these components are somehow of lesser quality, are less tested/less supported than the core components, or have worse performance.

If those reactions are not true, then I guess I'm failing to see the value of introducing the second tier and scattering implementations and documentation for similar components into separate locations.

If any of these items are true, wouldn't these components be better off implemented as go plugins? I think I could see value in that approach:
- The benthos core remains small and focused on common streaming platform integrations and use cases.
- The core components are well supported, well documented, and well tested.
- The core release binaries and docker images are lightweight and ready to use off the shelf.

If plugins (like the mysql/postgres input types, which you could argue might be heavier than core components and/or fill more of a niche use case) were developed as go plugins, they could be published in their own repositories, have separate artifacts, and first class documentation.

In this scenario, plugins would hopefully have a minimal integration footprint. It could be something like adopting a naming convention similar to terraform providers (`benthos-input-mysql`, `benthos-condition-jsonschema`, etc), introducing new `plugin` configuration for component types and a top level `plugins` configuration section which specifies a `plugins` directory where the benthos core would look for plugin artifacts when plugin configuration is specified (of course defaults could be provided, similar to benthos configuration). It could evolve to support version compatibility requirements, etc.

```yaml
input:
  type: mysql
  plugin:
    batch_size: 0
    buffer_timeout: ""
    cache: ""
    ... etc ...

plugins:
  dir: /etc/benthos/plugins
```

production use should be easy: download the benthos binary and any used plugin artifacts, or the docker equivalent:
```dockerfile
FROM jeffail/benthos:latest

# install plugins
RUN wget -O /etc/benthos/plugins/benthos-input-mysql https://github.com/benthos-plugins/benthos-input-mysql/releases/download/v1.0.0/benthos-input-mysql-1.0.0-linux-amd64.so
```

The above approach is just a hypothetical implementation. I have not yet used any go projects that utilize native plugins, so there may be additional pain points that I'm unaware of. Also, I do see how the above approach makes benthos less adoptable, with a scattered ecosystem of components. That is definitely not an issue that I would want to overlook.

For example, being a heavy terraform user, I've seen how quality vastly differs between non core terraform providers, which ways on adoption decisions, and in an ideal world, benthos would avoid those pitfalls. I do think terraform mitigates this somewhat by differentiating between core/official plugins/providers and third party by including the core implementations in the official distribution channels. The biggest difference between terraform and the current implementation is that there are not two tiers of bundled terraform providers. All of the official terraform providers are included in the terraform binary, follow the same implementation/documentation patterns, and are expected to be of the same quality. Third party providers, on the other hand, require users to download provider binaries into the terraform plugins directory (`~/terraform.d/plugins`) and have the `YMMV` disclaimer.

I know that terraform and benthos have very different use cases, but I think they share similarities in that they aim to seamlessly integrate with a wide variety of providers (or in benthos' case, inputs and outputs). Maybe it would be helpful to hear what pain points your aiming to solve with the current implementation and your thoughts on my initial reaction?

Jeffail:
@cludden Thanks for the feedback, this is great. I agree on all of your points here. I'm trying to avoid fragmentation by placing them in the main project binary, but this is very unlikely to scale gracefully and it's confusing from a users perspective.

I'm definitely open to trying out Go plugins, from what I understand about plugins the API we already have should work just fine. I'll take a stab at making an example plugin this way and see what it looks like in practice.

Jeffail:
I've had a play around, the implementation is the exact same and works. However, there are lots of caveats when it comes to builds. I'm posting further findings here: https://github.com/Jeffail/benthos/issues/83

PR #77, opened 2018-08-31
feat: adds dynamodb cache type

cludden:


Jeffail:
Amazing, thanks @cludden!

PR #78, opened 2018-08-31
fix: adds endpoint config field for kinesis input type to support vpc…

cludden:
… endpoints


Jeffail:
Ace, thanks @cludden!

PR #79, opened 2018-09-02
feat: adds switch output type

cludden:


Jeffail:
Hey @cludden, I like the pattern but I'm not sure I'd want to add this to the broker type as the configuration spec would get muddied. I could imagine this as a new type, maybe called `switch`, which a spec something like this:

``` yaml
type: switch
switch:
- output:
    type: foo
    foo: {}
  condition:
    type: foo
    foo: {}
- output:
    type: bar
    bar: {}
  condition:
    type: bar
    bar: {}
```

I've created an issue for it where we can discuss further: https://github.com/Jeffail/benthos/issues/81

lib/output/switch.go
```diff
@@ -0,0 +1,341 @@
...
+}
+
+//------------------------------------------------------------------------------
+
+// loop is an internal loop that brokers incoming messages to many outputs.
+func (o *Switch) loop() {
+	defer func() {
+		for _, c := range o.outputTsChans {
+			close(c)
+		}
```
Jeffail: Would be good to block here until all child outputs are closed.

lib/output/switch.go
```diff
@@ -0,0 +1,341 @@
...
+			if oConf.cond == nil || oConf.cond.Check(ts.Payload) {
+				outputTargets = append(outputTargets, o.outputNs[i])
+				if !oConf.Fallthrough {
+					break
+				}
+			}
+		}
+		if len(outputTargets) == 0 {
+			select {
+			case ts.ResponseChan <- response.NewError(ErrSwitchNoConditionMet):
```
Jeffail: I think in this event we should probably return `response.NewAck()` and be sure to expose the drop via metrics, it's possible that users might intentionally set up their outputs to entirely drop certain message types.

cludden: I was hoping to find someway to avoid situations where messages are unintentionally dropped by forcing users to use some combination of `filter` processors and/or `/dev/null` outputs for messages where dropping is the desired behavior. what are your thoughts on adding an additional config field to tune this behavior, either `ack_unmatched: true`, `require_match: true`, `min_match: 1`, or something similar?

```yaml
output:
  type: switch
  switch:
    ack_unmatched: true
    outputs:
    - output:
        type: foo
        foo:
          foo_field1: blah
      condition:
        type: text
        text:
          operator: contains
          arg: foo
    - output:
        type: bar
        bar:
          bar_field1: blah
      condition:
        type: text
        text:
          operator: contains
          arg: bar
```

vs

```yaml
output:
  type: switch
  switch:
    outputs:
    - output:
        type: foo
        foo:
          foo_field1: blah
      condition:
        type: text
        text:
          operator: contains
          arg: foo
    - output:
        type: bar
        bar:
          bar_field1: blah
      condition:
        type: text
        text:
          operator: contains
          arg: bar
    - output:
        type: stdout
        processors:
          - type: filter
            filter:
              type: static
              static: true
```

lib/output/switch.go
cludden: I agree. It's not clear to me how to achieve this though, without changing the behavior to call each child outputs `CloseAsync` and `WaitForClose` methods, but even with that approach, it's not clear how to pass the wait timeout duration to child outputs.. how is the `fan_out` broker accomplishing this? This output borrowed heavily from that implementation.

Jeffail: Looks like the output brokers are missing this behaviour which I should look at, the dynamic fan out does it this way:

``` go
for _, ow := range d.outputs {
	ow.output.CloseAsync()
}
for _, ow := range d.outputs {
	if err := ow.output.WaitForClose(time.Second); err != nil {
		for err != nil {
			err = ow.output.WaitForClose(time.Second)
		}
	}
}
```

It's fine to use a hardcoded `time.Second` here, as the intention is to block forever regardless until it exits.

lib/output/switch.go
Jeffail: I'd prefer `ack_unmatched` (maybe instead called `drop_unmatched`) out of those options. But I'm also happy to leave it out entirely, I think it's reasonable to expect the messages to be filtered out if you haven't created an output that matches them.

lib/output/switch.go
```diff
@@ -0,0 +1,378 @@
...
+	throt *throttle.Type
+
+	transactions <-chan types.Transaction
+
+	outputTsChans  []chan types.Transaction
+	outputResChans []chan types.Response
+	outputs        []types.Output
+	outputNs       []int
+	confs          []switchConfigOutputs
+	typeStr        string
```
Jeffail: Doesn't look like this gets set anywhere.

Jeffail: Actually it's also worth simplifying your metrics to only increment values for `broker.switch.foo`, the idea is that only a final output target increments the shorter `output.foo` paths, otherwise we're double counting messages that leave the service.

Jeffail:
Actually I wouldn't worry about unit testing the processors. It's outside of the scope of your implementation and hard to test with mock output types. Just documenting the behaviour is fine.

Jeffail:
In fact I wouldn't bother documenting it either, it's expected behaviour. To summarise my morning ramblings: just simplify the metrics to be `broker.switch.foo` and this is ready to merge.

Jeffail:
Awesome, thanks @cludden!

PR #80, opened 2018-09-02
feat: adds regexp_partial and regexp_exact operators to metadata cond…

cludden:
…ition


Jeffail:
Ace, thanks @cludden!

PR #85, opened 2018-09-03
feat: add hdfs writer

ajbosco:


Jeffail:
Hey @ajbosco, looks great. Is there any chance you could update the unit tests to use a docker instance? Not sure how realistic it would be to do it that way but it would be nice to have.

Jeffail:
Ace, thanks @ajbosco!

Issue #84, opened 2018-09-03
proposal: refactor conditions to support operator configuration


cludden:
I am working on adding additonal operators to both the `metadata` and `jmespath` (which currently does not support operators) conditions. One of my use cases is to match against a long list of enumerations (to go along with the switch output). It currently feels awkward to force config for an `enum` operator into the string `arg` field. I'm proposing changing the way condition configuration is currently implemented to mirror the configuration structure of other benthos components (namely inputs, outputs, and processors). Below is an example to illustrate the proposal:

change this:
```yaml
condition:
  type: metadata
  metadata:
    operator: less_than
    key: foo
    part: 0
    arg: "3"
```

to this:
```yaml
condition:
  type: metadata
  metadata:
    operator: less_than
    key: foo
    part: 0
    less_than: 3
```

The difference may seem subtle, but is better illustrated with more complex (hypothetical) operators:

```yaml
condition:
  type: metadata
  metadata:
    operator: enum
    key: foo
    part: 0
    enum:
      - "a"
      - "very, long"
      - "list"
      - "of"
      - "values!"
```

or

```yaml
condition:
  type: jmespath
  jmespath:
    operator: json_schema
    query: foo
    part: 0
    json_schema:
      type: object
      properties:
        bar:
          type: string
          pattern: date-time
        baz:
          type: integer
          minimum: 5
          exclusiveMinimum: true
        qux:
          type: array
          uniqueItems: true
          items:
            type: string
            enum: ["quux","quz","corge"]
      required:
        - bar
        - baz
```

We could maintain backwards compatibility by continuing to support the `arg` field for existing operators. @Jeffail what are your thoughts?

Jeffail:
Hey @cludden, I'm open to it but would be worried about the rate at which the config spec might grow as we add more operators. As an alternative suggestion, what if we changed the type of `arg` to be an `interface{}`?

Also for the example of adding a `json_schema` operator to `jmespath` I think it might be more generally useful to have something similar to the `process_field` processor for conditions, maybe called something like `check_field`, where the config might look something like this:

``` yaml
condition:
  type: check_field
  check_field:
    path: foo
    conditions:
    - type: json_schema
      json_schema:
        type: object
        ... etc ...
```

That way we could recycle it for doing things like `text` conditions on a JSON field.

cludden:
[...]

PR #86, opened 2018-09-04
feat: adds 'enum' and 'has_prefix' operators to 'metadata' condition

cludden:


cludden:
took a stab at part I of #84, i feel like there has to be a better way of handling the interface{} `arg` field..

Jeffail:
Hey @cludden, the rawJSONValue type is useful if we wanted to do function interpolation on the structure, or if the value could potentially be an arbitrary map (YAML would parse it as a map[interface{}]interface{} which is problematic). 

I think for this use case we can get away with just having an straight up `interface{}` type. When we construct the operator we either try to cast to `string`, or to `[]interface{}` and convert it into `[]string`.

I was thinking this morning that an alternative to a generic type here might be to have `arg string`
and `arg_list []string`, and then maybe `arg_map map[string]string` in the future if we ever needed that. This would allow us to have an explicit type (still limited to only strings) and also a cap on the number of fields as the operator list grows. What would be your thoughts on that approach?

cludden:
i refactored this to use the empty `interface{}` type for arg, and included github.com/spf13/cast to allow individual operator factories to cast arg to complex types and avoid cluttering each operator with thorough [casting logic](https://github.com/spf13/cast/blob/master/caste.go#L1040). I debated the addition of this package, but with the bulk of configs likely being yaml, and all of the weirdness that comes with yaml parsing, it seemed like this could be useful to more operators if they require complex args. if you're opposed, let me know and I can take it out and inline all of the casting logic..

Jeffail:
Looks great, I think that's reasonable. We ought to be aware that `cast` is very liberal when it comes to input validation, for example I believe the string "foo bar baz" would be parsed into a `[]string` without error, resulting in `[]string{"foo","bar","baz}`, but I'm not convinced this is likely to hurt anyone.

Are you ready for this to be merged? My only nag would be to change the tabs in your `yaml` snippets to spaces.

cludden:
I did see that, but agreed that the impact is probably not harmful. I've gone ahead and fied the yaml whitespace as well. consistency is always good!

Jeffail:
@cludden just noticed that the `has_prefix` operator parses into a slice. That would mean that if I have an `arg` of `"foo bar baz "` then this operator would pass for messages I wouldn't expect it to, such as "bargraph".

Got example test cases here:

``` go
		{
			name: "has_prefix neg 2",
			fields: fields{
				operator: "has_prefix",
				key:      "foo",
				part:     0,
				arg:      "foo bar baz",
			},
			arg: map[string]string{
				"foo": "barley",
			},
			want: false,
		},
		{
			name: "has_prefix pos 2",
			fields: fields{
				operator: "has_prefix",
				key:      "foo",
				part:     0,
				arg:      "foo bar baz",
			},
			arg: map[string]string{
				"foo": "foo bar bazley",
			},
			want: true,
		},
```

cludden:
I refactored the operator to handle both `string` and `[]string` arg variants, which handles your test cases and but still supports checking against a list of prefixes. Does that work?

Jeffail:
Yep, this is great. Ready for me to merge?

cludden:
:+1:

Jeffail:
Thanks @cludden!

PR #87, opened 2018-09-05
feat: add hdfs reader

ajbosco:


Jeffail:
Looks great but can you change the metadata key being added to `hdfs_path`? I'd like to try and keep the metadata we add prefixed with the input type when appropriate.

Jeffail:
Dope, thanks @ajbosco. I added another metadata field so we now have both `hdfs_path` and `hdfs_name`. I also moved and combined the integration tests as there was a port collision.

PR #90, opened 2018-09-05
Feature: More Flexible TLS Config

ghballiet:
We need to load raw certs as strings as well as from files in some cases, and this lets us do that without any breaking changes.


ghballiet:
Relates to: #91

Jeffail:
Dope, thanks @ghballiet!

Issue #95, opened 2018-09-07
New `bounds_check` condition type


cludden:
use case: a `bounds_check` condition would be useful for routing messages to outputs. certain outputs (eg Kinesis), have a maximum message size, and certain inputs produce messages that may exceed those message size limits. a `bounds_check` condition would allow messages that exceed to bounds to be routed to an alternative output.

```yaml
output:
  type: switch
  switch:
    outputs:
    # send all messages to kinesis, as long as they do not exceed payload limit of 1mib
    - output:
        type: kinesis
        kinesis:
          # ...
      condition:
        type: bounds_check
        bounds_check:
          max_part_size: 1.049e+6
    # send any messages that are not sent to kinesis to s3 for additional processing
    - output:
        type: s3
        s3:
          # ...
```

PR #96, opened 2018-09-07
feat: adds `bounds_check` condition

cludden:
implements #95


Jeffail:
Cool, thanks @cludden!

PR #100, opened 2018-09-11
Adding nats queue as an option

nwest1:
This does change default behavior, but it does match the nats_stream behavior. Not sure which should win.


Jeffail:
Hey @nwest1, looks good, can you run `make docs`?

Jeffail:
Cool, thanks @nwest1!

PR #103, opened 2018-09-17
Support timeout for combine processor

andytruong:
Hi there,

Please review my idea for https://github.com/Jeffail/benthos/issues/101

Thanks

### DIFF
--- a/lib/processor/combine.go
+++ b/lib/processor/combine.go
@@ -26,6 +26,8 @@ import (
 	"github.com/Jeffail/benthos/lib/metrics"
 	"github.com/Jeffail/benthos/lib/response"
 	"github.com/Jeffail/benthos/lib/types"
+	"sync"
+	"time"
 )
 
 //------------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


lib/processor/combine.go
```diff
@@ -115,6 +123,22 @@ func NewCombine(
...
+		c.started = true
+
+		go func() {
+			ticker := time.NewTicker(c.timeout)
+
+			for {
+				<-ticker.C
+
+				if len(c.parts) >= c.n {
+					c.flush()
```
Jeffail: Messages flushed here are lost forever

Jeffail:
Hey @andytruong, I think I'd rather add batching by message count to the `batch` processor rather than add timeouts to the `combine` processor, then I can deprecate and eventually remove `combine` entirely since they basically achieve the same goal. Also there's a few issues with this pull request, the main one being that your timed flush calls don't have a way of propagating messages down the pipeline.

I've made an issue for adding message count to the `batch` processor: https://github.com/Jeffail/benthos/issues/104

PR #107, opened 2018-09-18
fix(kinesis-output): adds new metrics for counting throttled writes

cludden:


Jeffail:
Cool, looks good. Thanks @cludden!

PR #116, opened 2018-10-10
fix(input/s3): Use paged object list method

### LINKED ISSUE #115: AWS S3 input capped at 1000 items
When the [AWS S3](https://github.com/Jeffail/benthos/blob/master/lib/input/reader/amazon_s3.go#L139) input is used without an SQS queue the input is meant to download every object within the bucket under a specified prefix. However, it appears as though the API is capped at 1,000 objects, so if the bucket exceeds that we lose data.

It looks as though [the API now allows pagination](https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#S3.ListObjectsV2) when iterating keys, so we need to start using it.

ChristianAlexander:
Fixes #115 

Allows for more than 1,000 objects to be processed.

### DIFF
--- a/lib/input/reader/amazon_s3.go
+++ b/lib/input/reader/amazon_s3.go
@@ -143,15 +143,19 @@ func (a *AmazonS3) Connect() error {
 		if len(a.conf.Prefix) > 0 {
 			listInput.Prefix = aws.String(a.conf.Prefix)
 		}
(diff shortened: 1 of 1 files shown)


Jeffail:
Ace, thanks @ChristianAlexander!

PR #120, opened 2018-11-01
add extra configuration to the gcp_pubsub input

iluminae:
Added a few extra configuration options to the google pubsub input. This will allow for the following:
```yaml
input:
  type: gcp_pubsub
  gcp_pubsub:
    project: myproject
    subscription: my-subscription
    max_outstanding_messages: 2000
pipeline:
  threads: 1
  processors:
  - type: batch
    batch:
      count: 2000
      period_ms: 1000
output:
  type: "stdout"
```
before, this will deadlock, as the default `MaxOutstandingMessages` is 1000 in the golang pubsub bindings.


Jeffail:
Cool, thanks @iluminae!

Issue #38, opened 2018-07-27
Metadata fields


Jeffail:
I've added support for metadata fields within Benthos, that means three things:

- Inputs can add arbitrary metadata to messages
- The `metadata` processor can be used to modify the metadata of a message
- Function interpolation can be used to access metadata fields with `${!metadata:foo}`

This means we can now start rolling out metadata support to each input and output. Inputs can be modified to add metadata. Outputs can be modified to both allow function interpolation in more fields and add all metadata fields to messages if there's an appropriate way to do so.

I'll list all inputs and outputs here that are candidates for supporting metadata, and I'll remove the ones that have been done.

Outputs:
- amazon_s3
- amazon_sqs
- elasticsearch
- http_client
- http_server
- kafka
- mqtt
- nats
- nats_stream
- nsq
- websocket

Jeffail:
Reviewed all the inputs and added metadata from the ones I felt made sense.

chrisriley:
I've been working on a version of the amqp input that includes all but Acknowledger and MessageCount from the [amqp.Delivery](https://github.com/streadway/amqp/blob/master/delivery.go#L28) struct. Still working out the nesting of [amqp.Table](https://godoc.org/github.com/streadway/amqp#Table) in Delivery.Headers. I'll send you a PR when I'm done to see if you can use any of it.

Jeffail:
Hey @seanlaff, how would you intend on using the metadata? For cases like that I would usually either suggest setting and referring to the batch wide metadata on the first part, or just set the field on all parts.

seanlaff:
@Jeffail Inspecting the metadata on the first part is the solution I opted for in the end, and works pretty well. My use case may be a little contrived, but this is what it looks like:

Data is streamed from pubsub to a benthos batch processor, and then to a database sink. However the database table that the data is written into depends on a dynamic field in the message. Additionally, the database does not support writing to multiple tables in one request.

My first solution was to spin up goroutines in the output to handle writing to the N discrete tables concurrently- but that felt like work I could relegate to benthos itself. Instead, I opted to create a processor (executed after the batch processor) that creates N new multi-part message, with each message containing all parts that share the same keying field. Then in my output I inspect the metadata of the first part, and make a singular database write for the correct table for all parts of that message at once (thus leaning on benthos to do the concurrent work I was doing by hand before)

This may be a non-idomatic use of benthos- and is only a minor inconvenience since I can achieve the same by checking the metadata on the first part

[...]

PR #128, opened 2018-11-20
add metadata to kafka headers (#38)

ghballiet:
Copies message metadata into Kafka message headers for Kafka output type (partial implementation of #38).


Jeffail:
Ace, thanks @ghballiet!

PR #145, opened 2019-01-11
fix(awk-processor): should fail when interpreter exit with non-zero status code

jessynt:


Jeffail:
Awesome, thanks @jessynt!

PR #153, opened 2019-01-30
Adding option to set Transport in HTTP Client

asiragusa:
This PR adds the `OptSetHTTPTransport` function in `lib/util/http/client/type.go`. It is useful when using Benthos as a framework to set a custom HTTP Transport. 

In my case I have to whitelist the IP addresses which I connect to, banning the local addresses.

### DIFF
--- a/lib/util/http/client/type.go
+++ b/lib/util/http/client/type.go
@@ -234,6 +234,13 @@ func OptSetManager(mgr types.Manager) func(*Type) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


Jeffail:
Looks good. Just a heads up though, using this option would override any config fields referring to TLS settings, since they get applied here: https://github.com/Jeffail/benthos/blob/master/lib/util/http/client/type.go#L139

asiragusa:
Tanks for the quick reply! I know, but given that you are explicitly setting the Transport it shouldn't be a problem

Issue #146, opened 2019-01-18
Metric white/blacklisting


Jeffail:
It's possible to create a metric target type that shields a child metric target type with a whitelist of blacklist of metric prefixes. It would look in config something similar to this:

```
type: whitelist
whitelist:
  prefixes:
  - pipeline.processor
  child:
    type: statsd
    statsd:
      ...
```

This would result in only metrics with the path prefix `pipeline.processor` being sent to statsd. This would make it a lot easier to prevent the flooding of metric aggregators for when cost is a constraint.

danrubenstein:
Hi @Jeffail!

I've started to take this on - you can see an initial cut of the whitelist version of this [here](https://github.com/danrubenstein/benthos/commit/50d5d6b2ef9424e9fc1f8ab92e9cbc7c47d83632). It's not ready to go yet, there's still yet more to do with blacklist and the configuration management.

One scope question that I have for this is whether or not you'd be willing to accept regular expressions as a mode for white or blacklist - e.g., 

```
type: whitelist
   prefixes:
   - output.broker.outputs.*.output
   - ...
   mode: regex|direct
   child:
     ...
```
This would help with things like broker / switch etc outputs, where you might want to keep some but not all of the metrics in a sub-output or input.

Let me know what you think here - thanks!

Jeffail:
Hey @danrubenstein, awesome! I think regular expressions are a good idea, in which case I would probably have another config field `patterns` that contains them rather than having a `mode` field, it would potentially allow someone to have both prefix based and regexp based white/blacklists if they wanted to.

Your branch looks good, but I would suggest using `*Config` instead of `[]Config` here: https://github.com/danrubenstein/benthos/commit/50d5d6b2ef9424e9fc1f8ab92e9cbc7c47d83632#diff-2dba46c44615dcd5bf6140a61921e30bR49. There's an example of a component with a single child here: https://github.com/Jeffail/benthos/blob/master/lib/input/read_until.go#L81, but it requires some nasty trickery so I'm happy to implement that part for you when you're happy with the rest.

danrubenstein:
Glad you brought that up -- I was struggling with the single child bit, so I decided to just punt and go with a list (basically modeleted after the `broker` output) until I could make enough progress to get to a proof of concept.

I'll look at the example you posted and see if I can get through implementing a single child component, and then I'll push towards `patterns` for regex. 

Thanks for the pointers here - I'll post back once I've either hit a wall or made enough progress to be of note.

[...]

PR #160, opened 2019-02-12
Add metric white and blacklisting

danrubenstein:
Continued conversation from #146.

I have a specific few questions here about this, as well as looking for general feedback.

1. How should these classes handle regex match errors? 

Looking through the [regex source](https://golang.org/src/regexp/regexp.go), I started reading the docs for `compile` (line 169) before my eyes started to glaze over. Since metrics namespaces are dynamically allocated (e.g., we don't have a natural hook to all of them at the create time of the metric counter). A few options for consideration:

- Compare it against an empty string (or `"benthos"` if we wanted to be clever), and if it threw any errors, we could reject the metric creation entirely.
- Silently reject (in the case of whitelist) or accept (for blacklist) any metrics, and rely on other metrics, and then spit out a log line saying "WARNING this path comparison failed" (or something to that effect).

2. How should these things be tested?

I'm not super familiar with how in Go, we would test that the Whitelist and Blacklist collectors are returning the right struct - I got a little ways into that rabbit hole (ad86dfc), but I got through down in reading about the [Reflect package](https://golang.org/pkg/reflect/#Kind.String) and didn't see anything super helpful - it looks to me that this library is mostly for Go's primitive types, rather than structs.

Anyways, thanks for the time and consideration.

### DIFF
--- a/lib/metrics/blacklist.go
+++ b/lib/metrics/blacklist.go
@@ -1,0 +1,245 @@
+package metrics
+
+import (
+	"encoding/json"
+	"errors"
+	"net/http"
+	"regexp"
+	"strings"
+
+	"github.com/Jeffail/benthos/lib/log"
+)
+
+// BlacklistConfig allows for the placement of filtering rules to only allow
+// metrics that are not matched to be displayed or retrieved. It has a set of
+// prefixes (direct string comparison) that are checked as well as a set of
+// regular expressions for more precise control over metrics. It also has a
+// metrics configuration that is wrapped by the Blacklist.
+
+//------------------------------------------------------------------------------
+
+func init() {
+	constructors[TypeBlackList] = typeSpec{
+		constructor: NewBlacklist,
+		description: `
+Blacklist a certain set of paths around a child metric collector.
+
+### Patterns and paths
+
+Blacklists can be one of two options, paths or regular expression patterns.
+A metric path's eligibility is strictly additive - it only has to pass a
+single path or a single pattern for it to be not included.
+
+An entry in a Blacklist's ` + "`paths`" + `field will check using prefix
+matching. This can be used, for example to allow none of the  metrics from the
+` + "`output`" + `stats object to be pushed to the child metric collector.
+
+An entry in a Blacklist's ` + "`patterns`" + `field will check using Go's
+` + "`regexp.MatchString`" + ` function, so any submatch in the final path will
+result in the metric being allowed. To anchor a pattern to the start or end of
+the word, you might use the ` + "`^`" + ` or ` + "`$`" + ` regex operators.`,
+	}
+}
+
+//
+
+//------------------------------------------------------------------------------
+
+// BlacklistConfig is a list of configs...
+type BlacklistConfig struct {
+	Paths    []string `json:"paths" yaml:"paths"`
+	Patterns []string `json:"patterns" yaml:"patterns"`
+	Child    *Config  `json:"child" yaml:"child"`
+}
+
+// NewBlacklistConfig returns the default configuration for a Blacklist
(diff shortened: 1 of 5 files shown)


Jeffail:
Hey @danrubenstein, looking good.

You can compile the regular expression patterns during construction using [`Compile`](https://golang.org/pkg/regexp/#Compile), which would expose broken patterns during service initialization. Then when metrics are registered you would only have to call [`MatchString`](https://golang.org/pkg/regexp/#Regexp.MatchString) which doesn't return errors.

For unit tests it might be easier to create a mock implementation of a metric type that simply records metric paths that are registered to a map.

danrubenstein:
Thanks for the feedback - this was very helpful. 

I went the `compile` route for patterns, so I can store them in the struct at initialization.

I also went the recommended route for unit tests -- creating a somewhat fake `http_server` in the Whitelist and Blacklists. This does mean that the test set up is a little bit extensive, but without this I'm not sure I can get to the underlying `Local` map.

Let me know what you think - thanks!

Jeffail:
Looks good so far, I'll do a proper comb through this tomorrow if that's okay.

I've also implemented the plumbing needed to get nicer metrics documentation and config examples, got it on a branch here: https://github.com/Jeffail/benthos/tree/feature/metrics-docs, you don't need to do anything as I'll merge after this and will fix anything that needs to change as I go.

danrubenstein:
No rush, thanks for taking a look whenever.

The plumbing looks good too - now that there's getting to be a few more metrics options, having the docs and example ready to go is great.

lib/metrics/blacklist.go
```diff
@@ -0,0 +1,246 @@
...
+
+// SetLogger sets the logger used to print connection errors.
+func (h *Blacklist) SetLogger(log log.Modular) {
+	h.s.SetLogger(log)
+}
+
+// Close stops the Statsd object from aggregating metrics and cleans up
+// resources.
+func (h *Blacklist) Close() error {
+	h.s.Close()
```
Jeffail: There might be an error returned from the child so change this to `return h.s.Close()`

lib/metrics/blacklist.go
```diff
@@ -0,0 +1,246 @@
...
+
+// HandlerFunc returns an http.HandlerFunc for accessing metrics for appropriate
+// child types
+func (h *Blacklist) HandlerFunc() http.HandlerFunc {
+	if wHandlerFunc, ok := h.s.(WithHandlerFunc); ok {
+		return wHandlerFunc.HandlerFunc()
+	}
+
+	return func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(400)
```
Jeffail: Return a 501 here instead

lib/metrics/blacklist_test.go
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) 2014 Ashley Jeffs
```
Jeffail: Change name and year

danrubenstein:
Thanks for the feedback - I think I caught all of the updates you flagged, but if I missed any, please let me know.

Jeffail:
Hey @danrubenstein, 

This is merged but I used the command tools to squash it in a way that preserves you as the author, but the consequence of that is this pull request doesn't get the satisfying `merged` tag.

Commit: https://github.com/Jeffail/benthos/commit/a1004070b3569d988a07be906e55fafa34191c5a

Docs are being generated now: https://github.com/Jeffail/benthos/tree/master/docs/metrics#blacklist (made some small tweaks to clarify prefix behaviour)

These components are awesome, thank you!

danrubenstein:
Haha no worries - glad we were able to get this feature out there. Thanks for your help!

PR #165, opened 2019-02-20
adds nats sequence number to metadata

thinktainer:
In our use case being able to add original sequence to dead lettered messages is useful. I hope this is uncontroversial.


thinktainer:
Apologies for the seemingly unnecessary import modification, my tooling seems to add that automatically. Happy to look into removing if it's a bother to you.

Jeffail:
Hey @thinktainer, no problem with the import change. Thanks!

PR #170, opened 2019-03-05
Redis function interpolation

### LINKED ISSUE #169: Add function interpolation to redis_pubsub output
Adding function interpolation to the `channel` field of the `redis_pubsub` output would allow using processors to alter the data destination via setting (for example) metadata on each message.

This appears to be a feature that I will be needing, and have implemented. I don't see a downside, so I will be contributing the code up to go with this issue.

I do not currently _need_ the `redis_list` or `redis_stream` output to behave the same way, but I see the value in having them be consistent.

iluminae:
@Jeffail here is a PR giving function interpolation to the redis_pubsub output.

Closes #169

### DIFF
--- a/lib/output/writer/redis_pubsub.go
+++ b/lib/output/writer/redis_pubsub.go
@@ -26,8 +26,10 @@ import (
 	"time"
 
 	"github.com/Jeffail/benthos/lib/log"
+	"github.com/Jeffail/benthos/lib/message"
 	"github.com/Jeffail/benthos/lib/metrics"
 	"github.com/Jeffail/benthos/lib/types"
+	"github.com/Jeffail/benthos/lib/util/text"
 	"github.com/go-redis/redis"
 )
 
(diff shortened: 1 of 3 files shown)


lib/output/writer/redis_pubsub.go
```diff
@@ -44,7 +46,7 @@ type RedisPubSubConfig struct {
 func NewRedisPubSubConfig() RedisPubSubConfig {
 	return RedisPubSubConfig{
 		URL:     "tcp://localhost:6379",
-		Channel: "benthos_chan",
+		Channel: "benthos_chan_${!metadata:channel}",
```
Jeffail: Unfortunately, this needs to stay as `benthos_chan` by default otherwise it's breaking change.

iluminae: reverted this, pushed

Jeffail:
Awesome, thanks @iluminae!

Issue #171, opened 2019-03-08
gcp_pubsub input runs administrative API call


iluminae:
The call of `sub.Exists(existsCtx)` in the `Connect()` function of the `gcp_pubsub` input requires use of the administrative call quota. In our use case, we were met with this quota being filled (by a bug in dataflow, we believe) and our benthos based pipelines would not run. However, with a quick removal of that call, they were able to successfully start, as no calls were made that used the quota: `Administrator operations per minute`

Would we be interested in merging in a option to run that `Exists()` call in the `Connect()` for that input only if a boolean is set to true? We could default it to true for backwards compatibility - or make the boolean negative (eg: `skipExistsCheck: false`). I would be happy to contribute the solution.

Jeffail:
Hey @iluminae, since the call consumes quota I'd support simply removing it. From what I understand the call to `Exists` is just a nicety and if the subscription doesn't exist Benthos should report and recover as you would expect.

Since you are testing this already would you be willing to test what happens when the subscription doesn't exist when Benthos starts up with the Exists check removed? If it still begins consuming once the subscription does exist without needing a restart then that's ideal.

iluminae:
@Jeffail yep, I would support that. Im sure that will work, but I will test it to make sure. I will submit the change soon after.

iluminae:
@Jeffail it would appear that a failure from the `Connect()` code properly engages the throttling that is desired (normal throttling is set to 1/s i believe) but when a failure occurs in the read loop, the `Connect()` is called again, but without a throttle. I believe this is a bug, correct? im talking about this loop here in `lib/inputs/reader.go`

```go
		msg, err := r.reader.Read()

		// If our reader says it is not connected.
		if err == types.ErrNotConnected {
			mLostConn.Incr(1)
			atomic.StoreInt32(&r.connected, 0)

			// Continue to try to reconnect while still active.
			for atomic.LoadInt32(&r.running) == 1 {
				if err = r.reader.Connect(); err != nil {
					// Close immediately if our reader is closed.
					if err == types.ErrTypeClosed {
						return
					}

					r.log.Errorf("Failed to reconnect to %v: %v\n", r.typeStr, err)
					mFailedConn.Incr(1)

					if !r.connThrot.Retry() {
						return
					}
				} else if msg, err = r.reader.Read(); err != types.ErrNotConnected {
					mConn.Incr(1)
					atomic.StoreInt32(&r.connected, 1)
					r.connThrot.Reset()
					break
				}
[...]

PR #172, opened 2019-03-08
Skip subscription check on gcp_pubsub

iluminae:
see #171 for discussion. Includes connect throttle on `Read()` errors.


Jeffail:
Awesome, thanks @iluminae!

PR #176, opened 2019-03-14
NATS function interpolation

### LINKED ISSUE #175: add function interpolation to the nats output
This is the same thing that #169 implemented for the `redis_pubsub` output, and will allow setting the subject of the message based on eg: metadata on the message. I will make a PR with this in the same implementation as #169 if thats ok.

iluminae:
see #175. This adds function interpolation to the `subject` field in the NATS output which will allow flexibility in setting outputs via steps in the pipeline. Implemented like #170.

Closes #175


Jeffail:
Looks great, thanks @iluminae!

PR #184, opened 2019-04-17
Add DynamoDB Streams Lambda input

asiragusa:
Hi @Jeffail, as promised I am releasing the input source I did for DynamoDB Streams running in a Lambda.

The setup is quite straightforward:
- Enable Streaming in the DynamoDB table
- Compile & zip the executable
- Properly setup IAM & VPC (if needed)
- Create the lambda & add a DynamoDB trigger 

This input has no config parameters and it's as simple to use as:
```
input:
  type: dynamodb_streams_lambda
```

In my setup I use environment variables to pass the configuration, it should be possible to pass a cli flag containing the config file path (which should be shipped together with the .zip)

This code runs in production and I have tested it with few hundreds events/s.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/Shopify/sarama v1.21.0
 	github.com/armon/go-metrics v0.0.0-20180917152333-f0300d1749da // indirect
 	github.com/armon/go-radix v1.0.0
+	github.com/aws/aws-lambda-go v1.10.0 // indirect
 	github.com/aws/aws-sdk-go v1.17.10
 	github.com/benhoyt/goawk v1.4.1
 	github.com/bradfitz/gomemcache v0.0.0-20180710155616-bc664df96737
--- a/lib/input/constructor.go
+++ b/lib/input/constructor.go
@@ -61,104 +61,107 @@ var Constructors = map[string]TypeSpec{}
 
 // String constants representing each input type.
 const (
-	TypeAMQP          = "amqp"
-	TypeBroker        = "broker"
-	TypeDynamic       = "dynamic"
-	TypeFile          = "file"
-	TypeFiles         = "files"
-	TypeGCPPubSub     = "gcp_pubsub"
-	TypeHDFS          = "hdfs"
-	TypeHTTPClient    = "http_client"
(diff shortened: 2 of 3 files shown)


Jeffail:
Hey @asiragusa, thanks! This is very interesting. Your config ticket makes a lot more sense to me now. I think rather than having lambda invocations exist as an input type this might be better suited as a separate distribution of Benthos that specializes in this type of deployment. I'm going to create a ticket for dumping my thoughts about this.

Jeffail:
Hey @asiragusa, I'd like to proceed with the approach we have in https://github.com/benthosdev/benthos-lambda, for your specific use case here it means you'd need to handle dynamodb events (converting records into messages of a batch) using Benthos processors:

``` yaml
pipeline:
  processors:
  - type: jmespath
    jmespath:
      query: Records
  - type: unarchive
    unarchive:
      format: json_array
  - type: awk
    awk:
      program: |
        {
          metadata_set("dynamodbstreams_event_id", json_get("eventID"));
          metadata_set("dynamodbstreams_event_name", json_get("eventName"));
          metadata_set("dynamodbstreams_event_source", json_get("eventSource"));
          metadata_set("dynamodbstreams_event_source_arn", json_get("eventSourceArn"));
          metadata_set("dynamodbstreams_event_version", json_get("eventVersion"));
          if (json_get("userIdentity") != "null") {
            metadata_set("dynamodbstreams_user_identity_type", json_get("userIdentity.type"));
            metadata_set("dynamodbstreams_user_identity_principal_id", json_get("userIdentity.principalID"));
          }
        }
```

asiragusa:
Ok, looks good. Performance-wise which is the impact of doing this?

Jeffail:
It should be negligable, the `awk`  processor is the highest impact, if your throughput is an issue we can always look at more specialised processors.

asiragusa:
I set the metadata in order to filter the messages by eventType, so probably I will be able to avoid the awk processor.
I was thinking about the double json encode/decode, or even to avoid the json encode at all if the message is filtered out later. This could benefit some Benthos pipelines too, but I understand that it can be confusing to work with interfaces in some pipelines and json (or `[]byte`) in others.

Issue #186, opened 2019-04-17
Quoted Text Decoder


danrubenstein:
Much how the [Benthos Decode Processor](https://docs.benthos.dev/processors/#decode) has a base64 decoding type, a "quoted" decoder would be helpful for those who are processing quoted data. I played around a bit with JMESPath, but I'm not sure that you can replicate this functionality with that tooling (although if I can I would just as soon close this issue).

The encoder would be powered by [strconv's `Unquote` function](https://golang.org/pkg/strconv/#Unquote).

Example:

```
`{\"hello\": \"world\"}`
// is tranformed to 
`{"hello": "world"}`
```

If you like, while I'm at it, I could add a 'quoted' encoder as well.

Jeffail:
Hey @danrubenstein, looks good and seems a worthy addition. I would want to add this to the `text` processor instead though, the implementation in your PR looks good, would you be happy to move it?

Jeffail:
I would probably call the operators `quote` and `unquote` so that they're similar to the existing `escape_url_query` and `unescape_url_query` operators.

danrubenstein:
Ah, that's a much more suitable place for it. I'll update accordingly, thanks!

Jeffail:
Thanks @danrubenstein!

PR #187, opened 2019-04-17
Add quoted type to decode processor

danrubenstein:
Hi @Jeffail! 

This is my sketch for what I had proposed in #186. I know that I should probably still add some docs and such, but I wanted to make sure I could show you where I was headed before I got too far in the weeds. 

Thanks for all your time and consideration here.


danrubenstein:
Closing in favor a new branch with cleaner history, per discussion in #186.

PR #188, opened 2019-04-17
Add quote and unquote operators to text processors

danrubenstein:
Some git struggling later, here's a single commit to the effect of the discussion in #186


Jeffail:
Awesome, looks great. Thanks @danrubenstein!

PR #190, opened 2019-04-18
Add env var injection for streams API input

### LINKED ISSUE #189: Env Vars For Streams API
When using Benthos in the standard mode it will load a configuration file and inject environment variables into placeholders matching the `${}` pattern. However, this only happens as a part of `main.go` and only if the configurations are provided at startup time ( <https://github.com/Jeffail/benthos/blob/master/cmd/benthos/main.go#L228> ).

When running the service in "streams" mode, the configurations are accepted by the HTTP server which uses a different set of methods to parse and sanitize the configuration. ( <https://github.com/Jeffail/benthos/blob/master/lib/stream/manager/api.go#L230> ) The net result is that the streams API does not inject environment variables.

It seems like it would be easy enough to add a call to `text.ReplaceEnvironmentVariables` in streams but wanted to get your opinion of where/how this should be done or if you meant for streams to have this feature as unsupported.

kconwayatlassian:
The standard Benthos runtime allows for the use of "${}" as a
placeholder for environment variables and injects the value of those
variables in to a configuration at startup time. This patch adds the
same behavior to the YAML/JSON documents given to the streams API
endpoints when creating or updating a stream.

Closes #189

### DIFF
--- a/lib/stream/manager/api.go
+++ b/lib/stream/manager/api.go
@@ -35,6 +35,7 @@ import (
 	"github.com/Jeffail/benthos/lib/output"
 	"github.com/Jeffail/benthos/lib/pipeline"
 	"github.com/Jeffail/benthos/lib/stream"
+	"github.com/Jeffail/benthos/lib/util/text"
 	"github.com/Jeffail/gabs"
 	"github.com/gorilla/mux"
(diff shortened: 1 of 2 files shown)


Jeffail:
Looks great, thanks @kconwayatlassian!

PR #201, opened 2019-05-03
Fix sync lambda output if multiple parts

kconwayatlassian:
The original inserted the first part N times. This adjust the loop to use the
part provided by the iterator instead.


Jeffail:
Woops! Thanks @kconwayatlassian!

Jeffail:
I'm away from pc but I'll kick off a tag/build in the next few hours.

Jeffail:
Released: https://github.com/Jeffail/benthos/releases/tag/v1.17.0
Sorry for the delay. Thanks again!

PR #206, opened 2019-05-06
WIP [docs] improvements to docs related to JSON references.

spier:
WIP / don't merge :)

In your latest version of the docs, `config/json_references/README.md` contains a great example of how to use the config fragments in practice. Very handy and something I was missing in the previous docs. This complements `docs/configuration.md` very well now.

Not sure if I am supposed to make changes directly in the `.md` files but I found it was easier to read this way. If you like any of these ideas we can migrate them into the right place later.

# Changes I made

* adding links to jq website
* explain environment variable interpolations the other way around (I found it more important to know what to use them for, then to explain why this works.)
* removed unused json-pointer reference

# Questions about the docs

* `These references can be nested`. What does this mean, and what would be a use case for this? 
* Is the documentation easier to read when `foo.yaml` references `bar.yaml`? Or the other way around? I am undecided myself :)
* "fragment" and "snippet" are used interchangeably in the docs. Snippet more often than fragment. Would it be easier to always use "fragment", as that is the name of the section as well ("Fragmenting Your Configuration")
* do we need the directory ./config/ in the example, or does it only confuse the reader? I suspect you got it in there so that you can talk about the relative file paths?

# Questions about benthos and/or fragmenting

* does the config fragment that I use in $ref have to be a fully functional config on its own? or can it just a fragment that only works once it is referenced from a larger config?
    * I am asking because `$ref: ./modules/group.yaml` would be a more straightforward ref example than `$ref: ./modules/group.yaml#/pipeline/processors/0`
* What happens if A references B and B references A? (circular)
* What does benthos do when I load a config that is not functional?
* You had some documentation about json-pointers before. Did you remove that completely?

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -138,9 +138,11 @@ pipeline:
       cache: objects
 ```
 
-These references can be nested. Also, environment variable interpolations within
-configurations are resolved _before_ references are resolved. Therefore, it's
-possible to use them to specify which snippet to load:
+These references can be nested. 
+
+It is further possible to use environment variables to specify which snippet 
+to load. This works because environment variable interpolations within 
+configurations are resolved _before_ references are resolved.
 
 ``` yaml
 pipeline:
@@ -242,7 +244,7 @@ benthos --print-yaml --all > conf.yaml
 And simply delete all lines for sections you aren't interested in, then you are
 left with the full set of fields you want.
 
-Alternatively, using tools such as `jq` you can extract specific type fields:
+Alternatively, using tools such as [jq][jq] you can extract specific type fields:
 
 ``` sh
 # Get a list of all input types:
@@ -321,7 +323,7 @@ config.
 
 If your configuration is complex, and the behaviour that you notice implies a
 certain section is at fault, then you can drill down into that section by using
-tools such as `jq`:
+tools such as [jq][jq]:
 
 ``` sh
 # Check the second processor config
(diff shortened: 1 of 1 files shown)


Jeffail:
Ace, changes look good.

- "These references can be nested." is meant to mean that if config `foo.yaml` references config `bar.yaml`, which in turn references `baz.yaml`, then Benthos will resolve all of these references as it traverses them
- I did reconsider the ordering of `foo.yaml` and `bar.yaml` but I don't have strong feelings either way
- Good point, I got cold feet on the term 'fragment' at some stage because I also needed to talk about the 'fragment' of the path URI, which is something else entirely. I think it probably makes sense to change all cases of 'fragment' to 'snippet', will need a section title change, maybe 'Reusing Configuration Snippets'
- Yeah that's in there just to make it clear that the path is relative, I don't think it's needed necessarily so if it adds confusion then I'm happy for it to be removed

- No need for it to be fully functional on its own, but it does need to be valid YAML. I wanted to show in the main example that it's possible to access fields within the snippet as I consider it best practice to have functioning snippets that can be tested in isolation. I'm not opposed to simplifying the example but in that case I would at least want to follow it with a section explaining the paths and benefits of doing so.
- Circular references get caught and reported as an error ('caught' in a very crude sense, I have a ceiling of reference layers that prevents it from happening)
- If you try and reference a config that is broken then Benthos will print an error message similar to if the root config file was broken in the same way
- I removed the `json-pointers` spec link as for most people the path 'is what it looks like' and the json references spec links to it itself in the relevant section. Not opposed to adding it back in but I'd be a little worried of sending people down a rabbit hole if they mistake it for being necessary reading.

PR #211, opened 2019-05-13
Add config type inference

Jeffail:


Jeffail:
Closed, decided that the possibility for regression means we should bump the major version, adding this to a larger PR with V2 changes.

PR #217, opened 2019-05-27
Use gokini to read Kinesis

### LINKED ISSUE #127: Dynamically discover Kinesis Shards + automatic horizontal scaling
[The Kinesis configuration requires you to specify a shard](https://github.com/Jeffail/benthos/blob/master/docs/inputs/README.md#kinesis) to receive records from. Ideally we should discover shards automatically.

I maintain a [Kinesis library](https://github.com/patrobinson/gokini) written purely in Golang (i.e. not using the Java MultiLangDaemon) and was wondering if you'd be interesting in integrating it into this project? It currently supports automatic shard discovery.

Currently the library doesn't support dynamically distributing shards across workers to scale horizontally automatically but I see that as the next feature to be added.

If you think this is worthwhile I will put up a Pull Request.

patrobinson:
Closes #127


Jeffail:
Hey @patrobinson, this looks good, I'll need a bit of time to look over it in more detail.

There are some issues with backwards compatibility with things like the AWS session fields being removed as well as the explicit shard field no longer being used. Since this library is relatively new and in a beta stage it seems sensible to either offer it in an extended build (similar to the ZMQ distribution) or just place it under a different input name (like the `kafka` versus `kafka_balanced` inputs.) Either way I'm happy to try this out in the main project in some form as it's a great usability improvement over the existing offering.

I'll think it over and come back.

patrobinson:
Hey @Jeffail thanks, sorry I did raise this intended to ask you how we will handle breaking changes. Happy to make it a seperate input if that helps

lib/processor/subprocess_test.go
```diff
@@ -136,7 +136,7 @@ func TestSubprocessLineBreaks(t *testing.T) {
 	})
 	msgs, res := proc.ProcessMessage(msgIn)
 	if len(msgs) != 1 {
-		t.Fatal("Wrong count of messages")
+		t.Fatalf("Wrong count of messages %d", len(msgs))
```
patrobinson: This test was failing locally for some reason. Not sure why

Jeffail: Yeah these are bad unit tests as they rely on installed binaries. If you rebase on master I've got them disabled.

Jeffail:
Okay, so I have a plan of action:

- Instead of modifying the current `kinesis` input let's add a new `kinesis_balanced` input, I'll leave that to you.
- In the documentation for this input we will call this a Beta component and clarify what's different about it (specifically the automatic distribution of shards.) I'm happy to write this out.
- I'll add a note to the input and also make a ticket specifically calling for testers and feedback.

The only major missing feature I see right now is that we don't have the ability to explicitly set AWS credentials in config fields. That's not a deal breaker but it's useful for creating pipelines that access multiple AWS accounts, and would be good to add eventually just to keep it consistent with other components. We can add that at any time.

The actual implementation of this input looks fine to me but I'll do another review once the PR is updated.

patrobinson:
@Jeffail I've updated to use a different name for this input. I'll look at adding the ability to pass in a custom session upstream.

Jeffail:
Awesome, thanks @patrobinson! I should be able to take care of the rest this evening.

PR #219, opened 2019-06-03
Added simple Amazon SNS output source

donaherc:
Hello! I wanted to add an Amazon SNS output source to benthos -- it's a very simple implementation at the moment, but am open to expanding the coverage to a greater portion of what SNS supports (SMS, Email, etc.). I also had some questions about how you'd like to test/verify AWS specific input/output components. We could use something like a https://github.com/localstack/localstack docker container to test integration, although I'm unsure of the brittleness of that project.

### DIFF
--- a/lib/output/constructor.go
+++ b/lib/output/constructor.go
@@ -87,6 +87,7 @@ const (
 	TypeRedisStreams  = "redis_streams"
 	TypeRetry         = "retry"
 	TypeS3            = "s3"
+	TypeSNS           = "sns"
 	TypeSQS           = "sqs"
 	TypeSTDOUT        = "stdout"
 	TypeSwitch        = "switch"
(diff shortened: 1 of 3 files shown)


Jeffail:
Hey @donaherc, looks good. In this case, since we can only send one message at a time there's no need to include the `retries.Config` config fields, might as well remove those. I'm happy to use localstack for the integration tests, it's already being used for S3 (https://github.com/Jeffail/benthos/blob/master/lib/test/integration/s3_test.go), it's slow AF to start but we can always move to another option later.

donaherc:
Interesting -- I'm willing to incorporate localstack here. I'll remove the retries.Config fields as well.

Jeffail:
Hey @donaherc, still working on this? I'm happy to add integration tests and get this merged if you haven't got time.

donaherc:
Sorry yeah, work has really spiked up. I'll try to get around to it, but you're welcome to add tests.

Jeffail:
Done, thanks @donaherc, in the end I haven't hooked up integration unit testing as it was giving me a headache trying to wire up webhooks from localstack, but I ran some manual tests.

PR #235, opened 2019-07-18
Use `PSUBSCRIBE` instead of `SUBSCRIBE`

dicksontung:
https://github.com/Jeffail/benthos/issues/234


Jeffail:
Hey @dicksontung,

This is a nice addition but in order to be backwards compatible it'll have to be a config option so that `Subscribe` is the default behaviour. If you add a bool field `use_patterns` which is default `false`, then:

```go
if r.conf.UsePatterns {
	r.pubsub = r.client.PSubscribe(r.conf.Channels...)
} else {
	r.pubsub = r.client.Subscribe(r.conf.Channels...)
}
```

That should be enough, then it won't change anything for existing users.

dicksontung:
Sure, added `use_patterns` config

Jeffail:
Ace, thanks @dicksontung!

PR #237, opened 2019-07-25
Add brew installation option for macOS users

jcalonso:


Jeffail:
Ace, thanks @jcalonso!

PR #253, opened 2019-08-14
Add support for grok pattern definitions

### LINKED ISSUE #252: Support grok pattern definitions
At the moment benthos doesn't support custom pattern definitions, it has only [default ones](https://github.com/trivago/grok/blob/master/patterns.go#L5).

It would be useful if we can specify our specific definitions to use in grok patterns.

ilyaglow:
Turns out, it was very easy to add this feature.
Closes #252

### DIFF
--- a/config/processors/grok.yaml
+++ b/config/processors/grok.yaml
@@ -20,6 +20,7 @@ pipeline:
       named_captures_only: true
       output_format: json
       parts: []
+      pattern_definitions: {}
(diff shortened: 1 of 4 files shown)


Jeffail:
Awesome, thanks @ilyaglow, probably going to have to wait until tomorrow evening as I'm travelling but I'll get it merged and released as soon as I can.

PR #255, opened 2019-08-15
Fix array overwrites in prepend/append operators in text processor

adamwasila:
This fixes: https://github.com/Jeffail/benthos/issues/254

It also fixes similar issue with append (no ticket for that one, I guess?).

Both have similar root cause - having slice with capacity larger than lenght then calling "append" on such slice and both are operators from same processor - that's why both are included in single PR.


Jeffail:
Nice find, thanks @adamwasila!

PR #265, opened 2019-08-26
[metrics/prometheus]: Add default summary objectives

asiragusa:
Default objectives were removed in https://github.com/prometheus/client_golang/pull/600. This PR restores default quantiles for `metrics.Timer`


Jeffail:
Thanks @asiragusa, this one completely slipped under my radar.

asiragusa:
Sh*t happens :)

Btw have you seen https://github.com/go-kit/kit/tree/master/metrics? You might leverage this package to easily implement all the supported backends. https://github.com/go-kit/kit/tree/master/tracing could be useful too.

A problem I have with the current implementation is that the metrics namespacing is done via the metric name and not by labels (or tags), this makes harder to aggregate metrics from different processors, as the metric name differ (a processor used twice generates two different metrics). If the namespacing is done via labels it's easier to do math on such metrics

PR #267, opened 2019-08-28
New output metrics: latency & bytes

adamwasila:
As previously discussed on slack channel this PR adds two new generic metrics to (almost?) all outputs:

* `output.batch.sent.bytes` - total number of bytes sent to output. Of course different protocols will have different wrappings around payload, so this doesn't equal to number of bytes sent via network (maybe documentation should mention that, what you think?) but it is a quite valuable measure anyway
* `output.batch.sent.latency` - summary type (sum + count) that measures total time spent by output to write message. Again - it is not raw time spent sending message and waiting for reply as it may include overhead of preparation and/or retries upon error. Probably this could be explained somewhere in docs, eg. should each output document if it does anything unusual in Write() (dynamodb is a perfect example)?

### DIFF
--- a/docs/metrics/paths.md
+++ b/docs/metrics/paths.md
@@ -55,6 +55,8 @@ prefixed with `pipeline.processor.N`, where N is the index.
 - `output.count`: The number of times the output has attempted to send messages.
 - `output.sent`: The number of messages sent.
 - `output.batch.sent`: The number of message batches sent.
+- `output.batch.bytes`: The total number of bytes sent.
+- `output.batch.latency`: Latency of message batch write in nanoseconds. Includes only sucessful attempts.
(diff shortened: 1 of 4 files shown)


Jeffail:
Hey @adamwasila, looks great but I think it might be worth renaming to metric paths to avoid having `output.batch.sent` as a subpath. Would you be happy with `output.batch.bytes` and `output.batch.latency`?

adamwasila:
Ok, no problem. Less is better, I guess. I will fix it in place and rebase.

adamwasila:
Ok done. Looks like I forgot to measure latency upon reconnect. I've fixed it too.

Jeffail:
Very cool, thanks @adamwasila!

PR #275, opened 2019-09-10
Fix nats-io import path

asiragusa:
Hi @Jeffail,

Recently nats changed their repositories to https://github.com/nats-io/nats.go and https://github.com/nats-io/stan.go and this breaks go mod. I updated the dependencies, but still can't run `go mod tidy` because of https://github.com/nats-io/stan.go/issues/280

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,17 +1,14 @@
 module github.com/Jeffail/benthos/v3
 
 require (
-	cloud.google.com/go v0.45.1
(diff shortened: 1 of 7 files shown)


asiragusa:
Ok also tests are failing, we must wait for https://github.com/nats-io/stan.go/issues/280

Jeffail:
Cool, thanks @asiragusa!

PR #276, opened 2019-09-11
Redis processor

DpoBoceka:
Hey, look what I've done.

It is a draft and I'm still confused of some things but it is something after all.

- I'm not sure how I should calculate specific metrics for different operators - should I call a function with a `r *Redis` as a parameter so inside the function I could do `r.mRedisSAddCount.Incr(1)` ?
- It would be fair to have an arguments `path` and `metadata_path` so user could deliberately write result in it. As for now the result's in metadata field `redis_returned_value`
- of course there's tests to be done 

Guide me and I shall strike down these defects with great vengeance and furious anger.

Or just make some changes yourself, if you please..


lib/processor/redis.go
```diff
@@ -0,0 +1,290 @@
...
+
+// Redis is a processor that performs redis operations
+type Redis struct {
+	parts []int
+	conf  Config
+	log   log.Modular
+	stats metrics.Type
+
+	key               *text.InterpolatedString
+	valueBytes        []byte
```
Jeffail: You can do the same for `valueBytes` as you do for `key` and use https://godoc.org/github.com/Jeffail/benthos/lib/util/text#NewInterpolatedBytes

lib/processor/redis.go
```diff
@@ -0,0 +1,290 @@
...
+		key := r.prefix + r.key.Get(message.Lock(newMsg, index))
+
+		res, err := r.operator(key, valueBytes, r.client)
+		if err != nil {
+			r.log.Debugf("Operator failed for key '%s': %v\n", key, err)
+			return err
+		}
+		//
+
+		part.Metadata().Set("redis_returned_value", res)
```
Jeffail: It's more consistent with other processors if you set the contents of the message with the result. Most of the time a processor like this is going to be used within https://docs.benthos.dev/processors/#process_field or https://docs.benthos.dev/processors/#process_map which already gives you the ability to store the result within metadata or a particular JSON field.

lib/processor/redis.go
```diff
@@ -0,0 +1,290 @@
...
+//------------------------------------------------------------------------------
+
+// RedisConfig contains configuration fields for the Redis processor.
+type RedisConfig struct {
+	URL         string `json:"url" yaml:"url"`
+	Parts       []int  `json:"parts" yaml:"parts"`
+	Operator    string `json:"operator" yaml:"operator"`
+	Key         string `json:"key" yaml:"key"`
+	Value       string `json:"value" yaml:"value"`
+	Prefix      string `json:"prefix" yaml:"prefix"`
```
Jeffail: Looks like most of these config fields aren't used and can be removed.

lib/processor/redis.go
```diff
@@ -0,0 +1,290 @@
...
+// ProcessMessage applies the processor to a message, either creating >0
+// resulting messages or a response to be sent back to the message source.
+func (r *Redis) ProcessMessage(msg types.Message) ([]types.Message, types.Response) {
+	r.mCount.Incr(1)
+	newMsg := msg.Copy()
+
+	valueBytes := r.valueBytes
+	if r.interpolatedValue {
+		valueBytes = text.ReplaceFunctionVariables(msg, valueBytes)
+	}
```
Jeffail: If the `value` field is left empty then this processor ought to default to using the full contents of the message.

DpoBoceka: I've done it in a `func NewRedisConfig() RedisConfig` but I suspect that isn't what you wanted.

DpoBoceka:
Have a look, please. 
And also would you be kind enough to comment my issue here:
https://lab.benthos.dev/l/DAHoJsY6Rh4

Jeffail:
So for example, in the case where you want to get the length of an array with the key based on the field `foo` and embed that length in the original document, I would do something like this: https://lab.benthos.dev/l/c8HFGYmgcVZ

Jeffail:
This is looking good enough for me to merge with minor changes so let me know when you're ready.

DpoBoceka:
> So for example, in the case where you want to get the length of an array with the key based on the field `foo` and embed that length in the original document, I would do something like this: https://lab.benthos.dev/l/c8HFGYmgcVZ

That I knew, but if my key has values of two or more fields, then I would have to create new key with a json in it which contains all those keys and after that I would use `process_field + redis` processor with functions `json_get` to properly make a query to redis an store the result. And I would be forced to do such trick as many times as much calculation I have during pipeline and to clean unnecessary anymore fields afterwards. Isn't it too `pricey`? Perhaps there is another way to it?

DpoBoceka:
> This is looking good enough for me to merge with minor changes so let me know when you're ready.

Right, will do that tomorrow.

DpoBoceka:
I added some docs and changed the approach to set the default value for a `value` field.I think it can be merged. Only if the absence of tests isn't a stop factor.

Jeffail:
Thanks @DpoBoceka, I made some minor changes. I've removed the `value` field in favour of just always using the message contents, this isn't necessarily permanent but I'd like to see how it plays out before adding more fields. I also removed the prefix field as you can just place your prefix in the key interpolated field.

PR #278, opened 2019-09-15
adding syslog processor

DpoBoceka:
Hey, @Jeffail , I have a little something for benthos.

I used `github.com/influxdata/go-syslog ` to parse a message and I decided to implement just `rfc5424 ` support. Shouldn't be a real problem to include another rfc under processor's wing but it is quite good for a starter. I am diabolically sad that `rfc3164` (one of the popular) can't be parsed because the library hasn't implemented it yet, but then again I can't hold a grudge for longer than a few minutes, so..

I think, there is no need to provide functionality via config letting users not to create some of final JSON structure (not all of us need those field and those might be better off in the metadata), but every use case is unique and it's not fair to throw everyone in the same barrel, especially if it's limited space inside. So, overall, it should be easy to unset some of those JSON fields afterwards.

I also used `bestEffort` option which implies processor won't crash during the parsing if there would be some problem in an original message as long as it is bound, in a certain degree, to rfc.


Jeffail:
Hey @DpoBoceka, I'm not opposed to adding this in but have you tried using `grok` for this same purpose? Would be good to get a performance comparison, unless there's a significant gain from using this I'd rather strengthen the docs around grok than add another processor.

DpoBoceka:
I haven't tried grok, I dislike it for some reason. Isn't it going to be a bit painful to describe in grok every case out there (some fields can be omitted, timestamp can be formed differently, etc)? I'm really not sure if I can manage to write all these groks. Well, I guess that PR is going to hold on for some time

DpoBoceka:
However, how would you in general describe a proper benchmark in that case?

Jeffail:
Wow yeah, I spent a bit of time trying to get this working with grok and regret it. I'm going to think this one over for a few days as it'd be nice to make this processor a little more generic.

DpoBoceka:
Thank god, I've spent an hour and a half to gather some evidence to prove that it's better not to use grok in that case. And it wasn't delightful. 

As for making it more generic, I, then, would love to remake it in such a way as long as I find some time and my limited skills won't go in a way.

DpoBoceka:
Let's have a check list with things that need to be done in order to merge it. After all, syslog is a cornerstone transport for some use cases and we should cover that, I think

Jeffail:
Let's add this as a generic parser, steps should be:

- Come up with a good name (`convert_log`, `parse_log`, `process_log`)
- Create general processor with name
  - Has an output `codec` field (default `json` for now)
  - Has a `format` field (`syslog`)

I think it's reasonable for now to have a static output format based on the chosen input `format`, if the fields aren't what the user wants they can choose to mutate it however they like. We can always revisit later on with new fields.

DpoBoceka:
so current `xml` processor should be covered with that new processor, shouldn't it?

Jeffail:
Potentially, but let's leave `xml` where it is for now. There's always the possibility of expanding it in the future to support some of the same operations as `json`.

DpoBoceka:
Right, back to it. Still want all these steps to follow or we are changing strategy?

Jeffail:
Yeah I think a `parse_log` processor fits best.

DpoBoceka:
```
  - type: parse_log
     format: syslog
     rfc: rfc5424
     codec: json
```
like so?

Jeffail:
It'll be easier to document if we just have a `format` field with `syslog_rfc5424`, that way we're documenting a linear list of options.

DpoBoceka:
Right, what types of codec will be in use?

Jeffail:
For now just `json`, similar to the `grok` processor. The field itself isn't actually used as JSON is the only option, but I like to add it just to make it clear that there's a possibility of other options if people ask for them (no one has yet).

DpoBoceka:
Just noticed that I need to remove some tails of old "syslog processor"  from config/env/* files, I'll make another commit for it

Jeffail:
Ace, thanks @DpoBoceka!

Issue #279, opened 2019-09-16
local Metrics drops labels


bigkevmcd:
Hi,

I've been using metrics.Local for testing metrics from custom processors, we have a use-case for labels for a couple of fields...but...

https://github.com/Jeffail/benthos/blob/master/lib/metrics/local.go#L154

This drops the labels, and the raw stats that come through don't quite tell us enough.

Would you be open to a change to support labels on the `LocalStat` values?

Jeffail:
Hey @bigkevmcd, sounds good. How would you imagine the resulting structure should look?

Jeffail:
Released: https://github.com/Jeffail/benthos/releases/tag/v3.1.0

PR #281, opened 2019-09-18
Don't discard labels for local stats

bigkevmcd:
This addresses #279 by adding recording the labels / values on the `LocalStat` type in `lib/metrics`.

As I alluded to in the godoc, I used this for testing a custom processor that could count metrics in different ways depending on the messages it was processing.

### DIFF
--- a/lib/metrics/blacklist.go
+++ b/lib/metrics/blacklist.go
@@ -217,10 +217,10 @@ func (h *Blacklist) GetCounter(path string) StatCounter {
 }
 
 // GetCounterVec returns a stat counter object for a path with the labels
-// discarded.
+// and values.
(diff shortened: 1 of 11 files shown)


Jeffail:
Hey @bigkevmcd, looking good. There's technically a breaking API change here in `GetCounters()` and `GetTimings()`, I doubt anyone is using those methods externally but it would be good to find a backwards compatible work around.

bigkevmcd:
@Jeffail Hi, thanks for the review, yeah, that wasn't a good idea, I've reverted and extended the API.

Jeffail:
Ace, thanks @bigkevmcd! I'll take a look later at exposing this via the `http` metrics output.

PR #282, opened 2019-09-20
Add kinesis_firehose output

aidan-:
In typical fashion, I read the `CONTRIBUTING.md` as I created the PR, so hopefully there is appetite for accepting this additional output.

This PR add support for using [AWS Kinesis Firehose](https://aws.amazon.com/kinesis/data-firehose/) as an output destination and follows the existing `kinesis` output implementation.  The API is very similar, but different enough that it requires a separate implementation.  `kinesis_firehose` was chosen over `firehose` to differentiate between the bit of 'Twitter Firehose' that exist in the docs.

My use case for adding this output is using benthos to create a serverless router that can route subsets of a Kinesis Data Stream to different destinations, a feature you cannot do with native Kinesis Firehose.  Using Firehose for writing to the destinations allows us to decouple the different buffering requirements each target may have.


lib/output/writer/kinesis_firehose.go
```diff
@@ -0,0 +1,260 @@
...
+
+	sess, err := a.conf.GetSession()
+	if err != nil {
+		return err
+	}
+
+	a.session = sess
+	a.firehose = firehose.New(sess)
+
+	if _, err := a.firehose.DescribeDeliveryStream(&firehose.DescribeDeliveryStreamInput{
```
aidan-: It might be worth adding a more thorough check to ensure that the particular Kinesis Firehose is not just deployed, but also is deployed in a way that we can actually send messages directly to it (instead of being attached to service directly like Kinesis etc).  This would just check the response for `DeliveryStreamDescription.DeliveryStreamType == "DirectPut"`.

I will hold off on this in case there is additional feedback.

Jeffail:
Hey @aidan-, this looks great. I think checking that the stream exists is fine for now since we have transactions to fall back on.

PR #286, opened 2019-09-25
Add regex test matcher

woosteln:
Adds a regular expression matcher to pipeline testing to allow matching of dynamic ( ie time ) values and partial message excerpts.


Jeffail:
Looks great, thanks @woosteln!

PR #289, opened 2019-09-26
Add user and password to mqtt input output

woosteln:
Adds user and password fields to config and uses them if present in setting up mqtt client, both for input and output.


Jeffail:
Looks great, thanks @woosteln.

PR #290, opened 2019-09-27
Add stdout as metrics output type

aidan-:
**Overview**
This adds a new metric type called 'stdout' which allows metrics to be published to stdout for ingestion into further downstream systems (Elasticsearch was the destination intended during implementation).

**Use case**
We are currently running Benthos in a serverless (AWS Lambda) environment and without having a `statsd` or `prometheus` endpoint to push metrics to, there's currently no way to get metrics out.  By outputting the metrics to stdout (as well as general Benthos logs) we then have another Lambda function (also Benthos ;)) triggered from Cloudwatch Logs that allows us to ingest these events into Elasticsearch. 

**Example**
Configuration:
```
  "metrics": {
    "type": "stdout",
    "stdout": {
      "static_fields": {
        "foo": "${COLORTERM:defaulty}",
        "bar": {
          "bana": "baz"
        }
      }
    }
  }
```
Some example metrics output:
```
{"@service":"benthos","@timestamp":"2019-09-27T14:38:03+10:00","bar":{"bana":"baz"},"component":"input","foo":"truecolor","input":{"batch":{"received":1},"connection":{"failed":0,"lost":1,"up":1},"count":1,"latency":548965374,"received":1,"running":0},"metric":"input"}
{"@service":"benthos","@timestamp":"2019-09-27T14:38:03+10:00","bar":{"bana":"baz"},"component":"output","foo":"truecolor","metric":"output","output":{"batch":{"bytes":70962,"latency":543866729,"sent":2},"count":2,"error":0,"messages":{"received":1,"sent":2},"sent":2}}
{"@service":"benthos","@timestamp":"2019-09-27T14:38:03+10:00","bar":{"bana":"baz"},"batch":{"bytes":35481,"latency":543866729,"sent":1},"component":"output","count":1,"foo":"truecolor","metric":"output.broker.outputs.0","sent":1}
{"@service":"benthos","@timestamp":"2019-09-27T14:38:03+10:00","bar":{"bana":"baz"},"component":"system","foo":"truecolor","metric":"system","system":{"goroutines":5,"uptime":559}}
```

**Further**
- Metrics are currently 'grouped' by the component that the individual metrics belong to.  This is done by traversing the metric path backwards and assuming an 'integer' in the path is an instance of a component.
- During the implementation I originally only had the metrics output when the `Close()` method was called, but it turns out this is never the case in a serverless environment.  Because of this, I changed to also include an optional `push_interval`.  This seems to work nicely.
- Ideally, I'd also like the counters to be reset to 0 after they have been 'pushed' to stdout.  When looking at the metric interfaces, there doesn't seem to be a reset method (though I guess you could decrement the metric by its value..?).  Would adding a reset function to each of the counter types be palatable? Or is there a better way to achieve this?
- I noticed that some processors have 'labels' - it would be cool if all outputs/inputs/etc supported this as well.  We are 'dynamically' building Benthos configurations in a `broker` output using `fan_out`, so changing indexes is a bit difficult to track.
- What about Cloudwatch Metrics you ask? Cloudwatch Metrics really suck.  I tried having a look to see if I could easily fit the number of metrics Benthos outputs into a model using Cloudwatch Metrics but the cost and usability doesn't make it seem like a good place to put them.

Happy to open issues for separate discussion on some of those points, but wanted to share some of the reasoning for why this is implemented this way.


Jeffail:
Hey @aidan-, I like this idea. It would be cool to use the system wide logger for writing these out so we can recycle the formatting fields it has, but that would require changing the way we're printing the metrics. Adding a `Flush` method to the `Local` metrics aggregator should be straight forward and would get you what you want.

Adding metric labels to components within Benthos is possible but slightly delicate as for some aggregators there is a very strict requirement that every series of a path have exactly matching label keys, so we'd need to standardise on it.

aidan-:
> It would be cool to use the system wide logger for writing these out so we can recycle the formatting fields it has, but that would require changing the way we're printing the metrics. 

How were you thinking this would be done?  I was originally using the `log.Modular` package to do the printing of the metrics, but the current structure doesn't cleanly allow the 'message' to be JSON.  We would possibly want to add another `level` called `METRIC` to avoid having to use the `INFO`/`WARN`/etc

> Adding a `Flush` method to the `Local` metrics aggregator should be straight forward and would get you what you want.

I've just done this and rebased!  Testing now but seems it works nicely.

> Adding metric labels to components within Benthos is possible but slightly delicate as for some aggregators there is a very strict requirement that every series of a path have exactly matching label keys, so we'd need to standardise on it.

No worries, that can be left as a future enhancement then :)

aidan-:
@Jeffail The flush methods look to be working well!

Jeffail:
Cool, yeah if we went with the logger then it would limit you to only writing out metrics using fields, which is the wrong way of going about it. I think ultimately this should be interacting somehow with the logger but let's get this added in as an experimental component and we can work out the kinks later. 

Can you add a line `EXPERIMENTAL: This component is considered experimental and is therefore subject to change outside of major version releases.` to the top of the documentation?

aidan-:
@Jeffail , disclaimer added, docs generated and rebased from master to make the merge clean!

Jeffail:
Ace, thanks @aidan-!

PR #300, opened 2019-10-12
add jsonschema condition

aidan-:
This PR adds a new condition for validating JSON events using a configured [JSONSchema](https://json-schema.org/) document.  This is useful for ensuring events flowing through a pipeline are consistent.

The underling library that does the schema loading and event validating is [xeipuuv/gojsonschema](https://github.com/xeipuuv/gojsonschema).  It is not the fastest of the jsonschema libraries but it does seem to be the most feature complete based on these [benchmarks](https://github.com/WildBlueGuy/validator-benchmarks).  I'm still on the fence about whether a faster library would be more beneficial for this use case.

Also, the `schema_path` parameter allows for loading a schema from a 'remote' source (file on disk or http endpoint).  The actual retrieving of this configuration is handled by the underlying library.  Are there internal Benthos utilities/functions that should do this instead?  Using those might make it easier to switch to a different underlying validator if necessary. Thoughts?


Jeffail:
Looks good, I think `schema_path` is fine for now. If we end up replacing it with a new lib it wouldn't be a huge task to implement our own resolver for it. Thanks @aidan-!

Jeffail:
Hope you don't mind but I renamed the component to `json_schema` to make it more consistent with other names.

aidan-:
That's no issue at all, thanks!

PR #303, opened 2019-10-20
Add KMSKeyID option to AWS S3 writer

mwain:
When KMSKeyID is set objects will be uploaded with server side
encryption using AWS KMS


Jeffail:
Looks great, thanks @mwain!

PR #317, opened 2019-11-21
Support tags in statsd metric target

zacatac:
In this PR I've migrated from `quipo/statsd` to the `dtatadog-go/statsd` library. The datadog client supports a particular version of tag syntax in their statsd implementation. There is some discussion here (https://github.com/statsd/statsd/issues/619) about the divergence of tag syntax among statsd clients. 

This PR also adds support for string interpolation for the payload field in the http_client input type.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/Jeffail/benthos/v3
 require (
 	cloud.google.com/go/pubsub v1.0.1
 	github.com/Azure/go-ansiterm v0.0.0-20170929234023-d6e3b3328b78 // indirect
+	github.com/DataDog/datadog-go v3.2.0+incompatible
 	github.com/DataDog/zstd v1.4.1 // indirect
 	github.com/Jeffail/gabs/v2 v2.1.0
 	github.com/Microsoft/go-winio v0.4.14 // indirect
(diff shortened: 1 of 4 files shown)


Jeffail:
Hey @zacatac, I'd like to support Datadog tagging but it looks as though this PR would break for users relying on TCP traffic. As an alternative I'd be happy to add this as a specific `datadog` metric target instead of replacing the existing `statsd` version.

zacatac:
Sounds good, will update the PR this weekend!

assada:
Also We have problems with quipo/statsd package
This package does too much, for example, it considers aggregation for Timing metric is a problem if you use telegraph (https://github.com/influxdata/telegraf)

We replaced completely the github.com/Jeffail/benthos/v3/lib/metrics package with our implementation that uses https://github.com/smira/go-statsd and does not implement aggregations

Usually there's some kind of aggregation between statsd client and time-series database which might affect the queries you're using.

assada:
Maybe should implement metrics as a plugin? Like processors

assada:
@Jeffail Maybe should implement metrics as a plugin? Like processors

Jeffail:
Interesting, I wasn't aware of the problems we had with the existing lib, I've opened https://github.com/Jeffail/benthos/issues/324 to tackle the statsd aggregation part separately.

Jeffail:
@zacatac given the issues with the existing statsd client outlined by @assada I think it might be worth replacing it. I've added details to https://github.com/Jeffail/benthos/issues/324, but in short I like the look of https://github.com/smira/go-statsd and it also supports datadog tagging, so maybe we can get away without needing a separate `datadog` metric target.

Jeffail:
@zacatac, got a new PR that handles backwards compatibility https://github.com/Jeffail/benthos/pull/334, would you be willing to try it out? I'd like to make sure it works for your usecase before moving forwards.

Jeffail:
Hey @zacatac, I've merged a backwards compatible version. The way to use the new lib is outlined in the docs: https://github.com/Jeffail/benthos/tree/master/docs/metrics#statsd, basically you set `tag_format` to `none`, `datadog` or `influxdb` and it'll use the new library.

PR #319, opened 2019-11-24
JSONSchema processor

stephanebruckert:
This is to solve https://github.com/Jeffail/benthos/issues/313

Most of the processor comes from the JSONSchema condition and was adapted:
- https://github.com/Jeffail/benthos/blob/master/lib/condition/jsonschema.go
- https://github.com/Jeffail/benthos/blob/master/lib/condition/jsonschema_test.go

However I am not sure how to test that the ${!error} will contain the message I need? Other processor tests don't seem to ever test the error field.

Thx for your feedback!


lib/processor/constructor.go
```diff
@@ -163,6 +164,7 @@ type Config struct {
 	InsertPart   InsertPartConfig   `json:"insert_part" yaml:"insert_part"`
 	JMESPath     JMESPathConfig     `json:"jmespath" yaml:"jmespath"`
 	JSON         JSONConfig         `json:"json" yaml:"json"`
+	JSONSchema   JSONSchemaConfig   `json:"jsonschema" yaml:"jsonschema"`
```
Jeffail: The tags here need to match the type name `json_schema`

Jeffail:
Hey @stephanebruckert, error messages are stored as a metadata field, so to check the contents of the failure flag you can use `part.Metadata().Get(FailFlagKey)`, where `FailFlagKey` is the static string `benthos_processing_failed`.

stephanebruckert:
Hey @Jeffail, thanks, that helped! PR is now ready for full review

Jeffail:
Awesome, thanks @stephanebruckert!

stephanebruckert:
@Jeffail thanks a lot for merging this. Going to use it from today! 

However even though a tag hasn't been created yet, should it not have automatically updated the `latest` docker image? https://hub.docker.com/r/jeffail/benthos/tags

Jeffail:
@stephanebruckert, the latest tag only gets updated with a new tag or release candidate. I'm happy to put a build out today though, it'll be v3.5.0.

PR #320, opened 2019-11-26
implement redis_hybrid cache

iluminae:
@Jeffail here is a pattern for a using a in-memory cache ([risretto](https://github.com/dgraph-io/ristretto), its great) as a read-through onto redis, coupled with a invalidation channel so it can be used as a distributed in-memory cache using redis as a broker.

I would have loved to do this without requiring a redis specifically, but the invalidation channel is important for the distributed nature of this cache type, and the other caches do not implement a similar mechanism.

### DIFF
--- a/docs/caches/README.md
+++ b/docs/caches/README.md
@@ -49,7 +49,8 @@ cache is the same for both inputs.
 3. [`memcached`](#memcached)
 4. [`memory`](#memory)
 5. [`redis`](#redis)
-6. [`s3`](#s3)
+6. [`redis_hybrid`](#redis_hybrid)
+7. [`s3`](#s3)
 
 ## `dynamodb`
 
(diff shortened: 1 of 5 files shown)


Jeffail:
Thanks @iluminae this is really interesting.

I had a plan for eventually implementing cache layering within Benthos so that you could potentially have any number of read-throughs, but this takes that further with the distributed invalidation part. It would be awesome to find a clean way of pulling this out into generic components.

I'll take some time to think this through.

iluminae:
Sounds good to me - I tried to make this very much like your read through design, but I really wanted that invalidation channel.

kkessler:
another couple thoughts on this:
* we could implement risretto as another cache type or replace the `memory` implementation with risretto - then use the `memory` and `redis` one here.
* we could make several interfaces: [cache,localCache,remoteCache,invalidatingCache] and have this readthrough implementation take an array of caches as config params, then all the other cache types implement either {local,remote,invalidating}Cache. If the type is invalidatingCache, then we run the invalidatingCache methods of `Listen()` and `Invalidate(key string)`

Jeffail:
That could work but I prefer keeping the interfaces common and simpler for plugin authors. An alternative would be to implement a read-through cache type for the simple leveled caches implementation:

```yaml
resources:
  caches:
    levels:
      read_through: [foo, bar]
    foo:
      memory:
        ttl: 60
    bar:
      redis:
        ...
```

And then complement with a `read_through_invalidated` type that could use real Benthos inputs and outputs for broadcasting and listening for invalidations:

```yaml
resources:
  caches:
    levels:
      read_through_invalidated:
        caches: [foo, bar]
        invalidator_input:
          redis_pubsub:
            ...
        invalidator_output:
          redis_pubsub:
            ...
    foo:
      memory:
        ttl: 60
    bar:
      redis:
        ...
```

Will need some more thought I think.

iluminae:
that sounds awesome - I am all about that. Im not sure the infrastructure within the cache type you would need to run inputs and outputs in a vacuum like that, but I love the design.

feel free to close this if you have that implementation in the works.

Issue #322, opened 2019-11-27
Kafka headers generated when targeting version that doesn't support them


ollystephens:
If you set `target_version` on a kafka output below 0.11.0.0, the underlying Sarama library will refuse to send messages because they contain headers. The error you get is:
> Producing headers requires Kafka at least v0.11

https://github.com/Shopify/sarama/blob/49a563683f906b0ddbf5b607e8e056d3ed94bde0/async_producer.go#L335

Unfortunately, even stripping out all the metadata doesn't help, as benthos sends an empty array rather than a nil, which is what Sarama checks for.

The following code change makes Benthos silently omit the headers if the version of kafka doesn't support them:

https://github.com/ollystephens/benthos/commit/111c99a6ce4ac5c5e88428ff31d5a615eafa1941

If you're happy with this approach, I can raise a PR.

An alternative approach would be to force the user to explicitly strip out the metadata, and tweak the code to send nil rather than an empty array in this case; this would remove the need for the version check. But I'm happier with the ignore approach.

Olly

Jeffail:
Hey @ollystephens, nice find. A PR would be great!

Jeffail:
This'll be released in v3.6.0.

PR #323, opened 2019-11-28
only send kafka headers if the version supports them

ollystephens:
As per issue #322, this PR prevents headers being added to a message if the target version does not support them. Unfortunately, there's no bool "version.HeadersSupported()" method so I've copied the version number "version.IsAtLeast(...)" test from the Sarama codebase instead.


Jeffail:
Looks good, thanks @ollystephens!

PR #325, opened 2019-11-29
add MQTT output topic interpolation

### LINKED ISSUE #298: Add topic metadata to mqtt output
Would you consider to add metadata topic to mqtt output (with default from config) to be able to set topic based on message content in pipeline please ?

liangxiaobo2018:
close #298


Jeffail:
Looks great, thanks @liangxiaobo2018!

PR #328, opened 2019-12-03
Datadog metrics type proposal

assada:
This is just proposal for temporary resolving https://github.com/Jeffail/benthos/issues/324

But the name of the datadog is not the best choice. Maybe someone will have better ideas


Jeffail:
Can you summarize the difference in metric output there'll be when we switch from the current lib to this one? If not I'll spend some time investigating, if it's minimal then I'd be happy updating the existing `statsd` type and simply adding a special case for when the network is set to `tcp`.

assada:
General isue is that the current library make aggregations for timings. https://github.com/quipo/statsd/blob/master/event/timing.go#L44 The new library does not. And it sends just the value of the timer. https://github.com/smira/go-statsd/blob/master/client.go#L226

In general case you have some server tool on the between storage and statsd for calculate metric aggregations. In this case we got results metric like this: `prefix_output_batch_latency_avg_avg`. The first avg from lib and second avg is average of `lib avg` calculated by `telegraph`.

Jeffail:
Hey @assada, I've merged a backwards compatible version. The way to use the new lib is outlined in the docs: https://github.com/Jeffail/benthos/tree/master/docs/metrics#statsd, basically you set `tag_format` to `none`, `datadog` or `influxdb` and it'll use the new library.

assada:
Thanks. Sorry, I cannot check it now. We are release benthos flow to production and i am has lot of work in this regard :(
We plan to carry out an update in the first half of January.

Jeffail:
No problem, I tested it by capturing the UDP traffic so I'm confident with it. Good luck with the deployments!

PR #339, opened 2019-12-13
Add support for SASL access tokens to Kafka input/outputs

gabrielg:
This adds support for [Sarama's SASL access token providers](https://godoc.org/github.com/Shopify/sarama#AccessTokenProvider) to the Kafka inputs and outputs. Since access token providers are callbacks a custom build of Benthos that registers the access token provider under a given name must be used; that name is then used in the config to specify which access token provider to use.

For example, we have a library that generates auth tokens based on our cloud provider's IAM infrastructure, the `main.go` against this branch looks something like this:

```go
package main

import (
	"os"

	"github.com/Jeffail/benthos/v3/lib/input/reader"
	"github.com/Jeffail/benthos/v3/lib/service"
	"example.com/iam_auth"
)

func main() {
	provider := iam_auth.NewJWTTokenProvider()
	reader.KafkaRegisterAccessTokenProvider("iam", provider)
	service.Run()
}
```

This has a corresponding input configuration along the lines of:

```yaml
input:
  type: kafka
  kafka:
    addresses:
      - "localhost:9094"
    sasl:
      enabled: true
      token_provider: "iam"
    start_from_oldest: false
    topic: "example"
    tls:
      enabled: true
```

A static access token can also be specified via `access_token`, this is useful for testing, or for general purpose use if your tokens are long-lived and you can get away without worrying about refreshes and don't want to build your own Benthos binary.

----

Benthos is useful, thank you!


Jeffail:
Hey @gabrielg, this looks good but I have two items I'd like to solve:

1. There's potential in the future for sharing these token providers with other components (NATS, etc), so I think it makes sense to have the manager own this and components can access token providers similar to how they might access rate limits or caches.
2. I'm not a fan of leaking Go API instructions in the regular YAML config docs. If we're implementing a mechanism for shared token provider implementations I either want it to be 100% configurable or not exposed in the regular docs at all.

We can solve item 1 by adding a new `access_token_provider` (I'll think about a more appropriate name) type to the `resource` section where we can add implementations of the interface:

```go
type AccessToken struct {
    // Token is the access token payload.
    Token string

    // Extensions is a optional map of arbitrary key-value pairs that can be
    // sent with the SASL/OAUTHBEARER initial client response.
    Extensions map[string]string
}

type AccessTokenProvider interface {
    // Token returns an access token. The implementation should ensure token
    // reuse so that multiple calls at connect time do not create multiple
    // tokens. The implementation should also periodically refresh the token in
    // order to guarantee that each call returns an unexpired token.  This
    // method should not block indefinitely--a timeout error should be returned
    // after a short period of inactivity so that the broker connection logic
    // can log debugging information and retry.
    Token() (*AccessToken, error)
}
```

This interface might need to be adjusted to make it more generally useful for other components.

Once item 1 is dealt with we can look at adding configurable implementations of JWT, etc, so there's no need to use the Go API for working with this stuff, which solves item 2.

gabrielg:
Thank you for the review and suggestions, @Jeffail. I can rework this to address `1`.

`2` is interesting. One idea that wouldn't involve the Go API is that it could work with a credential helper kind of like the `subprocess` processor, when the input requests a new token it invokes the credential helper which returns the token on `stdout`. I don't think it would be necessary to have it be a constantly running process that reads from `stdin`. Config could be something like:

```yaml
resources:
  token_generators:
    aws_iam:
      type: subprocess
      subprocess:
        command: generate-token -jwt
```

I think some level of customisation is likely requisite because folks have all sorts of weirdware ways of generating these tokens (including our use case). Thoughts?

Jeffail:
I like the subprocess idea. I'm maybe thinking we could also solve this with the existing cache type, which would save us adding a new component. If we add a cache type `subprocess` that executes an arbitrary command for each cache operation (get initially, ignoring set, delete, etc for now) then we can complement it with more cache components:

```yaml
input:
  kafka:
    addresses:
      - "localhost:9094"
    sasl:
      type: jwt
      token_key: ""
      token_cache: "iam_levels"
    start_from_oldest: false
    topic: "example"
    tls:
      enabled: true

resources:
  caches:
    iam_levels:
      tiered: [ iam_cached, iam ]
    iam_cached:
      memory:
        ttl: 60
    iam:
      subprocess:
        command: generate-token -jwt
```

We can tackle implementing the cache levels separately, which means we'd need to implement a `sasl` configuration block that (ideally) would be flexible enough to use elsewhere, similar to the `tls` options we have. We don't even need to write the `subprocess` cache type initially as you can plug those in.

gabrielg:
That makes sense. I could also see it potentially playing nicely with config interpolation for a few different use cases, for example:

```yaml
input:
  type: stdin

pipeline:
  processors:
    - noop:
      type: noop

output:
  http_client:
    headers:
      Authorization: Bearer ${!cache:oauth2_token,value}
      Content-Type: application/json
    url: https://storage.googleapis.com/storage/v1/b/${BUCKET_NAME}/o/${!json_field:key}.json
    verb: POST

resources:
  caches:
    oauth2_token:
      tiered: [cached_token, generate_token]
    cached_token:
      memory:
        ttl: 60
    generate_token:
      subprocess:
        command: gcloud auth application-default print-access-token
```

In this case something along these lines would allow one to offload GCP credential generation to the `gcloud` command line tool and use GCS as a sink, without having to write an output plugin right away (ignoring the S3-compatible GCS API for a moment). This could be used to read credentials that rotate transparently from sources like Vault via the `vault` CLI, etc.

The idea does make `cache` more of a general purpose `kv`/`keyvalue` resource and I wonder if it would make sense to migrate to that name in a future release, were this to be the direction we go in.

gabrielg:
@Jeffail With some time for this to settle and in light of the appearance of #348, do you have any further thoughts on the right direction forward? Happy to rework this into either a new component type (e.g. `token_provider`) or a new cache type, I think they both have pros and cons. I'm leaning new component type myself.

Jeffail:
Hey @gabrielg, I think the cache plan should be fine. If the concept doesn't fit well and we change our minds it's easy enough to add a new component to replace it. If we add a new component it's a lot of work to implement, requires a whole new documentation section, and it's there for good, so I'd rather avoid it.

gabrielg:
@Jeffail I reworked this based on the feedback to instead allow keys to be fetched from a cache, let me know if it needs any further changes. I also have [another branch](https://github.com/Jeffail/benthos/compare/master...gabrielg:add-subprocess-cache-type) that adds a `subprocess` cache type, if you like I can open a PR for it in the future. Thanks!

Jeffail:
Ace, thanks @gabrielg. In the end I re-added the `enabled` field for backwards compatibility but it's hidden from docs.

PR #340, opened 2019-12-18
Kafka partitioners

adamwasila:
Main motiviation for this PR was to add default partitioner type as implemented by java driver. This differs from sarama defaul prtitioner. Former uses `"murmur2"` hash while latter - some variation of fmv (most probably because Go implements it in default lib). I wonder should `hash` be the name though or should it be called with exact name of the hash (`"fmv-1a"` perhaps)? Or maybe both names should be supported for default hashing? Speaking about aliases - maybe we should have `"java-default-hash"` and `"sarama-default-hash"` or similar?

Regarding implementation: you may also notice that `murmur2` implementation is quite straightforward yet not the perfect one in terms of memory efficiency. All because of hash.Hash32 interface limitations (it requires it to be incremental while for this hash we need to know input length in advance). Lets ignore this problem as sarama tends to do all hashing with single Write call anyway.

Documentation is just a draft - I'm not a native speaker, so I humbly accept all rewrites. :)

Last but not least - linter and fmt are passing but I have some troubles to run tests. Yet, same issue arise when I try to run from `master` so it is not my fault, I hope.

### DIFF
--- a/config/env/README.md
+++ b/config/env/README.md
@@ -589,6 +589,7 @@ OUTPUT_KAFKA_BACKOFF_MAX_ELAPSED_TIME                 = 5s
 OUTPUT_KAFKA_BACKOFF_MAX_INTERVAL                     = 1s
 OUTPUT_KAFKA_CLIENT_ID                                = benthos_kafka_output
 OUTPUT_KAFKA_COMPRESSION                              = none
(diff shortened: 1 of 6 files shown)


Jeffail:
Awesome, thanks @adamwasila, this looks great, I'll hopefully get this merged today.

PR #341, opened 2019-12-18
Kafka refresh coordinator

ollystephens:
Force a refresh of the coordinator if the current one fails.
If the underlying library returns the same coordinator, ensure we don't close it.


Jeffail:
Interesting, thanks for investigating @ollystephens!

PR #344, opened 2019-12-18
the sqs message attribute regex is not inclusive to defined spec

pkfiyah:
some allowed values defined in the sqs message attribute documentation are excluded from permission in the regex here. (Numerics and spaces being the most noticable, spaces only for the message value not key)

I don't have a go environment locally as I don't use it so I couldn't run the makes. Here are some links to an online go playground that show the before and after with some examples:
Before: https://play.golang.org/p/gZ4tdgmI9il
After: https://play.golang.org/p/e_IjPR_7l7R

I can't wait to be able to use the new stuff we've been stuck on v3.0.0-rc2 for awhile cause of this one hahaha

### DIFF
--- a/lib/output/writer/sqs.go
+++ b/lib/output/writer/sqs.go
@@ -148,8 +148,8 @@ type sqsAttributes struct {
 	dedupeID *string
 }
 
-var sqsAttributeKeyInvalidCharRegexp = regexp.MustCompile(`(^\.)|(\.\.)|(^aws\.)|(^amazon\.)|(\.$)|([^a-z_\-\.]+)`)
-var sqsAttributeValueInvalidCharRegexp = regexp.MustCompile(`(^\.)|(\.\.)|(\.$)|([^a-z_\-\.]+)`)
+var sqsAttributeKeyInvalidCharRegexp = regexp.MustCompile(`(^\.)|(\.\.)|(^aws\.)|(^amazon\.)|(\.$)|([^a-z0-9_\-\.]+)`)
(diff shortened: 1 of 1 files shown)


razor-x:
Thanks @pkfiyah. This has been a significant regression we are just catching now when we need to update for some new features / fixes.

I'm wondering if the value check can be even more permissive, or even removed entirely. The SQS docs say the value can have any characters that would be valid in an SQS body. 

> The message attribute value. For String data types, the attribute values has the same restrictions as the message body. 

The main restriction for an attribute is the total size (which does not appear to be checked anyway right now)

> All components of a message attribute are included in the 256 KB message size restriction.

See: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-attributes.html

For example, with the updated regex an ISO date would still get blocked:

```go
var sqsAttributeValueInvalidCharRegexp = regexp.MustCompile(`(^\.)|(\.\.)|(\.$)|([^a-z_\-\.]+)`)
// ...
isValidSQSAttribute("jobDate", "1997-07-16T19:20:30.45+01:00") // => false

Jeffail:
Thanks @pkfiyah, I'll get another release candidate out today.

razor-x:
@Jeffail thanks! Can you confirm if my comment above regarding was also addressed? In particular I see the regression test cases you added are still pretty limited and do not include a test for iso dates.


Ultimately if this check it's supposed to match the sqs restrictions then it would need a regex that matches the character ranges, i.e.

https://stackoverflow.com/questions/16688523/aws-sqs-valid-characters

razor-x:
See https://github.com/Jeffail/benthos/pull/345 for follow up on this still blocking too many attrs.

PR #345, opened 2019-12-19
Add additional tests for valid SQS attr values

razor-x:
Following the SQS docs, attr values can have any valid SQS body char. Those are shown here:

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html

```
#x9 | #xA | #xD | #x20 to #xD7FF | #xE000 to #xFFFD | #x10000 to #x10FFFF 
```

Since `Ѱ = #x470 = 0x470` is in the range `#x20 to #xD7FF` the new tests should pass.

### DIFF
--- a/lib/output/writer/sqs.go
+++ b/lib/output/writer/sqs.go
@@ -149,11 +149,9 @@ type sqsAttributes struct {
 }
 
 var sqsAttributeKeyInvalidCharRegexp = regexp.MustCompile(`(^\.)|(\.\.)|(^aws\.)|(^amazon\.)|(\.$)|([^a-z0-9_\-\.]+)`)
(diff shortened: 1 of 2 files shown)


razor-x:
@Jeffail Both the new tests for ISO date and the valid character `Ѱ` failed the build. Thus this check is still dropping valid SQS attr values.

I am going to remove the check on the value. The only other option I see is implimenting the proper unicode range check, and I'm unsure how to do that in Go.

razor-x:
@Jeffail This should be good to merge now. Even you want to implement the char code validation I suggest you merge this first to restore working SQS functionality and add the new feature independently so it can pass against the new regression tests.

Jeffail:
Looks good, thanks @razor-x!

PR #349, opened 2019-12-30
Cache enhancements

ollystephens:
Exposes a "delete" operator for the cache processor. The underlying cache implementations already had the delete method, but it wasn't exposed.
Adds a "cache" condition, which can test for a hit (key present), miss (key not present) or presence of a specific key/value combination.


Jeffail:
Hey @ollystephens, looks good. The `delete` operator for the processor is great, happy to merge that.

I need to think through the implications of the condition though. Currently there are no IO bound conditions, and I've made a conscious effort so far to keep that since it makes composing them much simpler. My recommendation would be to use the processor to query a cache and error handling to set a metadata field which you refer to with conditions.

ollystephens:
Hi @Jeffail; happy to have the cache condition rejected if I'm honest. Seemed like a useful thing to create when I started, but I was less enthusiastic by the time I finished. I was thinking it would be useful for simplifying our current merge logic, which is:
1. duplicate the part
2. replace the duplicate using cache get
3a. if successful, merge the two parts and cache set the merged result
3b. if not successful, delete the duplicate part
But on reflection, you don't get much in the way of simplification from the condition.
Do you want me to modify the PR to remove those changes and just leave the "delete" operator (which I really need)?

Jeffail:
Yeah let's take it out so we're just merging the `delete` operator.

ollystephens:
Closing; will propose the cache deletion operator separately.

ollystephens:
Replaced with https://github.com/Jeffail/benthos/pull/351

PR #351, opened 2020-01-02
enhance "cache" processor with a "delete" operator

ollystephens:
The cache abstraction already had a delete method; this just exposes it as an operator option.


Jeffail:
Awesome, thanks @ollystephens!

PR #356, opened 2020-01-09
add explode logic for maps

ollystephens:
As discussed on Slack, this PR extends the new json `explode` operator to work with maps as well as arrays.


Jeffail:
Awesome, thanks @ollystephens

PR #357, opened 2020-01-09
extend unarchive with support for JSON maps

ollystephens:
The PR extends the `unarchive` processor with the ability to deal with JSON maps.
The input message is expected to be a JSON map.
Each top-level value is converted to a separate message; the corresponding key is added to the message metadata.


Jeffail:
Thanks again @ollystephens!

PR #363, opened 2020-01-29
Add JSON Condition

DpoBoceka:
Hey. Closes https://github.com/Jeffail/benthos/issues/280
Looks a bit nasty in `Contains` operator since we have to handle an interface `Arg`. Also, didn't implement the case when `Arg` is a JSON object itself, don't know if we need that as well.
And no `Equals` operator, it is not so much difficult to check it old-fashioned way, I surmise.
`Exists` looks alright, though.
 I could forget something important regarding managing json, so please, have a closer look and help me with that new shiny doc.

### DIFF
--- a/lib/condition/constructor.go
+++ b/lib/condition/constructor.go
@@ -48,6 +48,7 @@ var (
 	TypeCheckInterpolation = "check_interpolation"
 	TypeCount              = "count"
 	TypeJMESPath           = "jmespath"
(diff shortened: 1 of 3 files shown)


DpoBoceka:
I did as I was told

Jeffail:
Ace, looks good. Thanks @DpoBoceka!

PR #365, opened 2020-01-31
add geoip processor

DpoBoceka:
hey, look what I've done.
Closes https://github.com/Jeffail/benthos/issues/277

Not sure If I handled closing handler to db right. 

Have a look at it, please, and I'll add some tests if everything in order.

```
pipeline:
  processors: 
  - geoip:
      arg: "city"  # or "asn"
      base: "path/to/db"
```

### DIFF
--- a/lib/processor/geoip.go
+++ b/lib/processor/geoip.go
@@ -1,0 +1,171 @@
+package processor
+
+import (
+	"fmt"
+	"net"
+	"time"
+
+	"github.com/Jeffail/benthos/v3/lib/log"
+	"github.com/Jeffail/benthos/v3/lib/metrics"
+	"github.com/Jeffail/benthos/v3/lib/types"
+	"github.com/opentracing/opentracing-go"
+	"github.com/oschwald/geoip2-golang"
+)
+
+//------------------------------------------------------------------------------
+
+func init() {
+	Constructors[TypeGeoIP] = TypeSpec{
+		constructor: NewGeoIP,
+		Description: `
+		// todo some docs
+		`,
+	}
+}
+
+//------------------------------------------------------------------------------
+
+// GeoIPConfig contains configuration fields for the GeoIP processor.
+type GeoIPConfig struct {
+	Parts []int  `json:"parts" yaml:"parts"`
+	Arg   string `json:"arg" yaml:"arg"`
+	Base  string `json:"base" yaml:"base"`
+}
+
+// NewGeoIPConfig returns a GeoIPConfig with default values.
+func NewGeoIPConfig() GeoIPConfig {
+	return GeoIPConfig{
+		Parts: []int{},
+		Arg:   "",
+		Base:  "",
+	}
+}
+
+//------------------------------------------------------------------------------
+
+type geoipReader func(db *geoip2.Reader, ip net.IP) (interface{}, error)
+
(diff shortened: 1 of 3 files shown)


Jeffail:
Hey @DpoBoceka, this looks good but I would rename the fields to be `database` and `database_type`.

DpoBoceka:
@Jeffail, Is there anything else I should fix?

lib/processor/geoip.go
```diff
@@ -0,0 +1,171 @@
...
+
+type geoipReader func(db *geoip2.Reader, ip net.IP) (interface{}, error)
+
+func cityReader() geoipReader {
+	return func(db *geoip2.Reader, ip net.IP) (interface{}, error) {
+		res, err := db.City(ip)
+		if err != nil {
+			return nil, err
+		}
+		return res, nil
```
Jeffail: Returning a `*City` type might cause issues downstream as it's a struct. If you don't want to map the entire struct into `map[string]interface{}` then I'll forgive the lazier approach of `json.Marshal(res)` into raw bytes and use `part.Set`. Considering we're doing a DB call here I don't think the performance impact is something to worry about.

Same goes for `asnReader`.

lib/processor/geoip.go
```diff
@@ -0,0 +1,171 @@
...
+	return nil, fmt.Errorf("unknown database type: %s", dbtype)
+
+}
+
+//------------------------------------------------------------------------------
+
+// GeoIP is a processor that looks up IP addresses.
+type GeoIP struct {
+	parts   []int
+	handler *geoip2.Reader
```
Jeffail: If you change this for a interface with the methods `City` and `ASN` then we can use a mock type for unit tests.

Jeffail:
Hey @DpoBoceka, left some comments. Does a unit test here make any sense? Would be good to get some coverage here even if it's wrapping a mock implementation. I'm happy to do the docs when I have time.

lib/processor/geoip.go
DpoBoceka: I don't exactly understand, could you clarify a couple of things for me, please?
I could use an interface for replacing current `GeoIP.reader` an mock it but in order for that test/code to work I would be forced to explicitly declare `processor.Geoip`, not by calling `NewGeoIP`, because this would fail https://github.com/Jeffail/benthos/pull/365/files#diff-a4df8555a7687abaed0682a67d09ea3dR114
Is it ok to omit test for `NewGeoIP` or there is another way to refactor that code?

Jeffail: Yeah that's fine, I'm mostly interested in testing `ProcessMessage` so I'm fine with declaring the struct instead of using `NewGeoIP`.

Jeffail:
Hey @DpoBoceka, I think it makes sense to rework this into a bloblang method if you're open to that as it's much more flexible. I can take over if you don't have time.

DpoBoceka:
Hey @Jeffail, that would be just great. 

If you don't mind implementing that yourself I'd be delightful as I need to invest some time in figuring out the bloblang internals( I didn't use benthos for quite a while now, hence absence of my activity)

Jeffail:
Merged from https://github.com/Jeffail/benthos/pull/1068

PR #366, opened 2020-02-03
Clarification of `sub_filters` description for zmq4 and nanomsg

ollieparsley:
When going through the docs I initially confused `sub_filters` with being synonymous with an array of filter processors specific to zmq4


Jeffail:
Nice one, thanks @ollieparsley!

PR #369, opened 2020-02-07
add syslog_rfc3164 format for parse_log processor

DpoBoceka:
Hi. Added ability to parse syslog messages formatted by rfc3164.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -38,7 +38,7 @@ require (
 	github.com/gotestyourself/gotestyourself v2.2.0+incompatible // indirect
 	github.com/hashicorp/go-uuid v1.0.2 // indirect
 	github.com/hashicorp/golang-lru v0.5.3 // indirect
-	github.com/influxdata/go-syslog v1.0.1
+	github.com/influxdata/go-syslog/v3 v3.0.0
 	github.com/jcmturner/gofork v1.0.0 // indirect
 	github.com/jmespath/go-jmespath v0.0.0-20180206201540-c2b33e8439af
 	github.com/jtolds/gls v4.20.0+incompatible // indirect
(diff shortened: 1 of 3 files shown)


DpoBoceka:
Hey @Jeffail, is it okay if we'd merge this one?

Jeffail:
Hey @DpoBoceka, are the parameters all features you need? I'm fine with adding them but I think the naming needs some work so they aren't as cryptic. I'd also rather not prefix them with `syslog` since there's potential cross over with other formats (we might want `best_effort` for other types for example).

DpoBoceka:
Well, with my experience in processing syslog from various vendors we should have as much flexibility as possible since they don't in fact play by "rules" quite often and you are, eventually, to end up with ugly hacks here and there to cover up some laziness which shouldn't be there in the first place. So I strongly suggest we keep these options. 
I do agree the exemption of complicated prefixes would benefit us, but I was afraid back then that someone would occasionally misuse some of them. But if the doc has it covered, no harm done, I suppose. 

Would you find the following as an appropriate choice?
```
best_effort
allow_rfc3339
default_yer
default_timezone
```

Jeffail:
Cool, sounds good. Yeah those names are fine. Let's also put them in the root of the config structure rather than under the `parameters` namespace.

DpoBoceka:
@Jeffail, I made changes you requested. At the moment, doc isn't quite generic as it could be, but it should be alright for now provided we don't have other parser options at hand apart from syslog cases.

Jeffail:
Thanks @DpoBoceka, if you get a chance to add unit tests for the new parameters that would appreciated but I've merged as is for now just to get this out the way.

Issue #396, opened 2020-03-06
Kafka input with SASL only sets user/password if mechanism is PLAIN


hav:
```input:
  kafka_balanced:
    addresses:
      - localhost:9094
    topics:
      - topic
    client_id: client-id
    consumer_group: consumer-group
    sasl:
      mechanism: SCRAM-SHA-256
      user: ${USER}
      password: ${PASSWORD}
```

Minimal config. User and Password will not get passed through to be set since mechanism is not to `PLAIN`.

PR #397, opened 2020-03-06
Setting SASL user and password for SCRAM mechanisms as well.

hav:
This solves #396.


Jeffail:
Looks great, thanks @hav! I'll aim to get a release out this weekend.

PR #404, opened 2020-03-10
Adding Interpolations to json processor 

DpoBoceka:
Hey @Jeffail. This adds interpolations to both `path`(including `copy` and `move` operators now) and `value` fields of `json` processor.
All tests have passed, I tagged this as a draft as I'd like your general review first and then I'll add more unit tests.
Also, I would really love to extend support of interpolation across many processors/conditions one by one as it is helpful to achieve functional-based config support which I'm after.

### DIFF
--- a/lib/processor/json.go
+++ b/lib/processor/json.go
@@ -311,12 +311,12 @@ func NewJSONConfig() JSONConfig {
 
 //------------------------------------------------------------------------------
 
-type jsonOperator func(body interface{}, value json.RawMessage) (interface{}, error)
+type jsonOperator func(body interface{}, path []string, value json.RawMessage) (interface{}, error)
 
-func newSetOperator(path []string) jsonOperator {
-	return func(body interface{}, value json.RawMessage) (interface{}, error) {
+func newSetOperator() jsonOperator {
+	return func(body interface{}, path []string, value json.RawMessage) (interface{}, error) {
+		var data interface{}
 		if len(path) == 0 {
-			var data interface{}
 			if value != nil {
 				if err := json.Unmarshal([]byte(value), &data); err != nil {
 					return nil, fmt.Errorf("failed to parse value: %v", err)
@@ -327,7 +327,6 @@ func newSetOperator(path []string) jsonOperator {
 
 		gPart := gabs.Wrap(body)
 
-		var data interface{}
 		if value != nil {
 			if err := json.Unmarshal([]byte(value), &data); err != nil {
 				return nil, fmt.Errorf("failed to parse value: %v", err)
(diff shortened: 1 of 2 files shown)


Jeffail:
Hey @DpoBoceka, similar to the metadata PR I'd like to see some benchmarks if possible comparing these operators to awk program equivalents.

DpoBoceka:
I wrote a not accurate benchmark just to see the overhead of calling `awk`. 
On my machine I got the following:
```
/usr/local/bin/go test -benchmem -run=^$ github.com/Jeffail/benthos/v3/lib/processor -bench ^(BenchmarkJSONMove)$
goos: darwin
goarch: amd64
pkg: github.com/Jeffail/benthos/v3/lib/processor
BenchmarkJSONMove/awk-16         	   27001	     44536 ns/op	   89793 B/op	     111 allocs/op
BenchmarkJSONMove/json-16        	  254396	      4821 ns/op	    3003 B/op	      50 allocs/op
```

Perhaps, we should compare `json` with interpolation and without, but I think there'll be not much of a difference, though

DpoBoceka:
Actually, I switched to master and tried to run similar benchmark for regular move of `json` processor, without interpolated `path` and `value` fields. Benchmark related to `awk` was constant.

```
BenchmarkJSONMove/awk-16         	   26587	     47444 ns/op	   89652 B/op	     111 allocs/op
BenchmarkJSONMove/json-16        	  507470	      2304 ns/op	    2096 B/op	      25 allocs/op
```

DpoBoceka:
Added benchmark to case when no interpolation is used

```
BenchmarkJSONMove/awk-init-16         			   26907	     44032 ns/op	   89850 B/op	     111 allocs/op
BenchmarkJSONMove/json-with-interpolations-16         	  252193	      4609 ns/op	    3001 B/op	      50 allocs/op
BenchmarkJSONMove/json-no-interpolations-16           	  444342	      2876 ns/op	    2666 B/op	      34 allocs/op```

**Comment by Jeffail** [comment]
Hey @DpoBoceka, just to give you an update, I'm currently working on an improved interpolation syntax, which would eventually include arithmetic:

```yaml
pipeline:
  processors:
  - text:
      operator: set
      value: "${!json(foo) + json(bar)}"
```

And my longer term goal is to then extend that syntax to support a general mapping transform, that would allow you to express long mappings with the same performance as the regular processors. That would look something like this:

```yaml
pipeline:
  processors:
  - mapping: |
      json.foo = json(thing1) - json(thing2)
      json.bar = meta(kafka_key)
      meta.baz = meta(error) | "no error"
```

And I think this would be a lot cleaner overall than mutating the existing transforms.

DpoBoceka:
Hey @Jeffail, thank you for the info, sounds promising.

PR #405, opened 2020-03-10
add mutliset operator for metadata processor

DpoBoceka:
Hey @Jeffail, is that conceptually alright? Or do you want to change `key` and `value` types to slice, but that would be a breaking change.

### DIFF
--- a/lib/processor/metadata.go
+++ b/lib/processor/metadata.go
@@ -1,6 +1,8 @@
 package processor
 
 import (
+	"encoding/json"
+	_ "encoding/json"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


Jeffail:
Hey, I think if we were going to lean into multiset we would need to do it similarly to the [metadata condition](https://github.com/Jeffail/benthos/blob/master/lib/condition/metadata.go) and use :

```yaml
metadata:
  operator: multiset
  value:
    foo: bar
    foo2: bar2
```

Although maybe hold off on that for now, I'd like to see some evidence that the performance of the `awk` processor for setting multiple keys is significantly worse before expanding metadata.

Jeffail:
Hey @DpoBoceka, closing this as well as https://github.com/Jeffail/benthos/pull/404 in favour of the new [bloblang processor](https://www.benthos.dev/docs/components/processors/bloblang). Thanks for your effort here as this PR along with our discussions helped a lot with planning bloblang out.

PR #409, opened 2020-03-12
add contains_any contains_cs_any for text condition

DpoBoceka:
Closes https://github.com/Jeffail/benthos/issues/406

### DIFF
--- a/lib/condition/text.go
+++ b/lib/condition/text.go
@@ -48,6 +48,16 @@ Checks whether the content contains the argument (case sensitive.)
 Checks whether the content contains the argument under unicode case-folding
 (case insensitive.)
 
+### ` + "`contains_cs_any`" + `
+
+Checks whether the content contains any of the list of arguments under unicode case-folding
+(case sensitive.)
+
+### ` + "`contains_any`" + `
+
+Checks whether the content contains any of the list of arguments under unicode case-folding
+(case insensitive.)
+
 ### ` + "`is`" + `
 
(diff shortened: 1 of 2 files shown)


DpoBoceka:
I think we could perform a bit of a refactor job here as `case_sensitive` could be offered as a standalone option in config. Same goes to new-came `any` logic for `contains` operator which could adopt `all` as well.
Also, it is not possible to add interpolation to that condition's codebase as it is.

I would like to propose:
-  Creation of new branch `v4` which would accumulate all breaking changes in it up to the point we are safe to switch to it and prepared some instructions and scripts, perhaps
- Conduct a refactor of some components to provide as much interpolation as possible and while we're at it we could replace some `string` types to `[]byte` to avoid excessive operation in memory
- Consider a possibility in allowing for some fields to be expected as arrays and values - so with that operator under this examination would become:
```
text:
    operator: contains
    case_sensitive: true
    policy: any
    arg: test # could be an array as well  
```

@Jeffail If you are in favour with these and would like to add something on we could outline some sort of a best-practises we would like to stick to in the code base(e.g. the third proposal here) and launch a milestone to `v4`
I am so very much interested in helping you with it.

Jeffail:
Thanks this looks good for now, I'm just going to rename `contains_cs_any` to `contains_any_cs`. On the subject of case sensitivity I would rather keep them as separate operations. The problem with a general `case_sensitive` field is that it's ambiguous as to how it would interact with operators like the `regexp_exact`.

PR #413, opened 2020-03-23
Make records chan so ProcessRecords doesn't block on nil chan

avdempsey:
Make the `*gokini.Records` channel so we don't have a zero value (nil) channel created by NewKinesisBalanced. When fetching records from Kinesis they were getting sent to a nil channel in ProcessRecords. This blocks.


Jeffail:
Thanks @avdempsey, looks good.

PR #419, opened 2020-04-02
redis_list: Add support for databases other than 0

relistan:
What
-----
The `redis_list` input reads a list key from Redis but currently only supports the default database `0`. This adds support for the Redis standard of supplying the database number in the URL path. For backward compatibility it defaults to `0` if not supplied. ~~The defaults have been updated~~ to explain the added database support and the docs have been regenerated.

New URLs support a syntax like this:

```
tcp://localhost:6379/0
or
tcp://localhost:6379
or
tcp://localhost:6379/12
```

### Tests
I did not include any additional tests since this input is currently untested and this is in the connection logic so without adding a dependency on Redis it's not clear how to test it properly. It has been tested locally and works as expected for me both with a specified DB and without one.

Why
-----
We've got a number of Redis databases and need to support the others. The other Redis input types don't need this support since they don't reference keys.


go.mod
```diff
@@ -47,8 +47,6 @@ require (
 	github.com/linkedin/goavro/v2 v2.9.7
 	github.com/mailru/easyjson v0.7.0 // indirect
 	github.com/microcosm-cc/bluemonday v1.0.2
-	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
```
relistan, line 50: These are removed when running the build. I committed them since they are required to run the build now.

Jeffail:
Thanks @relistan, this looks great. 

Would you have time to add this to the [redis cache](https://www.benthos.dev/docs/components/caches/redis) and the `redis_list` output (if it makes sense to do that)? Otherwise I'm happy to add it.

Also, it's probably best to leave the existing default value for `url` unchanged for now since it's functionally the same, I plan to remove all default url and address fields in the next major version release anyway (https://github.com/Jeffail/benthos/issues/392).

For testing there are some integration tests in https://github.com/Jeffail/benthos/tree/master/lib/test/integration which use docker. As long as they pass in their current state we're fine, I'll run it myself before the next release anyway so unless you want to add more tests then you don't need to worry about it.

relistan:
Sure, @Jeffail I'll take a look tomorrow.

relistan:
Hi @Jeffail, I made all those updates. I am unable to get the integration tests to run at all. They create Redis Docker containers, which are avialable, but from both macOS and Linux I get:

```
--- FAIL: TestRedisHashIntegration (33.78s)
    redis_hash_test.go:63: Could not connect to docker resource: read tcp [::1]:53180->[::1]:32780: i/o timeout
--- FAIL: TestRedisStreamsIntegration (33.86s)
    redis_streams_test.go:60: Could not connect to docker resource: read tcp [::1]:41100->[::1]:32781: i/o timeout
--- FAIL: TestRedisPubSubIntegration (37.15s)
    redis_pubsub_test.go:60: Could not connect to docker resource: read tcp [::1]:48568->[::1]:32782: i/o timeout
```

Executed with the following CLI from inside `lib/test/integration`:
```
go test -tags integration -run ".*Redis.*"
```

Jeffail:
Thanks @relistan, this is great, and thanks for trying out the integration tests. I pulled the changes locally and the tests run fine so it's most likely just a networking problem against docker. I'm hoping to make the integration tests more portable soon (and finally have them run in CI).

Issue #381, opened 2020-02-20
Improvements to the avro processor


gregorskii:
HI again,

I am working with the Avro processor. We are trying to use it to process Avro textual records (or pure JSON) as the event input, and marshal it to Avro a record to send to GCP as bytes.

Currently the process we have been debugging in #377 is working, but we are trying to move Avro validation and processing closer to our event processing side (after GCP).

In our ideal flow our event publishers would send either pure JSON or textual Avro records to Benthos, Benthos would process those records into Avro records and transmit them to our pipeline as a Avro byte object. The pipeline would be able to validate the bytes with the Avro reader and writer.

Is this possible with Benthos and the Avro processor currently, or am I misunderstanding what the Avro processor is doing?

I have added it to a test pipeline with a schema defined and it seems to have no affect.

A second question is whether the Avro.schema property can be loaded from a file dynamically, or does it have to come from a single Avro schema definition?

Here is our example Avro schema:

```
{
  "type" : "record",
  "name" : "Event_TestEvent",
  "fields" : [ {
    "name" : "event_type",
    "type" : {
      "type" : "enum",
      "name" : "EventType",
      "symbols" : [ "TestEvent" ]
    }
  }, {
    "name" : "metadata",
    "type" : {
      "type" : "record",
      "name" : "Metadata",
      "fields" : [ {
        "name" : "schema_version",
        "type" : "string"
      } ]
    }
  }, {
    "name" : "event_properties",
    "type" : {
      "type" : "record",
      "name" : "TestEventProperties",
      "fields" : [ {
        "name" : "source",
        "type" : [ "null", "string" ],
        "default" : null
      } ]
    }
  }, {
    "name" : "event_id",
    "type" : "string"
  }, {
    "name" : "timestamp",
    "type" : {
      "type" : "long",
      "logicalType" : "timestamp-millis"
    }
  } ]
}
```

Benthos config:

```
http:
  address: 0.0.0.0:1337
  read_timeout: 5s
  root_path: /benthos
  debug_endpoints: true

input:
  type: broker
  broker:
    copies: 25
    inputs:
    - type: http_server
      http_server:
        address: ""
        cert_file: ""
        key_file: ""
        path: /track
        timeout: 5s
        ws_path: /track/ws

buffer:
  type: memory
  memory:
    limit: 1_000_000_000

pipeline:
  processors:
    - avro:
        operator: from_json
        encoding: textual
[...]

PR #420, opened 2020-04-06
Adding schema_path option to avro processor

dscleaver:
Attempted to address part of #381 by adding a schema_path option to the avro processor.

Similar to the jsonschema processor, schema_path allows the configuration of a file or http url that is used to load the schema.


Jeffail:
Awesome, looks great, thanks @dscleaver!

PR #423, opened 2020-04-10
gcp_pubsub: Remove client context cancellation

### LINKED ISSUE #422: gcp_pubsub input/output cannot authenticate with user application default credentials
Currently, the gcp_pubsub input and output clients are created with a context that has a timeout and is cancelled shortly after being used. This conflicts with how the Pub/Sub client's authentication works when using application default credentials associated with a user (e.g., for testing a configuration locally).

What happens is basically this:

1. Benthos creates the Pub/Sub client with the following (effectively the same for both input and output):
    ```go
    ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
    defer cancel()
    client, err := pubsub.NewClient(ctx, conf.ProjectID)
    ```
2. Down the chain, application default credentials are used -- that is, this eventually hits [google.CredentialsFromJSON](https://github.com/golang/oauth2/blob/bf48bf16ab8d/google/default.go#L129).
3. This leads to [credentialsFile.tokenSource](https://github.com/golang/oauth2/blob/bf48bf16ab8d622ce64ec6ce98d2c98f916b6303/google/google.go#L130) being called with a [user credentials key](https://github.com/golang/oauth2/blob/bf48bf16ab8d622ce64ec6ce98d2c98f916b6303/google/google.go#L130) (i.e., `"authorized_user"` type), which creates a token source using a refresh token, passing it and the original context from (1) to cfg.TokenSource.
4. At this point, [cfg.TokenSource hangs onto the context](https://github.com/golang/oauth2/blob/bf48bf16ab8d622ce64ec6ce98d2c98f916b6303/oauth2.go#L240) (because it needs to refresh the token).

You can reproduce this by authenticating via the [gcloud tool's application-default authentication](https://cloud.google.com/sdk/gcloud/reference/auth/application-default). What you should end up with is roughly the following output from Benthos for any combination of project and topic (if using a gcp_pubsub output -- the failure might be delayed for an input):

```
{"@timestamp":"2020-04-10T10:31:57-07:00","@service":"benthos","level":"ERROR","component":"benthos.output","message":"Failed to connect to gcp_pubsub: rpc error: code = Unauthenticated desc = transport: context canceled"}
```

I think the simplest solution is to just remove the timeout and cancellation for the pubsub.NewClient calls. Hopefully the above isn't too confusing -- it skips most of the hops through the pubsub client and vkit, since it's a bit of a rabbit hole to go down.

nilium:
Remove the client cancellation / timeout because this affects
user-level application default credentials (i.e., `gcloud auth
application-default ...`). Service account credentials don't appear to
need to refresh tokens, but user-level credentials do.

Through the Pub/Sub client's vkit client, and further under
golang.org/x/oauth2, the token refresher hangs onto the context passed
into pubsub.NewClient. As a result, when the context is cancelled (or
inevitably exceeds its deadline), the client becomes unable to refresh
tokens. This, in turn, results in an error that looks like "rpc error:
code = Unauthenticated desc = transport: context canceled".

So, in short, the Pub/Sub inputs and outputs only function with service
account keys unless the timeout is removed, because the context has to
survive.

Closes #422 (the accompanying issue describing this a bit more).

### DIFF
--- a/lib/input/reader/gcp_pubsub.go
+++ b/lib/input/reader/gcp_pubsub.go
@@ -67,9 +67,7 @@ func NewGCPPubSub(
 	log log.Modular,
 	stats metrics.Type,
 ) (*GCPPubSub, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
-	defer cancel()
-	client, err := pubsub.NewClient(ctx, conf.ProjectID)
+	client, err := pubsub.NewClient(context.Background(), conf.ProjectID)
 	if err != nil {
 		return nil, err
 	}
--- a/lib/output/writer/gcp_pubsub.go
+++ b/lib/output/writer/gcp_pubsub.go
@@ -54,9 +54,7 @@ func NewGCPPubSub(
 	log log.Modular,
 	stats metrics.Type,
 ) (*GCPPubSub, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
-	defer cancel()
-	client, err := pubsub.NewClient(ctx, conf.ProjectID)
+	client, err := pubsub.NewClient(context.Background(), conf.ProjectID)
 	if err != nil {
 		return nil, err
 	}


Jeffail:
Hey @nilium, this is great, thanks for investigating/fixing and for the great write up!

I'll get another release candidate out this weekend.

PR #425, opened 2020-04-14
Ensure we checkpoint the right shard

patrobinson:
While testing some other changes I discovered a bug in the code I wrote for talking to my library.
Because Init can be called with different Shard IDs, as we may be processing multiple shards, then we may checkpoint the wrong shard.
This causes checkpoints to be lost and benthos to crash, as Kinesis API throws an error when we try and retrieve a Sequence number that doesn't match the Shard ID.

Also fixed some incorrect errors, when nil records are received that just means no new records are available so a Timeout error is raised. This similar to how SQS consumer behaves.

### DIFF
--- a/lib/input/reader/kinesis_balanced.go
+++ b/lib/input/reader/kinesis_balanced.go
@@ -59,7 +59,6 @@ type KinesisBalanced struct {
 
 	session *session.Session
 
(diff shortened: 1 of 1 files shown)


Jeffail:
Looks good, thanks @patrobinson

PR #427, opened 2020-04-17
Accept http requests without content-type

dscleaver:
The http-server input now accepts requests that do not contain the
"Content-Type" header. This solves an issue when using some existing
HTTP logging tools. Specifically, Benthos can now receive http requests
from the Splunk logging driver for docker.


Jeffail:
Looks good, thanks @dscleaver!

PR #434, opened 2020-05-02
Add hmac algorithms to hash processor

cludden:


Jeffail:
Looks great, thanks @cludden!

PR #440, opened 2020-05-15
Add ascii85 and z85 options to decode/encode

ollystephens:
This PR adds ascii85 and z85 options to the encode/decode processors.

ascii85 is another encoding available in the standard Go library.

z85 is a "better" version as it's easier to embed in source code, json; is defined here: http://rfc.zeromq.org/spec:32/Z85


Jeffail:
Thanks @ollystephens, I'll be adding an `encode` and `decode` function to bloblang soon so I'll make sure to copy these new options over as well.

PR #445, opened 2020-05-21
redis_streams input: do not copy message body to metadata

citrin:
Documentation on redis_streams input says that the `body_key` key from received entry is extracted as raw message. All **other** keys will be stored in the message as metadata.
https://www.benthos.dev/docs/components/inputs/redis_streams#body_key

This change should make behavior to confirm the documentation.

Also it will allow to log an entry metadata without logging the body (which can be too large to log) using following config:
```
pipeline:
  processors:
    - log:
        message: '${! meta() }'
```

Thanks in advance.


Jeffail:
Looks good, thanks @citrin!

PR #448, opened 2020-06-01
Add type property for AMQP_0.9 output

jimgus:
In order to be able to use the amqp_0.9 output to send messages to an existing consumer that expects the Type property to be set on the messages, a configuration option is added to be able to set that property. It also support interpolations.

More info on the Type property: https://www.rabbitmq.com/consumers.html#message-properties


Jeffail:
Looks great, thanks @jimgus!

PR #452, opened 2020-06-04
feat: add base64url variant to decode and encode

ollystephens:
Adding a base64url variant to the bloblang decoder/encoder.
I have a need to be able to decode a JWT token, which is formatted this way.


Jeffail:
Thanks @ollystephens, looks great!

PR #454, opened 2020-06-11
Use Go Modules for lambda package

patrobinson:
When trying to install the lambda package, as per the [documentation](https://www.benthos.dev/docs/guides/serverless/lambda)
it fails with:
```
09:53 $ go get github.com/go-redis/redis/v7
package github.com/go-redis/redis/v7: cannot find package "github.com/go-redis/redis/v7" in any of:
/usr/local/Cellar/go/1.14.2_1/libexec/src/github.com/go-redis/redis/v7 (from $GOROOT)
/Users/patrickrobinson/go/src/github.com/go-redis/redis/v7 (from $GOPATH)
```

This [suggested solution](https://github.com/go-redis/redis/issues/1174) is to use Go Modules

### DIFF
--- a/cmd/serverless/benthos-lambda/go.mod
+++ b/cmd/serverless/benthos-lambda/go.mod
@@ -1,0 +1,5 @@
+module github.com/Jeffail/benthos/cmd/serverless/benthos-lambda
+
+go 1.14
+
+require github.com/Jeffail/benthos/v3 v3.17.0


Jeffail:
Hey @patrobinson, you can get away with simply including the version number in the path:

```
go build github.com/Jeffail/benthos/v3/cmd/serverless/benthos-lambda
```

That way we don't need to maintain multiple `go.mod` files. I've fixed the docs with the new URL.

PR #458, opened 2020-06-23
add strict mode for switch conditions

zlozano:
We have a use case where our switch outputs are an exhaustive list, and the condition is derived from user input. If the input is incorrect, we want to communicate that back to the input, and ultimately the caller. This PR is a proposal to satisfy this need. Happy to pivot if there is a better way to go about this.


Jeffail:
Thanks @zlozano, this looks great! I'm planning to put a release out this weekend.

PR #459, opened 2020-06-25
expose max_in_flight for switch

zlozano:
👋 

me again.

This PR is short, sweet, and hopefully make sense. I think there are a couple of use cases where exposing this makes sense.

1. Where the child outputs have max_in_flight > 1
2. Where different outputs are mutually exclusive

I have a use case where I am routing multiple inputs to multiple outputs. If one of those outputs fails, and has an aggressive retry policy, then other messages are blocked and/or contending for the transaction channel until that head of line message is dropped, skipped (which may not be possible depending on the input), or succeeds. If we have the ability to spin up additional consumer loops here then we can mitigate this problem quite a bit I believe for distinct, mutually exclusive outputs/conditions.

### DIFF
--- a/config/switch.yaml
+++ b/config/switch.yaml
@@ -19,6 +19,7 @@ pipeline:
 output:
   type: switch
   switch:
+    max_in_flight: 1
     outputs: []
     retry_until_success: true
     strict_mode: false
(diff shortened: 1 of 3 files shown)


Jeffail:
Thanks @zlozano, looks good. I was a little hesitant to expose this at first but I don't see any valid reason not to.

PR #466, opened 2020-07-06
Azure Blob Storage output

mfamador:
Hi!
I've created an Azure Blob Storage output that I've been running smoothly for a few days now.
It allows specifying dynamically the container which is created if it does not exist.

Can you guys please have a look?

Example:
```
output:
  switch:
    outputs:
    - condition:
        bloblang: meta("msg_type") == "event"
      fallthrough: false
      output:
        blob_storage:
          storage_account: ${AZURE_STORAGE_ACCOUNT}
          storage_access_key: ${AZURE_STORAGE_ACCESS_KEY}
          container: 'test-events-${!timestamp("2006")}'
          path: '${!timestamp("2006-01-02")}/event-${!timestamp_unix_nano()}.json.gz'
          max_in_flight: 5
    - output:
        blob_storage:
          storage_account: ${AZURE_STORAGE_ACCOUNT}
          storage_access_key: ${AZURE_STORAGE_ACCESS_KEY}
          container: 'test-readings-${!timestamp("2006")}'
          path: '${!timestamp("2006-01-02")}/reading-${!timestamp_unix_nano()}.json.gz'
          max_in_flight: 5
```

Thanks!


lib/output/writer/blob_storage.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// Write attempts to write message contents to a target Azure Blob Storage container as files.
+func (a *AzureBlobStorage) Write(msg types.Message) error {
+	return a.WriteWithContext(context.Background(), msg)
+}
+
+func (a *AzureBlobStorage) getContainer(name string) (*azblob.ContainerURL, error) {
+	accountName, accountKey := a.conf.StorageAccount, a.conf.StorageAccessKey
+	if len(accountName) == 0 || len(accountKey) == 0 {
```
Jeffail, line 107: This check ought to be done within `NewAzureBlobStorage` so that users can see the error during start up.

lib/output/writer/blob_storage.go
```diff
@@ -0,0 +1,178 @@
...
+		return nil, fmt.Errorf("invalid azure storage account credentials")
+	}
+	credential, err := azblob.NewSharedKeyCredential(accountName, accountKey)
```
Jeffail, line 110: Would it make sense to initialize this once during construction and reuse the `credential` struct? Also I'm wondering if we should actually allow users to leave the access key empty in order to use an anonymous credential.

lib/output/writer/blob_storage.go
```diff
@@ -0,0 +1,178 @@
...
+
+	return msg.Iter(func(i int, p types.Part) error {
+		c, err := a.getContainer(a.container.String(i, msg))
+		if err != nil {
+			return err
+		}
+		if err := a.uploadToBlob(ctx, p.Get(), a.path.String(i, msg), a.blobType.String(i, msg), c); err != nil {
+			a.log.Errorf("Error uploading blob: %v.", err)
+			if containerNotFound(err) {
+				if _, cerr := c.Create(ctx, azblob.Metadata{}, azblob.PublicAccessNone); cerr != nil {
```
Jeffail, line 150: If this is successful would it make sense to attempt `uploadToBlob` once more?

lib/output/writer/blob_storage.go
mfamador: You're right. I'm checking now within `NewAzureBlobStorage`

lib/output/writer/blob_storage.go
mfamador: Agreed, reusing `credential` now and using anonymous credential if key is empty.

lib/output/writer/blob_storage.go
mfamador: Agreed, retrying to upload to blob after the container is created successfully.

Jeffail:
Thanks for the updates @mfamador, this looks great. I'll include this in the next release but I'm going to add a BETA tag to the docs for now just in case we want to adjust the config fields eventually.

One thing I'd be interested in is whether there's a common use case for uploading to the storage of one account whilst using the credentials of another, do you know if that's possible? If so it might be worth working that into the config structure.

mfamador:
Great! 

I see, like the case when an AWS account has a trusted relationship with another AWS account and assumes a role which allows it to write in an s3 bucket, right? 
I'm fairly new to Azure, but I'm pretty sure it should be possible besides using access keys. I'll investigate it and try to come up with a solution.

Thanks @Jeffail !

PR #469, opened 2020-07-09
Set correct case for property fields within jsonschema validation errors description

### LINKED ISSUE #468: jsonschema validation errors description in lowercase
The `jsonschema` validation errors description is in lowercase, changing the real name of the fields.

Ex:
```myarray.0 issomething is required```

when in fact the field is called `isSomething`

mfamador:
Hi!
Please confirm that this would resolve #468 
Thanks!


Jeffail:
Thanks @mfamador, looks great. I'll aim to kick off a release this weekend.

PR #470, opened 2020-07-12
Add possibility to define user headers for kafka output

markwiat:
This changes allow user to define own header that is send to kafka topic. Config would looks like:
```
output:
  kafka:
    addresses:
      - localhost:9092
    topic: ot
    client_id: benthos_kafka_output
    key: ${! json("a") }
    headers:
      my-first: ${! json("a") }
      my-second: direct-value
```
EDIT 1
After changes it would looks like:
```
output:
  kafka:
    addresses:
      - localhost:9092
    topic: ot
    client_id: benthos_kafka_output
    key: ${! json("a") }
    static_headers:
      my-first: some-value
      my-second: other-direct-value
```

### DIFF
--- a/lib/output/kafka.go
+++ b/lib/output/kafka.go
@@ -29,9 +29,7 @@ acknowledgement from all replicas or just a single broker.
 Both the ` + "`key` and `topic`" + ` fields can be dynamically set using
 function interpolations described [here](/docs/configuration/interpolation#bloblang-queries).
 When sending batched messages these interpolations are performed per message
-part.
-
-The value of header under ` + "`headers`" + ` can be given in the same format as under ` + "`key` and `topic`.",
+part.`,
 		sanitiseConfigFunc: func(conf Config) (interface{}, error) {
 			return sanitiseWithBatch(conf.Kafka, conf.Kafka.Batching)
 		},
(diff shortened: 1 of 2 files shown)


Jeffail:
Hey @markwiat, I'm not opposed to adding a static map of headers like this, but you can also currently express this with:

```
output:
  kafka:
    addresses:
      - localhost:9092
    topic: ot
    client_id: benthos_kafka_output
    key: ${! json("a") }
  processors:
    - bloblang: |
        meta "my-first" = a
        meta "my-second" = "direct-value"
```

Although since this has a performance overhead of needing to read and write metadata headers for each message then there's likely a valid use case for a `headers` field.

My concern here is that the name `headers` is potentially misleading as I can imagine someone reading your example config and assuming that `my-first` and `my-second` are the _only_ headers that will be sent, and missing the fact that message metadata is included as well.

Maybe we could rename the field `static_headers` to make it more obvious that there are also dynamic headers from the message. I also think we ought to remove the interpolation of the values since that's already solved with metadata.

markwiat:
@Jeffail, Thank you for explanation.
I've renamed `headers` into `static_headers` and remove interpolation.

Jeffail:
Thanks @markwiat, looks good. I'll add this to todays release.

PR #475, opened 2020-07-17
AES encrypt/decrypt

eduardodbr:
Hi!
I've created a bloblang method to do AES encryption/decryption using either CTR, OFB or CBC method. The implementation was based on the encode/decode methods. The method receives the key and an initialization vector/nonce as arguments, can you take a look?

Example:
```
pipeline:
  processors:
    - bloblang: |
        decryptedData = value.decrypt_aes("ctr", "key", "iv").string()
```

### DIFF
--- a/lib/bloblang/x/query/methods_strings.go
+++ b/lib/bloblang/x/query/methods_strings.go
@@ -2,6 +2,8 @@ package query
 
 import (
 	"bytes"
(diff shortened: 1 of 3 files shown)


Jeffail:
Thanks @eduardodbr, this is very cool.

PR #481, opened 2020-07-28
Azure Table storage batch output

eduardodbr:
Hi!

The table storage output (https://github.com/Jeffail/benthos/pull/480) only allowed for single message inserts so we decided to add the batch insert functionality to improve performance. Following the requirements defined [here](https://docs.microsoft.com/en-us/rest/api/storageservices/performing-entity-group-transactions#requirements-for-entity-group-transactions), the output groups the messages by table name and partition key and does batch inserts to every combination.
Please take a look.


Jeffail:
Thanks @eduardodbr this is very cool, I'll take a proper look this weekend but at a glance it looks good to me. @mfamador any chance you have time to take a look at this as well? I'm mostly interested in the higher level stuff as I'm not familiar with azure services.

mfamador:
Hi @Jeffail! I and @eduardodbr actually did this PR together, so from my side, it also looks good :) Thanks!

Jeffail:
@mfamador awesome, in that case thanks both! I'll aim to get this released this weekend.

PR #489, opened 2020-08-14
Add bloblang parse_timestamp_unix method

mfamador:
Hi!
Please have a look at this `bloblang` method to parse a timestamp from a string and return it as a unix epoch.
Thanks!


Jeffail:
Thanks @mfamador, looks great!

PR #493, opened 2020-09-05
feat: Add jq processor

Jeffail:
Hey @nilium, hope you don't mind but I saw this fork and figured it would be cool to add. I can take care of fixing the merge conflicts.


Jeffail:
Done, I've removed the slurp option for now and changed the behaviour when multiple values are emitted (they're now added to messages as an array) in order to help with error handling. Thanks @nilium for kicking this off, and thanks to @itchyny for the cool lib!

nilium:
@Jeffail I don't mind, cool to have another processor. I think I remember there being some issues with the slurp option as implemented in the fork that I never found a good answer for as well, so removing it seems like the best thing to do. Also not sure of the performance of the processor since I never benchmarked it and Bloblang seemed more promising (started the fork around the same time as its introduction).

Anyway, happy to take a look at anything you're not sure about in the code, though it's been a while since I last touched it. May be a few hours though, since I'm going to get some sleep for now.

PR #497, opened 2020-09-15
Enable RSS and Atom feed generation for blog

napkindrawing:
Howdy!

This Docusaurus config tweak enables RSS and Atom feeds for the blog.

I tried to subscribe to the blog to stay up-to-date with Benthos ... but I couldn't as there were no feeds! 😭

I verified locally that this change correctly generates `atom.xml` and `rss.xml` in `website/build/blog/`

Cheers,
Mike


Jeffail:
Thanks @napkindrawing! Shouldn't take long for the site to update and then the feed frenzy can begin.

PR #499, opened 2020-09-15
website/blog: fix link to outputs

markus-wa:


Jeffail:
Nice catch, thanks @markus-wa!

PR #504, opened 2020-09-25
new bloblang method parse_unix_timestamp

mfamador:
Hi!

I've created this bloblang method to convert an unix epoch to a timestamp.

Not sure about the naming, though, since I'm not really parsing the input number, it's converting it to the timestamp, basically it was an attempt to make it closer to the already existent `parse_timestamp_unix`.

Cheers!

### DIFF
--- a/internal/bloblang/query/methods.go
+++ b/internal/bloblang/query/methods.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/Jeffail/gabs/v2"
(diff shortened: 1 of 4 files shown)


Jeffail:
Hey @mfamador, I think this is already covered by the new `format_timestamp`: https://github.com/Jeffail/benthos/commit/6c33c232587fbd886635fb479965533afd9216d8, looks like we went down pretty much the same path 😛

mfamador:
Great! I like yours better 🤣 Let's close this one

PR #509, opened 2020-09-28
AMQP 1.0: disconnect and reconnect on unknown error

### LINKED ISSUE #507: AMQP 1.0 - connection reset by peer
Hi,

we're using Benthos in production for a week now, connecting to Azure Service Bus via amqp 1.0 input with the following configuration:

```
input:
  broker:
    inputs:
    - amqp_1:
        url: "...."
        source_address: "..."
    batching:
      count: 180
      period: 220s
```

it works well except for two times where we got the following error:

```
OUR_APP_NAME.input | Failed to read message: read tcp 172.20.0.3:38854->51.107.71.205:5671: read: connection reset by peer
```

Once this error occurs, it's then repeated every second until a reboot of the application is made.
Any idea of what could it be? Checked on Azure's side and there's no error there.. might be a server issue as well i guess... but all other services on the same machine are running fine 
We're still investigating as it's not reproducible, but maybe it rings a bell to you @Jeffail..?

ianaz:
As suggested by @Jeffail 

Fixes #507

Co-authored-by: Ashley Jeffs <ash@jeffail.uk>

### DIFF
--- a/lib/input/reader/amqp_1.go
+++ b/lib/input/reader/amqp_1.go
@@ -170,13 +170,13 @@ func (a *AMQP1) ReadWithContext(ctx context.Context) (types.Message, AsyncAckFn,
 		if err == amqp.ErrTimeout {
 			err = types.ErrTimeout
 		} else {
-			if dErr, isDetachError := err.(*amqp.DetachError); isDetachError {
-				if dErr.RemoteError != nil {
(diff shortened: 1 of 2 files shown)


Jeffail:
Awesome, thanks @ianaz!

PR #514, opened 2020-10-01
add connection string to azure blobstorage output

mfamador:
We can create a connection to the Azure Storage Account through a connection string.
The Azure API in this output is in preview and doesn't support connection string directly yet, so we need the extract the Storage Account Name and Access Key from the connection string. 
Using a connect string allows us to use the output on Azure Blob Storage emulators, like https://github.com/Azure/Azurite.


lib/output/blob_storage.go
```diff
@@ -19,13 +19,16 @@ Sends message parts as objects to an Azure Blob Storage Account container. Each
...
+The output can be authenticated through a connection string or a storage account name with 
+the respective access key. If both authentication methods are set, the connection string is used.
 In order to have a different path for each object you should use function
 interpolations described [here](/docs/configuration/interpolation#bloblang-queries), which are
 calculated per message of a batch.`,
 		Async: true,
 		FieldSpecs: docs.FieldSpecs{
 			docs.FieldCommon("storage_account", "The storage account to upload messages to."),
 			docs.FieldCommon("storage_access_key", "The storage account access key."),
+			docs.FieldCommon("storage_connection_string", "The storage account connection string."),
```
Jeffail, line 31: This description ought to specify that only one of `storage_connection_string` or `storage_account` + `storage_access_key` is required.

lib/output/writer/blob_storage.go
```diff
@@ -46,24 +48,47 @@ func NewAzureBlobStorage(
...
+
 	var credential azblob.Credential
-	if len(conf.StorageAccessKey) == 0 {
+	storageAccount := conf.StorageAccount
+	storageAccessKey := conf.StorageAccessKey
+	storageURLFormat := ""
+	if len(conf.StorageConnectionString) != 0 {
+		pcs := extractValues(conf.StorageConnectionString)
+		if sa := pcs["AccountName"]; sa == "" {
+			return nil, fmt.Errorf("invalid azure storage connection string")
```
Jeffail, line 59:
```suggestion
			return nil, errors.New("invalid azure storage connection string (missing AccountName)")
```

lib/output/writer/blob_storage.go
```diff
@@ -46,24 +48,47 @@ func NewAzureBlobStorage(
...
+		} else if sk := pcs["AccountKey"]; sk == "" {
+			return nil, fmt.Errorf("invalid azure storage connection string")
```
Jeffail, line 61:
```suggestion
			return nil, errors.New("invalid azure storage connection string (missing AccountKey)")
```

Jeffail (commented):
Looks good, just a couple of minor comments. https://github.com/Azure/Azurite looks great, if you have time it'd be awesome to try setting up an integration test https://github.com/Jeffail/benthos/tree/master/lib/test/integration, otherwise I can take a look when I get a chance.

lib/output/blob_storage.go
mfamador: I put it at the beginning of the output description, but perhaps it's not clear enough, I'll try to rephrase it.

Jeffail: I think the wording further up is fine, but I'd expect users to skip straight to the field docs quite often so it's probably best there instead.

mfamador: I add it now to the field docs

mfamador:
The https://github.com/Azure/Azurite is nice indeed, I'll try to have a look at the integration tests over the weekend.

mfamador:
Hi! I replaced `azure-storage-blob-go` library by `azure-SDK-for-go`, the same library already used by `table_storage` output, since `azure-storage-blob-go` is in preview (for a long time) and doesn't support connection string, which makes us to parse the connection string in some hacky ways.
Also added additional parameter do define the container's access level: private, blob or container.
Changed `blob_type` values to lowercase, It's a breaking change, but since this is a beta output, and I believe most of the people don't set this parameter anyway (defaults to `block`).

mfamador:
I'll try to add the integration tests, I already tested it with Azure and Azurite and seems to work fine.
Not sure how to run individual integration tests, though

Jeffail:
Looking good @mfamador, I don't think this needs to be blocked on integration tests, they're more of a nice-to-have, are these PR otherwise ready to merge?

mfamador:
Yeah, I think they're good to go, I already tested in Azure and Azurite. But I'll try to add the integration tests when I have the time, but probably it'd help to have the corresponding `inputs` as well, right?

Jeffail (approved):

PR #528, opened 2020-10-17
OAuth2 token processor

mfamador:
Hi!
I've created a processor to retrieve an OAuth2 token using the `client credentials` token flow.

We can use it within a `branch` processor, for example:
```
pipeline:
  processors:
  - branch:
      processors:
      - oauth2_token:
          endpoint: https://my-identiy-server.foo.com
          client_id: ${!json("client_id")}
          client_secret: ${CLIENT_SECRET}
          scopes:
          - my-scope
      result_map: |-
        token = this
  - bloblang: |
      root = this
      access_token = token.access_token
      token_type = token.token_type
      expiry = token.expiry
      ttl = token.expiry.parse_timestamp_unix()-timestamp_unix()
```
or getting the token marshalled in a metadata key:
```
pipeline:
  processors:
  - oauth2_token:
      endpoint: https://my-identiy-server.foo.com
      client_id: ${!json("client_id")}
      client_secret: ${CLIENT_SECRET}
      scopes:
      - my-scope
      metadata_key: token
  - bloblang: |
      root = this
      marshalled_token = meta("token")
```


Jeffail:
Hey @mfamador, looks interesting, what would be the typical use for this? Wondering if it could be something we bake into the https://www.benthos.dev/docs/components/processors/http#oauth2 functionality.

mfamador:
In our case, we're going to use this processor to retrieve the OAuth token that we need to use afterwards for authorization on a few other custom plugins calling gRPC services. We're going to combine it with the cache processor to store the token.

mfamador:
Regarding the `http` processor I'm not sure if we could use it in our use case, we just need to retrieve the OAuth2 token, we don't really want to perform an HTTP request.

mfamador:
I guess this one got stale :) Closing it.

PR #537, opened 2020-10-23
add cache per-key ttl

mfamador:
Hi!

I've tried to jump-start https://github.com/Jeffail/benthos/issues/531. I hope it will help.

It's now possible to set a per-key TTL for the resources which have that functionality, afaik, Ristretto, Redis and Memcache.
```
pipeline:
  processors:
  - branch:
      processors:
      - cache:
          resource: token_cache
          operator: get
          key: token
      - catch:
        - oauth2_token:
            endpoint: https://my-identity-server
            client_id: my-client-id
            client_secret: a-secret
            scopes:
            - my-scope
        - cache:
            resource: token_cache
            operator: set
            key: token
            ttl: '${! this.expiry.parse_timestamp_unix()-timestamp_unix() }'s
            value: '${! json() }'
      result_map: |
        token = this

resources:
  caches:
    token_cache:
      ristretto:
        ttl: 600s
```

Not sure if it's compliant to what you were thinking about the interfaces. 
I have some doubts in `Multilevel` cache:
https://github.com/anovateam/benthos/blob/22145b33de0e4e0f7c761babec290a08b4717c95/lib/cache/multilevel.go#L134
In `Multilevel` `Get`  methods, I'm not sure what to set in the per-key TTL since I can't determine it within a `Get`.

### DIFF
--- a/lib/cache/redis.go
+++ b/lib/cache/redis.go
@@ -212,7 +212,7 @@ func (r *Redis) SetWithTTL(key string, value []byte, ttl *time.Duration) error {
 		r.log.Errorf("Set command failed: %v\n", err)
 		<-time.After(r.retryPeriod)
 		r.mSetRetry.Incr(1)
-		err = r.client.Set(key, value, r.ttl).Err()
+		err = r.client.Set(key, value, t).Err()
 	}
 	if err != nil {
 		r.mSetFailed.Incr(1)
@@ -278,7 +278,7 @@ func (r *Redis) AddWithTTL(key string, value []byte, ttl *time.Duration) error {
 		r.log.Errorf("Add command failed: %v\n", err)
 		<-time.After(r.retryPeriod)
 		r.mAddRetry.Incr(1)
-		if set, err = r.client.SetNX(key, value, r.ttl).Result(); err == nil && !set {
+		if set, err = r.client.SetNX(key, value, t).Result(); err == nil && !set {
 			r.mAddFailedDupe.Incr(1)
 
 			latency := int64(time.Since(tStarted))
--- a/lib/cache/redis.go
+++ b/lib/cache/redis.go
@@ -195,13 +195,19 @@ func (r *Redis) Get(key string) ([]byte, error) {
 }
 
 // Set attempts to set the value of a key.
-func (r *Redis) Set(key string, value []byte) error {
+func (r *Redis) Set(key string, value []byte, ttl *time.Duration) error {
 	r.mSetCount.Incr(1)
 	tStarted := time.Now()
 
 	key = r.prefix + key
 
-	err := r.client.Set(key, value, r.ttl).Err()
+	var t time.Duration
+	if ttl != nil {
+		t = *ttl
+	} else {
+		t = r.ttl
+	}
+	err := r.client.Set(key, value, t).Err()
 	for i := 0; i < r.conf.Redis.Retries && err != nil; i++ {
 		r.log.Errorf("Set command failed: %v\n", err)
 		<-time.After(r.retryPeriod)
@@ -223,10 +229,10 @@ func (r *Redis) Set(key string, value []byte) error {
 
 // SetMulti attempts to set the value of multiple keys, returns an error if any
 // keys fail.
-func (r *Redis) SetMulti(items map[string][]byte) error {
+func (r *Redis) SetMulti(items map[string][]byte, t *time.Duration) error {
 	// TODO: Come back and optimise this.
 	for k, v := range items {
-		if err := r.Set(k, v); err != nil {
+		if err := r.Set(k, v, t); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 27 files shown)


mfamador:
I just realized that there are breaking changes if we're using the cache within a custom plugin. 
I'll try to fix it.

Jeffail:
Hey @mfamador, the problem comes from two sides, we need to be backwards compatible from the side of users of the `types.Cache` interface, so that interface needs to stay the same and still work with the existing caches. However, we also need to make sure that cache plugins that implement `types.Cache` will still work against the cache processor, where if the configuration specifies a per-item ttl and the cache plugin doesn't support these extended interfaces then we catch and log the error appropriately.

mfamador:
I was thinking of adding `AddWithTTL`, `SetWithTTL` and `SetMultiWithTTL` to the interface, and return an error in the ones that don't support it.
But that would still have breaking changes to the plugins implementing `types.Cache`, right?

Jeffail:
You can work around that until V4 by defining a separate interface and casting at the edge of where the methods are needed, a good example is how I expanded the `types.Manager` interface to support processors: https://github.com/Jeffail/benthos/blob/master/lib/processor/resource.go#L81, it's a bit of a headache but it gets the job done.

mfamador:
I'm pretty sure it still needs some work, but I think there are no breaking changes now.
When you have the time can you please have a look, @Jeffail ? Thanks

Jeffail:
Awesome, this looks good. I'm hoping to put a release candidate out this weekend so I might hold off on the merge until then as it's already quite a large release, but this looks ideal in terms of compatibility.

lib/cache/redis.go
```diff
@@ -194,14 +194,20 @@ func (r *Redis) Get(key string) ([]byte, error) {
...
 	key = r.prefix + key
 
-	err := r.client.Set(key, value, r.ttl).Err()
+	var t time.Duration
+	if ttl != nil {
+		t = *ttl
+	} else {
+		t = r.ttl
+	}
+	err := r.client.Set(key, value, t).Err()
```
Jeffail, line 210: The next call within the retry loop needs the provided ttl as well.

lib/cache/redis.go
```diff
@@ -221,27 +227,44 @@ func (r *Redis) Set(key string, value []byte) error {
...
 	key = r.prefix + key
 
-	set, err := r.client.SetNX(key, value, r.ttl).Result()
+	var t time.Duration
+	if ttl != nil {
+		t = *ttl
+	} else {
+		t = r.ttl
+	}
+	set, err := r.client.SetNX(key, value, t).Result()
```
Jeffail, line 267: Same with the next retry block.

lib/cache/ristretto.go
```diff
@@ -117,29 +117,61 @@ func (r *Ristretto) Get(key string) ([]byte, error) {
...
+func (r *Ristretto) SetMulti(items map[string][]byte) error {
+	return r.SetMultiWithTTL(items, nil)
+}
+
+// AddWithTTL attempts to set the value of a key only if the key does not already exist
+// and returns an error if the key already exists.
+func (r *Ristretto) AddWithTTL(key string, value []byte, ttl *time.Duration) error {
+	//if _, err := r.Get(key); err == nil {
+	//	return types.ErrKeyAlreadyExists
+	//}
```
Jeffail, line 167: Can remove this

lib/output/writer/cache.go
```diff
@@ -87,6 +94,15 @@ func (c *Cache) WriteWithContext(ctx context.Context, msg types.Message) error {
 // Write attempts to write message contents to a target Cache.
 func (c *Cache) Write(msg types.Message) error {
 	if msg.Len() == 1 {
+		var ttl *time.Duration
+		var err error
+		if ttls := c.ttl.String(0, msg); ttls != "" {
+			*ttl, err = time.ParseDuration(ttls)
```
Jeffail, line 100: I think this is an invalid pointer, it's also not used below.

Jeffail (changes requested):
v3.32.0 is out so this is no longer blocked, I've left a couple of comments, would be good to get unit tests for the cache processor and output to cover this stuff.

lib/cache/redis.go
mfamador: Done

lib/cache/redis.go
mfamador: Done

lib/output/writer/cache.go
mfamador: Oh, I forgot this one, I think it's done now.

mfamador:
All comments were addressed, I'll try to create the tests over the weekend.

mfamador:
Added a unit test for ristretto and an integration test for Redis, and it seems to work fine.
Also added an integration test for Memcache but it looks like per-key TTL it's not working as expected, probably I'm doing something wrong.

Jeffail (approved):
Looks good! The memcached test error is because of a pre-existing bug, I can fix that.

PR #540, opened 2020-10-29
add context aware client functions

zlozano:
I have a plugin that uses the client type, and in order to get some of my `http.RoundTripper` tooling to work, I require context propagation.

I've seen some comments where the message interface is defined, and noticed that this may be a feature coming in V4. I obviously don't want to interfere with that, so if there is a better way to go about this, I'm open to your input.


Jeffail:
Hey @zlozano, yeah this is fine to add. I'm slowly in the process of adding context to everything so this fits nicely. One thing we might want to consider later on is that messages themselves can have a context (acquired with [`message.GetContext`](https://github.com/Jeffail/benthos/blob/master/lib/message/part_with_context.go#L11)), which we could fall back to instead of `context.Background()`, but let's stick with this for now.

PR #541, opened 2020-10-30
add grpc client output.

brenol:
Ok, I left a few things open as I'm not sure on how to proceed:

- Should we use gRPC Retries instead of Benthos Retries? (Not sure how the project works with this one)
- Same question for backoffs
- There is a JSON input limitation. Is that OK?
- I'm not sure how I can set the response. Can you please assist me?

Also, if input has some 'incorrect data' that does not exist in the proto the response is slow. I'm not sure how I can improve that one yet but if you have any ideas, I'll be glad to implement them.

Also, I know there is no issue for this but I thought this would be a fun thing to implement (dynamic gRPC message creation!) and we might need it in our company. Also, adding it to Benthos was a breeze :)


Jeffail:
Hey @brenol, this is very cool!

Being strict about messages having a JSON structure is fine as users can add `try` brokers to capture bad data. In terms of retries I think using the internal retry mechanisms of the grpc lib is fine as long as it has a logical end _outside_ of context cancellation (something like a cap of retrying for X seconds), otherwise we might want to provide a context with our own timeout just in case.

For setting the response you can copy the way the `http_client` output does it: https://github.com/Jeffail/benthos/blob/master/lib/output/writer/http_client.go#L118.

It might be easier to start with to implement this as a processor similar to the http processor: https://github.com/Jeffail/benthos/blob/master/lib/processor/http.go, as processors are generally a little better for dealing with request/reply semantics, although an output is still useful.

Also a thing to keep an eye on is that [`protoreflect`](https://github.com/jhump/protoreflect/issues/301) (which is very cool) has a WIP branch that I believe would allow us to use `.proto` files as an alternative to `.pb` files. But I don't see any need to block this waiting for that.

brenol:
Hi @Jeffail,

thanks for your feedback. I saw http_client but I was sleepy and did not follow the whole chain of calls - I'll fix that.

About the timeouts/retries, I'll have a look at how gRPC handles those. 

Also, I really wanted to use .proto files instead of .pb files (did not even know those existed!) but I was happy it worked with .pb files :)

One more thing, (and this is a question) - I thought this would be a output (there are still things that I don't understand about Benthos!). Should we have a output AND a processor? (As we can reuse most code, I think that's OK?)

Oh, btw, I'll add tests, just wanted this to get this PR out of my way as there were things I had questions and I think the PR would be a nice place to evolve the implementation :)

Thank  you

Jeffail:
Hey @brenol, if we add a processor then yeah we can reuse most of the code, I would say start off with whichever you personally find more useful/interesting and then if you want to add the other afterwards then that's great otherwise I'd be happy to help out. 

> but I was sleepy and did not follow the whole chain of calls

No problem at all, Benthos is well versed in sleepy programming :stuck_out_tongue:, take your time and give me a ping whenever you want feedback or if you have any questions.

PR #543, opened 2020-11-04
Initial implementation of a Cassandra output.

codingconcepts:
Hi all,

Thanks to @adamwasila and @Jeffail, here's an implementation of a Cassandra sink that I've tested with the following configuration:

``` yaml
input:
  stdin: {}

output:
  cassandra:
    keyspace: ks
    table: tb
    password_authenticator:
      enabled: true
      password: cassandra
      username: cassandra
```


lib/output/cassandra.go
```diff
@@ -0,0 +1,33 @@
...
+
+//------------------------------------------------------------------------------
+
+// NewCassandra creates a new Cassandra output type.
+func NewCassandra(conf Config, mgr types.Manager, log log.Modular, stats metrics.Type) (Type, error) {
+	c, err := writer.NewCassandra(conf.Cassandra, log, stats)
+	if err != nil {
+		return nil, err
+	}
+	return NewWriter(
```
Jeffail, line 28: If we use `NewAsyncWriter` here instead then we can replace the `async` parameter with `max_in_flight` similar to other outputs like nats: https://github.com/Jeffail/benthos/blob/master/lib/output/nats_stream.go#L43, although you don't need to bother with the check for `1` just always create a `NewAsyncWriter` with whatever the field is set to.

lib/output/writer/cassandra.go
```diff
@@ -0,0 +1,253 @@
...
+		msg.Iter(func(i int, p types.Part) error {
+			g.Go(func() error {
+				return c.writePart(session, p)
+			})
+			return nil
+		})
+		return g.Wait()
+	}
+
+	return msg.Iter(func(i int, p types.Part) error {
```
Jeffail, line 171: We can probably speed things up here when there are >1 messages by using https://godoc.org/github.com/gocql/gocql#Batch

Jeffail (commented):
Hey @codingconcepts, looks great. I've left two comments but I'm happy to quickly do those myself this week. It would be good if you could get started on the documentation just with a general description of what the component is doing and a note that we only support JSON documents currently.

I've also just pushed an integration test suite for cache based outputs which should be useful here, but I can add those myself once this is merged.

codingconcepts:
Hey @Jeffail, I've added the documentation as requested, good shout!

Jeffail:
Thanks, this is next on my list, if you don't mind I might make a few changes here and there while I merge it in. I've just added a new sql output (https://github.com/Jeffail/benthos/blob/master/lib/output/sql.go) and I think it might make sense to open the configuration fields up a little bit here to be similar, so if you want you can specify your own query and what the parameters are.

codingconcepts:
Hey @Jeffail, not at all, you know the code better than anyone, so do what you think is best!

PR #544, opened 2020-11-09
Add .tgz extension to inference of input codec

nicktelford:
`.tgz` is a common contraction for `.tar.gz`, and should be inferred as a `tar-gzip` file.


Jeffail:
Thanks @nicktelford!

PR #557, opened 2020-11-17
Fixed float and double marshalling for CQL

codingconcepts:
Was previously attempting to insert floats as integers.

### DIFF
--- a/lib/output/cassandra.go
+++ b/lib/output/cassandra.go
@@ -329,17 +329,17 @@ func (s stringValue) MarshalCQL(info gocql.TypeInfo) ([]byte, error) {
 			return []byte{0}, nil
 		}
 	case gocql.TypeFloat:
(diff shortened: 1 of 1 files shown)


codingconcepts:
Hey @Jeffail,

A small fix to some float/double parsing issues in the Cassandra output for you!


Cheers,

Rob

Jeffail:
Thanks @codingconcepts! Is this issue blocking you? I'll probably put out a patch release this week but I can bring that forward if you need.

codingconcepts:
Hey @Jeffail, thanks for merging in.

Unfortunately, this is quite blocking and is in the last part of the stream processing architecture I'm building, so annoying close to it being ready to show :D

If you wouldn't mind cutting a patch release/Docker image, that would be brilliant.


Thanks again,

Rob

PR #559, opened 2020-11-18
Update zmq4_config.go

CodingJzy:
hello，I haven't come into contact with zmq4 much. I tried it for the first time today and even benthos was the first to use it. Then tried some publish and subscribe models. I found out that benthos can't be subscribed, but the small demo can be subscribed. Reading the source code of benthos, I found that there is no default filter set. The yaml file that causes benthos create zmq4 to default to []string{}
image

### DIFF
--- a/lib/input/reader/zmq4_config.go
+++ b/lib/input/reader/zmq4_config.go
@@ -18,7 +18,7 @@ func NewZMQ4Config() *ZMQ4Config {
 		URLs:          []string{"tcp://localhost:5555"},
 		Bind:          false,
 		SocketType:    "PULL",
-		SubFilters:    []string{},
+		SubFilters:    []string{""},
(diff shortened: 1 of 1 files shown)


Jeffail:
Hey @CodingJzy, thanks for raising this. Rather than adding a default empty string I think it might be more intuitive to simply subscribe to the empty string when the filters array is empty. However, this would also technically be a breaking change, as odd as it might sound there could be users relying on the current empty array == no subscriptions behaviour. As a temporary alternative I'm going to return an error when the array is empty on a SUB socket, and log a message describing how to resolve the situation.

However, on the next major version this is probably something we can look at doing.

Jeffail:
Temporary solution message fix: https://github.com/Jeffail/benthos/commit/348990221aaa4f0e271d6c6033285ac82c251168

PR #561, opened 2020-11-18
add：connect nats userinfo option

CodingJzy:
* When the input type is nats, support username and password authentication

### DIFF
--- a/lib/input/reader/nats.go
+++ b/lib/input/reader/nats.go
@@ -19,7 +19,7 @@ import (
 // NATSConfig contains configuration fields for the NATS input type.
 type NATSConfig struct {
 	URLs          []string `json:"urls" yaml:"urls"`
-	UserName      string   `json:"user_name" yaml:"username"`
+	Username      string   `json:"username" yaml:"username"`
 	Password      string   `json:"password" yaml:"password"`
 	Subject       string   `json:"subject" yaml:"subject"`
 	QueueID       string   `json:"queue" yaml:"queue"`
--- a/lib/input/reader/nats.go
+++ b/lib/input/reader/nats.go
@@ -91,7 +91,7 @@ func (n *NATS) ConnectWithContext(ctx context.Context) error {
 
 	opts := []nats.Option{}
 	if n.conf.Username != "" || n.conf.Password != "" {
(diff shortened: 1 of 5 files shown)


lib/input/reader/nats.go
```diff
@@ -19,6 +19,8 @@ import (
 // NATSConfig contains configuration fields for the NATS input type.
 type NATSConfig struct {
 	URLs          []string `json:"urls" yaml:"urls"`
+	UserName      string   `json:"user_name" yaml:"username"`
```
Jeffail, line 22:
```suggestion
	Username      string   `json:"username" yaml:"username"`
```

lib/input/reader/nats.go
```diff
@@ -87,7 +89,12 @@ func (n *NATS) ConnectWithContext(ctx context.Context) error {
 	var natsSub *nats.Subscription
 	var err error
 
-	if natsConn, err = nats.Connect(n.urls); err != nil {
+	opts := []nats.Option{}
+	if n.conf.UserName != "" || n.conf.Password != "" {
```
Jeffail, line 93:
```suggestion
	if n.conf.Username != "" || n.conf.Password != "" {
```

Jeffail (changes requested):
Hey @CodingJzy, looking good. Added a couple of comments just to make the field names more consistent. If we're adding auth fields to one nats component then I'd also want to make sure we add them to all others (the `nats` output and also the `nats_steam` input and output).

You can also delete `nats_test.go` as the integration tests are here: https://github.com/Jeffail/benthos/blob/master/lib/test/integration/nats_test.go

CodingJzy:
Ok ! thanks for your guidance

CodingJzy:
@Jeffail  hey :grin:  In fact. I found that nats and nats-streaming officially support url semantics. e.g. nats://root:pass@localhost:4222

Jeffail:
> @Jeffail hey grin In fact. I found that nats and nats-streaming officially support url semantics. e.g. nats://root:pass@localhost:4222

Ahhh, in that case it's probably cleaner to document that syntax rather than add more fields, I've added examples with: https://github.com/Jeffail/benthos/commit/64ad8d480b9eb99e458ed543375b84c63c8211ea

Closing this PR for now but thanks for raising it and for the investigation.

PR #564, opened 2020-11-21
added the prepared statement syntax for postges

cuotos:
This tripped me up for way too long before I realised there was a different in syntax.


Jeffail:
Thanks @cuotos, sorry you got caught by this, you have a place in Valhalla reserved for coming back to improve the docs :heart:

PR #565, opened 2020-11-22
Fix missing content type in streams api

FabianTe:
Hi,

just a small change but this will improve the lives of devs working with the streams API.

Adding the JSON content type to the given routes enables browsers or Postman to display the server response in a structured format rather as raw text. This is also technically "more correct" 😉

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ All notable changes to this project will be documented in this file.
 
 - New field `retry_as_batch` added to the `kafka` output to assist in ensuring message ordering through retries.
 
+### Fixed
+
+- Add `Content-Type` headers in streams API responses
+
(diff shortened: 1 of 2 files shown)


FabianTe:
Uhm, that are a lot of errors. I tried testing locally but I get *a lot* of different errors that are surely not related to this change. Could someone give me a pointer.

Jeffail:
Thanks @FabianTe! PR looks good, the fail looks like a race condition somewhere in the tests so I've restarted it but your changes look fine, if the tests fail again I'll take care of it

FabianTe:
Alright, thank you for reviewing!

I'll hopefully add this for further API routes. I actually wanted to test some functions while I noticed this. Expect more to come up some time 😉

PR #578, opened 2020-12-02
Prometheus basic auth

razor-x:
For https://github.com/Jeffail/benthos/issues/577

1. Should be everything needed but let me know if I missed anything.
2. Not sure how to test this. I am assuming that all that is needed to to call `pusher.BasicAuth` (this mutates the pusher right, we don't need the new ref?)
3. Only thing I wasn't really sure about is if we need anything extra around the err handling or if this makes sense as-is.
4. Please squash commits.


Jeffail:
Hey @razor-x, looks good! You should be fine with `pusher.BasicAuth` as `pusher` is a pointer type so it'll be mutated. I think the error handling (just printing logs) is fine for now, but ideally we'd create the pusher outside of the goroutine and try out an initial push just to test the connection, if you fancy adding that then go ahead, otherwise I think this is good. For testing we unfortunately don't really have any integration suite for metrics so the best we can do is try it out locally for now.

razor-x:
I thought about if the pusher needs to be created outside that scope. Even so maybe better to put that in its own PR. I don't have much Go experience so this is about as far as I can get without learning something new. Normally would be fine with that haha but since this particular section of the code is impossible for me to test I'm not sure it's the best spot for me to experiment.

Jeffail:
Cool, I'm testing it with a local set up and it needs a bit of work but I can get this over the finish line.

PR #585, opened 2020-12-09
Adding https to influxdb metrics output

nwest1:
Missed this option.. 🐑


Jeffail:
Nice, looks great, thanks @nwest1!

Issue #505, opened 2020-09-28
Add cron schedule support to the `bloblang` input


Jeffail:
The [bloblang input](https://www.benthos.dev/docs/components/inputs/bloblang) currently supports an interval field, allowing you to generate data and trigger pipelines at a given time interval. However, it would be nice to expand this to support proper cron scheduling, allowing users to either specify a static `interval`, a `cron` schedule, or neither (generate as fast as we can).

The name of this input is also not particularly useful as users are more likely to be searching either for a scheduled input or a generated input type, not necessarily a bloblang mapping input type. We should consider renaming it to something like `generate`.

Edit: Useful library giving us most of what we need: https://godoc.org/github.com/robfig/cron#hdr-Alternative_Formats

Jeffail:
Once this is done we should also consider expanding the cron input concept by allowing users to specify a cron schedule for any child input, similar to how the `read_until` input works. The main challenge with this is that most inputs aren't designed to indefinitely block to a schedule and we might find that many inputs lose their connections during the waiting period.

dilipkk-foyernet:
@Jeffail currently bloblang **interval** is kind blocks the ReadWithContext, it should ok for smaller intervals but if the interval is bigger then it gets blocked several hours right? I am checking this issue to see if I can contribute to this so I want to know if understanding right or not.

Jeffail:
Hey @dilipkk-foyernet, yeah exactly, blocking `ReadWithContext` for a long time (hours, even days) is fine as long as you return `types.ErrTimeout` when `case <-ctx.Done():` unblocks.

dilipkk-foyernet:
When we implement cron based schedule there shouldn't be schedules missed due to process restart etc otherwise it will be incomplete. I guess that would be a pretty tough problem right? we need to maintain state/history or some kind of distributed scheduler is needed with shared configuration and bookkeeping so that there is no single point of failure and then there is less chance of missing schedules. @Jeffail what do you think about this?

Jeffail:
Hey @dilipkk-foyernet, for that we'd need coordination between benthos instances which I would say is out of scope. It's still possible to add resiliency with the scheduling by adding a queue system in between, so multiple redundant benthos instances writing cron events into something like kafka and then a layer of benthos consumers reading the topic with deduplication of the redundant trigger events.

dilipkk-foyernet:
Thanks for your response. I will look into it and get back to you as soon
as possible.

Jeffail:
[...]

PR #586, opened 2020-12-10
Add cron expression to bloblang input params

dilipkk-foyernet:
Hi @Jeffail,

I have added support for cron in bloblang input as discussed in [#505](https://github.com/Jeffail/benthos/issues/505).

This is my first PR for this repo so let me know if I have to do any else.

Thanks,
Dilip


lib/input/bloblang.go
```diff
@@ -123,10 +146,40 @@ func newBloblang(conf BloblangConfig) (*Bloblang, error) {
 		exec:        exec,
 		remaining:   remaining,
 		timer:       timer,
+		schedule:    schedule,
+		location:    location,
 		firstIsFree: true,
```
Jeffail, line 151: When set `true` this field ensures that a message is generated immediately regardless of the timer, but in the case where we have a cron schedule I wouldn't expect this so we should probably set it to `false` when `schedule` is non-nil.

Jeffail (commented):
Looking good @dilipkk-foyernet! I've left one comment around `firstIsFree`.

I regret making `interval: 1s` the default as it'll force users to explicitly set `interval: ""` here if they want cron expressions. I'll think this one over as we have the ability to swap the default without adding a breaking change but it might be better to identify a cron expression in the `interval` field and remove the `cron_expression` field. Leave it with me.

dilipkk-foyernet:
> Looking good @dilipkk-foyernet! I've left one comment around `firstIsFree`.
> 
> I regret making `interval: 1s` the default as it'll force users to explicitly set `interval: ""` here if they want cron expressions. I'll think this one over as we have the ability to swap the default without adding a breaking change but it might be better to identify a cron expression in the `interval` field and remove the `cron_expression` field. Leave it with me.

I have done this change. @Jeffail

Jeffail:
Awesome, thanks @dilipkk-foyernet!

Issue #588, opened 2020-12-11
Convert redis client to UniversalClient to support simple, cluster, and sentinel deployments


nwest1:
I was thinking about adding support for https://pkg.go.dev/github.com/go-redis/redis/v7#UniversalClient and converting the existing redis clients to use that. 

Without digging too deeply, I believe this would be backward-compatible. With some optional parameters added to the config, could enable cluster or sentinel support. The single address -> multiple could be handled a few different ways: single string, comma-separated urls, separate list of urls - whatever is most aligned with existing components. 

Their code looks pretty simple (https://github.com/go-redis/redis/blob/master/universal.go#L199-L206) and looks to be a wrapper around the same client type that benthos is currently using.
```go
func NewUniversalClient(opts *UniversalOptions) UniversalClient {
	if opts.MasterName != "" {
		return NewFailoverClient(opts.Failover())
	} else if len(opts.Addrs) > 1 {
		return NewClusterClient(opts.Cluster())
	}
	return NewClient(opts.Simple())
}
```

Any thoughts/opinions/experience along these lines would be great to have!

Jeffail:
Hey @nwest1, if it's backwards compatible then this looks like it'd be a great change. Since we only have a single `url` field on all of the redis components we could start off by parsing out a comma separated list, but in future if we wanted it to work similarly to other components we'd have a `urls` array and _also_ allow individual elements of the `urls` field to be a comma separated list (which is less readable but works better with environment variables).

nwest1:
Did a little bit more digging here - it looks as though there are some Redis configurations that may not be as straightforward: 
https://github.com/go-redis/redis/issues/1488

In light of that proposal, I'm thinking to accommodate that situation without worrying about changes in the go-redis library and expect the user to know what kind of cluster they're connecting to. 

I'm not sure the best way to define some forward-compatible config, but here's a shot - correct me if I misunderstood. 

```yaml
redis:
  url: tcp://localhost:6379  ## or "tcp://localhost:6379,tcp://localhost:6380"
  urls:  ## new optional field
    - tcp://localhost:6379
    - tcp://localhost:6380
  kind: [simple|cluster|failover] ## default simple
  tls:
    enabled: false
    skip_cert_verify: false
    root_cas_file: ""
    client_certs: []
  operator: scard
  key: ""
  retries: 3
  retry_period: 500ms
  parts: []
```

Jeffail:
I'll aim to cut a release this week :heart:

[...]

PR #594, opened 2020-12-15
Convert redis.Client to redis.UniversalClient

nwest1:
I think I've taken into account the existing possible url structures and this satisfies. 

I'm attempting at getting this integration test to work properly on osx, but I'm guessing there's something I'm missing from docker networking on mac. 

The cluster creates correctly, but nodes aren't connectable from the unit test. Maybe it works in the CI server?

### DIFF
--- a/internal/service/redis/config.go
+++ b/internal/service/redis/config.go
@@ -4,7 +4,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net/url"
-	"strconv"
+	"strings"
 
 	"github.com/Jeffail/benthos/v3/internal/docs"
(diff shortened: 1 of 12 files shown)


nwest1:
related #588

nwest1:
Hey! I was able to get the integration tests working locally, but I'm unsure of how exposing ports plays with the CI pipeline. 

The original redis integration tests passed locally for me, so I'm fairly confident in the conversion to the new client. 

Let me know if you'd like any other changes. Thanks!!

Jeffail:
Thanks @nwest1, this looks great! Trying to get the docker tests to work with networked contains is total a pain so thanks for getting that done. I haven't set the CI up to run them yet so nothing to worry about there (for now).

PR #595, opened 2020-12-17
Binary safe subprocess processor

stergiotis:
This PR extends the existing subprocess processor to enable binary safe, efficient communication with os processes.

The config is extended with two new params `format_send` (affecting `stdin` of the subprocess) and `format_recv` (affecting `stdout` of the subprocess) with allowed values
- `line`: existing behavior,
- `binary`: a uint32 big-endian length prefixed format (as it is used in the archive/unarchive processor),
- `netstring`: D. J. Bernstein's netstring format (see https://cr.yp.to/proto/netstrings.txt).

The handling of `stderr` remains unchanged.

I find benthos documentation wonderfully consistent in style. The assumption that communication with a subprocess is line-based seems to root deep. Therefore I did not edit the documentation and would suggest that @Jeffail would update it himself.

### DIFF
--- a/lib/processor/subprocess.go
+++ b/lib/processor/subprocess.go
@@ -204,21 +204,22 @@ func newSubprocWrapper(name string, args []string, maxBuf int, formatRecv string
 var maxInt = (1<<bits.UintSize)/2 - 1
 
 func binarySplitFunc(data []byte, atEOF bool) (advance int, token []byte, err error) {
+	const prefixBytes int = 4
 	if atEOF {
 		return 0, nil, nil
 	}
-	if len(data) < 8 {
+	if len(data) < prefixBytes {
 		// request more data
 		return 0, nil, nil
 	}
 	l := binary.BigEndian.Uint32(data)
-	if l > uint32(maxInt)-8 {
+	if l > (uint32(maxInt)-uint32(prefixBytes)) {
 		return 0, nil, errors.New("number of bytes to read exceeds representable range of go int datatype")
 	}
 	bytesToRead := int(l)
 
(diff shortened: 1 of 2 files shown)


Jeffail:
Hey @stergiotis, this is great. I think I'd rename the fields to use the word codec rather than format as it's more consistent with other components. Which of these formats are you planning to use? Asking because I think it might be worth omitting the `binary` type if you're not looking to use it as it's a pain to document (and understand).

stergiotis:
Hey @Jeffail, many thanks for looking into this! I am currently using the `binary` codec. In my opinion, `binary` has higher performance potential and is easier to implement than `netstring`. I added `netstring` as it is much easier to read when used with text based formats. A technical accurate term for `binary` would be `lengthPrefixedUInt32BE`.

Jeffail:
Yeah that makes sense, I think I'd prefer calling it `length_prefixed_uint32_be` (I'm team snake case) as I've seen `binary` cause a lot of confusion in the past.

Jeffail:
Hey @stergiotis, looks good. I'm in a rush to get a release out this evening so I'll pick this up after the holidays. It might be nice to add the same codecs to the subprocess input and output as well eventually but this PR doesn't need to include that.

Jeffail:
Thanks @stergiotis, not sure when I'll put the next release out, but it'll probably be next week.

stergiotis:
Thanks, @Jeffail it has been a pleasure to make this (tiny) contribution to benthos! I will see how this holds up in practice and may return with a pull-request to obtain feature parity between subprocess input, output, and the processor.

PR #596, opened 2020-12-17
tcp -> redis for url examples

nwest1:
This one last thing was bothering me - example url changes to conform with how the go-redis library expects them to be. 

They still remain backward compatible with the parsing changes I added as the `tcp` is swapped out for `redis`


Jeffail:
Thanks @nwest1, looks good.

PR #597, opened 2020-12-18
Add Azure Blob Storage input

### LINKED ISSUE #582: Azure Blob Storage input
Thank you for this awesome project! I have a use case where I need Benthos to have the Azure Blob Storage input, but, for now, it only seems to have the output, which was added in #466 and refined in #514.

As @mfamador mentioned [here](https://github.com/Jeffail/benthos/pull/514#issuecomment-703555640), this would also be beneficial for having integration tests (using [Azurite](https://github.com/Azure/Azurite)).

I am thinking of writing the code to add this input and upstreaming the changes in a PR, if you're interested in having such a contribution.

Would that be of interest to you?

mihaitodor:
Fixes #582.

As promised, here is the code for adding the Azure Blob Storage input. I used the [`aws_s3`](https://github.com/Jeffail/benthos/blob/master/lib/input/aws_s3.go) input as a template, but, for now, without the Azure Event Grid + Storage Queue notification functionality. We will probably need this addition in the near future, so I'll let you know if I'll start working on it.

I noticed that @mfamador used the [azure-sdk-for-go](https://github.com/Azure/azure-sdk-for-go) framework instead of the newer [azure-storage-blob-go](https://github.com/Azure/azure-storage-blob-go) as mentioned [here](https://github.com/Jeffail/benthos/pull/514#issuecomment-703099899) so I used the same one. Unfortunately, this older framework doesn't support contexts and has some strange semantics with regards to timeouts. For example, I noticed that @mfamador's output currently ignores the `timeout` config parameter (should we document this?). Not sure if this parameter going to be very useful in practice for the current implementation, but when setting it to something != 0 for the input in this PR, the server closes the connection after the timeout expires. azure-sdk-for-go doesn't return a custom error message, though, so all we get is a "read: connection reset by peer" error. I tried to add the same functionality to the output, but it looks like azure-sdk-for-go does some [exponential retries](https://github.com/Azure/azure-sdk-for-go/blob/928b9cbac68670876b583465f789bd3ba655777c/storage/client.go#L124) when getting error 500 from Azure, which is what it returns when a timeout happens. I guess this is why it wasn't hooked up in the first place.

I also added a roundtrip integration test for Azure Blob Storage using [Azurite](https://github.com/Azure/Azurite). Unfortunately, it requires a nasty hack due to [this bug](https://github.com/Azure/Azurite/issues/663). Since I had to do that http Transport hack, I decided that it would also be useful to have Azurite bind to a random port, rather than port 10000, but that requires an extra http Transport hack because azure-sdk-for-go [ignores the port value from the connection string](https://github.com/Azure/azure-sdk-for-go/blob/928b9cbac68670876b583465f789bd3ba655777c/storage/client.go#L466) if I use the default account. If I add a new account to Azurite (via `AZURITE_ACCOUNTS`), then azure-sdk-for-go [doesn't support](https://github.com/Azure/azure-sdk-for-go/blob/928b9cbac68670876b583465f789bd3ba655777c/storage/client.go#L473) using `localhost` 😕


lib/input/blob_storage.go
```diff
@@ -0,0 +1,449 @@
...
+func NewAzureBlobStorageConfig() AzureBlobStorageConfig {
+	return AzureBlobStorageConfig{
+		Codec:   "all-bytes",
+		Timeout: "0s",
+	}
+}
+
+//------------------------------------------------------------------------------
+
+type azureObjectTarget struct {
```
mihaitodor, line 121: Should I rename `objectTarget` to `s3ObjectTarget` in [`aws_s3.go`](https://github.com/Jeffail/benthos/blob/2955a506cc48cf2afafdddea4eac4eb8ea8e9d5d/lib/input/aws_s3.go#L168)?

lib/input/blob_storage.go
```diff
@@ -0,0 +1,449 @@
...
+	return object, nil
+}
+
+func blobStorageMsgFromPart(p *azurePendingObject, part types.Part) types.Message {
+	msg := message.New(nil)
+	msg.Append(part)
+
+	meta := msg.Get(0).Metadata()
+
+	meta.Set("blob_storage_key", p.target.key)
```
mihaitodor, line 365: Would adding `meta.Set("blob_storage_filename", path.Base(p.target.key))` be of any use?

lib/input/blob_storage.go
```diff
@@ -0,0 +1,449 @@
...
+	if p.obj.Container != nil {
+		meta.Set("blob_storage_container", p.obj.Container.Name)
+	}
+	meta.Set("blob_storage_last_modified", time.Time(p.obj.Properties.LastModified).Format(time.RFC3339))
+	meta.Set("blob_storage_last_modified_unix", strconv.FormatInt(time.Time(p.obj.Properties.LastModified).Unix(), 10))
+	meta.Set("blob_storage_content_type", p.obj.Properties.ContentType)
+	meta.Set("blob_storage_content_encoding", p.obj.Properties.ContentEncoding)
+
+	for k, v := range p.obj.Metadata {
+		meta.Set(k, v)
```
mihaitodor, line 375: I am a bit concerned that a user might define custom metadata with key names identical to the ones we used above, in which case we'd override them. I saw the same pattern [here](https://github.com/Jeffail/benthos/blob/2955a506cc48cf2afafdddea4eac4eb8ea8e9d5d/lib/input/aws_s3.go#L669), so I copied it over, but I wonder if it might be worth namespacing these somehow.

lib/input/blob_storage.go
Jeffail: Please, that would make things a bit clearer.

lib/input/blob_storage.go
Jeffail: Nah, it's pretty easy to rename metadata fields with bloblang if anyone needs.

lib/input/blob_storage.go
Jeffail: I think this is probably fine, it means in a world where someone goes from blob storage -> something else -> blob storage -> something else again then the metadata would (depending on the "something elses") match the very first blob storage, which I think is probably a more useful default.

If in future this becomes a major problem for someone we can always add an optional `metadata_prefix` config field that would rename these keys, but as a user I would expect the default to be that my pre-existing metadata keys are unchanged.

Jeffail:
Hey @mihaitodor, this looks great!

Sounds like you navigated some rough seas trying to get the integration tests to work so kudos for coming out unscathed. It's definitely hacky but also looks very clean so I'm happy to add it in, since it has the potential to stink up other tests I think it'd be sensible to disable the test by default (happy to just slap a `t.Skip()` in there for now, or add a build tag for it).

If the timeout is ignored in the output level then let's remove the fields, they're more of a quality-of-life feature for speeding up shutdowns when the output is unresponsive, but it has a fallback and it's an edge case so I think we're okay without. When the new client lib is stable (assuming it has feature parity) we can swap it out and lean on contexts and do it the proper way.

mihaitodor:
@Jeffail Thanks for the feedback! I believe I addressed all your comments:
- Removed the timeout field in both the input and the output
- Skip Azure Blob Storage integration test
- Namespace objects in aws_s3 and blob_storage inputs

Please let me know if you spot more issues.

Jeffail:
Looks great, thanks @mihaitodor, just fyi I'm in the process of namespacing all of the aws components so that things like `s3` become `aws_s3` etc. I think it makes sense to do the same with the azure components so I'm going to rename this to `azure_blob_storage`.

mihaitodor:
Awesome, thank you!

PR #598, opened 2020-12-21
Add Azure Queue Storage output

mfamador:
Hi!
I've created this Azure Queue Storage output. 
On this one, I had to use the "new" specific Azure Queue SDK go lib since the generic one used in blob and table storage doesn't implement the message enqueuing. 

We can use interpolation for the `queue_name` and for the `ttl` fields. If `ttl` is not defined the messages won't have any retention period.
Ex:
```
output:
  queue_storage:
    storage_connection_string: "UseDevelopmentStorage=true;"
    ttl: '${! json("ttl") }'
    queue_name: '${! json("queue") }'
```
The Azure Queue Storage `queue` is created if it does not exist.
I'll try to have another PR soon with corresponding integration tests and Azure Queue Storage input as well.


Jeffail:
Hey @mfamador, is your implementation of parsing of the connection string here the main reason that we went with the older client library before? If so then given the issues with timeout awkwardness we're seeing I think it might be worth using the new library in all the azure components since we're parsing this out anyway. What do you think?

mfamador:
Hi @Jeffail, yes, at least for the Blob Storage input/output it might work, not sure if everything is implemented in the new SDK, like access level and so on, but I'll check. If I remember correctly the connection string parsing was the main reason for changing it back to the "old" SDK. 
Regarding Table Storage, I don't think there's a specific implementation for it, I guess we'll have to use the generic SDK for this one, AFAIK.

Jeffail:
Ah okay, in that case if we need to keep the old lib around anyway let's leave the existing components as-is for now. I've decided it's probably best to keep all the azure components prefixed with `azure_` as I'm (gradually) doing the same with AWS components, I've already updated the existing components: https://github.com/Jeffail/benthos/commit/0ff54f67463de7516afb25307785684b23595693, could you update this new one to match?

The old names (`blob_storage` and `table_storage`) can still be used for backwards compat.

mfamador:
The prefix `azure_` is a good idea. I've changed in this one too.

Jeffail:
Thanks @mfamador looks good!

PR #602, opened 2020-12-23
Add support for snappy in compress processor

dilipkk-foyernet:
Implements: https://github.com/Jeffail/benthos/issues/601


Jeffail:
Looks good, thanks @dilipkk-foyernet!

PR #605, opened 2021-01-01
add go.sum back

chenrui333:
followup of f33bad1c141ad0f4d357a2daa8626166379f7405

in fact, go 1.16 would make go modules `on` by default.


Jeffail:
Thanks @chenrui333, seems it's about the right time. Sorry for the unrelated test fail, I've fixed that separately.

chenrui333:
yeah, it is about the time, thanks for the merge! 👍

PR #608, opened 2021-01-06
Add example of how to send data to streams endpoint

patrobinson:
The http endpoint is namespaced, but this is not clear from the documentation.


Jeffail:
Looks good, thanks @patrobinson

PR #610, opened 2021-01-07
#609 Update documentation to use Go 1.15

### LINKED ISSUE #609: Compilation error due to time.Reset()
`ticker.Reset()` have been introduced only in 1.15 and Benthos makes use of this.

Building with previous versions fails with `b.timer.Reset undefined (type *time.Ticker has no field or method Reset)`

ianaz:
Should close #609

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ It is possible to select fields inside a configuration file to be set via [envir
 
 ## Build
 
-Build with Go (1.11 or later):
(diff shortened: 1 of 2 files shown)


Jeffail:
Thanks @ianaz, sorry that this impacted you. I'm assuming you're unblocked now, but if you need compatibility with 1.14 then let me know as I think it should be easy to get it building again as the features we're using from 1.15 are very minor.

PR #613, opened 2021-01-12
tracing - expose jaeger collector endpoint

nwest1:
- Allows user to specify `jaeger_collector` endpoint that overrides `agent_address` (based on client functionality)
- Adds docker-compose with jaeger ports exposed that work with current benthos configurations (agent and collector)

### DIFF
--- a/lib/tracer/jaeger.go
+++ b/lib/tracer/jaeger.go
@@ -127,13 +127,13 @@ func NewJaeger(config Config, opts ...func(Type)) (Type, error) {
 		cfg.Reporter = reporterConf
 	}
 
-	if i := config.Jaeger.CollectorAddress; len(i) > 0 {
-		reporterConf.CollectorEndpoint = i
-	}
-
 	if i := config.Jaeger.AgentAddress; len(i) > 0 {
 		reporterConf.LocalAgentHostPort = i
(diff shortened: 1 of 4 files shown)


lib/tracer/jaeger.go
```diff
@@ -20,7 +20,9 @@ func init() {
 		Summary: `
 Send spans to a [Jaeger](https://www.jaegertracing.io/) agent.`,
 		FieldSpecs: docs.FieldSpecs{
-			docs.FieldCommon("agent_address", "The address of a Jaeger agent to send tracing events to."),
+			docs.FieldCommon("agent_address", "The address of a Jaeger agent to send tracing events to.", "jaeger-agent:6831"),
+			docs.FieldCommon("collector_address", "The address of a Jaeger collector to send tracing events to. If set, this will override `agent_address`.",
```
Jeffail, line 24: Looks like this field accepts full URLs so it should be called `collector_url` in order to fit in with other Benthos fields.

Jeffail (commented):
Hey @nwest1, looks good, left one comment as I think the field name ought to be changed.

nwest1:
Thanks again @Jeffail !

Jeffail:
Looks good, nice one @nwest1!

PR #614, opened 2021-01-12
Add SAS token support for Azure Blob Storage

mihaitodor:
Hey @Jeffail, Azure Storage supports this extra SAS token authentication mechanism (details [here](https://docs.microsoft.com/en-us/rest/api/storageservices/create-account-sas)) which we are using in production. It can be passed in as a connection string, but it's more straightforward to support it directly.

I tested both the input and output with one of our accounts and these changes seem to work correctly.


Jeffail:
Awesome, looks good, thanks @mihaitodor!

PR #617, opened 2021-01-13
Adds md5 hash algorithm

cludden:


Jeffail:
Looks great, thanks @cludden!

PR #620, opened 2021-01-14
Allow Prometheus PushGateway to work if PushInterval not set

### LINKED ISSUE #619: Prometheus Push Gateway configuration
I was trying to configure Benthos to use the Prometheus Push Gateway and, according to [the docs](https://www.benthos.dev/docs/components/metrics/prometheus#push-gateway), I need to set the `push_url` field:

> The field push_url is optional and when set will trigger a push of metrics to a Prometheus Push Gateway once Benthos shuts down. It is also possible to specify a push_interval which results in periodic pushes.

Since Benthos would be running only briefly, I reasoned that I don't need to set the `push_interval` too and was assuming that Benthos will simply emit metrics just before shutting down. Unfortunately, that's not the case, because [this code](https://github.com/Jeffail/benthos/blob/9cf647c8c6d2eb978c49881272ee9703be51b0e2/lib/metrics/prometheus.go#L151-L152) prevents the pusher from being initialised if `push_interval` is not set, so then no metrics will be emitted [on shutdown](https://github.com/Jeffail/benthos/blob/9cf647c8c6d2eb978c49881272ee9703be51b0e2/lib/metrics/prometheus.go#L415-L423).

I think it's desirable to adjust the code to allow the pusher to work even if `push_interval` is not set, but, otherwise, I'd also be OK with adjusting the docs to explicitly mention that `push_interval` is mandatory when setting `push_url`. WDYT?

mihaitodor:
Fixes #619 and also adds some crude unit tests for Prometheus PushGateway.

### DIFF
--- a/lib/metrics/prometheus.go
+++ b/lib/metrics/prometheus.go
@@ -148,29 +148,31 @@ func NewPrometheus(config Config, opts ...func(Type)) (Type, error) {
 		return nil, fmt.Errorf("failed to init path mapping: %v", err)
 	}
 
-	if len(p.config.PushURL) > 0 && len(p.config.PushInterval) > 0 {
+	if len(p.config.PushURL) > 0 {
 		p.pusher = push.New(p.config.PushURL, p.config.PushJobName).Gatherer(prometheus.DefaultGatherer)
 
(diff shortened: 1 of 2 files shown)


Jeffail:
Awesome, thanks @mihaitodor!

Issue #583, opened 2020-12-09
Add json_equals unit test predicate


nicktelford:
The `content_equals` predicate does an exact string match on the output document. This has several drawbacks when working with JSON:
  * structural whitespace, which is irrelevant in JSON, is included in the test
  * ordering of fields, which is also irrelevant in (most) JSON, is strictly enforced

This can be problematic for a number of reasons:
  * Very large JSON fixtures have to be compacted and inlined to a single, very long, line in the test YAML document.
  * If the field ordering of a mapping behaves non-deterministically, it may not be possible to write a test for it.

The solution is to create a `json_equals` predicate, that behaves similarly to `content_equals`, but with the following differences:
  * the test fixture must be a valid JSON document
  * the output message must be a valid JSON document
  * structural whitespace is ignored during comparison
  * field ordering is ignored during comparison

ollystephens:
See https://github.com/Jeffail/benthos/pull/621 for incoming PR that implements this.
Also adds `json_contains` predicate which passes if condition is structural subset of message.
Also adds `json_content` so can specify input messages structurally too.

PR #621, opened 2021-01-15
feat: initial support for JSON test helpers

ollystephens:
- can now use `json_content` instead of `content` to specify test
  messages structurally
- can now use `json_equals` and `json_contains` as test conditions to
  check messages structurally (avoiding formatting and order variation
      issues that are prone when testing JSON verbatim)


ollystephens:
resolves #583

Jeffail:
Very cool, thanks @ollystephens!

PR #629, opened 2021-01-19
added interpolation functions to redis_list key

lorenz-qbees:
Adopt interpolation functions to redis_list key from redis_hash.


Jeffail:
Looks great, thanks @lorenz-qbees! And :heart: for updating the docs.

Issue #626, opened 2021-01-18
golangci-lint as a linter?


mihaitodor:
I prefer to use [golangci-lint](https://github.com/golangci/golangci-lint) as a linter and was wondering if you considered using it for this project. I did a quick run on on the current codebase (42cb803c) using default settings and it took only a few seconds. It reveals a bunch of potential issues and some code improvements (see below), but it can also be noisy / annoying sometimes and it's quite strict about error checks (although this check can be disabled at your own peril 😅 ). The `unused` check emits false positives if there are build flags in the codebase, but I think that can be fixed. I also checked if enabling the extra `bodyclose`, `gofmt` and `goimports` checks would work and they didn't cause slowness or noise.

<details>
  <summary>golangci-lint output</summary>

```

> golangci-lint run

lib/service/run.go:205:20: Error return value of `cli.ShowAppHelp` is not checked (errcheck)
				cli.ShowAppHelp(c)
				               ^
lib/service/run.go:341:19: Error return value of `cli.ShowAppHelp` is not checked (errcheck)
			cli.ShowAppHelp(context)
			               ^
lib/service/run.go:353:19: Error return value of `cli.ShowAppHelp` is not checked (errcheck)
			cli.ShowAppHelp(context)
			               ^
lib/service/run.go:356:14: Error return value of `flags.Parse` is not checked (errcheck)
		flags.Parse(os.Args[1:])
		           ^
lib/service/run.go:366:9: Error return value of `app.Run` is not checked (errcheck)
	app.Run(os.Args)
	       ^
lib/service/service.go:327:23: Error return value of `httpServer.Shutdown` is not checked (errcheck)
			httpServer.Shutdown(context.Background())
			                   ^
lib/service/service.go:341:37: Error return value of `(*runtime/pprof.Profile).WriteTo` is not checked (errcheck)
			pprof.Lookup("goroutine").WriteTo(os.Stderr, 1)
			                                 ^
lib/service/service.go:355:37: Error return value of `(*runtime/pprof.Profile).WriteTo` is not checked (errcheck)
			pprof.Lookup("goroutine").WriteTo(os.Stderr, 1)
			                                 ^
lib/service/deprecated.go:62:4: S1038: should use fmt.Fprintf instead of fmt.Fprintln(fmt.Sprintf(...)) (but don't forget the newline) (gosimple)
			fmt.Fprintln(os.Stderr, fmt.Sprintf("Configuration sanitise error: %v", err))
			^
[...]

PR #630, opened 2021-01-19
Add golangci-lint and enable some checks

mihaitodor:
Also remove some dead code and fix some tests to use the test name.

This is the first step to addressing #626. At a bare minimum, I think the following extra linters should be enabled:
- deadcode
- errcheck
- gosimple
- ineffassign
- staticcheck
- unused
- varcheck
- bodyclose (not run by default by golangci-lint, but should be handy)


Jeffail:
Thanks @mihaitodor this is glorious, so much red :heart_eyes:

mihaitodor:
Whoops, just realised I missed the `internal` folder 🤦  Thanks for adding it to the linter!

Issue #539, opened 2020-10-29
SFTP input and output


Jeffail:
Similar to file input and outputs it'd be nice to have an SFTP equivalent input and output.

PR #634, opened 2021-01-20
Add SFTP input + output

nmbryant:
Here is code for adding an input for files that are on an SFTP server. There is an option for processing an individual file or all files within a directory. I followed what was done for the Azure Blob Storage input, thanks @mihaitodor for all the help with this!

Also, have an SFTP test file that has tests that are being skipped. Initially I just removed it since it causes drone tests to fail due to using dockertest, but added it back with skips since I think it could be helpful for testing new features or changes. Although I wouldn't mind removing it if it's not necessary or helpful.

Closes: #539


lib/input/sftp.go
```diff
@@ -0,0 +1,495 @@
...
+			CategoryNetwork,
+		},
+	}
+}
+
+//------------------------------------------------------------------------------
+
+// SFTPConfig contains configuration fields for the SFTP input type.
+type SFTPConfig struct {
+	Server                string          `json:"server" yaml:"server"`
```
Jeffail, line 103: In most places we have a single field called `address` for host:port combos, it'd be good to follow that here for consistency, you can then use `url.Parse` to pull out the individual values.

lib/input/sftp.go
```diff
@@ -0,0 +1,495 @@
...
+	Port                  int             `json:"port" yaml:"port"`
+	Filename              string          `json:"filename" yaml:"filename"`
+	Credentials           SFTPCredentials `json:"credentials" yaml:"credentials"`
+	Path                  string          `json:"path" yaml:"path"`
+	MaxConnectionAttempts int             `json:"max_connection_attempts" yaml:"max_connection_attempts"`
+	RetrySleepDuration    string          `json:"retry_sleep_duration" yaml:"retry_sleep_duration"`
+	Codec                 string          `json:"codec" yaml:"codec"`
+	DeleteObjects         bool            `json:"delete_objects" yaml:"delete_objects"`
```
Jeffail, line 111: This ought to be named `delete_files`

lib/input/sftp.go
```diff
@@ -0,0 +1,495 @@
...
+	}
+
+	s := &SFTP{
+		conf:              conf,
+		log:               log,
+		stats:             stats,
+		objectScannerCtor: objectScannerCtor,
+	}
+
+	err = s.initSFTPConnection()
```
Jeffail, line 169: Normally the connection of a reader wouldn't be established until `ConnectWithContext` is called, this allows Benthos to manage retries and reconnects as/when needed.

lib/input/sftp.go
```diff
@@ -0,0 +1,495 @@
...
+}
+
+type sftpPendingObject struct {
+	target    *sftpObjectTarget
+	obj       *sftp.File
+	extracted int
+	scanner   codec.Reader
+}
+
+func newSFTPTargetReader(
```
Jeffail, line 227: Looks like the inspiration for this was taken from the s3 input, but it might be worth checking out the file input as well: https://github.com/Jeffail/benthos/blob/master/lib/input/file.go as it's simpler.

The reason for the complexity in the s3 version, and needing a generic object reader type, is because we need to support multiple strategies for obtaining a list of files to download, whereas here it looks like we're always walking a static list of files so we can probably reduce a lot of the code here.

lib/input/sftp.go
```diff
@@ -0,0 +1,495 @@
...
+		target.ackFn(ctx, err)
+		return nil, err
+	}
+
+	s.object = object
+	return object, nil
+}
+
+// SFTPServer contains connection data for connecting to an SFTP server
+type SFTPServer struct {
```
Jeffail, line 442: Since this is duplicated for the output as well it might be worth pulling it out into a shared package similar to https://github.com/Jeffail/benthos/tree/master/internal/service/redis

lib/input/sftp.go
```diff
@@ -0,0 +1,495 @@
...
+	MaxConnectionAttempts int             `json:"max_connection_attempts" yaml:"max_connection_attempts"`
+	RetrySleepDuration    string          `json:"retry_sleep_duration" yaml:"retry_sleep_duration"`
+	Codec                 string          `json:"codec" yaml:"codec"`
+	DeleteObjects         bool            `json:"delete_objects" yaml:"delete_objects"`
+}
+
+// SFTPCredentials contains the credentials for connecting to the SFTP server
+type SFTPCredentials struct {
+	Username string `json:"username" yaml:"username"`
+	Secret   string `json:"secret" yaml:"secret"`
```
Jeffail, line 117: I might be wrong but I think it's probably more common for this to be referred to as "password" in the world of ssh/sftp.

Jeffail (changes requested):
Hey @nmbryant, looking good, left some comments mostly for cosmetic stuff.

One thing I'm considering is how this compares with the current `file` input: https://www.benthos.dev/docs/components/inputs/file, ideally I'd like them to be consistent, as sftp has more in common with files on disk than s3 and azure blob storage.

They mostly look the same right now but the major distinction is in how the path works, the `file` input allows you to specify any number of paths and supports wildcards, I think ideally it'd be good to support the same flow here.

nmbryant:
@Jeffail Thanks for the feedback, I think I have covered all of your comments with my latest changes. 
Mainly, I reworked the SFTP input so that it looks more like the file input instead of the Azure blob one.

Jeffail (approved):

Jeffail:
Looks good, thanks @nmbryant! Just a heads up that I might make a few tweaks to some of the field defaults.

PR #636, opened 2021-01-22
Added tags to S3 objects

codingconcepts:
Hey @Jeffail,

I've added the ability to tag objects in S3. Please give me a shout if you'd like me to make any changes!

Cheers,

Rob

### DIFF
--- a/lib/output/aws_s3.go
+++ b/lib/output/aws_s3.go
@@ -34,12 +34,12 @@ Tag fields will be stored as object tags against the S3 object. For example:
 
 ` + "```yaml" + `
 output:
-	aws_s3:
-		bucket: TODO
-    	path: ${!count("files")}-${!timestamp_unix_nano()}.tar.gz
-		tags:
-			Key1: Value1
-			Timestamp: ${!meta("Timestamp")}
+    aws_s3:
+        bucket: TODO
+        path: ${!count("files")}-${!timestamp_unix_nano()}.tar.gz
+        tags:
+            Key1: Value1
+            Timestamp: ${!meta("Timestamp")}
 ` + "```" + `
 
 ### Credentials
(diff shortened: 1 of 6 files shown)


lib/output/writer/s3.go
```diff
@@ -162,9 +174,17 @@ func (a *AmazonS3) WriteWithContext(wctx context.Context, msg types.Message) err
 			contentEncoding = aws.String(ce)
 		}
 
+		// Prepare tags.
+		tags := []string{}
+		for k, v := range a.tags {
+			tags = append(tags, k+"="+v.String(i, msg))
```
Jeffail, line 180: I might be wrong as I haven't read up on S3 tags but it looks like we're building a url query string here, so it's probably worth escaping the keys and values with https://golang.org/pkg/net/url/#QueryEscape

lib/output/aws_s3.go
```diff
@@ -123,6 +144,20 @@ calculated per message of a batch.
 
 Metadata fields on messages will be sent as headers, in order to mutate these values (or remove them) check out the [metadata docs](/docs/configuration/metadata).
 
+### Tags
+
+Tag fields will be stored as object tags against the S3 object. For example:
+
+` + "```yaml" + `
+output:
+	aws_s3:
```
Jeffail, line 153: Looks like your IDE has mixed tabs and spaces, mine does this all the time and it drives me crazy embedding yaml configs in a Go string literal.

lib/output/writer/s3.go
codingconcepts: This is fixed, great catch 👍

lib/output/aws_s3.go
codingconcepts: And lo! My changes seem to have no effect, I'll spend some more time on this tomorrow 😄

codingconcepts: Righto, that's fixed!

codingconcepts:
Apologies, I accidentally resolved the conversation about the query escaping! Please re-open the conversation if you're not happy with the fix, @Jeffail.

Jeffail:
Ace, thanks @codingconcepts, I might make a small tweak as it might be useful to ensure consistent ordering of tags.

codingconcepts:
Awesome, thanks, @Jeffail!

Want me to sort them by keys?

Jeffail:
@codingconcepts, done https://github.com/Jeffail/benthos/commit/228b4377ba13d9b2994efd2aa0c23130706e325c, would be good if you could test it at some point though.

Issue #618, opened 2021-01-13
Allow disabling consumer groups with `kafka` input


Jeffail:
The kafka input should have an option to support ignoring any stored offsets and instead consume from the latest (or oldest) offset each time the connection is established.

codingconcepts:
Would making consumer_group an interpolation field help here, perhaps?

Jeffail:
An alternative is that you allow users to specify an empty consumer group which disables partition balancing and offset persistence. A minor annoyance here is that Benthos sets it to a default value, so you'd need to add an explicit `consumer_group: ""` to your config, but we can have it empty by default in V4.

Allowing consumer group interpolations has the advantage of being able to resume a previous group if you later decide to, but we're not really offering any tools for enabling that, so you'd need to manually dig out the random identifier and make it explicit in the Benthos config. It would also mean persisting an unbounded number of group offsets, meaning if the retention period (`offsets.retention.minutes`) of the broker is high we'd potentially end up taking a significant amount of space and resources for information that's likely being unused.

The consumer group interpolation approach is definitely cleaner for us to implement (kudos to @DeeJayBro) but I'm a little worried that once the cats out of the bag we're then responsible for any negative repercussions, and this is something that could end up slowing down brokers if a user isn't careful.

DeeJayBro:
Another alternative would be to add an option to reset the offset to oldest or newest on startup, but this could potentially lead to problems with other consumers already online, but completely disabling offset persistence would prob be the cleanest solution

Jeffail:
Yeah let's go with disabling the consumer group. The logic for the kafka input is already quite complex specifically around commits due to the checkpointer but hopefully this should be fairly quick and easy to add.

mfamador:
And what about to have a parameter like the `-o` (offset) in `kafkacat` where we can set if we want to read from the end (`-o end`), from the beginning (-o begin) or from an absolute or relative value:
```  -o <offset>        Offset to start consuming from:
                     beginning | end | stored |
                     <value>  (absolute offset) |
                     -<value> (relative offset from end)
                     s@<value> (timestamp in ms to start at)
                     e@<value> (timestamp in ms to stop at (not included))```

DeeJayBro:
What about disabling `auto.commit.offset`, so we can use the existing `start_from_oldest` to define the offset at each startup. Because no last offsets are saved in the Kafka, thus the user doesn't accidentally flood resources etc

yev-dimo:
[...]

PR #637, opened 2021-01-23
Enable consumer_group function interpolation in Kafka Input

DeeJayBro:
Address #618 by enabling function interpolation for consumer_group and thus allowing unique consumer_groups per session.

### DIFF
--- a/lib/input/kafka.go
+++ b/lib/input/kafka.go
@@ -10,6 +10,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/Jeffail/benthos/v3/internal/bloblang"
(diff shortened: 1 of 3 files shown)


Jeffail:
Thanks @DeeJayBro, this implementation looks good but I've been thinking about the implications of allowing interpolations here and there's a few things I want to understand before we proceed, going to follow up in https://github.com/Jeffail/benthos/issues/618

Jeffail:
Went with enabling empty consumer groups: https://github.com/Jeffail/benthos/commit/76264dfe8c6afba31b5383093ee97041578fc92b

PR #640, opened 2021-01-25
Add chunker codec

### LINKED ISSUE #599: S3 / Azure input codecs and memory usage
Say, for example, somebody wants to use Benthos to transfer arbitrary binary files from S3 to Azure that can vary in size by orders of magnitude.

I was looking into the [downloading large files](https://www.benthos.dev/docs/components/inputs/aws_s3/#downloading-large-files) section of the docs and started wondering how would one constrain Benthos' memory in this particular case. Using `all-bytes` as a codec will make it blow up when encountering a large file, since it tries to read the whole file in memory, if I understand [the code](https://github.com/Jeffail/benthos/blob/0ff54f67463de7516afb25307785684b23595693/internal/codec/reader.go#L170) correctly.

Would adding something like a `chunker:size` codec make sense? It would be similar to `delim:x`, but `size` would be in a format which can be parsed into bytes (using a helper lib such as [this one](https://github.com/alecthomas/units)).

mihaitodor:
This fixes #599. Note that appending the chunks into a single file in the output will only be supported by the [`file`](https://www.benthos.dev/docs/components/outputs/file) and [`azure`](https://www.benthos.dev/docs/components/outputs/azure_blob_storage) outputs. For S3, one can still redirect each chunk to a separate file via bloblang (i.e. `path: out-${!count("files")}.txt`), if that's an acceptable way to transfer very large files without blowing up memory. Refer to #599 for more details.

### DIFF
--- a/internal/codec/reader.go
+++ b/internal/codec/reader.go
@@ -12,6 +12,7 @@ import (
 	"io"
 	"io/ioutil"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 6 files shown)


Jeffail:

PR #642, opened 2021-01-26
Add append output codec

### LINKED ISSUE #641: Output codec without delimiter
Looking at the [available output codecs](https://www.benthos.dev/docs/components/outputs/file), I noticed that `all-bytes` [truncates](https://github.com/Jeffail/benthos/blob/d10ce3d96f0f313c79af7a7264f016b78d295c5f/internal/codec/writer.go#L71) on each message and `delim` [doesn't support](https://github.com/Jeffail/benthos/blob/d10ce3d96f0f313c79af7a7264f016b78d295c5f/internal/codec/writer.go#L58-L60) an empty delimiter.

Would you consider removing the non-empty delimiter restriction in the `delim` output codec? I need this when using the `chunker` input codec that I added in #640 for the use case of transferring a large file from Azure to a local file.

mihaitodor:
This fixes #641.

Additionally, I noticed during some testing that Azure sends an EOF error when the last byte of a blob is read, which caused the new chunker input codec to skip the last chunk. I didn't bump into it before because I used the file input for testing and that one seems to only return EOF after everything was read and a subsequent call to Read is made.

The test I added enforces that all the benthos input readers will only return EOF if the underlying reader returns 0 bytes and EOF. This mandates that the underlying readers obey the spec once EOF has been returned and return 0, EOF during subsequent calls to Read.

This is the relevant paragraph from [the docs](https://golang.org/pkg/io/#Reader):

> When Read encounters an error or end-of-file condition after successfully reading n > 0 bytes, it returns the number of bytes read. It may return the (non-nil) error from the same call or return the error (and n == 0) from a subsequent call. An instance of this general case is that a Reader returning a non-zero number of bytes at the end of the input stream may return either err == EOF or err == nil. The next Read should return 0, EOF.

### DIFF
--- a/internal/codec/reader.go
+++ b/internal/codec/reader.go
@@ -499,17 +499,23 @@ func (a *chunkerReader) Next(ctx context.Context) (types.Part, ReaderAckFn, erro
 	}
 
 	n, err := a.r.Read(a.buf)
-	if err == nil {
+	if err != nil {
+		if err == io.EOF {
+			a.finished = true
+		} else {
+			a.sourceAck(ctx, err)
+			return nil, nil, err
+		}
+	}
+
+	if n > 0 {
 		a.pending++
 
 		bytesCopy := make([]byte, n)
 		copy(bytesCopy, a.buf)
 		return message.NewPart(bytesCopy), a.ack, nil
-	} else if err == io.EOF {
-		a.finished = true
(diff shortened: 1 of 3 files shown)


Jeffail:
Nice, thanks @mihaitodor!

PR #645, opened 2021-01-26
Enable gosiple linter

mihaitodor:
Most of these changes are straightforward, but please have a look at the following:
- 3e1101b - While there are fewer function calls, it kinda' stinks that `\n` has to be inserted in the format pattern. Might be worth considering switching to the `log` package for these, because a) it provides `log.Fatalf` and b) it always outputs the newline char, even if it's not added to the format pattern. Let me know if you'd like me to take a stab at that.
- 632d615 - Collapse for loop in single append statement: Intuitively, *I think* `values := make([]interface{}, 0, len(m)); values = append(values, m...)` performs better than `values := make([]interface{}, len(m)); copy(values, m)` because it avoids zero-initialising the memory when allocating the slice, but I haven't tested it. It looks like the latter pattern is used in several places throughout the codebase (`ag -s -B3 "append\("` should reveal most of them). I can change it to append if you think it's worth the effort.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -17,7 +17,7 @@ linters:
     # Default linters reported by golangci-lint help linters` in v1.35.2
     # - deadcode
     # - errcheck
-    # - gosimple
+    - gosimple
     - govet
     - ineffassign
     # - staticcheck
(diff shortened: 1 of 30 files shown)


Jeffail:
>  +47 −77

:heart_eyes: 

These changes all look good to me, I wouldn't worry too much about the copy vs append performance as I doubt it'd show up on a service-wide profile.

PR #647, opened 2021-01-27
feat: Add format "all-bytes" to archive processor

stergiotis:
This PR introduces the `all-bytes` codec to the `archive` processor. It has the same semantic as the `all-bytes` codec used e.g. in the `file` input and output.

I found it to be especially useful in conjecture with the `http_client`.

### DIFF
--- a/lib/processor/archive.go
+++ b/lib/processor/archive.go
@@ -44,7 +44,7 @@ of the batch.`,
 		},
 		UsesBatches: true,
 		FieldSpecs: docs.FieldSpecs{
-			docs.FieldCommon("format", "The archiving [format](#formats) to apply.").HasOptions("tar", "zip", "binary", "lines", "json_array"),
+			docs.FieldCommon("format", "The archiving [format](#formats) to apply.").HasOptions("tar", "zip", "binary", "lines", "json_array", "all-bytes"),
 			docs.FieldCommon(
 				"path", "The path to set for each message in the archive (when applicable).",
 				"${!count(\"files\")}-${!timestamp_unix_nano()}.txt", "${!meta(\"kafka_key\")}-${!json(\"id\")}.json",
@@ -73,6 +73,10 @@ Archive messages to a binary blob format consisting of:
 ### ` + "`lines`" + `
 
 Join the raw contents of each message and insert a line break between each one.
+
+### ` + "`all-bytes`" + `
+
+Join the raw contents of each message.
 
 ### ` + "`json_array`" + `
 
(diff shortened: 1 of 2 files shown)


lib/processor/archive.go
```diff
@@ -198,6 +202,39 @@ func linesArchive(hFunc headerFunc, msg types.Message) (types.Part, error) {
...
+
+	retr := make([]byte, totalLen, totalLen)
+	pos := 0
+	for _, slice := range slices {
+		pos += copy(retr[pos:], slice)
+	}
+	return retr
+}
+func allBytesArchive(hFunc headerFunc, msg types.Message) (types.Part, error) {
+	tmpParts := make([][]byte, msg.Len())
```
Jeffail, line 225:
```suggestion
	var buf bytes.Buffer
```

lib/processor/archive.go
```diff
@@ -198,6 +202,39 @@ func linesArchive(hFunc headerFunc, msg types.Message) (types.Part, error) {
...
+	err := msg.Iter(func(i int, part types.Part) error {
+		tmpParts[i] = part.Get()
```
Jeffail, line 227:
```suggestion
		buf.Write(part.Get())
```

lib/processor/archive.go
```diff
@@ -198,6 +202,39 @@ func linesArchive(hFunc headerFunc, msg types.Message) (types.Part, error) {
...
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	newPart := msg.Get(0).Copy()
+	newPart.Set(concatenate(tmpParts))
```
Jeffail, line 234:
```suggestion
	newPart.Set(buf.Bytes())
```

Jeffail (commented):
Hey @stergiotis, I've added a suggestion for using a bytes buffer for the concat, I also think the name `concatenate` for the format itself is probably better suited.

lib/processor/archive.go
stergiotis: @Jeffail
Many thanks for looking into this. My choice has todo with my mental model of the performance of  languages with a genrational garbage collector. I assume that is less costly to allocate many fixed length small arrays for a short time than to dynamically grow a single buffer. I did not benchmark though and am very new to golang.

I therefore suggest that you choose what implementation feels more idiomatic to you and matches the existing codebase.

stergiotis: @Jeffail Regarding the naming `all-bytes`: I found it quite supristing that throughout benthos an empty `delimiter` option is synonymous to `'\n'`. I expected it to be what is labeled `all-bytes` here. 

You're right: `concatenate` seems a nice descriptive name for this operation.

Jeffail:
Updated, fixed and merged here: https://github.com/Jeffail/benthos/commit/2cb833e696adf42ab960c25019ca2039fbb79e72

Jeffail:
Whoops, amended that commit and force pushed as it wiped you as an author, new commit: https://github.com/Jeffail/benthos/commit/48244633964298c9a5e177c74695c4e40501206d

Thanks @stergiotis!

PR #657, opened 2021-02-08
SFTP Watcher Mode

nmbryant:
Adds a new option to the SFTP input to enable it to keep running and processing new files on the given paths as they are created. A concern that @mihaitodor pointed out was that the hash table on line 112 of input/sftp.go is that it will grow indefinitely as the pipeline is running and could get quite large. Any thoughts on how to handle this @Jeffail?

### DIFF
--- a/lib/input/sftp.go
+++ b/lib/input/sftp.go
@@ -3,6 +3,7 @@ package input
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"sync"
 	"time"
@@ -21,7 +22,7 @@ import (
 func init() {
 	Constructors[TypeSFTP] = TypeSpec{
 		constructor: fromSimpleConstructor(func(conf Config, mgr types.Manager, log log.Modular, stats metrics.Type) (Type, error) {
-			r, err := newSFTPReader(conf.SFTP, log, stats)
+			r, err := newSFTPReader(conf.SFTP, mgr, log, stats)
 			if err != nil {
 				return nil, err
 			}
@@ -74,6 +75,12 @@ You can access these metadata fields using [function interpolation](/docs/config
 
 //------------------------------------------------------------------------------
 
+type WatcherConfig struct {
+	Enabled bool `json:"enabled" yaml:"enabled"`
+	PollInterval string `json:"poll_interval" yaml:"poll_interval"`
+	Cache string `json:"cache" yaml:"cache"`
+}
+
 // SFTPConfig contains configuration fields for the SFTP input type.
 type SFTPConfig struct {
 	Address        string                `json:"address" yaml:"address"`
@@ -82,7 +89,7 @@ type SFTPConfig struct {
 	Codec          string                `json:"codec" yaml:"codec"`
 	DeleteOnFinish bool                  `json:"delete_on_finish" yaml:"delete_on_finish"`
 	MaxBuffer      int                   `json:"max_buffer" yaml:"max_buffer"`
-	WatcherMode    bool                  `json:"watcher_mode" yaml:"watcher_mode"`
+	Watcher        WatcherConfig		 `json:"watcher" yaml:"watcher"`
 }
 
 // NewSFTPConfig creates a new SFTPConfig with default values.
@@ -94,7 +101,9 @@ func NewSFTPConfig() SFTPConfig {
 		Codec:          "all-bytes",
 		DeleteOnFinish: false,
 		MaxBuffer:      1000000,
-		WatcherMode:    false,
+		Watcher:    WatcherConfig{
+			Enabled:      false,
+		},
 	}
 }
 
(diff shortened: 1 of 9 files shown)


nmbryant:
Updated the PR to squash the commits, but now the test TestRetryParallel is failing. I think it might be a flaky test since it is passing locally and passed before I squashed the commits. Any thoughts on this @Jeffail ?

Jeffail:
Yeah that's a flaky test, my bad, ignore it and I'll see if a rerun clears it. I love this concept, there's a similar issue being discussed in https://github.com/Jeffail/benthos/issues/271, the main thing that might apply here is that if you utilize a cache resource rather than store seen files within a map then it solves both your memory leak issue (with configurable TTLs that users can choose) as well as solving persistence across restarts.

Another thing is I think we should create a new configuration block for file watching as there's almost definitely going to be more fields to add over time. Maybe something like this:

```yaml
input:
  sftp:
    address: ""
    credentials:
      username: ""
      password: ""
    paths:
      - ./*.txt
    codec: all-bytes
    watcher:
      enabled: true
      poll_interval: 1s
      cache: files-memory

resources:
  caches:
    files-memory:
      memory:
        ttl: 900
```

mihaitodor:
FYI, I think I fixed the flakey test in #661.

nmbryant:
@Jeffail Updated the watcher config and switched to using a cache instead of a map for filesProcessed, what do you think?

lib/input/sftp.go
```diff
@@ -108,21 +140,43 @@ type sftpReader struct {
...
+	if conf.Watcher.Enabled {
+		if watcherPollInterval, err = time.ParseDuration(conf.Watcher.PollInterval); err != nil {
+			return nil, fmt.Errorf("failed to parse watcher poll interval string: %v", err)
+		}
+
+		if conf.Watcher.Cache == "" {
+			return nil, fmt.Errorf("watcher cache is required if watcher mode is enabled")
+		}
+
+		cache, err = mgr.GetCache(conf.Watcher.Cache)
```
Jeffail, line 167: Resources right now are static but there's no guarantee that they won't change over time, so this should stick around as a check that the cache exists but do not capture it in a variable. Each time it gets referenced below you need to call `mgr.GetCache` again.

```suggestion
		_, err = mgr.GetCache(conf.Watcher.Cache)
```

lib/input/sftp.go
```diff
@@ -200,6 +250,9 @@ func (s *sftpReader) ReadWithContext(ctx context.Context) (types.Message, reader
 			err = types.ErrTimeout
 		}
 		if err != types.ErrTimeout {
+			if s.conf.Watcher.Enabled {
+				err = s.watcherCache.Set(s.currentPath, []byte{})
```
Jeffail, line 254: Rather than `[]byte{}` we probably need to set a real >0 byte value as some cache types might not allow an empty value. I'd recommend just using a single char like `[]byte("@")`.

lib/input/sftp.go
```diff
@@ -240,3 +293,31 @@ func (s *sftpReader) CloseAsync() {
...
+			continue
+		}
+
+		for _, path := range paths {
+			if s.conf.Watcher.Enabled {
+				if _, err := s.watcherCache.Get(path); err != nil {
+					filepaths = append(filepaths, path)
+				} else {
+					// Reset the TTL for the path
+					err = s.watcherCache.Set(path, []byte{})
```
Jeffail, line 312: Same here as above.

lib/test/integration/integration_test_helpers.go
```diff
@@ -331,7 +332,21 @@ func initInput(t *testing.T, env *testEnvironment) types.Input {
...
-	input, err := input.New(s.Input, types.NoopMgr(), env.log, env.stats)
+	caches := map[string]types.Cache{}
+	for name, c := range s.Manager.Caches {
+		cacheObj, err := cache.New(c, types.NoopMgr(), env.log, env.stats)
+		require.NoError(t, err)
+		caches[name] = cacheObj
+	}
+
+	var mgr types.Manager
+	if len(caches) > 0 {
```
Jeffail, line 343: Rather than using a mock manager here it's probably worth instantiating a real one, you can just use noops for all the params (and nil for api) like this: `manager.New(s.Manager, nil, log.Noop(), metrics.Noop())`.

lib/test/integration/sftp_test.go
```diff
@@ -35,6 +49,27 @@ var _ = registerIntegrationTest("sftp", func(t *testing.T) {
...
+		Password: sftpPassword,
+	}
+
+	sftpPort = resource.GetPort("22/tcp")
+	sftpEndpoint := "localhost:" + sftpPort
+	deadline := time.Now().Add(dialTimeout)
+	for time.Now().Before(deadline) {
+		t.Logf("waiting for SFTP server to come up on '%v'...", "localhost:"+sftpPort)
+
+		sftpClient, err = creds.GetClient(sftpEndpoint)
```
Jeffail, line 63: This should go within the `pool.Retry` block above, it handles looping, backoff, etc for you.

Jeffail (changes requested):
Hey @nmbryant, looks good, left a few small bits to update and then I think it's ready to roll out.

nmbryant:
@Jeffail Thanks for the feedback, just updated the PR with changes requested.

Jeffail (approved):
Thanks @nmbryant, looks good. One thing to consider is that I believe there's a race condition here where if a file is partially written during a list files poll then it would be possible to only process some of the data and mark the file as done. This isn't likely with small files but as the files themselves get larger so does the likelihood of missing data.

This is listed as an experimental input so I don't think we need a proper solution to this yet, but calling it out now in case it compromises your use case. One potential quick solution would be to add a parameter for only consuming a file after a given time period since it was last written.

Jeffail:
@nmbryant want some more time on this or would you rather merge as it currently is and iterate?

nmbryant:
Thanks for pointing out that issue, our use case is with large files so it is something we will have to address. I will think about this and probably get back to it sometime this week, though I think it might be better to just merge what is in this branch and then do the fix in a new branch.

Jeffail:
Cool, thanks @nmbryant, merging as is and I'll add a note to the docs.

Jeffail:
@nmbryant had a quick attempt at this: https://github.com/Jeffail/benthos/commit/044741914332caf3b85f121ef3fedcbca54b19c8, I think this should be fine for now.

nmbryant:
@Jeffail Awesome, thanks for making the change. I looked through that commit and it looked good.

PR #661, opened 2021-02-10
Close HTTP response body in TestHTTPBasic

mihaitodor:
This test was leaking a ton of goroutines. Now it seems fine and doesn't seem to break with `go test -v -run=TestHTTPBasic -race -count=200 ./lib/output/`.

I also noticed that the request Body is closed explicitly in getHandler and streamHandler, but this doesn't seem necessary according to the [docs](https://golang.org/pkg/net/http/#Request):
> // For server requests, the Request Body is always non-nil
> // but will return EOF immediately when no body is present.
> // The Server will close the request body. The ServeHTTP
> // Handler does not need to.

I guess these Close() calls can be left in there if you think they're somehow needed.


Jeffail:
Thanks @mihaitodor! Those closes were probably just my paranoia at the time, I'm fine with deleting them.

PR #665, opened 2021-02-11
influx - fixes adding extra values in tagged metrics

nwest1:
When dealing attempting to utilize some simple additional tag values, I ran into an issue when using labeled metrics.

The issue is that we're appending to a variable that's not scoped within the returned function. This causes additional values to be appended as tag values and screws up reporting for any metrics that use tags. Essentially creating a static tag/field combo that does not change after first set. Not a great look from metrics that you'd expect to be increasing.

I have been struggling to write a test case that exposes this, as I think I'm lacking some understanding of how these are tested vs. how they are referenced and utilized in any particular processor. If you have any insight as to how to write something that exposes this, let me know. Otherwise through local testing I've validated this works way better than the previous version.

Thanks again!!

### DIFF
--- a/lib/metrics/influxdb.go
+++ b/lib/metrics/influxdb.go
@@ -361,8 +361,8 @@ func (i *InfluxDB) GetCounterVec(path string, n []string) StatCounterVec {
 	labels = append(labels, n...)
 	return &fCounterVec{
 		f: func(l []string) StatCounter {
-			values = append(values, l...)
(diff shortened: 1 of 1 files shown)


Jeffail:
Thanks @nwest1, I'm planning to do the next release early next week, but let me know if you need a patch release earlier than that and I'll get one sorted.

nwest1:
We have a workaround in place, so no worries at the moment. 🥂

PR #667, opened 2021-02-13
Add missing error handling on StorageErrors 

mfamador:
Hi!. 
I've created this tiny fix to return the error when the `azqueue.StorageError` is not `azqueue.ServiceCodeQueueNotFound`, for instance when the queue name has invalid characters.


Jeffail:
Thanks @mfamador!

PR #668, opened 2021-02-13
Add the Azure Queue Storage Input

mfamador:
Hello!

I've created a very simple Azure Queue Storage input. 
I've tested for a while, with multiple replicas, and seems to be working fine.

Ex:
```
input:
  queue_storage:
    storage_connection_string: "<REDACTED>"
    queue_name: 'foo'
 ```

The Azure Queue Storage queue is created if it does not exist.

Cheers

### CODE REVIEW

File: lib/input/azure_queue_storage.go [hunk 1]
```diff
@@ -0,0 +1,169 @@
...
+	azQueueEndpointExp  = "https://%s.queue.core.windows.net"
+	devQueueEndpointExp = "http://localhost:10001/%s"
+	azAccountName       = "accountname"
+	azAccountKey        = "accountkey"
+	devAccountName      = "devstoreaccount1"
+	devAccountKey       = "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=="
+)
+
+// parseConnectionString extracts the credentials from the connection string.
+func parseConnectionString(input string) (storageAccount, storageAccessKey string, err error) {
```
mfamador, line 43: This logic around parsing azure connection strings is repeated all over azure outputs and inputs. Not sure where to put it in order to improve this not so DRY approach.

Jeffail: There's a directly of packages for stuff like this growing here: https://github.com/Jeffail/benthos/tree/master/internal/service, I think a new `azure` package there is probably a good place for it.

Jeffail:
Hey @mfamador, looks good, any chance you can add an integration test now that we have both an input and output?

lib/input/azure_queue_storage.go
mfamador: Great, I'll try to move this logic to a new azure package.

mfamador:
Sure, I'll try to add those tomorrow, it should be feasible with azurite.

lib/input/azure_queue_storage.go
mfamador: Created the azure package to place the shared constants and functions used among azure inputs and outputs. Only put the related to Azure Queue Storage, for now, I think I can create a distinct PR to add the others (Blob and Table Storage) later.

mfamador:
Hey!
I added an integration test and it's working but I had to do a little hack and store the random queue service port in azurite in an env var, or it would try to use the service always on `10001`.

mfamador:
I also kept the `t.Skip()`, probably it would be worth to use the default ports `10000`, `10001`, despite the fact of having clashes when running multiple tests at the same time.

Jeffail:
Looks good, I'm planning to do a release this evening so I'll merge this afterwards for some edge level testing.

mfamador:
Cool, thanks!

PR #669, opened 2021-02-15
Add missing `try` broker pattern

nenad:
This was listed as an option in the [output broker pattern](https://www.benthos.dev/docs/components/outputs/broker/#pattern), but there was no docs for it.

### DIFF
--- a/lib/output/broker.go
+++ b/lib/output/broker.go
@@ -86,16 +86,11 @@ potentially disproportionate message allocations to those outputs. Each message
 is sent to a single output, which is determined by allowing outputs to claim
 messages as soon as they are able to process them. This results in certain
 faster outputs potentially processing more messages at the cost of slower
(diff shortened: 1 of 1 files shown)


Jeffail:
Hey @nenad, I removed the docs for the `try` pattern because it's now deprecated in favour of the new dedicated `try` output: https://www.benthos.dev/docs/components/outputs/try. The implementation will stay there for backwards compatibility but rather than add the docs for it back in I'd rather remove it from the list of options for the `pattern` field.

Jeffail:
Looks good, thanks @nenad! FYI the build failure is because the docs need updating with `make docs`, I can do that on my side. I'll update the build step so it's clearer in the CI output.

nenad:
Makes sense, yeah. Is there anything else to be done to get the PR build green?

PR #680, opened 2021-02-22
Add suport for codec in socket output

DeeJayBro:
Allows defining own message delimiter for the socket output.

The default is `\n` so this shouldn't break anything for existing users.
Also added new tests to check for the correct behavior


Jeffail:
Hey @DeeJayBro, I'm trying to migrate stream outputs to use codecs instead, could you copy the way that the file output works: https://github.com/Jeffail/benthos/blob/master/lib/output/file.go, ignore the `delim` field there as it's just for backwards compatibility.

DeeJayBro:
@Jeffail Should I just ignore the append/truncate specific fields?

DeeJayBro:
Migrated to Codec System

Jeffail:
Looks good, thanks @DeeJayBro!

PR #682, opened 2021-02-22
Add Google Cloud Storage input and output support

### LINKED ISSUE #652: GCS (Google Cloud Storage) support
Would there be any interest in having support for GCS in Benthos (both input and output)? Their [API](https://pkg.go.dev/cloud.google.com/go/storage) [looks](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/storage/objects/main.go) quite straightforward to use.

mihaitodor:
This should do for the first iteration of #652. I might look into adding support for bucket notifications in the future.

Fixes #652.


website/docs/components/outputs/gcp_gcs.md
```diff
@@ -0,0 +1,300 @@
...
+
+```yaml
+# All config fields, showing default values
+output:
+  gcp_gcs:
+    bucket: ""
+    path: ${!count("files")}-${!timestamp_unix_nano()}.txt
+    content_type: application/octet-stream
+    content_encoding: ""
+    chunk_size: 16777216
```
mihaitodor, line 58: This might trip some users, but it's the default value of `googleapi.DefaultUploadChunkSize` and it's an optional feature of the Google storage writer which can increase / decrease performance in some cases, depending on the message size. Might be useful for some people, but not sure. I can leave it out of the config if you think it's too confusing.

lib/util/gcp/storage/client/client.go
```diff
@@ -0,0 +1,41 @@
...
+
+// New creates a storage.Client instance and, if the GCP_GCS_EMULATOR_URL env
+// var is set, then a custom HTTPClient is used, which forwards all requests to
+// the provided URL.
+// Note: [this bug](https://github.com/googleapis/google-cloud-go/issues/2476)
+// prevents the `STORAGE_EMULATOR_HOST` environment variable from working
+// correctly, so this alternative approach is needed when using a GCS emulator.
+func New(ctx context.Context) (*storage.Client, error) {
+	var opts []option.ClientOption
+	if rawURL := os.Getenv("GCP_GCS_EMULATOR_URL"); len(rawURL) > 0 {
```
mihaitodor, line 31: I think hiding this behind an undocumented env var is better than exposing it as a config param. I haven't managed to find any other reasonable way to work around this bug (and I'm glad this works).

website/docs/components/outputs/gcp_gcs.md
Jeffail: I think this is fine, I'll probably mark this as an experimental component at first so we'll have an opportunity to get feedback.

lib/util/gcp/storage/client/client.go
Jeffail: I can live with it, sounds like some grossness we need to deal with for now. Might be worth opening an issue once this is merged to revisit this later.

Jeffail:
Looks good, I'll need to put some time aside this week to review it properly, the only thing I'd change right now is maybe rename these to `gcp_cloud_storage`.

mihaitodor:
@Jeffail Thanks a lot for that suggestion! Figuring out how to name stuff is very hard sometimes and I like`gcp_cloud_storage` a lot more. Are you OK with the `gcs_` prefix for the meta or would you prefer `gcp_cloud_storage`? We have `blob_storage_` for `azure_blob_storage` and `s3_` for `aws_s3`.

Please take your time with the review. No hurry!

mihaitodor:
PS: Would you consider having an optional field called `credentials_json` which takes a base64-encoded credentials JSON? It would supersede `GOOGLE_APPLICATION_CREDENTIALS` so people don't have to provide the creds as a separate file. The Google API provides a [`WithCredentialsJSON` option](https://pkg.go.dev/google.golang.org/api/option#WithCredentialsJSON).

mihaitodor:
PPS: Seeing some issues if I do `make docs` after I rebase this on latest master (3722694). Any idea? :)

```
make docs
# github.com/Jeffail/benthos/v3/lib/output
lib/output/gcp_cloud_storage.go:76:3: unknown field 'sanitiseConfigFunc' in struct literal of type TypeSpec
lib/output/gcp_cloud_storage.go:87:5: docs.FieldCommon("path", "The path of each message to upload.", "${!count(\"files\")}-${!timestamp_unix_nano()}.txt", "${!meta(\"kafka_key\")}.json", "${!json(\"doc.namespace\")}/${!json(\"doc.id\")}.json").SupportsInterpolation undefined (type docs.FieldSpec has no field or method SupportsInterpolation)
lib/output/gcp_cloud_storage.go:88:80: docs.FieldCommon("content_type", "The content type to set for each object.").SupportsInterpolation undefined (type docs.FieldSpec has no field or method SupportsInterpolation)
lib/output/gcp_cloud_storage.go:89:98: docs.FieldAdvanced("content_encoding", "An optional content encoding to set for each object.").SupportsInterpolation undefined (type docs.FieldSpec has no field or method SupportsInterpolation)
lib/output/gcp_cloud_storage.go:90:247: docs.FieldAdvanced("chunk_size", "An optional chunk size which controls the maximum number of bytes of the object that the Writer will attempt to send to the server in a single request. If ChunkSize is set to zero, chunking will be disabled.").SupportsInterpolation undefined (type docs.FieldSpec has no field or method SupportsInterpolation)
make: *** [target/bin/benthos] Error 2
```

Jeffail:
@mihaitodor, you don't need to define a sanitation func any more as it's all being handled by the field specs, 
the `SupportsInterpolation` method is now just `IsInterpolated`, without an argument (it was redundant now anyway).

I'm planning to re-work this and the mongo PRs myself to use the new stuff as a sort of test-run but if you feel adventurous and want to try it out there's an existing pulsar package that does things the new way: https://github.com/Jeffail/benthos/tree/master/internal/service/pulsar

There might still be some rough edges though. Also I think I'd prefer to omit `credentials_json` for now.

mihaitodor:
OK, thank you! That makes sense. I pushed a few small changes so this is now rebased on the latest master. I'll have a look at that updated Pulsar service, but feel free to adjust the PR as you see fit.

> Also I think I'd prefer to omit `credentials_json` for now.

Okie dokie!

Jeffail:
@mihaitodor, sorry this took so long, I've been bouncing around all over the place recently. Got it reworked with some of the new stuff: https://github.com/Jeffail/benthos/tree/gcp-gcs-input-output, it's pretty much ready to merge but there are one or two extra bits I want to add, it'll probably be tomorrow evening. I'm planning to put out a release candidate this week.

mihaitodor:
@Jeffail No worries and thank you so much for doing this work! There's no hurry on my side. I've been stretching myself a bit too thin lately...

I checked out the branch and did a few tests. All seems to be OK.

Jeffail:
Merged: https://github.com/Jeffail/benthos/commit/9a99026c9f8a0113c15480fbef9a6ca64e55c3a9, thanks @mihaitodor!

PR #687, opened 2021-02-27
Enable staticcheck linter

mihaitodor:
This is part of #626. Took me a while to get back to this...

Important changes:

- ~8ae85e6d: Remove unused retriesMut from socket_server input - This is an odd one. Looks like this mutex was introduced recently but it seems to be a leftover from some attempted refactor. I removed it here and created a follow-up issue (#688) to address remaining concurrency bugs in this input.~ My mistake for jumping the gun here. Thanks Ash for the detailed [explanation](https://github.com/Jeffail/benthos/pull/687#discussion_r584089306)!
- ~03000c47: Fix SA4006: Return error on deletion failure - This introduces a change in behaviour for the `azure_queue_storage` input, which will now barf if an error is return if deletion fails. Hope this is OK.~ Nope, this is wrong. Will drop this change once #689 is merged.
- 95f468c1: Ignore SA1019: Type is deprecated for elasticsearch: Since I'm not too familiar with Elasticsearch, I marked as `nolint:staticcheck` the deprecation warnings from `lib/output/writer/elasticsearch.go` and `lib/output/writer/elasticsearch_test.go`. If you have any ideas, please let me know. Otherwise, I'll open a follow-up issue to dig into it in detail.
- b97877ca: Ignore SA1019 for github.com/golang/protobuf/proto: The same as above goes for switching from the deprecated `github.com/golang/protobuf/proto` to `google.golang.org/protobuf/proto`. It seems to require more work than simply changing `github.com/jhump/protoreflect/dynamic` to `google.golang.org/protobuf/types/dynamicpb`.

### DIFF
--- a/lib/input/socket_server.go
+++ b/lib/input/socket_server.go
@@ -199,6 +199,7 @@ func (t *SocketServer) sendMsg(msg types.Message) bool {
 					return
 				}
 				if !hasLocked {
+					hasLocked = true
 					t.retriesMut.RLock()
 					defer t.retriesMut.RUnlock()
 				}
--- a/lib/input/socket_server.go
+++ b/lib/input/socket_server.go
@@ -171,6 +171,8 @@ func (t *SocketServer) sendMsg(msg types.Message) bool {
 
 	// Block whilst retries are happening
 	t.retriesMut.Lock()
+	// Ignore SA2001: empty critical section
+	// nolint:staticcheck
 	t.retriesMut.Unlock()
 
 	resChan := make(chan types.Response)
@@ -239,6 +241,8 @@ func (t *SocketServer) loop() {
 		wg.Wait()
 
 		t.retriesMut.Lock()
+		// Ignore SA2001: empty critical section
+		// nolint:staticcheck
 		t.retriesMut.Unlock()
 
 		t.listener.Close()
@@ -323,6 +327,8 @@ func (t *SocketServer) udpLoop() {
 
 	defer func() {
 		t.retriesMut.Lock()
+		// Ignore SA2001: empty critical section
+		// nolint:staticcheck
 		t.retriesMut.Unlock()
 
 		close(t.transactions)
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,7 +20,7 @@ linters:
     - gosimple
     - govet
     - ineffassign
-    # - staticcheck
+    - staticcheck
     - structcheck
     - typecheck
     # - unused
--- a/lib/condition/metadata.go
+++ b/lib/condition/metadata.go
@@ -114,7 +114,7 @@ func metadataEqualsOperator(key string, arg interface{}) (metadataOperator, erro
 		return nil, fmt.Errorf("failed to parse argument as string: %v", err)
 	}
 	return func(md types.Metadata) bool {
-		return strings.ToLower(md.Get(key)) == strings.ToLower(argStr)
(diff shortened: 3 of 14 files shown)


lib/input/azure_queue_storage.go
```diff
@@ -79,6 +79,9 @@ func (a *azureQueueStorage) ReadWithContext(ctx context.Context) (msg types.Mess
 			}
 			msgIDURL := messageURL.NewMessageIDURL(queueMsg.ID)
 			_, err = msgIDURL.Delete(ctx, queueMsg.PopReceipt)
```
Jeffail, line 81: Would this `Delete` call effectively be an acknowledgement? If so this ought to be placed within the ack func that gets returned below.

lib/input/socket_server.go
```diff
@@ -89,7 +89,6 @@ type SocketServer struct {
 	listener  net.Listener
 	conn      net.PacketConn
 
-	retriesMut   sync.RWMutex
```
Jeffail, line 92: There's definitely an issue with `hasLocked` not being set to true, but I'm surprised this has been flagged as not having effect. We definitely need a lock here because it's our only back pressure mechanism that ensures the queue of retried messages doesn't swell indefinitely. I'll spend some time playing around with this on a branch.

Jeffail: Okay this is definitely a false positive, it's assuming it's ineffective because it looks like a common bug https://staticcheck.io/docs/checks#SA2001 but in our case it's necessary. The `hasLocked` part is a real bug though, it should be:

```go
				if !hasLocked {
					hasLocked = true
					t.retriesMut.RLock()
					defer t.retriesMut.RUnlock()
				}
```

It shouldn't be high impact though as it'll just mean we add some extra read locks during an exhaustive retry loop.

lib/input/azure_queue_storage.go
mfamador: Hi! This one's on me, my bad.
Hopefully, #689 will fix it.

mihaitodor: Awesome, thanks for the quick investigation! Happy to drop this change and rebase once that's merged.

mfamador: Cool, thank you @mihaitodor

lib/input/socket_server.go
mihaitodor: Thank you so much Ash for looking into this so quickly! I apologize for my initial remark where I dismissed this as a potential refactoring leftover. I sometimes assume that what I don't understand is probably wrong and I really shouldn't do that. I'm sorry!

I fixed the hasLocked flag issue in 50e461d2 and I re-added the locks and instructed statickcheck to stop warning about them in 74bf1e1e. I got misled by the fact that there is nothing happening between lock and unlock, which I thought will be ineffective. I never implemented back pressure mechanisms before, TBH, and I know they're a pain to get right.

Jeffail: No worries, personally I'd rather you lean on the side of deleting code that looks unnecessary as that's usually going to be the correct approach, and when it's not there really ought to be a test failure to catch it anyway.

lib/input/azure_queue_storage.go
mihaitodor: Rebase done (cc @Jeffail)

Jeffail:
Awesome, thanks @mihaitodor!

PR #689, opened 2021-02-27
Fix azure storage queue input acknowledge

mfamador:
As identified in #687 the input wasn't acknowledging correctly. Here's a fix.


Jeffail:
Thanks @mfamador!

PR #690, opened 2021-03-01
Fix integration tests

mihaitodor:
There is a [breaking change](https://github.com/localstack/localstack/blame/c1a02d56ff6bdd552c568c94c382da4f4a626d3d/CHANGELOG.md#L10) in localstack which prevents the AWS integration tests from running. This PR fixes this by replacing the localstack deprecated service ports with the new unified one.

I see that the emulators for the integration tests are mostly ran with the implicit `latest` tag, but since DockerHub no longer retains images forever, it's probably better to not pin them to a specific version... You wouldn't happen to maintain an internal Docker image registry proxy, would you? :)

I also took the liberty of using a runtime check to figure out when the integration tests need to be executed. While a build flag does seem less ceremonious, it prevents the CI system from detecting build failures in the integration test files, since these tests are not being run there.

I got the idea from here: https://stackoverflow.com/questions/25965584/separating-unit-tests-and-integration-tests-in-go/53450257#53450257

Quick question: With this change, `make test-integration` will run the integration tests and **only** the integration tests. Would you rather preserve the old behaviour and run **all** tests?

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,8 +3,6 @@ run:
     # Extracted with `grep -rho '\+build.*' --include \*.go . | tr -d '!' | cut -d' ' -f2 | sort | uniq`
     # Can't enable ZMQ4 because golangci-lint gets confused due to CGo deps
     # - ZMQ4
-    - fuzz
-    - integration
     # If we enable these, then files starting with `+build !windows` and
     # `+build !wasm` will not be scanned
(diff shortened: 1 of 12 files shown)


Jeffail:
Thanks @mihaitodor, looks good and I'm down like a clown frown with all the changes.

Issue #701, opened 2021-03-15
Question about checkpoint.Capped behavior


greysond:
I've been reading through the code for the [`checkpoint.Capped`](https://github.com/Jeffail/benthos/blob/master/internal/checkpoint/capped.go#L20) implementation to understand how the `checkpoint_limit` configuration worked on the `kafka` input and I _think_ I found some undesirable behavior.

In kafka, compacted topics can have some pretty wild differences in offsets from one message to the next, but I don't think `checkpoint_limit` takes this into account, which can break batching.

Here's a test that fails where I would expect `checkpoint_limit: 2` to allow two "sequential" offsets `1000` and `2000` through:
```golang
func TestBigJumpsMore(t *testing.T) {
	c := NewCapped(2)
	assert.Equal(t, 0, c.Highest())

	ctx, done := context.WithTimeout(context.Background(), time.Second*5)
	defer done()

	require.NoError(t, c.Track(ctx, 1000))
	assert.Equal(t, 0, c.Highest())

	require.NoError(t, c.Track(ctx, 2000))
	assert.Equal(t, 0, c.Highest())

	v, err := c.Resolve(1000)
	require.NoError(t, err)
	assert.Equal(t, 1000, v)
	assert.Equal(t, 1000, c.Highest())

	v, err = c.Resolve(2000)
	require.NoError(t, err)
	assert.Equal(t, 2000, v)
	assert.Equal(t, 2000, c.Highest())
}
```

I just want to make sure this behavior isn't desired before I attempt a workaround.

Jeffail:
Hey @greysond, yeah this ought to be expanded to allow for any range between sequential offsets. It might need a bit of a refactor, it's been a while since I wrote the checkpointer but I probably did it this way for a reason. The only other component that uses this is the `aws_kinesis` input, but it won't be impacted since it has a (slightly gross) wrapper around this to account for the fact that the shard iterators are arbitrarily long numbers.

greysond:
Mind if I take a crack at it? And how would you feel about a linked-list implementation?

Jeffail:
Go for it, any implementation is fine but keep an eye on the benchmarks. It doesn't need to be super efficient but it'd be nice not too regress too far.

greysond:
Should it be assumed that `Capped.Track(ctx, offset)` can be called from more than one goroutine? If not and there will only ever be one goroutine waiting on that call (like a sarama partition consumer), I have a much simpler solution.

[...]

PR #702, opened 2021-03-15
Add more flexible support for non-sequential offsets from kafka

greysond:
fix for #701


Jeffail:
Hey @greysond, unfortunately the solution here needs to be a bit more involved as the number of pending offsets alone isn't enough to limit the number of uncommitted offsets. For example, if a hypothetical offset 1 was _never_ resolved but all other offsets were then only looking at the number of pending offsets would mean we'd continue consuming unlimited without ever being able to commit offset 1 or above.

greysond:
🤦 I was wondering why it seemed too simple! I'll try to add a test case to ensure my new solution accounts for this.

greysond:
This fix now uses a linked list of intervals to represent un-resolved offsets and stores the batch size ("offset weight") of each offset if they differ from 1.

greysond:
All the tests are passing locally for me and I'm not sure how to re-run these tests.

Jeffail:
Hey @greysond, I like this direction, I think a linked list is the right solution here since we no longer have a clear relation between an integer offset (call that the "payload" of a checkpoint) and the actual range between two checkpoints. I think we ought to lean into it a bit more and totally ditch my old linear checkpointer as it's just extra complexity.

The linked list itself ought to be enough to store the distance between each checkpoint in the chain as well as the payload of the checkpoint (I might suggest storing an `interface{}` here as it could be an int or a string):

```go
type node struct {
  position int64
  payload  interface{}
  next     *node
}
```

The field `position` can be an absolute offset of the node from 0 of messages checkpointed, calculated as the current last node position plus the batch size of this node. We don't really need to worry about roll-over since this number represents discrete messages from a single partition. This would mean we can also get rid of the `offsetWeight` map.

We can then calculate the current number of messages in flight as the position of the latest node minus the position of the earliest node. When a node is removed its position is given to the previous node in the chain, if it's the earliest node then its payload becomes the checkpointed value.

The API will need to be refactored slightly to make resolving easier. I'd suggest returning a `func() interface{}` closure from `Track` that uses the pointer of the node to identify it within the chain and then remove the old `Resolve` method.

If that makes any sense and you fancy trying it out then let me know, otherwise I'll try and put a proof of concept together when I get a chance.

greysond:
I can give it a go!

greysond:
Since `checkpoint.Type` is exported, are you wanting to keep its API backwards compatible?

Jeffail:
@greysond let's abandon `checkpoint.Type` and create a new one within `internal/checkpoint`. We'll need to leave the old version there unchanged until v4 allows us to nuke it :smiling_imp:.

greysond:
Alright, the implementation was pretty close to your spec, with these differences:

- Added a `prev *node` field in `type node`

  Since the resolve func returned by `Track` needs to know the previous node for rollbacks (and that can change with node deletions), I couldn't figure out how to do it without this.

- Added a `positionOffset int64` field to `checkpoint.Type` 
  
  If the first node is removed, we need a place to store it's position. This value resets to `0` when the list is empty.

greysond:
In retrospect, you could just iterate through the linked list instead of using `node.prev` (🤦), so just let me know if you want me to change it to that!

greysond:
Apologies for all the force pushes, I'm not getting these test errors locally.

Jeffail:
No worries, at a glance this looks great, I'll dig into it soon. The errors look like flaky tests so don't worry about them unless they specifically mention the checkpointer package.

Jeffail:
Merged: https://github.com/Jeffail/benthos/commit/7cccc3002b60b71c6efd316810f15af458ff6285, I did a small bit of refactoring to keep the capped part separate from the main checkpointer, and to allow batches that are larger than the cap as long as the number of pending messages is zero. Thanks for tackling this one @greysond it's definitely a certified doozy.

I've also run comparison benchmarks, the new version is less efficient as expected (mostly due to passing `interface{}` around), but given the calls are still measured in nanoseconds and correspond to a message delivery there's no question that this is a superior solution.

Jeffail:
Actually I should mention that I've added integration tests for checkpoints, but haven't actually tried this with a compacted topic. Please reopen https://github.com/Jeffail/benthos/issues/701 if you notice any issues.

PR #703, opened 2021-03-18
Mongo cache fix

nmbryant:
I was testing Mongo components and found an issue with using a Mongo cache with SFTP input in watcher mode. I think changing it to an Add instead of Update should fix it.

### DIFF
--- a/internal/service/mongodb/cache.go
+++ b/internal/service/mongodb/cache.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


Jeffail:
@nmbryant, ah, cache implementations are meant to be relaxed during `Set` and create the key if it does not exist already, so the fix here should be in the mongodb cache instead.

nmbryant:
@Jeffail Ah thanks, added the changes to move the fix to the mongo cache and undid the change in the SFTP input.

PR #712, opened 2021-03-25
Renew Azure Service Bus lock to allow long processing time

maxmati:
In Azure service bus user can configure default lock time up to 5
minutes. If processing or batching takes longer then the lock is lost
and the message is retransmitted. In order to allow any processing or
batching time, this commit adds renewing this lock when it comes close
to the expiration time.

In our case, we would like to batch for more than 5 minutes but there are other cases as well.

Unfortunately there is no docker image for Azure Service Bus so integration tests required credentials to actually service bus running on Azure

### DIFF
--- a/lib/input/reader/amqp_1.go
+++ b/lib/input/reader/amqp_1.go
@@ -3,6 +3,9 @@ package reader
 import (
 	"context"
 	"crypto/tls"
+	"errors"
+	"fmt"
+	"math/rand"
 	"sync"
 	"time"
 
@@ -15,12 +18,18 @@ import (
 	btls "github.com/Jeffail/benthos/v3/lib/util/tls"
 )
 
+const (
+	lockRenewResponseSuffix = "-response"
+	lockRenewRequestSuffix  = "-request"
+)
+
 //------------------------------------------------------------------------------
 
 // AMQP1Config contains configuration for the AMQP1 input type.
 type AMQP1Config struct {
 	URL           string      `json:"url" yaml:"url"`
 	SourceAddress string      `json:"source_address" yaml:"source_address"`
+	RenewLock     bool        `json:"renew_lock" yaml:"renew_lock"`
 	TLS           btls.Config `json:"tls" yaml:"tls"`
 	SASL          sasl.Config `json:"sasl" yaml:"sasl"`
 }
@@ -39,15 +48,19 @@ func NewAMQP1Config() AMQP1Config {
 
 // AMQP1 is an input type that reads messages via the AMQP 1.0 protocol.
 type AMQP1 struct {
-	client   *amqp.Client
-	session  *amqp.Session
-	receiver *amqp.Receiver
+	client            *amqp.Client
+	session           *amqp.Session
+	receiver          *amqp.Receiver
+	renewLockReceiver *amqp.Receiver
+	renewLockSender   *amqp.Sender
 
 	tlsConf *tls.Config
 
 	conf  AMQP1Config
 	stats metrics.Type
 	log   log.Modular
+
+	lockRenewAddressPrefix string
 
 	m sync.RWMutex
 }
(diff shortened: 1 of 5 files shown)


lib/input/reader/amqp_1.go
```diff
@@ -193,14 +246,137 @@ func (a *AMQP1) ReadWithContext(ctx context.Context) (types.Message, AsyncAckFn,
...
 	msg.Append(part)
 
+	var done chan struct{}
+	if a.conf.RenewLock {
+		done = a.startRenewJob(amqpMsg)
+	}
+
 	return msg, func(ctx context.Context, res types.Response) error {
+		if done != nil {
+			close(done)
```
Jeffail, line 256: It might be a bit paranoid but there's no explicit guarantee that the ack func is only called once. However, it's probably overkill to use something like `sync.Once` here so I'd say setting `done = nil` after the close is enough.

lib/input/reader/amqp_1.go
```diff
@@ -193,14 +246,137 @@ func (a *AMQP1) ReadWithContext(ctx context.Context) (types.Message, AsyncAckFn,
...
+	amqpUUID := amqp.UUID(lockTokenBytes)
+
+	return &amqpUUID, nil
+}
+
+func (a *AMQP1) renewWithContext(ctx context.Context, msg *amqp.Message) (time.Time, error) {
+	a.m.RLock()
+	var r *amqp.Receiver
+	var s *amqp.Sender
+	if a.renewLockReceiver != nil {
```
Jeffail, line 324: This and below can just be `r := a.renewLockReceiver`

lib/input/reader/amqp_1.go
```diff
@@ -15,12 +18,18 @@ import (
...
+	lockRenewRequestSuffix  = "-request"
+)
+
 //------------------------------------------------------------------------------
 
 // AMQP1Config contains configuration for the AMQP1 input type.
 type AMQP1Config struct {
 	URL           string      `json:"url" yaml:"url"`
 	SourceAddress string      `json:"source_address" yaml:"source_address"`
+	RenewLock     bool        `json:"renew_lock" yaml:"renew_lock"`
```
Jeffail, line 32: Since this is Azure specific I'd suggest we prefix the field name to `azure_renew_lock`.

Jeffail (commented):
Hey @maxmati, sorry for the wait. Looking good, left a few comments. I think given the increasing number of client related pointers being managed it's probably worth refactoring that into a single object but I can manage that outside of this PR.

maxmati:
Hi @Jeffail,
Thanks for the review. I think I addressed all your comments so please take an other look.

lib/input/reader/amqp_1.go
```diff
@@ -116,6 +129,29 @@ func (a *AMQP1) ConnectWithContext(ctx context.Context) error {
...
+		renewLockSender, err := a.session.NewSender(amqp.LinkSourceAddress(a.lockRenewAddressPrefix+lockRenewRequestSuffix), amqp.LinkTargetAddress(managementAddress))
+		if err != nil {
+			receiver.Close(context.Background())
+			session.Close(context.Background())
+			client.Close()
+			return err
+		}
+		renewLockReceiver, err := a.session.NewReceiver(amqp.LinkSourceAddress(managementAddress), amqp.LinkTargetAddress(a.lockRenewAddressPrefix+lockRenewResponseSuffix))
+		if err != nil {
+			a.renewLockSender.Close(context.Background())
```
Jeffail, line 145:
```suggestion
			renewLockSender.Close(context.Background())
```

Jeffail:
Thanks, looks good @maxmati! I've added a one line suggestion just as a reminder but I can do that myself when I merge this. It might need to wait for a day or two as I'd like to put some time aside to make sure I've fully understood how this works.

Jeffail:
Merged with a small refactor to make managing the client objects a bit easier: https://github.com/Jeffail/benthos/commit/2d362b8a92e03c64978500433f149023dba08b62

@maxmati would you mind running your integration test again to make sure I haven't broken anything? Running the standard `amqp_1` tests works.

maxmati:
@Jeffail thanks for merging it, I executed integration tests on latest master and it works just fine.

PR #719, opened 2021-03-31
Use IterateBatchedSend on azure queue storage output

mfamador:
Hi!
I've made this very tiny change to use IterateBatchedSend on Azure Queue Storage output


Jeffail:
Looks good, thanks @mfamador

PR #721, opened 2021-04-01
Bail out if Azure blob storage input creds become invalid

### LINKED ISSUE #684: Exiting if creds / tokens expire after a while
Is there any mechanism at the input/output level to instruct Benthos to stop trying to read / write messages and exit if it gets an invalid / expired credentials error? I'd be interested to implement this for Azure (for now). Or can it be achieved somehow with some try / catch config?

mihaitodor:
This fixes #684 for my use case. I think we're OK for now to hang in a retry loop if the creds are not valid upfront and `ConnectWithContext` can't succeed. Since the Benthos semantics are to keep retrying to send a message if the output is blocked, I haven't changed the behaviour on the output side, so it will keep trying forever. In such cases, it's best to watch metrics and have some external coordinator take action if needed.

I looked into making a similar change for Google Cloud Storage, but it seems that the keys I'm able to generate ([Application Default Credentials](https://cloud.google.com/storage/docs/authentication#libauth)) don't support having an expiration date. I did try deleting the keys while files are still being copied from a bucket and, after a while, the client library stops reading data, but I couldn't see any way to distinguish that from a context cancelled / timeout error. Will leave this as is for now.

### DIFF
--- a/lib/input/azure_blob_storage.go
+++ b/lib/input/azure_blob_storage.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"net/url"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


mihaitodor:
Hmmm... Struggling to replicate that test (`go test -v -run ^TestSequenceSad$ -count 2 -race ./lib/input`) crash on localhost, but no luck. I don't think it's my fault, but I'm looking to see why it's flakey.

Jeffail:
Don't worry about the sequences they don't own us and they certainly can't tell us what to do.

PR #724, opened 2021-04-06
Add path to HTTP server input metadata

### LINKED ISSUE #720: Capture path in http_server input
Is it possible to extract parts of the path of http_server input?
eg.

input:
  http_server:
    address: ""
    path: /post
    allowed_verbs:
      - GET
    timeout: 5s
    rate_limit: ""

I would like to call, http://localhost:4195/post/foo/123456 and to use the parts of the path ("foo" and "123456") in the following processing steps.

Thanks for your help.

mihaitodor:
Fixes #720.

Hope this is in line with what you'd expect to have @abasse. For getting the path components, something like `meta("http_server_request_path").split("/")` can be used.

I also added a basic test for the HTTP Server input.

### DIFF
--- a/lib/input/http_server.go
+++ b/lib/input/http_server.go
@@ -93,6 +93,7 @@ This input adds the following metadata fields to each message:
 
 ` + "``` text" + `
 - http_server_user_agent
+- http_server_request_path
 - All headers (only first values are taken)
 - All query parameters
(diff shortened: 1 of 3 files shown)


Jeffail:
Nice, thanks @mihaitodor!

PR #728, opened 2021-04-08
Use acknowledge `context` from parameter to prevent expired context error

mfamador:
Hi!
The ack function was using the read context instead of using the one from the parameters.
It would make the acks fail with context expired when the messages take too long to acknowledge.


Jeffail:
Looks good, thanks @mfamador!

PR #731, opened 2021-04-12
Add `dequeue_visibility_timeout` and `max_inflight` parameters to azure queue storage input

mfamador:
Hi!

I've been using this input as a "parent" of another custom input.
This custom input sometimes takes longer than the default 30s dequeue visibility timeout used to hide the messages from other consumers while being processed, which makes the message show up in the queue again before the first iteration acknowledges it. I've added the parameter `dequeue_visibility_timeout` to allow us to increase this timeout, keeping the 30s default as previously.

It was also added the `max_inflight` parameter to set the maximum number of unprocessed messages to fetch at a given time.

Cheers,
Marco

### DIFF
--- a/lib/input/azure_queue_storage_config.go
+++ b/lib/input/azure_queue_storage_config.go
@@ -51,6 +51,8 @@ Only one authentication method is required, ` + "`storage_connection_string`" + 
 			docs.FieldCommon(
 				"queue_name", "The name of the target Storage queue.",
 			),
+			docs.FieldAdvanced("dequeue_visibility_timeout", "The timeout duration until a dequeued message gets visible again, 30s by default"),
+			docs.FieldAdvanced("max_inflight", "The maximum number of unprocessed messages to fetch at a given time."),
 		},
 		Categories: []Category{
 			CategoryServices,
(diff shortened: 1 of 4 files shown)


lib/input/azure_queue_storage_config.go
```diff
@@ -64,15 +66,20 @@ Only one authentication method is required, ` + "`storage_connection_string`" +
...
-	StorageSASToken         string `json:"storage_sas_token" yaml:"storage_sas_token"`
-	StorageConnectionString string `json:"storage_connection_string" yaml:"storage_connection_string"`
-	QueueName               string `json:"queue_name" yaml:"queue_name"`
+	StorageAccount           string `json:"storage_account" yaml:"storage_account"`
+	StorageAccessKey         string `json:"storage_access_key" yaml:"storage_access_key"`
+	StorageSASToken          string `json:"storage_sas_token" yaml:"storage_sas_token"`
+	StorageConnectionString  string `json:"storage_connection_string" yaml:"storage_connection_string"`
+	QueueName                string `json:"queue_name" yaml:"queue_name"`
+	DequeueVisibilityTimeout string `json:"dequeue_visibility_timeout" yaml:"dequeue_visibility_timeout"`
+	MaxInFlight              int32  `json:"max_inflight" yaml:"max_inflight"`
```
Jeffail, line 75: Minor nit pick: this should be `max_in_flight` in order to match similar fields elsewhere.

mfamador: Ok, sorry, I've looked at the `nats`, it has it like this, I'll change it in 1 min

Jeffail:
Thanks @mfamador, looks good, left one minor comment to rename the field to `max_in_flight`.

lib/input/azure_queue_storage_config.go
mfamador: It's fixed. Thanks, @Jeffail

Jeffail:
Don't worry about the test, looks like a flaky one.

PR #730, opened 2021-04-11
Enable gocritic

mihaitodor:
This is the first small batch of changes for enabling the [`gocritic` linter](https://github.com/go-critic/go-critic) (part of #626).

One notable change is 1a4e15c. It threw me off a bit, but, as, far as I can tell, that backslash is indeed superfluous. At least https://regexr.com/ sees them as identical.

Please note that, due to [this bug](https://github.com/go-critic/go-critic/issues/1028) that got fixed in gocritic, you'll need a recent version of golangci-lint. v1.39.0 definitely has the fix.

I have more incoming changes (see below), but I think it's best to group them in smaller batches and send them as separate PRs after this one gets merged. Some of them might be a bit too nit-picky / annoying and you might not like them :)

<details>
  <summary>Remaining changes</summary>

```  
> git log --shortstat --oneline
01fbddfe Fix deprecatedComment: use `Deprecated: ` (note the casing) instead of `DEPRECATED: `
 3 files changed, 3 insertions(+), 3 deletions(-)
6056e810 Fix appendAssign: append result not assigned to the same slice
 1 file changed, 1 insertion(+)
fe56dba1 Fix typeUnparen: could simplify (...) to ...
 10 files changed, 11 insertions(+), 11 deletions(-)
89ecb392 Fix valSwap: can re-write as `array[indexOne], array[indexTwo] = array[indexTwo], array[indexOne]`
 1 file changed, 1 insertion(+), 3 deletions(-)
3dd5934c Fix exitAfterDefer: log.Fatal will exit, and `defer fSync.write()` will not run
 1 file changed, 2 insertions(+), 2 deletions(-)
83969818 Fix unnamedResult: consider giving a name to these results
 13 files changed, 27 insertions(+), 31 deletions(-)
995096b5 Fix yodaStyleExpr: consider to change order in expression to ...
 14 files changed, 78 insertions(+), 70 deletions(-)
b1af51d7 Fix unlambda: replace `func() hash.Hash { return ...() }` with `...`
 1 file changed, 2 insertions(+), 3 deletions(-)
acc2ad96 Fix paramTypeCombine: func(...)
 23 files changed, 27 insertions(+), 27 deletions(-)
bc8409e6 Fix wrapperFunc: use strings.ReplaceAll method in `strings.Replace(..., ..., ..., -1)`
 11 files changed, 22 insertions(+), 22 deletions(-)
4b706a8a Fix unlabelStmt: label ... is redundant
[...]

PR #732, opened 2021-04-12
More gocritic

mihaitodor:
Follow-up to #730 (part of #626).

There's a bunch of interesting stuff in here:

- Potential bug: 9f24fd4. I re-ran the HDFS integration tests (had to comment out the skip) and they seemed OK. I hope this change makes sense.
- 4befa62: that append and assignment to another slice seems correct, but might be worth looking at to make sure. Since this is an uncommon pattern, it's probably safer to have this linter enabled and silence it where needed.

LE: Looks like the tests fail because of some flakiness in `TestSyncBatcherTimeout`. I can replicate it by running `go test -count 1 -run TestSyncBatcherTimeout ./lib/input/reader` if I add a `time.Sleep(1 * time.Second)` at the beginning of the [goroutine](https://github.com/Jeffail/benthos/blob/e44e871b7fdec605d61541e3960bce41487699b4/lib/input/reader/sync_batcher_test.go#L278) from the test. I'm almost sure that the issue is the `time.Millisecond` timeout [here](https://github.com/Jeffail/benthos/blob/e44e871b7fdec605d61541e3960bce41487699b4/lib/input/reader/sync_batcher_test.go#L260) is too long and it gets a chance to do a retry [here](https://github.com/Jeffail/benthos/blob/50f99bd8ee73753e71b4b4d51368c741028938b2/lib/input/reader/sync_batcher.go#L94-L102), which makes it get stuck in the subsequent call to `mockSyncReader.ReadNextWithContext`. Will need to go through it in more detail to figure out a fix.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -22,32 +22,18 @@ linters-settings:
       - importShadow # Probably not worth the hassle...
       # TODO
       - sloppyReassign
-      - singleCaseSwitch
-      - regexpMust
-      - commentFormatting
-      - captLocal
       - elseif
-      - commentedOutCode
-      - emptyFallthrough
-      - typeAssertChain
-      - initClause
-      - unslice
       - appendCombine
       - emptyStringTest
       - assignOp
-      - octalLiteral
       - builtinShadow
       - unlabelStmt
       - wrapperFunc
       - paramTypeCombine
-      - unlambda
       - yodaStyleExpr
       - unnamedResult
-      - exitAfterDefer
-      - valSwap
       - typeUnparen
       - appendAssign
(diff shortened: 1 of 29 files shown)


Jeffail:
Thanks @mihaitodor, this all looks good. https://github.com/Jeffail/benthos/commit/4befa62aaf8c2b7da5ee255d59d97e292141260d definitely looks wrong but it's been ages since I last touched that stuff so I might be missing something. However, let's leave it as is as it's deprecated code and that particular part should never be hit anyway, if sqs isn't configured we handle the nacks downstream.

PR #738, opened 2021-04-15
Adds support to aws_dynamodb output for deleting items

cludden:
**What**
this PR adds support for deleting items to the `aws_dynamodb` output via a new `delete_mapping` bloblang mapping config field.

**Why**
For standard change-data-capture use cases, where there is a desire to sync data (including deletion markers) to DynamoDB.

@Jeffail, curious to hear your thoughts on this approach..

### DIFF
--- a/config/aws_dynamodb.yaml
+++ b/config/aws_dynamodb.yaml
@@ -25,6 +25,7 @@ output:
     ttl: ""
     ttl_key: ""
     max_in_flight: 1
+    delete_mapping: ""
     batching:
       count: 0
       byte_size: 0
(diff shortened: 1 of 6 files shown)


Jeffail:
Hey @cludden, I'd eventually like to deprecate the `string_columns` and `json_map_columns` fields and replace them with a bloblang mapping, similar to how to the new `mongodb` output works. Maybe we can use this as a starting point where your new `delete_mapping` field becomes the key map itself, so you could configure something like:

```yaml
output:
  aws_dynamodb:
    table: foobar
    delete_mapping: |
      root.id = this.document.id
      root.key = meta("kafka_key")
    max_in_flight: 10
```

Then eventually we'd add a field `put_mapping` with similar behaviour:

```yaml
output:
  aws_dynamodb:
    table: foobar
    put_mapping: |
      root.id = this.document.id
      root.key = meta("kafka_key")
      root.name = this.user.name
      root.topic = meta("kafka_topic")
    max_in_flight: 10
```

Also it's now been brought to my attention that the mongodb output fields have a `_map` suffix, but we have `mapping` fields elsewhere so maybe that should change. It's experimental so if there's stuff we learn with this output then we can update mongodb to follow it.

cludden:
closing in favor of #753

PR #741, opened 2021-04-18
Gocritic more extras

mihaitodor:
Here are the final bits of changes that I think are worth making as advised by gocritic. Sorry Yoda! 😢 

- `unnamedResult` is intended to capture [these rules](https://github.com/golang/go/wiki/codereviewcomments#named-result-parameters) from the official guidelines, however, I'm not sure how they came up with [the idea](https://github.com/go-critic/go-critic/blob/f599c814bd8d49e04e1ba5d1ced26782127e42f5/checkers/unnamedResult_checker.go#L55-L56) that the last type needs to be either `error` or `bool` to allow unnamed parameter returns. I might try to raise an issue upstream.
- `builtinShadow` is a bit over the top, I think, but will let you decide.
- `typeUnparen` made me wonder why `docs.SanitiseComponentConfig` requires that `map[string]interface{}` cast. I read a bit through the code, but it's hard to tell what the right design should be there.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -22,10 +22,6 @@ linters-settings:
       - importShadow # Probably not worth the hassle...
       # TODO
       - emptyStringTest
-      - builtinShadow
-      - yodaStyleExpr
-      - unnamedResult
-      - typeUnparen
 
 linters:
   disable-all: true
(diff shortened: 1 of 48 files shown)


Jeffail:
Yeah it all looks good to me. The `docs.SanitiseComponentConfig` type is deprecated and nothing should be actively using it any more, the cast was so that the called function could infer the `interface{}` param directly to `map[string]interface{}`, it'll get deleted in V4 once all the old sanitise APIs can be removed.

PR #744, opened 2021-04-20
Do not set `Timestamp` explicitly

mfamador:
Hi.
Just a very tiny change. 
We don't need to set the timestamp explicitly here since it's updated automatically when not set.
The way it is right now I imagine it can create some discrepancies when the clock is not correctly synced on the instance running benthos.


Jeffail:
Sounds good, thanks @mfamador!

PR #747, opened 2021-04-22
Adds reverse and index_of bloblang string methods

nmbryant:


Jeffail:
Looks good, thanks @nmbryant!

PR #757, opened 2021-05-04
Add format_timestamp_strftime and parse_timestamp_strptime bloblang methods

mihaitodor:
Note that `github.com/itchyny/timefmt-go` is already used indirectly through `github.com/itchyny/gojq`, so no new dependencies had to be added.

Annoyingly, this doesn't cover all the popular date layouts. 'YYYY-MM-dd HH:mm:ss' is a Java [SimpleDateFormat](https://docs.oracle.com/javase/10/docs/api/java/text/SimpleDateFormat.html) layout, which isn't covered by `strftime` / `strptime`. I found some libs for dealing with this format (see below), but not sure if it makes sense to add even more `format_timestamp` / `parse_timestamp` functions.

- https://github.com/viant/toolbox/blob/master/time_format.go (this one seems to support a wider variety of format specifiers)
- https://github.com/SumoLogic/sumo-otc/blob/main/datetranslate/java2go.go
- https://github.com/vigneshuvi/GoDateFormat
- https://github.com/newm4n/go-dfe


Jeffail:
Thanks @mihaitodor, I think we've got pretty decent coverage at this point so let's leave it as is for now. If there's enough demand out there for more format/parse syntaxes then I think it'd be worth reconsidering how these date methods can co-exist.

PR #772, opened 2021-05-23
Sign JSON Web Token (JWT) 

manute:
https://github.com/Jeffail/benthos/issues/769

### DIFF
--- a/lib/util/http/auth/jwt.go
+++ b/lib/util/http/auth/jwt.go
@@ -43,8 +43,9 @@ func (j JWTConfig) Sign(req *http.Request) error {
 		return nil
 	}
 
-	// Must parse private key only once
-	j.parsePrivateKey()
+	if err := j.parsePrivateKey(); err != nil {
+		return err
+	}
 
 	var bearer *jwt.Token
 	switch j.SigningMethod {
@@ -67,7 +68,7 @@ func (j JWTConfig) Sign(req *http.Request) error {
 	return nil
 }
 
-// parsePrivateKey parses only once the RSA private key.
+// parsePrivateKey parses once the RSA private key.
 // Needs mutex locking as Sign might be called by parallel threads.
 func (j JWTConfig) parsePrivateKey() error {
 	j.rsaKeyMx.Lock()
--- a/lib/util/http/auth/jwt.go
+++ b/lib/util/http/auth/jwt.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"sync"
 
 	"github.com/dgrijalva/jwt-go"
 )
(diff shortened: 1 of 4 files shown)


manute:
## WIP 
 First Implementation to see if the approach I took is the correct one.
Also I would like to test it but I could not see other tests for similar cases like OAuth/2, so I was not sure how to test it. 
Any feedback is really appreciated 👍

lib/util/http/auth/jwt.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// internal private fields
+	rsaKey *rsa.PrivateKey
+}
+
+// NewJWTConfig returns a new JWTConfig with default values.
+func NewJWTConfig() JWTConfig {
+	return JWTConfig{
+		Enabled:       false,
+		Claims:        nil,
```
Jeffail, line 29: This should be initialized as an empty map: `map[string]interface{}{}`, it's not a big deal but makes generated configs look a bit neater.

lib/util/http/auth/jwt.go
```diff
@@ -0,0 +1,77 @@
...
+//------------------------------------------------------------------------------
+
+// Sign method to sign an HTTP request for an JWT exchange.
+func (j JWTConfig) Sign(req *http.Request) error {
+	if !j.Enabled {
+		return nil
+	}
+
+	// Must parse private key only once
+	if j.rsaKey == nil {
```
Jeffail, line 44: This will need mutex locking as we can't guarantee `Sign` isn't called by parallel threads.

lib/util/http/auth/jwt.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+//------------------------------------------------------------------------------
+
+// JWTConfig holds the configuration parameters for an JWT exchange.
+type JWTConfig struct {
+	Enabled       bool          `json:"enabled" yaml:"enabled"`
+	Claims        jwt.MapClaims `json:"claims" yaml:"claims"`
+	SigningMethod string        `json:"signing_method" yaml:"signing_method"`
+	PrivateKey    string        `json:"private_key" yaml:"private_key"`
```
Jeffail, line 19: It often helps to make the fact that this is a file path explicit in the field name, so something like `private_key_file`.

Jeffail (commented):
Hey @manute, a couple of comments but this looks fine to me.

manute:
@Jeffail I think I've addressed all the feedback, thanks

Jeffail:
Thanks @manute, this looks great. The pipeline will fail as the docs need generating with `make docs`, but as long as the unit tests pass I'll merge and do that myself.

mihaitodor:
@manute @Jeffail I know this was merged, but I had a look at that JWT library and it looks like they have a preview release for a CVE as described [here](https://github.com/dgrijalva/jwt-go/issues/463). Not sure if it's an issue for this use case. From the comments, it looks like form3 adopted [a fork](https://github.com/form3tech-oss/jwt-go) of it and made a bunch of security fixes. Then again, it seems [this](https://github.com/golang-jwt/jwt) will be the new home for this library based on [this conversation](https://github.com/dgrijalva/jwt-go/issues/462).

manute:
@mihaitodor thanks for the heads up. I believe this should be addressed in other PR, and when the official new home for this lib has added that sec issue. Because base on its [README](https://github.com/golang-jwt/jwt#readme) : 
> NEW VERSION COMING: There have been a lot of improvements suggested since the version 3.0.0 released in 2016. I'm working now on cutting two different releases: 3.2.0 will contain any non-breaking changes or enhancements. 4.0.0 will follow shortly which will include breaking changes. See the 4.0.0 milestone to get an idea of what's coming. If you have other ideas, or would like to participate in 4.0.0, now's the time. If you depend on this library and don't want to be interrupted, I recommend you use your dependency mangement tool to pin to version 3.

mihaitodor:
I agree. I opened an issue to keep track of this situation: #779.

PR #773, opened 2021-05-24
Adds 'total bytes' metric to AWS S3 input

nmbryant:
Adds 'total bytes' metric that records the total number of bytes that the AWS S3 input will process.

### DIFF
--- a/lib/input/aws_s3.go
+++ b/lib/input/aws_s3.go
@@ -179,7 +179,7 @@ func newS3ObjectTarget(key, bucket string, notificationAt time.Time, ackFn codec
 }
 
 type s3ObjectTargetReader interface {
-	Pop(ctx context.Context) (*s3ObjectTarget, error)
+	Pop(ctx context.Context, stats metrics.Type) (*s3ObjectTarget, error)
 	Close(ctx context.Context) error
 }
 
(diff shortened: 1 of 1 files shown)


Jeffail:
Hey @nmbryant, my intention is to (eventually) have component metrics quite restricted and consistent so I'd rather not introduce new metrics, especially ones unique to a particular input. I believe you should be able to implement this in config by using a metric processor:

```yaml
input:
  label: ""
  aws_s3:
    bucket: foo
  processors:
    - metric:
        type: counter_by
        name: input.total.bytes
        value: "${! content().length() }"
```

nmbryant:
@Jeffail Gotcha, that makes sense especially since this metric isn't feasible to track in some of the inputs. Thanks for the config, I'll have to test to see how it behaves. Wanted the total bytes metric to be updated when the input makes the connection so that we can make an estimate on how long the process will take for a dashboard.

Jeffail:
Ohhh right, sorry I missed that. In that case my config won't work as a replacement on its own. We could potentially solve this by adding a new metadata field to messages of an object that includes the object size: https://www.benthos.dev/docs/components/inputs/aws_s3#metadata and then we could set a gauge metric by that value.

nmbryant:
Cool, I'll look into adding it as a metadata field

PR #775, opened 2021-05-25
Use SubscribeMultiple instead of Subscribe for MQTT input

nenad:
This PR handles cases where the broker doesn't work properly with multiple `Subscribe` commands but works with `SubscribeMultiple`.

This is the case with the RabbitMQ broker with MQTT plugin. It ignores subsequent subscriptions of a single client ID.

For example, from this configuration:
```
topics:
- "AU/+/+/+/+/+/Induct"
- "CA/+/+/+/+/+/Induct"
- "CH/+/+/+/+/+/Induct"
- "DE/+/+/+/+/+/Induct"
- "NL/+/+/+/+/+/Induct"
- "NZ/+/+/+/+/+/Induct"
- "SE/+/+/+/+/+/Induct"
- "UK/+/+/+/+/+/Induct"
```

Only the first topic would be handled for that Client ID and not the others, however they will have subscriptions registered. Then, the MQTT library will say the following for any message arriving on a non `AU/+/+/+/+/+/Induct` topic.
```
[DEBUG] [router]   matchAndDispatch received message and no handler was available. Message will NOT be acknowledged.
[DEBUG] [router]   matchAndDispatch received message and no handler was available. Message will NOT be acknowledged.
[DEBUG] [router]   matchAndDispatch received message and no handler was available. Message will NOT be acknowledged.
```
This blocks the broker from sending new messages and it starts accumulating them.

### DIFF
--- a/lib/input/reader/mqtt.go
+++ b/lib/input/reader/mqtt.go
@@ -135,23 +135,26 @@ func (m *MQTT) ConnectWithContext(ctx context.Context) error {
 			m.log.Errorf("Connection lost due to: %v\n", reason)
 		}).
 		SetOnConnectHandler(func(c mqtt.Client) {
+			topics := make(map[string]byte)
 			for _, topic := range m.conf.Topics {
-				tok := c.Subscribe(topic, byte(m.conf.QoS), func(c mqtt.Client, msg mqtt.Message) {
-					msgMut.Lock()
-					if msgChan != nil {
-						select {
-						case msgChan <- msg:
-						case <-m.interruptChan:
-						}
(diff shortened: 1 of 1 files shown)


nenad:
I've converted this to draft as I'm investigating `concurrent writes to map` panics in `bloblang` package.

Jeffail:
@nenad, that looks unrelated to your PR, if you're running an older branch then it's likely to be this: https://github.com/Jeffail/benthos/commit/70e4a9b68e5a1fbf1e0fa60937606389ae48e759

Jeffail:
Awesome, this looks great, thanks @nenad!

PR #783, opened 2021-06-01
Adds total bytes metadata field to AWS S3 messages

nmbryant:
Adds total bytes metadata that records the total number of bytes that are to be processed by the S3 input. Needed to add getTotalBytes method to sqsTargetReader and just have it return 0, but it doesn't add the metadata to messages since I don't think the total byte stat is relevant for SQS.

### DIFF
--- a/lib/input/aws_s3.go
+++ b/lib/input/aws_s3.go
@@ -76,6 +76,7 @@ This input adds the following metadata fields to each message:
 - s3_last_modified (RFC3339)
 - s3_content_type
 - s3_content_encoding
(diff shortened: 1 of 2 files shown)


Jeffail:
Hey @nmbryant, still thinking this one over. My main problem is that the value will potentially increase over time whilst the items in the bucket are walked, it's hard to describe this field in a way that makes it easy to understand what it is and how it's generally useful.

nmbryant:
No worries, I think you are right. It would probably end up being more confusing than helpful, so I'll close this PR.

PR #789, opened 2021-06-02
Update log level for non-final retries in Retry output

elliotwms:
This PR updates the logging behaviour in the Retry output type so that
* Non-final failures are logged as warnings
* Final failures are logged as errors

Closes https://github.com/Jeffail/benthos/issues/770

I haven't added tests for logging behaviour as we don't seem to have that elsewhere, but can add if needed.


Jeffail:
Thanks @elliotwms, looks good and makes sense.

Issue #786, opened 2021-06-01
Add `incrby` operator for redis cache


kvalv:
I suggest adding `incrby`  as an operator to the [redis processor](https://www.benthos.dev/docs/components/processors/redis/#operators).

**use case**
`incrby` can be used for counting, of which there are many situations where counting is involved. An example, let's say we have this data:
```
{"name": "morkel", "owes": "government", "amount": 100}
{"name": "ash", "owes": "morkel", "amount": 200}
{"name": "morkel", "owes": "elizabeth", "amount": 300}
```
If we want the total obligation for each `name`, we can simply do what's described in the [#real counter ](https://www.benthos.dev/docs/configuration/windowed_processing/#real-counter) documentation. However, if we want the running obligation we could use a cache. The desired output:
```
{"amount":100,"name":"morkel","owes":"government","running_debt":100}
{"amount":200,"name":"ash","owes":"morkel","running_debt":200}
{"amount":300,"name":"morkel","owes":"elizabeth","running_debt":400}
```

Right now, we would have to do something like this:

```yaml
pipeline:
  processors:
    - branch:
        request_map: |
          meta name = this.name
          meta amount = this.amount
        processors:
          - cache:
              resource: redis
              operator: get
              key: '${!meta("name")}'
          - catch:
              - noop: {}
          - bloblang: 'root = (this | 0) + meta("amount").number()'
          - cache:
              resource: redis
              operator: set
              key: '${!meta("name")}'
              value: '${!json()}'
        result_map: |
          root.running_debt = this
```

... which is awfully many lines of code. If we have `incrby` we can reduce to this:

```yaml
pipeline:
  processors:
    - branch:
        request_map: 'root = this'
        processors:
          - redis:
              operator: incrby
              key: 'json("name")'
              value: 'json("amount")'
        result_map: |
          root.running_debt = this
```
(there's probably better ways at solving this problem, but I hope I am able to show the utility of `incrby` in situations where running statistics are useful).

**Issues and notes**
* If the key is the same across runs, then on the first run (assuming the key does not exist) the initial value will be 0, and on the next run the same key will be used (assuming the key is not cleared before the next run) and a different value will be produced. This might be a bug depending on the situation. The documentation should probably highlight this potential issue. (On second thought, this same issue would also be present on the "current implementation", so this request wouldn't actually introduce this issue)
* The returned value of the `redis` processor should be the new value (just like described in the [redis docs](https://redis.io/commands/INCRBY))
[...]

PR #790, opened 2021-06-03
Redis incrby

kvalv:
Implements #786 

Please don't hesitate to nitpick if there's anything that should be changed. I don't have much experience with go either so there might be some things that isn't implemented optimally. The integration test runs fine for me, however, and the example also works, so I don't see any immediate issues.

Note 9266c5e isn't really related to this PR, but I guess you want that typo fixed.


Jeffail:
Thanks @kvalv looks great, bonus points for adding an example and double bonus points for fixing the typo.

PR #793, opened 2021-06-05
Enable gosimple and stylecheck linters

mihaitodor:
I noticed a discussion [here](https://github.com/golang/go/issues/38968) regarding the deprecation of golint and thought this might start causing weird inconsistencies soon. In the [golint repo](https://github.com/golang/lint), they say that `staticcheck` and `go vet` should be used instead and I also noticed that [`stylecheck`](https://github.com/dominikh/go-tools/tree/master/stylecheck) is branded as a "golint replacement".

I opened this PR to see what your thoughts are around dropping `golint`.


Jeffail:
Thanks @mihaitodor, yeah I'm definitely happy to drop it as long as we're covered by stylecheck.

PR #795, opened 2021-06-09
Moving to `golang-jwt/jwt`

oxisto:


Jeffail:
Awesome, thanks @oxisto!

PR #801, opened 2021-06-18
Add missing close calls

mihaitodor:
I think these might be redundant because the compressors and decompressors go out of scope immediately, but the documentation does say for example for `flate.NewReader` that "it is the caller's responsibility to call Close on the ReadCloser when finished reading": https://golang.org/pkg/compress/flate/#NewReader

Also:
- Switch from `outBuf.ReadFrom` to idiomatic `io.Copy` to skip `io.EOF` check. See https://golang.org/pkg/io/#Copy
- Do some small cleanup


Jeffail:
Nice, looks good, thanks @mihaitodor

PR #805, opened 2021-06-19
Skip scheduled GitHub actions in forks

mihaitodor:
With this change, forks of this repo should not attempt to run any scheduled GitHub actions. I hope!

I pushed this change to my master branch and will check if it does the right thing.

LE: Looks OK on my end: https://github.com/mihaitodor/benthos/actions
image


Jeffail:
Nice, thanks @mihaitodor!

PR #807, opened 2021-06-22
Use background context to avoid closing connection

nicktelford:
This was causing the `http_client` input to consistently fail to connect
to a remote server.


Jeffail:
Thanks bae :heart:

PR #798, opened 2021-06-13
Add MSSQL driver

mihaitodor:
Fixes #797.

I tested the output like so (might be worth adding some integration tests for this too, since only Postgres is covered  under `lib/test/integration`:

```shell
localhost> docker run --rm -it -e ACCEPT_EULA=Y -e SA_PASSWORD=ins4n3lyStrongP4ssword -p1433:1433 microsoft/mssql-server-linux
```

```shell
container> /opt/mssql-tools/bin/sqlcmd -l 30 -S localhost -h-1 -V1 -U sa -P ins4n3lyStrongP4ssword -Q "create table footable(data varchar(50) not null);"
```

```yaml
input:
  stdin:
    codec: lines
output:
  sql:
    driver: mssql
    data_source_name: "sqlserver://sa:ins4n3lyStrongP4ssword@localhost:1433?database=master"
    query: "insert into footable(data) values(?);"
    args:
      - ${! json("data") }
```

```shell
localhost> ./target/bin/benthos -c sql.yaml
{"data":"foo"}
```

```shell
container> /opt/mssql-tools/bin/sqlcmd -l 30 -S localhost -h-1 -V1 -U sa -P ins4n3lyStrongP4ssword -Q "select * from footable;"
foo

(1 rows affected)
```

PR #811, opened 2021-06-30
Fix MSSQL invalid data_source_name

mihaitodor:
This is a follow-up to the discussion in #798.

I think this change is more user friendly than documenting that users need to URL-encode the backslash in the username part of the `data_source_name`.

However, it's a hack and there might be other corner cases that we don't know about... Maybe this should be fixed in the `github.com/denisenkom/go-mssqldb` library instead?

### DIFF
--- a/lib/output/sql.go
+++ b/lib/output/sql.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


Jeffail:
Hey @mihaitodor, yeah that seems like a reasonable solution for now. Might be worth letting the upstream repo know if it's something they might be better off also handling themselves.

mihaitodor:
Yep, there's an issue about it https://github.com/denisenkom/go-mssqldb/issues/441 and I left a comment there to ask. Thanks! :)

PR #813, opened 2021-07-01
Add compaction_on_read option for memory cache

### LINKED ISSUE #812: Feature: Add compaction_on_read option for memory cache
I am busy with a flow where I am required to renew credentials every 15 minutes.

My current approach is to run a processor pipeline on every request, atempting to read from memory, and if the key does not exist, to renew credentials.

```yaml
pipeline:
  processors:
    - resource: get_memory_auth
    - catch:
        - resource: "do_auth_"
        - resource: "set_memory_auth"
        - resource: "get_memory_auth"
```

After reading the documentation more closely, I noticed that compaction only happens on writes, and not reads.

To solve this problem, I would like to add an optional argument `compaction_on_read` to allow for compaction to also happen on read events. This does not break backwards compatibility. The new default memory cache config will then look like this:

```yaml
memory:
  ttl: 300
  compaction_interval: "60s"
  compaction_on_read: false
  init_values: {}
  shards: 1
```

If you are happy with the feature request, I am more than happy to implement it.

hjblom:
Resolves #812
* Added `compaction_on_read` option to memory cache.
* Added unit test
* Updated documentation

~~Refactored `shouldCompact` logic to be reused within the `Get` function, to avoid unnecessarily RW locking.~~

### DIFF
--- a/lib/cache/memory.go
+++ b/lib/cache/memory.go
@@ -25,7 +25,8 @@ compaction.`,
 		Description: `
 A compaction only occurs during a write where the time since the last compaction
 is above the compaction interval. It is therefore possible to obtain values of
-keys that have expired between compactions.
+keys that have expired between compactions. If ` + "`compaction_on_read`" + `is
+set, compaction will also happen on read events.
 
 The field ` + "`init_values`" + ` can be used to prepopulate the memory cache
 with any number of key/value pairs which are exempt from TTLs:
@@ -42,6 +43,7 @@ TTL is respected as usual.`,
 		FieldSpecs: docs.FieldSpecs{
 			docs.FieldCommon("ttl", "The TTL of each item in seconds. After this period an item will be eligible for removal during the next compaction."),
 			docs.FieldCommon("compaction_interval", "The period of time to wait before each compaction, at which point expired items are removed."),
+			docs.FieldAdvanced("compaction_on_read", "Whether compaction should occur on read events."),
 			docs.FieldAdvanced("shards", "A number of logical shards to spread keys across, increasing the shards can have a performance benefit when processing a large number of keys."),
 			docs.FieldString(
 				"init_values", "A table of key/value pairs that should be present in the cache on initialization. This can be used to create static lookup tables.",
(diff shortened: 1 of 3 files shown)


lib/cache/memory.go
```diff
@@ -114,6 +116,16 @@ func (s *shard) compaction() {
 	s.lastCompaction = time.Now()
 }
 
+func (s *shard) shouldCompact() bool {
+	if s.compInterval == 0 {
+		return false
+	}
+	if time.Since(s.lastCompaction) < s.compInterval {
```
Jeffail, line 123: If we aren't protecting `lastCompaction` with a lock then it'll need to be made into an atomic otherwise the write at the end of compaction will race.

Jeffail (commented):
Thanks @hjblom! One small change needed but other than that I think we're good.

lib/cache/memory.go
hjblom: Nice find! 

After reading over the code a few times, the case where a race occurs should not be a problem as the compaction function should imediatly return within this section. (Under the safety of an rw lock)
```go
if shard.compOnRead && shard.shouldCompact() {
   shard.Lock()
   shard.compaction()
   shard.Unlock()
}
```
I am happy to remove `shard.shouldCompact()` from the `if` statement. The only downfall then is that we will always aquire a write lock on a read event in the case that `compact_on_read` is active. Keeping the current approach will 99% of the time only aquire a read lock and and for the 1% of the time aquire a write lock and imediatly return.

What do you think?

Jeffail: I'm pretty sure it'll still race as there could be two concurrent operations where one is actively compacting and has the lock, and another enters `Get` and reads `lastCompaction` at the point where the other writes to it. Using an atomic would let you access `shouldCompact` lock free so the hot path would still be fast, might look at bit ugly though as you'd need to convert to/from `time.Time`.

Another alternative that (I think) would also solve your problem would be to check the TTL on items during `Get`, so if the item exists we also follow with:

```go
if !k.ts.IsZero() && time.Since(k.ts) >= m.ttl {
  return nil, types.ErrKeyNotFound
}
```

And therefore if it has expired but we haven't compacted it yet we still treat it as though it doesn't exist.

hjblom: Great sugestion! I went with the second approach as it keeps the code clean and simplifies the locking logic.

Jeffail:
Awesome, thanks @hjblom! I might possibly remove the `compaction_on_read` config field and enable this by default as in my opinion it's a bug to rely on the internals to access expired keys intentionally.

PR #814, opened 2021-07-01
Add nats input replies

### LINKED ISSUE #513: NATS input replies
The NATS input has a stub for acknowledgements, yet the NATS system supports replies creating an acknowledgment system. From the [NATS documentation](https://docs.nats.io/nats-concepts/acks): 
> When you are publishing an event or data stream, one way to ensure message delivery is to turn it into a request-reply with the concept of an acknowledgement message, or ACKs. In NATS an ACK can simply be an empty message, a message with no payload.

So may I suggest that we add a conditional ACK if the `*nats.Msg` has a `Reply` string? Based on the Nats documentation and a quick peek at the code it would just be a `n.natsConn.Publish(msg.Reply, nil)`. That way a NATS publisher could confirm a message was processed in it's entirety.

Thanks!

seanlaff:
Resolves  #513

### DIFF
--- a/lib/input/reader/nats.go
+++ b/lib/input/reader/nats.go
@@ -193,7 +193,7 @@ func (n *NATS) ReadWithContext(ctx context.Context) (types.Message, AsyncAckFn, 
 		if res.Error() != nil {
 			return msg.Nak()
 		}
-		return msg.Nak()
+		return msg.Ack()
 	}, nil
 }
 
--- a/lib/input/reader/nats.go
+++ b/lib/input/reader/nats.go
@@ -48,6 +48,8 @@ type NATS struct {
 	stats metrics.Type
 	log   log.Modular
 
+	unAckMsgs []*nats.Msg
+
 	cMut sync.Mutex
 
 	natsConn      *nats.Conn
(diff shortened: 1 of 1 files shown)


Jeffail:
Hey @seanlaff, I might need to read some more into this but my understanding is that calling `msg.Ack()` on a regular NATS message is a no-op, so if we were to implement some sort of ack mechanism we would need to build it upon something like https://pkg.go.dev/github.com/nats-io/nats.go#Msg.Respond, where we'd respond with some pre-determined string for an ack/nack.

Ideally the response messages should be configurable (and optional) so maybe two fields like `ack_response` and `nack_response` with empty strings as defaults.

seanlaff:
Hey @Jeffail, as far as I can tell, it seems like ack/nack is no longer a noop in NATS?

https://github.com/nats-io/nats.go/blob/3b1f6fcc1e1014c838036367494c3012523166b0/js.go#L1924-L1926

It replies with these blessed strings

https://github.com/nats-io/nats.go/blob/3b1f6fcc1e1014c838036367494c3012523166b0/js.go#L2127-L2132

My test case was using https://pkg.go.dev/github.com/nats-io/nats.go#Conn.Request (may be relatively new?) which handles all the ack/nack replies behind the scenes for you. I was targeting this request-reply use-case as described here https://stackoverflow.com/a/47537696

Does this match your understanding? I am pretty new to NATS :)

Jeffail:
Ahhh okay that's awesome, thanks @seanlaff! I'm going to trim out the acks from `Read` and `Acknowledge` as those methods are only there for backwards compatibility, that'll also eliminate the lint error so don't worry about that.

PR #815, opened 2021-07-02
Upgrade MQTT to pull in some bug fixes

mihaitodor:
I saw a bunch of bug fixes in this lib and thought it might be worth upgrading it.

- Disconnect hang: https://github.com/eclipse/paho.mqtt.golang/issues/501
- Occasional deadlock when connections are lost: https://github.com/eclipse/paho.mqtt.golang/issues/509


Jeffail:
Sounds good, thanks @mihaitodor!

PR #816, opened 2021-07-02
Append to file in GCP output

nmbryant:
Updates the GCP output to append data to the output path file if the file already exists. Pipeline creates temporary files, uses compose to merge the data into the destination file and then removes the temporary files.
I tested this using a sample pipeline and verifying the output in GCP, was unable to test using the integration test because the emulator image does not currently support compose. Created an [issue in that repo](https://github.com/fsouza/fake-gcs-server/issues/527), so we can revisit updated the test once that's implemented.


internal/impl/gcp/cloud_storage_output.go
```diff
@@ -3,6 +3,7 @@ package gcp
 import (
 	"context"
 	"fmt"
+	"path/filepath"
```
mihaitodor, line 6: I think you might want to use `path` instead of `path/filepath`, because `path/filepath` is OS-dependent and GCP only uses Unix-style paths.

Jeffail:
Hey @nmbryant, haven't forgotten about this one, sorry it's taking a while. I like the functionality overall but I have a feeling there's some edge cases that need considering on this, so ideally I'd like to put some time aside to properly dig into it.

nmbryant:
@Jeffail No worries! Just added a mode option from @mihaitodor's request, that may cover some of the edge cases you were thinking of. It will default to overwrite so that behavior of existing pipelines won't change and has a mode for append. There is also a mode to throw an error if the file already exists or skip sending the message data if the file already exists.

nmbryant:
I added compose support to the GCS emulator so I added an append mode test to the GCP test file. I added the function integrationTestStreamIsolatedAppend to integration_test_definitions to test and I am confident that it is testing the append case, but not sure if the new function is needed or if it's the best way to test it. @Jeffail should I have added that new function or is there an existing one I can use to test the append functionality?

internal/impl/gcp/cloud_storage_output.go
```diff
@@ -191,17 +196,56 @@ func (g *gcpCloudStorageOutput) WriteWithContext(ctx context.Context, msg types.
 			return nil
 		})
 
-		w := client.Bucket(g.conf.Bucket).Object(g.path.String(i, msg)).NewWriter(ctx)
+		outputPath := g.path.String(i, msg)
+		_, err := client.Bucket(g.conf.Bucket).Object(outputPath).Attrs(ctx)
```
Jeffail, line 200: Is this needed? It looks like a network call but I'm not sure what it's for.

nmbryant: Yea, that's needed to check if the object at `outputPath` exists. The attributes object isn't needed from that call, just need to see if `err` is the ErrObjectNotExist error.

Jeffail: Ah okay, in that case I think we ought to skip it when we're set to overwrite on collision as it's an unused network call in those circumstances.

Jeffail (commented):
Thanks @nmbryant and sorry for the delay. I think I'm overall happy, just one minor comment left.

Jeffail:
Awesome, looks good, thanks @nmbryant!

Issue #817, opened 2021-07-07
http processor errors


alexstuckey:
I recently started using Benthos and have been blown away so far, but have hit a wall while building a slightly more complicated section of the pipeline.

I'm trying to enrich from an HTTP service while caching it to reduce load on the service. The service can 404 if not available (the branch should skip), but it can also respond without the correct field (branch should skip also). The URL for the http request is interpolated with a key from the message.

When the http request fails it issues an error. I'm unsure as to how exactly it is failing.

An `@service benthos` log is emitted from the HTTP request 
```
{"@timestamp":"", "@service":"benthos","component":"enrich_service.processor.1.0", "level":"ERROR", "message":"HTTP request to 'http://service/${! meta(\"msg_id\") }' failed: HTTP request returned unexpected response code (404): 404 Not Found"}
{"@timestamp":"", "@service":"benthos","component":"enrich_service.processor.1.1.0", "level":"WARN", "message":"getting from service", }
```
1. I thought this shouldn't appear because it is 'caught', although the catch routine is still run, because my own log is emitted.
2. The Benthos log's message displays an un-interpolated string. I couldn't tell if the interpolation was failing and that is what was requested, or whether it just logged the raw string. If that's the case, it would be useful to see the exact URL that was called.

I have a suspicion that the second catch is also catching cache get misses - I might need to `errored = deleted()` before the http request.

I was also wondering if there's a better way to handle boolean values in meta fields, as well as unset meta fields.
One option instead of the bracketed fallback is to have the opening `request_map` set them to the default value.

```yaml
- label: "enrich_service"
  branch:
    # only branch for messages that have a msg_id to enrich with
    request_map: |
      root = this.msg_id | deleted()
      meta msg_id = this.msg_id
    processors:
      - cache:
         resource: the_cache
         operator: get
         key: ${! meta("msg_id") }
      # catch empty cache => GET from http service and set if property set
      - catch:
        - http:
            url: 'http://service/${! meta("msg_id") }'
            verb: GET
        - catch:
          - log:
              level: WARN
              message: "getting from service"
              fields:
                key: '${! meta("msg_id") }'
                meta: '${! meta() }'
          - bloblang: 'meta failed_service = "failed"'
        - branch:
            request_map: 'root = json("property") | deleted()'
            processors:
              - cache:
                  resource: the_cache
                  operator: set
                  key: '${! meta("msg_id") }'
                  value: '${! content().number() }'
    result_map: |
[...]

PR #818, opened 2021-07-10
Fix HTTP processor logging to print the correct URL

mihaitodor:
This addresses item 2 from #817.

### DIFF
--- a/lib/processor/http.go
+++ b/lib/processor/http.go
@@ -1,6 +1,7 @@
 package processor
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 	"time"
@@ -200,12 +201,13 @@ func (h *HTTP) ProcessMessage(msg types.Message) ([]types.Message, types.Respons
 		resultMsg, err := h.client.Send(msg)
 		if err != nil {
 			var codeStr string
-			if hErr, ok := err.(types.ErrUnexpectedHTTPRes); ok {
+			var hErr types.ErrUnexpectedHTTPRes
+			if ok := errors.As(err, &hErr); ok {
 				codeStr = strconv.Itoa(hErr.Code)
 			}
 			h.mErr.Incr(1)
 			h.mErrHTTP.Incr(1)
-			h.log.Errorf("HTTP request to '%v' failed: %v\n", h.conf.HTTP.URL, err)
+			h.log.Errorf("HTTP request failed: %v\n", err)
 			responseMsg = msg.Copy()
 			responseMsg.Iter(func(i int, p types.Part) error {
 				if len(codeStr) > 0 {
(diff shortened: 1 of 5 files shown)


lib/processor/http.go
```diff
@@ -205,7 +214,7 @@ func (h *HTTP) ProcessMessage(msg types.Message) ([]types.Message, types.Respons
 			}
 			h.mErr.Incr(1)
 			h.mErrHTTP.Incr(1)
-			h.log.Errorf("HTTP request to '%v' failed: %v\n", h.conf.HTTP.URL, err)
+			h.log.Errorf("HTTP request to '%v' failed: %v\n", h.urlField.String(0, msg), err)
```
Jeffail, line 217: Unfortunately we can't assume that the interpolation here will be the exact same as the actual URL, since there could be an aspect of randomness to it. I think the best option here is to instead add the URL as a prefix to the error string returned from `h.client.Send`, so here we would just print `HTTP request failed: %v\n` and the returned error would be some thing like `http://example.com/foobar: big foot is real`.

mihaitodor: Oh, interesting! But I'm not sure I get it. For example for the 404 case, the error returned by `h.client.Send` doesn't contain the actual URL, just "HTTP request returned unexpected response code (404): 404 Not Found\n". I can add a defer inside `h.client.DoWithContext` which pulls the URL directly from the `req` object (if it's not nil) and prepends it to the error, but I'm not sure if that's what you're suggesting.

I looked through the code and the URL is constructed [here](https://github.com/Jeffail/benthos/blob/58453c21ad1bd09ab94532e4ab85f1a454f1ec36/lib/util/http/client/type.go#L356) where the `url` `field.Expression` is stringified using the current message, but since `CreateRequest` is only called within `h.client.DoWithContext`, I don't see what would cause it to add randomness to the URL. Does it have something to do with the retry logic? I'm thinking to add a test for this.

Jeffail: The randomness is because you can use functions like `now()` and `random_int()` that'll resolve differently each time the interpolation is made. I think the approach I'd lean towards is adding:

```go
	var req *http.Request
	if req, err = h.CreateRequest(msg); err != nil {
		h.mErrReq.Incr(1)
		h.mErr.Incr(1)
		logErr(err)
		return nil, err
	}
	defer func() {
		if err != nil {
			err = fmt.Errorf("%v: %w", req.URL, err)
		}
	}()
```

There are some `return nil, types.ErrTypeClosed` cases in there but those are probably fine to return without wrapping as it's unrelated to the URL. There's a line `return nil, err` near the bottom that should probably be changed for just `return` as I think that would also bypass the defer.

mihaitodor: Oh, that makes sense, thank you! Forgot that bloblang can do everything 😅 Changes pushed.

Jeffail:
Awesome, thanks @mihaitodor!

PR #822, opened 2021-07-19
Storage queue message lag

mfamador:
Hi!

We need to auto-scale our storage queue consumers, and for that, we need to create a gauge metric with the number of messages that are still in the queue.
In order to create that metric in our pipelines, I've added an additional metadata called `queue_storage_message_lag` which holds the number of messages not yet read.

Please let me know if the naming is ok.

Thanks

### DIFF
--- a/config/azure_queue_storage.yaml
+++ b/config/azure_queue_storage.yaml
@@ -16,6 +16,7 @@ input:
     queue_name: ""
     dequeue_visibility_timeout: 30s
     max_in_flight: 10
+    track_lag: false
 buffer:
   none: {}
 pipeline:
--- a/lib/input/azure_queue_storage.go
+++ b/lib/input/azure_queue_storage.go
@@ -63,8 +63,10 @@ func (a *azureQueueStorage) ConnectWithContext(ctx context.Context) error {
 func (a *azureQueueStorage) ReadWithContext(ctx context.Context) (msg types.Message, ackFn reader.AsyncAckFn, err error) {
 	messageURL := a.queueURL.NewMessagesURL()
 	var approxMsgCount int32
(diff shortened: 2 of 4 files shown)


mfamador:
I've also added `queue_storage_queue_name` metadata, (similat to `kafka_topic` on kafka input) and then we can create our metrics like:
```
- metric:
    name: storage_queue_message_lag
        type: gauge
        labels:
          queue: ${! meta("queue_storage_queue_name") }
          service: "my-service"
        value: ${! meta("queue_storage_message_lag") }          
```

Jeffail:
Hey @mfamador, looks good. Any idea what the performance implication is around: https://github.com/Jeffail/benthos/pull/822/files#diff-d10dfcf6b55f0ffad0ee7a5434f5c51a18af9e5f2d8fa5f91d3cd210a83b722bR67? It looks as though there's a network call so if it's possible that it'll have an impact on throughput I'd probably want to hide this around a config field, maybe something like `track_lag: true`.

mfamador:
Thanks, @Jeffail. I've added the `track_lag` parameter, which is set by default to `false`.

mfamador:
Another thing with this input is that I've set `max_in_flight` initially to 10, probably by mistake, not sure if it should have been 1 by default.

Jeffail:
Thanks @mfamador, looks good. I might change the name of the field to be more general as it's technically a toggle for obtaining fresh properties on the queue, so we could potentially expand it to include other metadata fields in addition to the lag.

I wouldn't worry about the default max in flight for now. The existing components have it set to 1 for backwards compatibility, but ideally all outputs would default to a "best guess" for performance. If you have an educated guess as to a better default then we can consider changing it as the component is still labelled beta which gives us some wiggle room, but I wouldn't consider it a priority.

PR #824, opened 2021-07-20
Add support for elasticsearch action and routing key

alecrubin:


Jeffail:
Looks good, fixing the test after merge. Thanks @alecrubin!

PR #834, opened 2021-08-04
feat: implement gcp_bigquery output #306

### LINKED ISSUE #306: BigQuery support
Would you be open to accepting a [BigQuery](https://cloud.google.com/bigquery/docs/) output to benthos?

lucasoares:
Fixes #306 

A working version of a JSON/CSV BigQuery output

Possible future improvements:

- Usage of interpolated fields to configure dataset and table, checking if they exist in the batch insertion
- JSON and CSV are the only supported formats by this plugin right now. We could in the future support Avro, Parquet, etc
- Maybe we could use the same approach of document mapping used in other plugins


lucasoares:
HELP needed @Jeffail: please help me with the best way to test this output.

Don't know also how to make integration tests since BigQuery doesn't provide an Emulator to use.

lucasoares:
> HELP needed @Jeffail: please help me with the best way to test this output.
> 
> Don't know also how to make integration tests since BigQuery doesn't provide an Emulator to use.

Now this version is better using BigQuery resources and provides all configurations needed using JSON and CSV formats.

I will try to make some tests for these implementations but the HELP is still needed hahaha

lucasoares:
> Thanks @lucasoares, looks good, sorry for the delay. If it's okay with you I'm going to merge this as is and then adapt it to work with the new plugin APIs.

Ok!

Jeffail:
Thanks @lucasoares, merged but I did it separately with the plugin API changes so unfortunately there's no satisfying `merged` flag on the PR: https://github.com/Jeffail/benthos/commit/85a4502b1680e8fb32c18849a261a65fc16bad5b

It's mostly unchanged from your implementation, the tests pass but if you get a chance to try it out before the next release let me know.

PR #835, opened 2021-08-04
docs: lint version doc supporting revive linter

lucasoares:


Jeffail:
Hey @lucasoares, I think we might as well remove the tag and advise pulling the latest unless there's a good reason to keep it.

Jeffail:
Thanks @lucasoares!

PR #839, opened 2021-08-07
feat(tls): add root_cas string field

ar3s3ru:
Currently, the only way to pass a root certificate authority for trusted communication is through the `tls.root_cas_file` parameter.

This means, the `.pem` certificate file must be available on the machine/container where Benthos is running. That, however, is not versatile in those cases where the secret might be instead injected in a Kubernetes Pod using environment variables (maybe through Helm charts + Vault... maybe... not saying that's my use case... :eyes:).

This PR adds support for a `tls.root_cas` string field, where the certificate can be passed explicitly, or through an environment variable instead.

The high number of modifications are due to `make docs` updating all inputs/outputs documentation files. I suggest to check the first commit for analyzing the changes 😄

### DIFF
--- a/config/amqp_0_9.yaml
+++ b/config/amqp_0_9.yaml
@@ -23,6 +23,7 @@ input:
       enabled: false
       skip_cert_verify: false
       enable_renegotiation: false
+      root_cas: ""
       root_cas_file: ""
       client_certs: []
 buffer:
@@ -53,6 +54,7 @@ output:
       enabled: false
       skip_cert_verify: false
       enable_renegotiation: false
+      root_cas: ""
       root_cas_file: ""
       client_certs: []
 logger:
--- a/config/amqp_1.yaml
+++ b/config/amqp_1.yaml
@@ -16,6 +16,7 @@ input:
       enabled: false
       skip_cert_verify: false
       enable_renegotiation: false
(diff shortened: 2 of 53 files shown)


Jeffail:
Awesome, looks good, thanks @ar3s3ru! Bonus thanks for sorting the docs out, that's 53 files now blessed with your avatar.

ar3s3ru:
Passing the certificate content as an env variable using this field (e.g. `CA_PEM="$(cat root.pem)"` and `tls.root_cas: "${CA_PEM}"`) is not working as expected (testing it with a `kafka` input).

Maybe it's the `\n` or the formatting of `cat` 👀 

Might also be the type casting from `string` to `[]byte`.

You just merged while I was typing this message, so I guess I'll figure it out and create a follow-up PR later 😄 

Thanks for reviewing @Jeffail ❤️

Jeffail:
Haha no worries, I can take a look this evening as well. The env var replace is a simple byte swap so it might be that the line breaks need escaping in order for them to be interpreted correctly during the YAML parse (https://github.com/Jeffail/benthos/blob/5b2eecee82ce5ec5602ebb35d579e4a054371bb8/lib/util/text/env_vars.go#L31).

PR #841, opened 2021-08-09
Escape newlines in env var swaps

Jeffail:
I own this repo I don't need to explain myself.


Jeffail:
@ar3s3ru let me know if this works for your use case and I'll merge it in.

ar3s3ru:
@Jeffail can confirm, it works 🎉 💯

rockwotj:
Thank you you beautiful person @Jeffail for this. I was panic earlier that I'd never be able to get secrets for PEM encoded private keys because of this

Issue #838, opened 2021-08-06
Feature Request: RSA authentication with SFTP input / output


lorenz-qbees:
Would be great to have an RSA key authentication method.
I'll have a look at it by myself and keep this issue updated.

Jeffail:
Hey @lorenz-qbees, yeah this would be a reasonable addition.

mihaitodor:
I can take a stab at this. There's some prior art in rclone here: https://github.com/rclone/rclone/blob/654575575810fb01f76e8b4ffb363adfe7473afe/backend/sftp/sftp.go

lorenz-qbees:
@mihaitodor looks great! I will try to implement RSA authentication today and keep this issue updated.

PR #842, opened 2021-08-11
RSA authentication with SFTP input / output regarding

lorenz-qbees:
Implementation of feature request #838:

Added RSA-based authentication for SFTP input and output with optional passphrase.

Successfully tested, formatted and linted (make test, make fmt, make lint)

Updated documentation of SFTP input and output with yarn

I tested the up- and download to a SFTP server with:
- password only
- private key only (without passphrase)
- private key only (with passphrase)
- password and private key (without passphrase)
- password and private key (with passphrase)

Error-Messages are provided when a password, key file or passphrase is necessary but not provided.

### DIFF
--- a/internal/impl/sftp/sftp.go
+++ b/internal/impl/sftp/sftp.go
@@ -2,6 +2,7 @@ package sftp
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net"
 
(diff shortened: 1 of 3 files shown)


Jeffail:
Awesome, thanks @lorenz-qbees!

PR #843, opened 2021-08-12
Add nanoid function to bloblang

razor-x:
Pretty straight forward, but I really can't grok how to add the optional arguments to the function so both tests can pass.

### DIFF
--- a/internal/bloblang/query/functions.go
+++ b/internal/bloblang/query/functions.go
@@ -781,8 +781,8 @@ var _ = RegisterFunction(
 	),
 	true,
 	nanoidFunction,
-	ExpectIntArg(1),
 	ExpectStringArg(0),
+	ExpectIntArg(1),
 )
 
 func nanoidFunction(args ...interface{}) (Function, error) {
--- a/internal/bloblang/query/functions.go
+++ b/internal/bloblang/query/functions.go
@@ -775,22 +775,28 @@ var _ = RegisterFunction(
 		FunctionCategoryGeneral, "nanoid",
 		"Generates a new nanoid each time it is invoked and prints a string representation.",
 		NewExampleSpec("", `root.id = nanoid()`),
-		NewExampleSpec("It is possible to specify a custom alphabet and length.",
-			`root.id = nanoid("abcde", 54)`,
-		),
+		NewExampleSpec("It is possible to specify a length.", `root.id = nanoid(54)`),
+		NewExampleSpec("It is possible to specify a length and custom alphabet.", `root.id = nanoid(54, "abcde")`),
 	),
 	true,
 	nanoidFunction,
-	ExpectStringArg(0),
-	ExpectIntArg(1),
+	ExpectBetweenNAndMArgs(0, 2),
+	ExpectIntArg(0),
+	ExpectStringArg(1),
 )
 
 func nanoidFunction(args ...interface{}) (Function, error) {
-	nid, err := gonanoid.New(args...)
-	if err != nil {
-		panic(err)
-	}
-	return nid, nil
+	return ClosureFunction("function nanoid", func(ctx FunctionContext) (interface{}, error) {
+		switch len(args) {
+		case 0:
+			return gonanoid.New()
+		case 1:
+			return gonanoid.New(args[0].(int))
+		case 2:
+			return gonanoid.Generate(args[1].(string), args[0].(int))
+		}
+		return nil, errors.New("nanoid cannot take more than two arguments")
+	}, nil), nil
 }
 
 //------------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


Jeffail:
Hey @razor-x, `registerSimpleFunction` doesn't allow arguments, use `RegisterFunction` instead: https://github.com/Jeffail/benthos/blob/master/internal/bloblang/query/functions.go#L448

internal/bloblang/query/functions.go
```diff
@@ -779,15 +779,20 @@ var _ = registerSimpleFunction(
...
-		}
-		return nid, nil
-	},
+	true,
+	nanoidFunction,
+	ExpectIntArg(1),
+	ExpectStringArg(0),
 )
 
+func nanoidFunction(args ...interface{}) (Function, error) {
```
razor-x, line 788: @Jeffail So I really just want to do the JS thing but I'm not sure if that's possible here with the type system or your conventions

```js
const x = (...args) => f(...args)
```

mihaitodor: Go doesn't allow us to do that. Since your new function requires between zero and two arguments, you'll want to call `ExpectBetweenNAndMArgs(0, 2)` followed by `ExpectStringArg(0)` and `ExpectIntArg(1)` like you have now. You'll also want to document and add a test for `nanoid(54)`, which corresponds to calling `gonanoid.New(54)` as described [here](https://github.com/matoous/go-nanoid/blob/b658bf464e31bd49c50b5173aaa41c7bfb2461d9/gonanoid.go#L81-L85). Then, you'll need to add a switch, something like this:
```
switch len(args) {
case 0:
	return gonanoid.New()
case 1:
	return gonanoid.New(args[0].(int))
case 2:
	return gonanoid.New(args[0].(string), args[1].(int))
}
```
Finally, instead of panicking when `err != nil`, you'll want to wrap the whole thing in a ClosureFunction like so: https://github.com/Jeffail/benthos/blob/53435661b1703e7100e8f07bcb37d70940792acd/internal/bloblang/query/functions.go#L649-L651

I haven't actually tested these suggestions, so apologies if I got something wrong.

razor-x:
Thank's to @mihaitodor this is now passing, but only if I remove the arg checks (https://github.com/Jeffail/benthos/pull/843/commits/153b9adb06329cd12544f186d95b307df233d4af)

I can't see why those would throw though, arg 0 should be int, and arg 1 should be string.

```
ok  	github.com/Jeffail/benthos/v3/internal/bloblang/parser	2.488s
?   	github.com/Jeffail/benthos/v3/internal/bloblang/plugins	[no test files]
--- FAIL: TestNanoidFunctionLength (0.00s)
    functions_test.go:226: 
        	Error Trace:	functions_test.go:226
        	Error:      	Expected nil, but got: &errors.errorString{s:"expected int argument, received int"}
        	Test:       	TestNanoidFunctionLength
--- FAIL: TestNanoidFunctionAlphabet (0.00s)
    functions_test.go:235: 
        	Error Trace:	functions_test.go:235
        	Error:      	Expected nil, but got: &errors.errorString{s:"expected int argument, received int"}
        	Test:       	TestNanoidFunctionAlphabet
FAIL
FAIL	github.com/Jeffail/benthos/v3/internal/bloblang/query	0.190s
```

mihaitodor:
@razor-x Oops, my bad. Forgot that ints need to be int64 in bloblang (see [here](https://github.com/Jeffail/benthos/blob/9427cc3367f268d3dc5a35d764d314b69e68ddd0/internal/bloblang/query/function_ctor.go#L245-L260)). So, to make those arg checks work, you'll need to use `int(args[0].(int64))`. Also, in your tests, you can use `int64(54)` and the tests should pass after that.

From an ergonomics perspective, I wonder if people might expect to call `nanoid("abcde", 54)` instead of `nanoid(54, "abcde")`, given what the `nanoid` library docs.

razor-x:
> @razor-x Oops, my bad. Forgot that ints need to be int64 in bloblang (see [here](https://github.com/Jeffail/benthos/blob/9427cc3367f268d3dc5a35d764d314b69e68ddd0/internal/bloblang/query/function_ctor.go#L245-L260)). So, to make those arg checks work, you'll need to use `int(args[0].(int64))`. Also, in your tests, you can use `int64(54)` and the tests should pass after that.

Thanks, all working now!

> From an ergonomics perspective, I wonder if people might expect to call `nanoid("abcde", 54)` instead of `nanoid(54, "abcde")`, given what the `nanoid` library docs.

I had it this way at first to match `Generate`, but then switched it when I saw there are actually two functions, `New` and `Generate` where `New` has one arg and `Generate` has two. I'm happy with either API but I'm a little bias to this one since it matches the original JS lib. I'll mark it as ready for review and we can make a final decision before it's merged.

image

Jeffail:
Awesome, looks good, thanks @razor-x! (and @mihaitodor)

PR #845, opened 2021-08-15
Enable extra linters

mihaitodor:
I noticed a bunch of superfluous converts that don't seem to be needed and the `unconvert` linter can warn about those. Go made the byte to uint8 conversion implicit, which can be a bit surprising, I think.

I think [durationcheck](https://github.com/charithe/durationcheck) and [wastedassign](https://github.com/sanposhiho/wastedassign) can help in some cases, so I enabled them too. Luckily, there weren't any bugs uncovered by durationcheck :)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -29,7 +29,7 @@ linters-settings:
 linters:
   disable-all: true
   enable:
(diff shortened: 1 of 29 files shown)


Jeffail:
Nice, I see red and it makes me feel warm and fuzzy inside.

PR #850, opened 2021-08-19
Add MQTT configuration for retained, keepalive and last will

### LINKED ISSUE #848: MQTT Output: Parameters for retain and last will
Hi!

We just started using Benthos and we love it. :+1: Thank you for this great project!

When using MQTT output, we would like to enable a few more options:
- Make retain flag for publish configurable: https://github.com/Jeffail/benthos/blob/68879ed28636b19bb3d4f15d486ac5fd470d19f6/lib/output/writer/mqtt.go#L166
- Make `Will`-settings configurable: https://github.com/eclipse/paho.mqtt.golang/blob/21ffd9190b7f3916a82abe907d5820feaff1ff46/options.go#L115-L119

Is this something you would think is OK? If you do, we'd love to provide a PR for review with the changes.

bittrance:
As promised, some additional MQTT options that opens up new use cases for Benthos:

- Retained flag is only relevant on output. It tells the broker to retain this message on the topic and pass it on to subsequent subscribers. Subsequent retain-flagged messages will replace the previous. This allows Benthos to maintain the status of some resource and subscribers will always have ready access to the latest status.
- Last will can be passed on both input and output. It passes a "canned" publish operation to the broker to perform in case the client should disconnect abnormally. This can be used to inform upstream and downstream actors that Benthos is no longer producing or consuming.
- Keepalive informs the client to send "PINGREQ" messages and informs the broker to expect them. Absence of such messages is a common critera for the broker to consider the client to have disconnected abnormally, thus triggering the last will.

Some notes on implementation:
- I extracted the shared will docs and types into an util module `mqtt_util`. I broke the style of util modules thus because the code follows the Paho tradition of aliasing the module as "mqtt". If you have a better idea on how to disambiguate here, I'm all ears.
- In order to keep the code simple, I have "copied" the underlying default keepalive value of 30 seconds, rather than introduce a separate value or type to know when not to set keepalive. This solution is vulnerable to future changes in the default value in Paho.

Fixes #848 . Ping @simongottschlag .


Jeffail:
Hey @bittrance, looking good and the docs are lit :fire:, only comment is that I'd rather put the shared util package within https://github.com/Jeffail/benthos/tree/master/internal, I'm putting new stuff in there as it lets us rename and move things without breaking changes in future. I can live with just having `internal/mqtt_util` for now, although I try and avoid underscores in package names so maybe `internal/mqttconf`, both are a bit wonky but at least within internal we can change it.

bittrance:
I managed to run the lints and they now pass locally. The helpers are now `internal/mqttconf`.

internal/mqttconf/package.go
```diff
@@ -0,0 +1,2 @@
+// Package mqtt contains supporting utils shared between MQTT reader and writer.
+package mqttconf
```
simongottschlag, line 2: Not 100% sure, but I think an empty file like this may not be required (?)

internal/mqttconf/will.go
```diff
@@ -0,0 +1,44 @@
...
+	QoS      uint8  `json:"qos" yaml:"qos"`
+	Retained bool   `json:"retained" yaml:"retained"`
+	Topic    string `json:"topic" yaml:"topic"`
+	Payload  string `json:"payload" yaml:"payload"`
+}
+
+// EmptyWill returns an empty will, meaning last will message should not be registered.
+func EmptyWill() Will {
+	return Will{}
+}
```
simongottschlag, line 21: Can this be a constant `const EmptyWill = Will{}` instead? I think that's what I've seen the most compared to a function.

lib/input/reader/mqtt.go
```diff
@@ -77,13 +82,19 @@ func NewMQTT(
...
 
+	var err error
 	if len(conf.StaleConnectionTimeout) > 0 {
 		var err error
 		if m.staleConnectionTimeout, err = time.ParseDuration(conf.StaleConnectionTimeout); err != nil {
 			return nil, fmt.Errorf("unable to parse stale connection timeout duration string: %w", err)
 		}
 	}
 
+	err = m.conf.Will.Validate()
```
simongottschlag, line 93: The first `var err error` should be possible to remove and then do `err := ..Validate()` ?

lib/output/mqtt.go
```diff
@@ -2,6 +2,7 @@ package output
 
 import (
 	"github.com/Jeffail/benthos/v3/internal/docs"
+	mqttconf "github.com/Jeffail/benthos/v3/internal/mqttconf"
```
simongottschlag, line 5: `mqttconf` before shouldn't be required here since it is the package name

lib/input/reader/mqtt.go
bittrance: The style is good when you expect to add additional error guards to the same function at a later time, since that future diff will not touch the first err guard. This style follow a common pattern in this code base. Ultimately, it is a YAGNI judgement call perhaps, but this style is a common pattern in this code base.

lib/output/mqtt.go
bittrance: Indeed. Strange that the linter didn't catch this one. Fixed.

internal/mqttconf/package.go
bittrance: I think they exist to hold the package-level documentation? All benthos packages seems to have them.

internal/mqttconf/will.go
bittrance: Only primitive values can be const in Go. If you could have a const struct, it would be the variable that would be const, not its content. Thus, if you had `const EmptyWill = Will{}` and `Will: mqttconf.EmptyWill`, it would presumably still be legal to do `m.conf.will.topic = "asdf"` only this would mutate a struct instance shared between all `MQTTConfig` objects which I think is a potential footgun.

PR #855, opened 2021-08-26
Redis streams input allows for automatic creation of streams

bittrance:
This is just enabling passing the Redis `XGROUP CREATE ... MKSTREAM` option. In some scenarios it is important that the consumer is present before producers arrive. This option facilitates that scenario. Given that a Redis stream does not have any configuration, automatic creation is rather harmless and this option should ideally be turned on by default, but I don't know enough about Benthos usage to discern whether such a change would be polite.

Also fixes the error messages where group and stream names were flipped.


bittrance:
Yes, I suppose. There is a small risk that existing producers assume that they own the stream and will get confused if it is created by the consumer, but that sounds remote. [Kafka defaults this option to true](https://kafka.apache.org/documentation/#brokerconfigs_auto.create.topics.enable), so it may be what users expect here. I've updated the PR.

Jeffail:
Thanks @bittrance, looks good!

Issue #854, opened 2021-08-26
Kafka output component with manual partitioning


cscorley:
Hi team,

I want to implement manual partitioning for the output component for Kafka.  For my use case, I need to be able to send messages to the correct Kafka partition based on some data within the message.  I have an experimental implementation that works with my manual testing and desired configuration, but wanted to open an issue first for discussion.

`sarama` supports `sarama.NewManualPartitioner`, which will use `sarama.ProducerMessage.Partition` from the sent record. 

The configuration I have experimentally implemented looks like the following:
```
output:
  kafka:
    topic: test
	partitioner: manual
	partition: '${! json("event_id") % 4 }'
...
```

Implementing this and adding an integration test was easy enough, but since I need to use a bloblang interpolated field the value can only return be interpolated as a string (I think).  Thus, I need to parse the result as an integer while building the messages for sarama and this will be error prone.  Further, any integer parsed needs to be cast down to 32 bit to meet the saram requirement for the field on the record.

Additionally, I am wondering what the behavior between the current `partitioner` field and a new `partition` field should be.  If a `partition` is specified, but no `partitioner`, is that enough to assume `partitioner=manual`?  What should happen when both a `partition` and non-manual `partitioner` is specified?

Any advise or guidance would be appreciated.  The branch I am working on can be found here: https://github.com/cscorley/benthos/tree/manual-partitioning

PR #857, opened 2021-08-27
Add manual partitioning to Kakfa output component

cscorley:
Hi team,

This PR implements the feature described in issue #854.  

Additionally, I have removed a broken link in the README.  Please let me know if I should separate this into another PR.


Jeffail:
Thanks @cscorley, looks good. I might remove the unit tests for checking config validation for now, it's good to see that it works but eventually the constructors will be migrated to the new plugin APIs and we can do cleaner tests that way.

PR #860, opened 2021-09-01
Passing real Go types into Cassandra driver.

codingconcepts:


lib/output/cassandra.go
```diff
@@ -382,6 +342,50 @@ func (c *cassandraWriter) writeBatch(session *gocql.Session, msg types.Message)
...
+
+		values := make([]interface{}, 0, len(args))
+		for _, arg := range args {
+			values = append(values, stringValue(arg.String(0, msg)))
+		}
+
+		return values, nil
+	}
+
+	argsMapping, err := bloblang.NewMapping("", c.conf.ArgsMapping)
```
Jeffail, line 365: It's possible that this parsing is actually quite expensive so we probably ought to do it only once in the constructor, and the same goes for the args fields. It's not as tidy but it also means we'd capture mapping errors earlier.

Jeffail (commented):
Thanks @codingconcepts, looks good but left one comment. I'm also going to mark the old `args` field as deprecated and remove it from the docs as the new mapping field is cooler, but I'll take care of that after merging.

codingconcepts:
Hey @Jeffail, great shout. That's done 🙂

Jeffail:
Awesome, looks great, thanks @codingconcepts!

PR #866, opened 2021-09-07
DRAFT: First pass at geoip support in benthos

jamesharr:


internal/bloblang/query/methods_geoip.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/oschwald/geoip2-golang"
+)
+
+var _ = registerSimpleMethod(
+	NewMethodSpec(
+		"geoip_city", "",
+	).Param(ParamString("path", "Path to city mmdb file")),
+	func(args *ParsedParams) (simpleMethod, error) {
+		// TODO: Ensure path is a static. Unsure how to accomplish this
```
Jeffail, line 14: Leave this one with me, I'll need to add a way to optionally memoize constructors for functions like this, and part of that will likely be destruction as well (in which we'll call `db.Close()`).

internal/bloblang/query/methods_geoip.go
```diff
@@ -0,0 +1,58 @@
...
+			case string:
+				ip = net.ParseIP(t)
+			default:
+				return nil, NewTypeError(v, ValueString)
+			}
+
+			// TODO: ip will be 'nil' if not a valid v4/v6 address. How to handle?
+			// It currently winds up reporting this when invalid:
+			//    "failed assignment (line 2): field `this.ip`: IP passed to Lookup cannot be nil"
+			// should we add context? IE what the lookup value is?
```
Jeffail, line 42: If you return a bit of context like `fmt.Errorf("value %v does not appear to be a valid v4 or v6 IP address", v)` then it'll be wrapped with the extra context, so the resulting error should be something like `(line x): field this.ip: value foobar does not appear...`, I'd say that's fine for now

Jeffail (commented):
Hey @jamesharr, looking good. I'd say the best approach would be to use something like https://github.com/mitchellh/mapstructure to convert the structured result into a generic `map[string]interface{}` form, at which point you'll be able to access those individual fields within bloblang.

In an ideal world the resulting structure would be snake case. I think we can likely do something like walk the tree of `map[string]interface{}` and rename all the keys by regular expressions without noticing a performance hit, but it maybe ought to be benchmarked. However, that also sounds like a drag so if you want to copy it through with camel case for now I'm happy to dig into that part myself.

helgeolav:
I have not looked into the code yet, but support for geoenrichment is something I need. I, however, need to use ip2location as my provider.

When working on this, will it be possible to make it generic in code so Maxmind, ip2location and others can be added with more common code?

internal/bloblang/query/methods_geoip.go
gallypette: It looks like this is done. 
stracing this PR but merged with df919d7e9c7c3689132a5cf80799d3b5d0290308 shows that the mmdb file is only read once.

Jeffail: I think it's good enough for now to proceed, ideally I want to add some sort of decaying cache but I don't think we need to block this PR on that.

gallypette: Decaying would be nice but is really unnecessary for this use case imho.
Is the plan to wrap each geoip2 method into a bloblang method?

gallypette: I am ok with it, it's just to be sure how to implement the rest ;)

Jeffail:
Merged from https://github.com/Jeffail/benthos/pull/1068

PR #873, opened 2021-09-11
Add addtional string meta fields for S3 output

razor-x:
This pretty much covers all the S3 fields, expect Expires since I'm not sure the best way to pass though time-type fields from the yaml config and parse into go

> // The date and time at which the object is no longer cacheable. For more information,
> // see http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.21 (http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.21).
> Expires *time.Time `location:"header" locationName:"Expires" type:"timestamp"`
>

### DIFF
--- a/config/aws_s3.yaml
+++ b/config/aws_s3.yaml
@@ -44,6 +44,9 @@ output:
     tags: {}
     content_type: application/octet-stream
     content_encoding: ""
+    cache_control: ""
+    content_disposition: ""
+    content_language: ""
     metadata:
       exclude_prefixes: []
     storage_class: STANDARD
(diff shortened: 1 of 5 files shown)


Jeffail:
Thanks @razor-x, I think for `*time.Time` the best option would probably be to use an interpolation string but enforce that users define an interpolation that creates either a valid ISO 8601 string or a unix timestamp. It'd need a bit of extra boilerplate to do the parse for each object but that's probably the best we can do for now.

If it's okay with you I'm going to modify the docs for these fields so that they're marked as advanced, it'll mean they don't show up in common config examples.

PR #878, opened 2021-09-16
Add rack_id field for Kafka components

### LINKED ISSUE #560: Rack-awareness in Kafka input
To reduce cross rack/DC/AZ bandwidth costs, the `kafka` input should support the rack-aware consumer API introduced in KIP-392.

[Sarama has landed support for this upstream](https://github.com/Shopify/sarama/pull/1822), although it has not yet been released. Once it has been released, it would be good for benthos to support this ASAP.

At the very least, benthos should add support via a new `rackId` configuration parameter on `kafka` inputs, which defaults to not being set.

mihaitodor:
Fixes #560.

I managed to run the `TestIntegration/kafka_redpanda` and `TestIntegration/kafka_old` integration tests in a DinD container, since `kafka_old` requires the [correct host IP](https://github.com/Jeffail/benthos/blob/46c86d067212dbe796ee439258a3d4139da0f392/lib/test/integration/kafka_test.go#L354) which doesn't work directly on Docker for Mac. There doesn't seem to be any easy way to add a specific test for this field, so I guess we'll just have to count on Sarama to do the right thing.

Annoyingly, since [`ListNetworks`](https://github.com/Jeffail/benthos/blob/46c86d067212dbe796ee439258a3d4139da0f392/lib/test/integration/kafka_test.go#L330-L336) doesn't work correctly in DinD, I hardcoded the "host" IP address in the code to run the `kafka_old` tests. There's a [bug report](https://github.com/moby/moby/issues/26799) for this since 2016 🤦‍♂️

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ All notable changes to this project will be documented in this file.
 
 - Fields `cache_control`, `content_disposition`, `content_language` and `website_redirect_location` added to the `aws_s3` output.
 - Field `cors.enabled` and `cors.allowed_origins` added to the server wide `http` config.
+- For Kafka components the config now supports the `rack_id` field which may contain a rack identifier for the Kafka client.
+
 
 ### Fixed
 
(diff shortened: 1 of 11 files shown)


Jeffail:
Looks good, thanks @mihaitodor! But yikes on that ancient issue.

PR #887, opened 2021-09-28
simplified the docker compose profiling example

SaberD:
The main point of this pull request is to enable running the resources/docker/profiling example with one docker-compose up command.

The previous setup crashed my laptop so made some changes to make running the setup easier:
- benthos added to the docker compose file
- reduced input bloblang interval
- removed traefik, prom/pushgateway
- added buffer to the pipeline so that Grafana dashboard no longer shows "no data"
- added some missing instructions to the readme

### DIFF
--- a/resources/docker/profiling/README.md
+++ b/resources/docker/profiling/README.md
@@ -5,12 +5,13 @@ This docker compose sets a Benthos instance up with a custom config, [Prometheus
 
 # Set up
 
-- Run Grafana and Prometheus with `docker-compose up`.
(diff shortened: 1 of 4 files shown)


Jeffail:
Hey @SaberD, this setup is often used to test locally built benthos instances so ideally I'd rather not force it to be run as a docker container, although I'm happy to support both in the same example. I'd want to at least keep the push gateway (traefik and labels aren't needed) as it's useful for that.

Makes sense to reduce the default threads and increase the generate period, but I'd rather not include a buffer in the default example as I generally recommend against using it.

PR #916, opened 2021-10-23
Enable depguard linter

mihaitodor:
Prevent "io/ioutil" imports.

This is a follow-up to @Juneezee's cleanup in #915. Many thanks for that!


Jeffail:
Nice, thanks @mihaitodor!

PR #921, opened 2021-10-28
Fix Table Storage output error when having batches bigger than 100

mfamador:
Hi!
Azure Table Storage has a hard limit of 100 records for a single batch.
To fix this issue we'll split the batch into subsets up to 100 records each.


Jeffail:
Looks good, thanks @mfamador! I'm assuming you've already tested it but would be good to verify it again after merge.

mfamador:
Thanks! Yes, I've tested and I'll do it again when merged.

mfamador:
Tested again and everything's working after merging.

PR #922, opened 2021-10-29
Build for lambda arm64

### LINKED ISSUE #895: Support arm64 for AWS lambda
See https://aws.amazon.com/blogs/aws/aws-lambda-functions-powered-by-aws-graviton2-processor-run-your-functions-on-arm-and-get-up-to-34-better-price-performance/

razor-x:
Is this really all we need to support lambda on arm64? Closes #895.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -22,7 +22,7 @@ builds:
     env:
       - CGO_ENABLED=0
     goos: [ linux ]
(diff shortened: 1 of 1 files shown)


Jeffail:
We can build the artifacts and give it a try, fingers crossed it'll "just work" but I haven't checked through the arm lambda docs yet.

PR #927, opened 2021-10-30
Add parse_duration_iso8601 bloblang method

### LINKED ISSUE #913: parse_duration vs ISO 8601 durations
parse_duration is only able to parse custom durations since that's what the (currently) underlying libgo/go/time/format.go is capable of.

It would be nice to be able to parse normal ISO 8601 durations.
Note that 8601 durations do not support sub-seconds durations which is of course a pity (and probably the reason go went this different way).

mihaitodor:
Fixes #913.

Note: I raised a bug upstream to harden the format check when multiple fractions are present (which, apparently, is not allowed for some reason): https://github.com/rickb777/date/issues/19


Jeffail:
Looks great, thanks @mihaitodor!

PR #929, opened 2021-11-01
Added support for NATS message headers when server is compatible

angiglesias:
- On NATS reader received message headers are exposed to message metadata
- On NATS writer a new interpolable map field `headers` is available to write headers on the outgoing message


Jeffail:
Looks great, thanks @angiglesias!

PR #930, opened 2021-11-01
Extended MQTT writer with retained interpolable field

angiglesias:
Retained field on MQTT writer is now interpolable to allow a more granular control in complex pipelines


lib/output/writer/mqtt.go
```diff
@@ -22,20 +23,20 @@ import (
 
 // MQTTConfig contains configuration fields for the MQTT output type.
 type MQTTConfig struct {
-	URLs                  []string      `json:"urls" yaml:"urls"`
-	QoS                   uint8         `json:"qos" yaml:"qos"`
-	Retained              bool          `json:"retained" yaml:"retained"`
```
Jeffail, line 27: Unfortunately this would be a breaking change. I think in order to support an interpolated field here we'd need to have a new field, maybe just `retained_interpolated`, which when specified would override the `retained` field. It's convoluted but wouldn't break existing configs.

Jeffail (changes requested):
Hey @angiglesias, looks good for the most part but the config approach would need to change, I've pointed out the breaking change.

lib/output/writer/mqtt.go
angiglesias: Oh I see. Does this means that interpolable fields cannot take and non interpolable value (like bool, in this case)?

angiglesias: I took this weird way with custom Marshal/Unmarshal to preserve compatibility with previous boolean definition of `retained` field and also allow the field to be interpolable when it was defined as a string on loaded config

Jeffail: It's because there are users that instanciate a pipeline using the old method of creating a config struct programmatically with `NewConfig()`, so we can't change the names or types of fields within the config struct without it being a breaking change.

My intention with v4 is to make all of this stuff purely internal and we won't have to provide backwards compat support for things like this but for now we need to leave the fields unchanged, so our only option is to add new fields.

angiglesias: @Jeffail thanks for the explanation!

Jeffail (approved):

Jeffail:
Great, thanks @angiglesias!

PR #935, opened 2021-11-02
update gcp pubsub package to v1.17.1

willie:
This PR updates the Google Cloud Pub/Sub and Google Cloud Storage packages to their latest versions.

This change is needed to fix a problem with the Pub/Sub library where `subscription.Receive` would return an err of `context deadline exceeded` after processing millions of messages even though the context deadline passed in was not exceeded. This is bad since it stops Benthos from processing any new messages from that point. Before the library update, I had tried putting the whole thing in a loop and retrying on a non-cancellation error, but that would get stuck in an `context deadline exceeded` error loop.

I cannot find documentation in the library where it addresses that specific issue, but I did have a [conversation with a Googler](https://gophers.slack.com/archives/C16HS5FV4/p1635697977001200?thread_ts=1634761124.049800&cid=C16HS5FV4) and he indicated that these errors were fixed in recent versions. I wrote a small Go program to reproduce the error and my testing indicates that updating the library fixes this error.

I ran `make test` and it passed. I did not run `make lint` or `make fmt` as there are no source changes in this PR. I did run `make test-integration` but it was horribly broken on my machine.

I created a PR instead of a message [per this suggestion](https://discord.com/channels/746368194196799589/746368194704572508/904729233123868722).


Jeffail:
Awesome, thanks @willie!

PR #937, opened 2021-11-03
Add benthos-lambda-al2

razor-x:
https://github.com/Jeffail/benthos/issues/895

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -22,6 +22,13 @@ builds:
     env:
       - CGO_ENABLED=0
     goos: [ linux ]
+    goarch: [ amd64 ]
+  - id: benthos-lambda-al2
+    main: cmd/serverless/benthos-lambda/main.go
+    binary: bootstrap
+    env:
+      - CGO_ENABLED=0
+    goos: [ linux ]
     goarch: [ amd64, arm64 ]
 archives:
   - id: benthos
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -22,6 +22,13 @@ builds:
...
       - CGO_ENABLED=0
     goos: [ linux ]
+    goarch: [ amd64 ]
+  - id: benthos-lambda-al2
+    main: cmd/serverless/benthos-lambda/main.go
+    binary: bootstrap
+    env:
+      - CGO_ENABLED=0
+    goos: [ linux ]
     goarch: [ amd64, arm64 ]
```
Jeffail, line 32: Does it make sense to include `amd64` here? I don't know anything about al2.

razor-x: Yes. To run go on AWS lambda you can choose `go1.x` runtime or `provided.al2` runtime. The `benthos-lambda` build is for `go1.x` which only supports `amd64`, and `benthos-lambda-al2` is for `provided.al2` which supports both `amd64` and `arm64`.

We could have a single build like below that works for  `go1.x` runtime and `provided.al2`, but then we would technically break backwards compatibility by renaming the binary (your call if that's ok):

```yaml
  - id: benthos-lambda
    main: cmd/serverless/benthos-lambda/main.go
    binary: bootstrap
    env:
      - CGO_ENABLED=0
    goos: [ linux ]
    goarch: [ amd64, arm64 ]
  - id: benthos-lambda
    builds: [ benthos-lambda ]
    format: zip
    name_template: "{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}"

razor-x: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html
image
image

Jeffail: Ah okay awesome, thanks for the clarification

Jeffail:
Awesome, thanks @razor-x!

PR #939, opened 2021-11-04
use go:embed to include the playground editor

dviramontes:
- as titled

### verified by running
- `make install`
- `/Users/<userland>/go/bin/benthos blobl server`
- playground still works 👍🏽


Jeffail:
Awesome, thanks @dviramontes!

PR #940, opened 2021-11-04
New lambda docs

razor-x:
Just updating the docs for al2 artifacts. I have NOT tested any of the examples here yet. One thing to note is I left in the handler arguments even though it's not used since (A) I think the tooling may still require it even though it's ignored and (B) users might be confused why it's missing otherwise.

Maybe best not to merge this or update the chagelog with the new artifacts until it's all properly tested.


Jeffail:
Awesome, thanks @razor-x!

PR #941, opened 2021-11-05
update /version endpoint doc

dviramontes:
[found this in the code ](https://github.com/Jeffail/benthos/blob/master/lib/service/run.go#L37)and thought it would be good to surface in the docs :)

### DIFF
--- a/website/docs/components/http/about.md
+++ b/website/docs/components/http/about.md
@@ -34,7 +34,9 @@ If the certificate is signed by a certificate authority, the `cert_file` should 
 
 The following endpoints will be generally available when the HTTP server is enabled:
 
-- `/version` provides version info.
+- `/version` provides version info.  
+    - can be set via build flags:
+        - `go build -ldflags "-X github.com/Jeffail/benthos/v3/lib/service.Version=$(VERSION) -X github.com/Jeffail/benthos/v3/lib/service.DateBuilt=$(DATE)" ...`
 - `/ping` can be used as a liveness probe as it always returns a 200.
 - `/ready` can be used as a readiness probe as it serves a 200 only when both the input and output are connected, otherwise a 503 is returned.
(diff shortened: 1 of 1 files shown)


mihaitodor:
@dviramontes Thanks for raising this! Instead of documenting it that way, I think it would be better to leverage the Makefile, since `make all` builds the `benthos` executable with those flags set [here](https://github.com/Jeffail/benthos/blob/0e389834f1f27d44c30c4b25babe7f82bab13ad4/Makefile#L42) and [here](https://github.com/Jeffail/benthos/blob/0e389834f1f27d44c30c4b25babe7f82bab13ad4/Makefile#L21-L22)

dviramontes:
> @dviramontes Thanks for raising this! Instead of documenting it that way, I think it would be better to mention that the Makefile does it automatically, since `make all` builds the `benthos` executable with those flags set [here](https://github.com/Jeffail/benthos/blob/0e389834f1f27d44c30c4b25babe7f82bab13ad4/Makefile#L42) and [here](https://github.com/Jeffail/benthos/blob/0e389834f1f27d44c30c4b25babe7f82bab13ad4/Makefile#L21-L22)

That's a great point @mihaitodor but what if you're using your own Makefile to build benthos for your own project. Would it not be a good idea to document what flags to use in that case ? Lmk if i'm missing something here. Thanks for your feedback.

Jeffail:
Yeah I'm a bit torn on this, it's worth documenting somewhere but this feels a bit out of place. Maybe a good alternative would be to put it in the plugin API documentation. Another thing to consider is that this is going to change relatively soon: https://utcc.utoronto.ca/~cks/space/blog/programming/GoVersionOfYourSource

dviramontes:
Got it thanks @Jeffail - closing in light of [news](https://utcc.utoronto.ca/~cks/space/blog/programming/GoVersionOfYourSource). Thank you both!

PR #944, opened 2021-11-08
aws_sns output: add support for attributes

alexthecarp:
This adds support for attributes to the `aws_sns` output. The AWS doc page is [here](https://docs.aws.amazon.com/sns/latest/dg/sns-message-attributes.html). The code is pretty much exactly the same [code as for SQS](https://github.com/Jeffail/benthos/blob/master/lib/output/writer/sqs.go#L162), with one small change (SQS is limited to 10 attributes whereas SNS doesn't have that limit).

There should be no breaking existing behavior because I'm only adding new optional fields (I've obviously tested locally and also all the tests and integrations pass). 

### Notes
- Using SNS without the ability to set MessageAttributes is quite limiting (for example one can't use [AWS SNS message filtering](https://docs.aws.amazon.com/sns/latest/dg/sns-message-filtering.html)). 
- The only "incidental" change is that `NewSNS` now needs a `types.Manager` parameter, but everything was already in place (`newAmazonSNS` is the only caller of `NewSNS` and it already had the `types.Manager` parameter, it just wasn't passing it to `NewSNS`). 
- Tested all the things pass locally (docs, test, test-integration, docker)

### DIFF
--- a/lib/output/aws_sns.go
+++ b/lib/output/aws_sns.go
@@ -1,6 +1,7 @@
 package output
 
 import (
+	"github.com/Jeffail/benthos/v3/internal/component/output"
 	"github.com/Jeffail/benthos/v3/internal/docs"
 	"github.com/Jeffail/benthos/v3/lib/log"
 	"github.com/Jeffail/benthos/v3/lib/metrics"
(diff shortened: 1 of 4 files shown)


alexthecarp:
@Jeffail I'm sorry to ping you but I noticed no activity on this PR and I can't seem to be able to add reviewers or assignees (but the other open PRs also don't have any so I'm assuming that's ok). Is there anything else I need to do to request a review?

Jeffail:
Hey @alexthecarp, sorry for the delay, this is on my radar. I had a quick look through a few days ago an everything looks great, just need to make some time to go through it properly.

alexthecarp:
> Hey @alexthecarp, sorry for the delay, this is on my radar. I had a quick look through a few days ago an everything looks great, just need to make some time to go through it properly.

Thank you! There's no real rush but I wasn't sure if I've done all the right steps for you to see it :). Awesome!

Jeffail:
Ready now, thanks @alexthecarp!

PR #947, opened 2021-11-11
Fix HTTP docs

mihaitodor:
The CORS config field references are outdated.


Jeffail:
Whoops, thanks @mihaitodor!

PR #948, opened 2021-11-15
Support mongodb upsert when update or replace a single document

BrenQ:
Hi ! I add a small configuration for mongodb that allows you to create an element when you use operations like replace_one or update_one and the element does not exist.

I have been running the integration tests within the mongodb scope and all have been successful. When I run all the integration tests there I have several timeouts. I hope it can be useful and any feedback is welcome.

https://docs.mongodb.com/manual/reference/method/db.collection.updateOne/#parameters

### DIFF
--- a/internal/impl/mongodb/output.go
+++ b/internal/impl/mongodb/output.go
@@ -172,7 +172,7 @@ func NewWriter(
 		return nil, fmt.Errorf("mongodb hint_map not allowed for '%s' operation", conf.Operation)
 	}
 
-	if !upsertAllowed {
+	if upsertAllowed && conf.Upsert != false {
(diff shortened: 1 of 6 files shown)


Jeffail:
Hey @BrenQ, this looks great but I'm planning to hold off on merging until v3.59.0 is out the door (hopefully early next week). However, if this is blocking your work then let me know and I'll squeeze it in.

BrenQ:
Hi @Jeffail ! Thank you very much for running the workflow. I will check the issues. At the moment,  it is not a blocker. My use case is processing data from a topic which has insert and update operations but I can split them by type. So, in conclusion, the approach just simplifies the pipeline.

Jeffail:
Those linting issues are quick to fix so I'm merging, thanks @BrenQ!

PR #949, opened 2021-11-16
Add Config Reader FileWatching tests

mihaitodor:
- Refactor the Reader slightly to allow adjusting `changeFlushPeriod` and `changeDelayPeriod` from tests.
- Move the original black box config tests to a package test file.

Test coverage isn't stellar, but we can add more in the future.


Jeffail:
Niceeee, thanks @mihaitodor!

PR #950, opened 2021-11-16
Decouple make and docker

Tapuzi:
Building Benthos through docker-compose or docker currently doesn't work, as it requires running `make` to solve dependencies.
(My setup: benthos is a git submodule in my project repo, I use docker-compose to build it from the root directory with some more components)

This is effectively decoupling Dockerfile and Makefile, allowing `docker build` to work without invoking `make`.
Hope it fits the project.


Jeffail:
Hey @Tapuzi, I'm in favour of removing the need for vendoring but ideally we'd need to replace it with something else, otherwise we'd end up re-downloading modules any time there's a minor code change.

I'm a bit rusty on docker stuff, but I believe there's a few different approaches to solving this, either we can add a layer to the build that contains only go.mod and go.sum and does the download, and that'll therefore only be rebuilt when those files change. Or alternatively we don't do a download within the docker build but we copy the mod cache from the host machine, either as an explicit copy or as a volume.

Tapuzi:
Good points.

I want to offer another solution:
To solve the issue in my projects, first I copy the dependency file and install the deps, then copy the entire directory.
So if go.mod file is left unchanged, docker build uses the previously cached layer with installed modules. Should look like this:
```
COPY go.mod /go/src/github.com/Jeffail/benthos/
RUN go mod tidy; go mod vendor
COPY . /go/src/github.com/Jeffail/benthos/
```
(I find this more elegant than multi stage builds, as it keeps things simple and only relies on docker's build cache.)

Please let me know what you prefer implemented.

resources/docker/Dockerfile.cgo
```diff
@@ -16,8 +23,6 @@ LABEL maintainer="Ashley Jeffs <ash@jeffail.uk>"
 
 WORKDIR /root/
 
-RUN apt-get update && apt-get install -y --no-install-recommends libzmq3-dev
```
Tapuzi, line 19: (Unrelated cleanup)

Tapuzi:
I am setting up a version with my proposed solution for now, and I have some minor issues to solve.
(I think they are minor)

resources/docker/Dockerfile.cgo
mihaitodor: I could be wrong, but since this container builds Benthos with `make TAGS=ZMQ4`, then `libzmq.so` and its dependencies need to be installed in the final container.

Tapuzi:
Hey.

After reading a bit about go module management, I propose this solution.
I was not enough aware of how go manages dependencies, so please disregard nonsense I may have said previously. (not sure how much of it came out).

This patch should reuse a cached module layer in case `go.mod` and `go.sum` are unchanged, so it relies on them being in sync at the build branch, i.e. `make deps` or `go tidy`, behaves well with the defined CI.


There is an additional optimization option here:
I read a bit about some [build optimizations that can be made](https://www.docker.com/blog/containerize-your-go-developer-environment-part-3/). 
The article mentions a small additional enhancement that can help speed the build process - caching the build directory. This approach can also be used for go's module cache.
I didn't include it because personally I think it's not a good idea, as a Docker user, because it requires some additional work from Docker users before running `docker build` that I find unintuitive and a bit frightening. So... its a bit different for each OS (requires user intervention which is bad), generally requires `make` to be installed and some environment variable to be set.
(in my experience, I expect `docker build` to work out of the box; Docker, however, in their docs say this new thing is the future "since version 18.09"... and I don't trust their docs.)
This is what an unsuspecting Docker user may stumble into with this option:
image
So now he will have to visit (!) a webpage (!!) instead of pressing enter.

Anyway if you feel this option is good for Benthos, I offer to implement it as well, its pretty simple, just let me know. [(reference)](https://github.com/chris-crone/containerized-go-dev/blob/main/Dockerfile)

resources/docker/Dockerfile.cgo
Tapuzi: Hmm, I have to say I am not sure about this, I will return it as its a no-brainer solution.
The reason I removed it is I thought it was useful for the build phase only, not sure now.

Jeffail:
Thanks @Tapuzi, this looks like the ideal solution!

PR #960, opened 2021-11-23
Add HTTP response header filtering

### LINKED ISSUE #899: Include metadata with specific prefixes as headers to http_client output request
I would like to provide a list of prefixes and all metadata with these prefixes to be included as headers in the http_client request.

### LINKED ISSUE #901: propogating headers from http_client response to http server input
In the documentation here https://www.benthos.dev/docs/guides/sync_responses#routing-output-responses-back, an illustration is given of Benthos acting as a reverse proxy.

Given 2 benthos processes:

```yaml
input:
  http_server:
    path: /post
    sync_response:
      headers:
        Content-Type: application/json
        blob: fish
output:
  broker:
    pattern: fan_out
    outputs:
    - file:
        path: data.txt
        codec: lines
    - type: sync_response
      processors:
      - text:
          operator: to_upper
```
and
```yaml
http:
  address: 127.0.0.1:4196
input:
  http_server:
    path: /post
output:
  http_client:
    url: http://localhost:4195/post
    verb: POST
    propagate_response: true
```
posting to 4195 with
```
echo {"address": {"city":"london"} } | httpie :4195/post
```
I get back `Content-Type: application/json`

posting to 4196, `Content-Type: application/octet-stream` is returned, and the `Blob: fish` is dropped.

I tried `copy_response_headers: true` but Content-Type remains application/octet-stream. Is it possible to preserve the headers of the response in the proxy configuration?

mihaitodor:
Fixes #901 and fixes #899.

These changes are inspired from https://www.benthos.dev/docs/components/outputs/kafka#metadata, but I wasn't sure if I should attempt to make https://github.com/Jeffail/benthos/blob/master/internal/component/output/metadata.go more generic, so I created a similar one in `internal/message/metadata/filter/filter.go`. Merging them into one would add some complexity, because the former is exclusive while the latter is inclusive and has extra functionality for regex patterns.

Please let me know if you'd like any adjustments :)

I tested it with the following two setups:

1. Original requirement from #901 with extra `extract_metadata` on the `http_client` output:

```yaml
input:
  http_server:
    path: /post
    sync_response:
      headers:
        Content-Type: application/json
        Blob: fish

output:
  broker:
    pattern: fan_out
    outputs:
      - stdout: {}
      - type: sync_response
        processors:
          - bloblang: root = this.string().uppercase()
```

```yaml
http:
  address: 127.0.0.1:4196
input:
  http_server:
    path: /post
    sync_response:
      extract_metadata:
        include_patterns:
          - ".*lo.*"

output:
  http_client:
    url: http://localhost:4195/post
    verb: POST
    propagate_response: true
    # copy_response_headers: true
    extract_metadata:
      include_prefixes:
        - "bl"
```

And I ran `curl -v -d '{"address": {"city":"london"} }' http://localhost:4196/post` to exercise the Benthos chain, getting the following output:
```
*   Trying ::1...
* TCP_NODELAY set
* Connection failed
* connect to ::1 port 4196 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 4196 (#0)
> POST /post HTTP/1.1
> Host: localhost:4196
> User-Agent: curl/7.64.1
> Accept: */*
> Content-Length: 31
> Content-Type: application/x-www-form-urlencoded
>
* upload completely sent off: 31 out of 31 bytes
< HTTP/1.1 200 OK
< Blob: fish
< Content-Type: application/octet-stream
< Date: Tue, 23 Nov 2021 02:05:41 GMT
< Content-Length: 29
<
* Connection #0 to host localhost left intact
{"ADDRESS":{"CITY":"LONDON"}}* Closing connection 0
```

2. `http_client` input with the previous `http_server` + output broker `sync_response`:

```yaml
input:
  http_server:
    path: /post
    sync_response:
      headers:
        Content-Type: application/json
        Blob: fish

output:
  broker:
    pattern: fan_out
    outputs:
      - stdout: {}
      - type: sync_response
        processors:
          - bloblang: root = this.string().uppercase()
```

```yaml
http:
  address: 127.0.0.1:4196
input:
  http_client:
    url: http://localhost:4195/post
    verb: POST
    payload: '{ "address": { "city": "london" } }'
    headers:
      Content-Type: application/json
    # copy_response_headers: true
    extract_metadata:
      include_prefixes:
        - "bl"
  processors:
    - sleep:
        duration: 1s

pipeline:
  processors:
    - bloblang: root = meta()

output:
  stdout: {}
```

This setup doesn't require an external driver and the output of the instance with the `http_client` input looks like this:
```
...
{"blob":"fish","http_status_code":"200"}
{"blob":"fish","http_status_code":"200"}
{"blob":"fish","http_status_code":"200"}
{"blob":"fish","http_status_code":"200"}
...
```

### DIFF
--- a/internal/http/client.go
+++ b/internal/http/client.go
@@ -156,7 +156,7 @@ func NewClient(conf client.Config, opts ...func(*Client)) (*Client, error) {
 		}
 	}
 
-	if h.metaFilter, err = h.conf.MetadataFilter.New(); err != nil {
+	if h.metaFilter, err = h.conf.ExtractMetadata.CreateFilter(); err != nil {
 		return nil, fmt.Errorf("failed to construct metadata filter: %w", err)
 	}
 
@@ -383,11 +383,11 @@ func (h *Client) ParseResponse(res *http.Response) (resMsg types.Message, err er
 				index := resMsg.Append(message.NewPart(buffer.Bytes()[bufferIndex : bufferIndex+bytesRead]))
 				bufferIndex += bytesRead
 
-				if h.conf.CopyResponseHeaders {
+				if h.conf.CopyResponseHeaders || h.metaFilter.IsSet() {
 					meta := resMsg.Get(index).Metadata()
 					for k, values := range p.Header {
 						normalisedHeader := strings.ToLower(k)
-						if len(values) > 0 && h.metaFilter.Match(normalisedHeader) {
+						if len(values) > 0 && (h.conf.CopyResponseHeaders || h.metaFilter.Match(normalisedHeader)) {
 							meta.Set(normalisedHeader, values[0])
 						}
 					}
@@ -406,11 +406,11 @@ func (h *Client) ParseResponse(res *http.Response) (resMsg types.Message, err er
 			} else {
 				resMsg.Append(message.NewPart(nil))
 			}
-			if h.conf.CopyResponseHeaders {
+			if h.conf.CopyResponseHeaders || h.metaFilter.IsSet() {
 				meta := resMsg.Get(0).Metadata()
 				for k, values := range res.Header {
 					normalisedHeader := strings.ToLower(k)
-					if len(values) > 0 && h.metaFilter.Match(normalisedHeader) {
+					if len(values) > 0 && (h.conf.CopyResponseHeaders || h.metaFilter.Match(normalisedHeader)) {
 						meta.Set(normalisedHeader, values[0])
 					}
 				}
--- a/internal/http/client.go
+++ b/internal/http/client.go
@@ -159,9 +159,6 @@ func NewClient(conf client.Config, opts ...func(*Client)) (*Client, error) {
 	if h.metaFilter, err = h.conf.MetadataFilter.New(); err != nil {
 		return nil, fmt.Errorf("failed to construct metadata filter: %w", err)
 	}
-	if !h.conf.CopyResponseHeaders && h.metaFilter.IsSet() {
-		h.log.Warnln("Metadata include_prefixes / include_patterns are ignored when copy_response_headers isn't set.")
-	}
 
 	h.mCount = h.stats.GetCounter("count")
 	h.mErr = h.stats.GetCounter("error")
@@ -390,7 +387,7 @@ func (h *Client) ParseResponse(res *http.Response) (resMsg types.Message, err er
 					meta := resMsg.Get(index).Metadata()
 					for k, values := range p.Header {
 						normalisedHeader := strings.ToLower(k)
-						if len(values) > 0 && (!h.metaFilter.IsSet() || h.metaFilter.Match(normalisedHeader)) {
+						if len(values) > 0 && h.metaFilter.Match(normalisedHeader) {
 							meta.Set(normalisedHeader, values[0])
 						}
 					}
@@ -413,7 +410,7 @@ func (h *Client) ParseResponse(res *http.Response) (resMsg types.Message, err er
 				meta := resMsg.Get(0).Metadata()
 				for k, values := range res.Header {
 					normalisedHeader := strings.ToLower(k)
-					if len(values) > 0 && (!h.metaFilter.IsSet() || h.metaFilter.Match(normalisedHeader)) {
+					if len(values) > 0 && h.metaFilter.Match(normalisedHeader) {
 						meta.Set(normalisedHeader, values[0])
 					}
 				}
--- a/internal/http/client.go
+++ b/internal/http/client.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/Jeffail/benthos/v3/internal/bloblang/field"
 	"github.com/Jeffail/benthos/v3/internal/interop"
+	"github.com/Jeffail/benthos/v3/internal/message/metadata/filter"
 	"github.com/Jeffail/benthos/v3/internal/tracing"
 	"github.com/Jeffail/benthos/v3/lib/log"
 	"github.com/Jeffail/benthos/v3/lib/message"
@@ -35,9 +36,10 @@ type Client struct {
 	dropOn    map[int]struct{}
 	successOn map[int]struct{}
 
-	url     *field.Expression
-	headers map[string]*field.Expression
-	host    *field.Expression
+	url        *field.Expression
+	headers    map[string]*field.Expression
+	host       *field.Expression
+	metaFilter *filter.Filter
 
 	conf          client.Config
 	retryThrottle *throttle.Type
@@ -152,6 +154,13 @@ func NewClient(conf client.Config, opts ...func(*Client)) (*Client, error) {
 				return nil, fmt.Errorf("failed to parse header '%v' expression: %v", k, err)
 			}
 		}
+	}
+
+	if h.metaFilter, err = h.conf.MetadataFilter.New(); err != nil {
+		return nil, fmt.Errorf("failed to construct metadata filter: %w", err)
+	}
+	if !h.conf.CopyResponseHeaders && h.metaFilter.IsSet() {
+		h.log.Warnln("Metadata include_prefixes / include_patterns are ignored when copy_response_headers isn't set.")
 	}
 
 	h.mCount = h.stats.GetCounter("count")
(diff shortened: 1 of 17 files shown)


lib/util/http/client/type.go
```diff
@@ -18,6 +18,7 @@ import (
```
mihaitodor, line 18: Should I also propagate the changes from the `ParseResponse ` method in `internal/http/client.go` to the same method here? I believe it got left in this file for backwards compatibility, in case someone depends on it in some plugin.

lib/input/http_server.go
```diff
@@ -510,7 +519,15 @@ func (h *HTTPServer) postHandler(w http.ResponseWriter, r *http.Request) {
 		}
 
 		if plen := responseMsg.Len(); plen == 1 {
-			payload := responseMsg.Get(0).Get()
+			part := responseMsg.Get(0)
+			part.Metadata().Iter(func(k, v string) error {
+				if h.metaFilter.Match(k) {
```
mihaitodor, line 524: Forgot to ask... Maybe I should do `h.metaFilter.Match(strings.ToLower(k))` to be consistent with the client? Now that I think about it, what is the reasoning for doing that in the client?

lib/util/http/client/type.go
Jeffail: I think we can safely leave it, I'm not sure how many people are actually using it in the wild but I don't think we need to go as far as updating it.

lib/input/http_server.go
Jeffail: I have a vague memory that this was because:

1. metadata keys created by benthos tend to be lowercase
2. headers are case insensitive so we can do what we want with them
3. I believe the Go client already normalizes headers in its own way (capitalizes them), so we're just overriding it with our own taste

I think your changes are fine as is as we're not bothered about normalizing the data that leaves Benthos (capital letters here will have been created by the user, so we can assume they're aware of them).

The patterns matched against metadata that we're adding ourselves from a response need to be tested against the keys the user is going to see in their data, so I'd say we leave that as well.

internal/http/client.go
```diff
@@ -381,8 +390,9 @@ func (h *Client) ParseResponse(res *http.Response) (resMsg types.Message, err er
 				if h.conf.CopyResponseHeaders {
 					meta := resMsg.Get(index).Metadata()
 					for k, values := range p.Header {
-						if len(values) > 0 {
-							meta.Set(strings.ToLower(k), values[0])
+						normalisedHeader := strings.ToLower(k)
+						if len(values) > 0 && (!h.metaFilter.IsSet() || h.metaFilter.Match(normalisedHeader)) {
```
Jeffail, line 394: I'm thinking through the use cases here and I reckon we should maybe flip the logic here so that rather than checking whether the filter is set we have an inclusive-default filter similar to the existing outputs where there's an exclude prefixes, exclude patterns, etc.

The reason for that is the field itself is sort of an indicator as to whether the component without filters defaults to including all metadata or excluding it all. We would see `exclude_*` and would be more likely to infer that the behaviour with zero filters in this case is to copy all headers as metadata.

mihaitodor: That's very thoughtful, nice catch! It also simplifies the code quite a bit. Thanks @Jeffail 

Just to see if I understood you correctly, there are 2 cases:
- http_client (input, processor, output): The user sets `copy_response_headers: true`, then all headers are included as metadata. The behaviour is the same if they additionally set `metadata_filter: {}`.
- http_server (input): There is no equivalent of `copy_response_headers`, so, after my latest changes, all headers are set as metadata unless there's a `metadata_filter` configured to limit them. Since this alters the existing behaviour, I'm not sure if it's always desirable. Shall I add add an extra flag, like `copy_request_headers`?

Jeffail: What we're aiming to end up with is two standards for metadata filtering, one where they are inclusive by default and one where they are exclusive by default. When we have inclusive fields and they aren't set then we shouldn't be adding any metadata fields at all, so the server implementation right now needs this block to be removed:

```go
if len(f.includePrefixes) == 0 && len(f.includePatterns) == 0 {
  return true
}
```

And then it works as I'd expect, with:

```yaml
# Config that doesn't copy any metadata in the sync response as headers:
http_server:
  sync_response:
    status: "200"
    headers:
      Content-Type: application/json

# Config that copies all metadata in the sync response as headers:
http_server:
  sync_response:
    status: "200"
    headers:
      Content-Type: application/json
      metadata:
        include_patterns: [ .* ]
```

However, the client cases are getting super confusing because of `copy_response_headers`. Ideally we wouldn't have that field at all and instead we'd have `include_*` fields working the exact same as `http_server`, as it's basically a more powerful version of it.

Since we _do_ have that field we can either embrace it and add `exclude_*` fields that complement it:

```yaml
# Config that doesn't copy any headers as metadata:
http:
  url: foo
  copy_response_headers: false

# Config that copies all headers as metadata:
http:
  url: foo
  copy_response_headers: true

# Config that copies a subset of headers as metadata:
http:
  url: foo
  copy_response_headers: true
  metadata_filter:
    exclude_patterns: [ foo_.* ]
```

Or, for the nicer more future proof case, we deprecate `copy_response_headers` entirely, and then add `include_*` fields that work in total isolation. If a user sets both `copy_response_headers` to `true` and also configures any `include_` patterns then we log a warning, and potentially also show that as a linting error.

That would give us this:

```yaml
# Config that copies a subset of headers as metadata keys:
http:
  url: foo
  extract_metadata:
    include_patterns: [ foo_.* ]

# Config that copies all headers as metadata keys:
http:
  url: foo
  extract_metadata:
    include_patterns: [ .* ]

# Config that copies all headers as metadata keys using the old deprecated style:
http:
  url: foo
  copy_response_headers: true

# Config that doesn't copy any headers as metadata keys:
http:
  url: foo
```

mihaitodor: Awesome! I really liked the idea of deprecating `copy_response_headers` and I wanted to get some experience with the linting functionality, so I went full in and implemented your 3rd proposal. For linting to work, I had to do some refactoring in each component which uses `lib/util/http/client/docs.go` because the lint needs to be set on the parent of the `FieldSpecs` and `docs.LintFunc` doesn't seem to provide any means to access the parent `FieldSpec` via the `LintContext`, but please correct me if I'm wrong here.

It might be a good idea to add some extra tests for the `http_client` input and output and the `http` processor that replicate some of the assertions from the `TestHTTPClientReceiveHeadersWithMetadataFiltering` test I added in `internal/http/client_test.go` if you think it's worth it.

One thing to note in the current code is that `include_prefixes: [ "" ]` includes everything, because `strings.HasPrefix()` returns `true` in this case. Somehow, I feel that maybe we should reject this pattern or skip it, but not sure. WDYT?

lib/output/writer/http_client.go
```diff
@@ -109,12 +109,12 @@ func (h *HTTPClient) WriteWithContext(ctx context.Context, msg types.Message) er
 				parts[i] = msgCopy.Get(0)
 			}
 			parts[i].Set(p.Get())
-			if h.conf.CopyResponseHeaders {
```
mihaitodor, line 112: Unless I'm missing something, this check doesn't seem to be needed at all because the filtering is done in `internal/http/client.go` already.

website/docs/components/processors/http.md
```diff
@@ -31,13 +31,13 @@ the original message parts with the body of the response.
...
 label: ""
 http:
-  parallel: false
   url: http://localhost:4195/post
   verb: POST
   headers:
     Content-Type: application/octet-stream
   rate_limit: ""
   timeout: 5s
+  parallel: false
```
mihaitodor, line 40: Maybe we should have a sort in the docs generator?

lib/util/http/client/docs.go
```diff
@@ -30,6 +34,19 @@ func FieldSpecs() docs.FieldSpecs {
 		docs.FieldInt("successful_on", "A list of status codes whereby the attempt should be considered successful, this is useful for dropping requests that return non-2XX codes indicating that the message has been dealt with, such as a 303 See Other or a 409 Conflict. All 2XX codes are considered successful unless they are present within `backoff_on` or `drop_on`, regardless of this field.").Array().Advanced(),
 		docs.FieldString("proxy_url", "An optional HTTP proxy URL.").Advanced(),
 	)
+	httpSpecs = append(httpSpecs, extraChildren...)
 
-	return httpSpecs
+	return docs.FieldComponent().WithChildren(httpSpecs...).
+		Linter((func(ctx docs.LintContext, line, col int, value interface{}) []docs.Lint {
```
mihaitodor, line 40: Loosely inspired from `lib/output/switch.go`

Jeffail:
Awesome @mihaitodor, I think this nails it, thanks for tidying up the docs stuff as well ❤️

PR #963, opened 2021-11-23
Support path component in schema registry base URL

### LINKED ISSUE #957: Can't use base URL with path for schema_registry_decode
Hi, thanks for maintaining this library!

If you use a base URL for `schema_registry_decode` with a path component, it gets ignored when benthos makes requests to Schema Registry.

For example, with the config
```yaml
pipeline:
  processors:
  - schema_registry_decode:
      url: "http://localhost:8080/apis/ccompat/v6"
```
a request is made to `http://localhost:8080/schemas/ids/1` rather than `http://localhost:8080/apis/ccompat/v6/schemas/ids/1` to get the schema.

I'm happy to open a PR if this seems like a valid bug.

mdowds:
Fixes #957 . Adds support for a Schema Registry URL with path components (eg `http://example.com/v1`) in the `schema_registry_decode` and `schema_registry_encode` processors.

I noticed there's some inconsistency between the two processors' implementations, so this also includes a little bit of refactoring so they handle the base URL in the same way.

### DIFF
--- a/internal/impl/confluent/processor_schema_registry_decode.go
+++ b/internal/impl/confluent/processor_schema_registry_decode.go
@@ -63,9 +63,10 @@ func init() {
 //------------------------------------------------------------------------------
 
 type schemaRegistryDecoder struct {
-	surl        string
 	client      *http.Client
 	avroRawJSON bool
+
+	schemaRegistryBaseURL *url.URL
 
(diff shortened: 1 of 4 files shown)


Jeffail:
Looks great, thanks @mdowds!

PR #967, opened 2021-11-25
Add support for arm v7 architecture

mariosant:
This PR adds support for arm v7 architecture. It will enable benthos to run on the new raspberry pi zero 2 :drooling_face:


Jeffail:
Nice, thanks @mariosant!

PR #968, opened 2021-11-25
Infer the Project ID when using the GCP BigQuery output

mihaitodor:
This is a minor annoyance that I bumped into a while back and was meaning to fix. No rush!

Support for this was added recently in the client libs: https://github.com/googleapis/google-cloud-go/issues/1294#issuecomment-689705537


Jeffail:
Very cool, thanks @mihaitodor!

PR #970, opened 2021-11-26
Set MQTT KeepAlive in seconds

htdvisser:
This is a small fix for the KeepAlive config of the MQTT reader and writer.

The documentation and default configuration both treat the KeepAlive config as Seconds. However, when creating the MQTT client, the KeepAlive just casted to a `time.Duration`, turning 30 seconds in 30 nanoseconds.

This pull request fixes that by multiplying the KeepAlive config with `time.Second`.

### DIFF
--- a/lib/input/reader/mqtt.go
+++ b/lib/input/reader/mqtt.go
@@ -159,7 +159,7 @@ func (m *MQTT) ConnectWithContext(ctx context.Context) error {
 		SetClientID(m.conf.ClientID).
 		SetCleanSession(m.conf.CleanSession).
 		SetConnectTimeout(m.connectTimeout).
-		SetKeepAlive(time.Duration(m.conf.KeepAlive)).
+		SetKeepAlive(time.Duration(m.conf.KeepAlive) * time.Second).
 		SetConnectionLostHandler(func(client mqtt.Client, reason error) {
 			client.Disconnect(0)
 			closeMsgChan()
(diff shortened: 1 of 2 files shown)


Jeffail:
Ouch! Thanks @htdvisser, this seems quite severe, do you know what the repercussions of this issue are in practice?

mihaitodor:
Looks like the notes [here](https://github.com/eclipse/paho.mqtt.golang/blob/5208ce8a95f1bde79ed2c7d45f48c27215c5724e/options.go#L60-L81) offer some insight:

> // WARNING: Create the below using NewClientOptions unless you have a compelling reason not to. It is easy
> // to create a configuration with difficult to trace issues (e.g. Mosquitto 2.0.12+ will reject connections
> // with KeepAlive=0 by default).

From [this](https://github.com/eclipse/paho.mqtt.golang/blob/5208ce8a95f1bde79ed2c7d45f48c27215c5724e/options.go#L257) code, I think the original implementation would always set the KeepAlive to 0.

htdvisser:
Indeed what @mihaitodor writes. The old implementation would set KeepAlive to 0, and the latest version of Mosquitto would reject the connection with the (unhelpful) error "Failed to connect to mqtt: identifier rejected".

Jeffail:
Okay thanks both, sounds like we want this out ASAP but we don't necessarily need an emergency patch release. I'll put a minor version release out on monday most likely, but let me know if you're blocked by this and I'll get one out sooner.

htdvisser:
That's fine. Until then I'll just use this workaround to set the KeepAlive to 30 seconds:

```yaml
keepalive: 30_000_000_000
```

PR #971, opened 2021-11-26
Crdb changefeed input

### LINKED ISSUE #964: CockroachDB Core Changefeed Support
I've been working on my own CRDB Changefeeds as a standalone repo, but I think it might fit really well in here.

Would want support for many of the options such as `WITH UPDATE`, as well as tracking the offset timestamp to disk for restart durability.

danthegoodman1:
Closes #964 

This is an initial pass at adding in CockroachDB Core Changefeed support. This only has simple config parsing testing.

Functionality has been manually tested, but otherwise all tests pass.

The goal in the next few PRs is to:

- Add better tests
- Add intelligent caching support for cursors (so we don't need to re-read the entire table on restart)
- Add external caching support so that the benthos cache system can be used as a cache


internal/impl/crdb/input_changefeed.go
```diff
@@ -0,0 +1,245 @@
...
+
+	logger  *service.Logger
+	shutSig *shutdown.Signaller
+}
+
+func newCRDBChangefeedInputFromConfig(conf *service.ParsedConfig, logger *service.Logger) (*crdbChangefeedInput, error) {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	c := &crdbChangefeedInput{
+		logger:     logger,
+		shutSig:    shutdown.NewSignaller(),
```
mihaitodor, line 70: Thanks for this work! I'm trying to understand the signalling logic and I was wondering if this `shutSig` was meant to be used somehow in `crdbChangefeedInput.Read()`. Also, instead of having a cancellable context as a struct member, I think a quit chan might be more idiomatic: https://go.dev/blog/context-and-structs

danthegoodman1: `shutSig` was something that I pulled from other inputs (as examples). I had assumed it was needed at a higher level. I can remove if it's not required from an input as the contexts handle the shutting down. Shall I remove?

For the cancelleable context, that is needed to kill the blocking query found in the `Read()` function. Without it we cannot gracefully exit the `c.pgPool.Query()` (blocks until first row is available) or `c.rows.Next()` (blocks until cancelled).

Unless I am missing something here. Still _relatively_ new to go semantics.

mihaitodor: Sure, no worries! I'm no expert either, but @Jeffail might be able to offer more insight into what would be a good way to make use of this `shutSig` in this input, if it makes sense to do so. Regarding the context part, yep, you can use it like you did, but the link I pasted above offers details about why one should avoid using contexts that way. If all you need is to signal from one goroutine to another to exit, you can simply `make` a channel of `struct{}`. If you close it in one goroutine, the other goroutine that's waiting to read from it will immediately unblock. See the ["channel axioms"](https://dave.cheney.net/2014/03/19/channel-axioms) for details.

danthegoodman1: I found that the shutdonw signal was not needed in my testing, so I have removed it. I can add back in if advised but it did not seem to have any impact on operation.

danthegoodman1: I guess my issue with the chan vs. context is how the query perceives it. I am assuming since the Read() context will be Done() when SIGTERM signal is sent, then the query should gracefully exit anyway. My worry was that the goroutine would still be consuming new rows while the Read()ing stopped, but I see this not to be the case...

I removed the context in the struct and use the function based on instead. Seems to have no impact on operations either. Thanks for the advice!

danthegoodman1: So we definitely need a context for that Query that is NOT shared with the Read(), seems like we need to keep it in the struct unfortunately due to the nature that the Read() context is cancelled after the read finishes.

danthegoodman1: I have found that after a single read, no more reads are parsed when using the Read() context

danthegoodman1: If we don't store this context somewhere we can cancel it (and not the same as the Read() context), then the Read() will continue to block and we will hit the timeout. The context needs to be cancelled and if we are in anything after the first read run we will no longer have access to the same context. I think this is a case where we have to put it in the struct due to the nature of the Read() context being cancelled after it is first called.

danthegoodman1: @mihaitodor Do you have any other suggestions? Otherwise I cannot access the context in the query and run the risk of messages getting consumed but dropped

mihaitodor: I'll defer to Ash for further feedback. I don't understand this well enough to offer proper guidance... Sorry about that!

danthegoodman1: @Jeffail Just wanted to send a friendly ping on whether you have any feedback here :)

Definitely don't think I can get away with not cancelling the context since that goroutine will block forever.

danthegoodman1: @mihaitodor One thing I've been thinking about is how we might be able to make compromises. My concern around not cancelling the query context has mostly been due to reading rows unnecessarily. I think this might be mitigated by 2 means, but I definitely need some confirmation:

1. benthos has already stopped reading, so returning new rows doesn't actually cause anything to happen
2. My other worry was when I v2 this intgeration and add in cursor caching. As long as the shutdown signal cuts of the cursor caching immediately, we don't need to worry about any subsequent hidden reads that go on in the background updating the cursor without emitting messages
3. Since benthos is shutting down it's easy enough for my to just kill the pool connection, that should shut things down similar to cancelling a query.

I am currently implementing these changes locally to see how they work.

danthegoodman1: *Edited the wording to be more accurate

danthegoodman1: I've just made the changes that remove the stored context, if we abort the shutdown and let the ending of the process terminate the DB connections (.Close() on row or pool will block due to the query) then everything works without stored context OR a shutdown channel.

danthegoodman1: Just checking in on this again :)

Jeffail: @danthegoodman1, the core mechanisms here look fine to me, the call to `Close` just needs to close any currently open `pgPool` and `rows`. Ideally we need some form of integration tests next as otherwise I can't really help out from my end.

danthegoodman1: @Jeffail I can only do that if I am able to cancel the long running query with a context, which means I need to keep a context in the struct that I can cancel across different Read() calls. This was something that was mentioned to avoid, but I can add that back in.

another way to handle that would be another goroutine but that would definitely involve more code. Would keep the context out of the struct though so I can try that way first.

If I can’t cancel that query then the close on rows and pgpool will hang.

danthegoodman1: @Jeffail @mihaitodor using nested goroutines I was able to do query cancellation without storing a context in the struct. Let me know if that works as an alternative to storing the context in the struct.

I will get an integration test working, will probably involve needing to run a CRDB container locally but otherwise should be quite simple. Will try to do that Sunday :)

danthegoodman1: Just want to note that I haven't forgotten about this, just have been super busy with work lately. Hoping to hit this by the end of the weekend.

mihaitodor: Yeah, I think that addresses my comment, thanks!

danthegoodman1:
@Jeffail In writing these integration tests, I am looking at the SQL Select example in trying to write the test for checking output.

The issue I am running in to is that since the changefeeds contain insert time timestamps, which I do not have exact control over, and I am not sure the output of `NOW()` would allign with the same timestamp that the changefeed would provide. As a result I cannot assert equal on an output row since the timestamp property changes every time it is run. Should I chop off the timestamp and validate the rest of the output?

internal/impl/crdb/integration_test.go
```diff
@@ -0,0 +1,93 @@
...
+		outBatches = append(outBatches, string(msgBytes))
+		return nil
+	}))
+
+	streamOut, err := streamOutBuilder.Build()
+	require.NoError(t, err)
+
+	require.NoError(t, streamOut.Run(context.Background()))
+
+	// This is where I get stuck because the output is going to change
```
danthegoodman1, line 89: @Jeffail here in particular

Jeffail:
@danthegoodman1 sorry I missed the update on this, I think it'd be enough to just `assert.Contains` on the parts you're expecting as we don't really need to test the output structure from CRDB, just that we're receiving it as expected.

danthegoodman1:
@Jeffail Great, thanks. I'll have that done this week, hopefully later today!

danthegoodman1:
@Jeffail scratching my head a bit on this one, running the test `go test -run TestIntegrationCRDB -v -count=1 ./...` I get the error:

```
=== RUN   TestIntegrationCRDB
=== PAUSE TestIntegrationCRDB
=== CONT  TestIntegrationCRDB
    integration_test.go:77:
        	Error Trace:	integration_test.go:77
        	            	lint errors: line 2: unable to infer component type
        	Test:       	TestIntegrationCRDB
--- FAIL: TestIntegrationCRDB (1.64s)
```

I don't think it is finding the crdb changefeed, which is strange since building and running the container locally works fine... any ideas?

Jeffail:
@danthegoodman1, the template given to `AddInputYAML` needs to be just the stuff within the `input` block, if you change it to:

```go
template := `
crdb_changefeed:
  dsn: postgresql://root@localhost:26257/defaultdb?sslmode=disable
  tables:
    - foo
`
	streamOutBuilder := service.NewStreamBuilder()
	require.NoError(t, streamOutBuilder.SetLoggerYAML(`level: OFF`))
	require.NoError(t, streamOutBuilder.AddInputYAML(template))
```

Then you should be fine

danthegoodman1:
@Jeffail no worries, I've been extremely busy as well so I've had nowhere near the time or dedication that I would have liked to put into this recently!

For some reason though it hangs at this test, unsure why. I put logs all over the test file and it doesn't seem to be logging any of them, although I can see it starting the docker container which is strange, because it only does that after I log... running with `go test -run TestIntegrationCRDB -v -count=1 ./...` and I don't even think the test is running...

danthegoodman1:
I'll add it also seems to want to run all the tests, even with the above command

danthegoodman1:
I have confirmed that the container works fine as well, can connect locally and run queries against it, so it is not an issue with CRDB starting

danthegoodman1:
Revisiting this I found how to run individual tests properly with `go test -run TestIntegrationCRDB -v -count=1 internal/impl/crdb/integration_test.go` 😅

danthegoodman1:
Adding some logging to the error, this is strange...
```
unable to infer input type, candidates were: [crdb_changefeed]
    integration_test.go:75:
        	Error Trace:	integration_test.go:75
        	            	lint errors: line 2: unable to infer component type
        	Test:       	TestIntegrationCRDB
```

danthegoodman1:
@Jeffail did you intend to close this? I’m not sure I understand why if so.

Jeffail:
@danthegoodman1 AHHHH, no that wasn't intentional and I'm an idiot, I deleted the `master` branch thinking these would automatically change to target the default `main` branch but they were automatically closed instead. Can you reopen on `main`?

danthegoodman1:
Will do

PR #973, opened 2021-11-27
Add MongoDB processor json_marshal_mode field

mihaitodor:
Received via user `beck` on the Golang Gophers Slack.

There's the documentation on `Canonical` vs `Relaxed` BSON to JSON marshalling: https://docs.mongodb.com/manual/reference/mongodb-extended-json/ I'm hoping that `bson.MarshalExtJSON` is functionally equivalent to the initial code in Canonical mode.

BTW, do you think we should have Relaxed as the default mode? It's a breaking change, but it might be less surprising to users.

I also did a bunch of refactoring that I hope won't be too painful to review. The MongoDB components could use some more tests, but I'll leave that for another time.


lib/output/mongodb.go
```diff
@@ -8,9 +8,9 @@ import (
 
 // MongoDBConfig contains config fields for the MongoDB output type.
 type MongoDBConfig struct {
-	MongoConfig client.Config `json:",inline" yaml:",inline"`
+	ClientConfig client.Config `json:",inline" yaml:",inline"`
 
-	Operation    string              `json:"operation" yaml:"operation"`
+	Operation    client.Operation    `json:"operation" yaml:"operation"`
```
Jeffail, line 13: I know it's super picky and extremely unlikely to be worth it, but unfortunately we can't change these field names/types without a major version release as they're public. It's lame, but in V4 it'll all be private and then we can refactor to our heart's content.

Jeffail: But just to clarify, I think it's good to have an explicit `client.Operation` type but we'll need to convert to/from string after config parse.

mihaitodor: Oops, indeed! I need to contribute more often, so I don't forget the gotchas 😁  Thanks for spotting that!

internal/impl/mongodb/output.go
```diff
@@ -126,54 +129,41 @@ func NewWriter(
...
+	if isHintAllowed(db.operation) && conf.HintMap != "" {
 		if db.hintMap, err = bEnv.NewMapping(conf.HintMap); err != nil {
 			return nil, fmt.Errorf("failed to parse hint_map: %v", err)
 		}
 	} else if conf.HintMap != "" {
-		return nil, fmt.Errorf("mongodb hint_map not allowed for '%s' operation", conf.Operation)
+		return nil, fmt.Errorf("mongodb hint_map not allowed for '%s' operation", db.operation)
 	}
 
-	if upsertAllowed && conf.Upsert {
```
mihaitodor, line 175: I believe the `upsertAllowed` should've been negated here. I fixed it.

lib/processor/mongodb.go
```diff
@@ -28,10 +29,11 @@ func NewMongoDBConfig() MongoDBConfig {
 	rConf.Backoff.MaxElapsedTime = "30s"
 
 	return MongoDBConfig{
-		MongoDB:      client.NewConfig(),
-		Parts:        []int{},
-		Operation:    "insert",
```
mihaitodor, line 33: The default value here was incorrect (but maybe we shouldn't have any default set for the operation field?

internal/impl/mongodb/output.go
mihaitodor: Rebased after #987 fixed this

beck:
I think the wrong `beck` has been @-mentioned

mihaitodor:
@beck Oops, my bad! It wasn't intentional. Your username matches somebody else's on the Golang Gophers Slack. I updated the notes.

Jeffail:
@mihaitodor, awesome and very neat

PR #983, opened 2021-12-01
fix nats auth problem that its configuration not found

ParkJunha:
Currently I am using Benthos and using nats as output.
However, there is a problem that the configuration cannot be found in the current authentication part.
I've confirmed that there is a problem in the following part.
Currently, I have modified this part to build benthos and use it well as follows.
```
   nats_jetstream:
     urls:
     - 'my_urls'
     subject: 'subjects'
     max_in_flight: 1024
     auth:
       user_credentials_file: "/nkeys/user.creds"
```


Jeffail:
Awesome, thanks for putting the fix together @ParkJunha!

PR #987, opened 2021-12-03
fix(output-mongodb): negate upsert allowed in error validation

pcha:
I found this bug in mongoDB output. The upsert validation is wrong and I can use it because I get an error saying that the config isn't allowed for replace_one.


pcha:
Could it be included like a patch?

mihaitodor:
Indeed, I noticed the same when working on https://github.com/Jeffail/benthos/pull/973. I’m happy to rebase my PR if @Jeffail needs more time to review it, since it has a bunch more changes besides this bug fix.

Jeffail:
Whoops, thanks @pcha, yeah let's get this merged and a patch out today.

PR #990, opened 2021-12-05
Disable acking a nats message when no reply is available. Closes #989

### LINKED ISSUE #989: Annoying nats lots when acks are disabled
When consuming a nats subject that doesn't support replies, I get an error:

```
{"@timestamp":"2021-12-05T18:35:57+02:00","@service":"benthos","component":"benthos.input","level":"ERROR","message":"Failed to acknowledge message: nats: message does not have a reply"}
```
Steps to reproduce:
* Create a nats producer as a benthos stream
* Create a nats consumer as a benthos stream

Run the above via benthos cli.

mariosant:
I think this fixes the issue with nats complains referred at #989.

Take a look and let me know if anything else is needed.

### DIFF
--- a/lib/input/reader/nats.go
+++ b/lib/input/reader/nats.go
@@ -182,7 +182,12 @@ func (n *NATS) ReadWithContext(ctx context.Context) (types.Message, AsyncAckFn, 
 		if res.Error() != nil {
 			return msg.Nak()
 		}
-		return msg.Ack()
+
+		if len(msg.Reply) > 0 {
+			return msg.Ack()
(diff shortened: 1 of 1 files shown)


Jeffail:
Thanks @mariosant but I've opted for a slightly different approach: https://github.com/Jeffail/benthos/commit/48c1c5b222e80939dec92ff24ac3920c28f63986, my worry here was that I want to let the library determine whether it's possible to ack, so instead I'm just surpressing the error we aren't bothered about.

PR #1004, opened 2021-12-10
fix: batch size limit calculation

mfamador:
Tiny fix to calculate correctly when the limit of a batch in azure table storage is reached.


Jeffail:
Good catch, thanks @mfamador!

PR #1012, opened 2021-12-15
Trigger batching after inactivity period

disintegrator:
## Overview

In this change, I'm proposing the addition of a new `inactivity_period` batching policy that triggers flushing a batch if no part is added within some configured period. 

1. The inactivity timer only starts after at least one part is added to a batch
2. The inactivity timer is disabled if a batch is empty
3. The `inactivity_period` policy must be combined with other policies to ensure a hard limit on batching

## Why

I'm currently using a batching config that combines the `byte_size` and `period` policies and I want to maximise my batching behaviour especially when there is a slow drip of messages. This experience has been in the context of working on a proof of concept and not real production experience with Benthos with a real workload.

## TODO

- [ ] Update docs

### DIFF
--- a/lib/buffer/constructor_test.go
+++ b/lib/buffer/constructor_test.go
@@ -93,7 +93,7 @@ func TestSanitise(t *testing.T) {
 	exp = `{` +
 		`"type":"memory",` +
 		`"memory":{` +
-		`"batch_policy":{"byte_size":0,"check":"","count":0,"enabled":false,"period":"","processors":[]},` +
+		`"batch_policy":{"byte_size":0,"check":"","count":0,"enabled":false,"inactivity_period":"","period":"","processors":[]},` +
 		`"limit":20` +
 		`}` +
 		`}`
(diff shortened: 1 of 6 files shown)


Jeffail:
I'm maybe thinking we should modify this so that it's a new field rather than a mutation of the existing `period` field. That would allow users to specify both fields and also possibly make the documentation clearer. We could call it an `inactivity_period`, which is an optional time duration that triggers a batch after a specified period of inactivity (no new messages are added).

disintegrator:
@Jeffail I agree that's much clearer. Since this doesn't change the meaning of `period`, I'm also going to keep the hard limited check as-is. That way, you are still required to set one of period, byte_size, count alongside this new policy.

disintegrator:
@Jeffail I agree that's much clearer. Since this doesn't change the meaning of `period`, I'm also going to keep the hard limited check as-is. That way, you are still required to set one of period, byte_size, count alongside this new policy.

disintegrator:
Closed this in favour of a different approach: https://github.com/Jeffail/benthos/issues/1031

PR #1015, opened 2021-12-17
Adding ability to create ksuids in bloblang.

zer0stars:
Adding ksuid generation to bloblang functions. https://github.com/segmentio/ksuid is an alternative to uuids.


Jeffail:
Thanks @zer0stars, don't worry about the check failures, just needs `make fmt && make docs` so I can do that on my end.

PR #1019, opened 2021-12-20
Correcting the example for the http result map

AndrewAdarga:
This will give the result for the example output

### DIFF
--- a/lib/processor/branch.go
+++ b/lib/processor/branch.go
@@ -108,7 +108,7 @@ pipeline:
           - http:
               url: https://hub.docker.com/v2/repositories/jeffail/benthos
               verb: GET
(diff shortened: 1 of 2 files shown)


Jeffail:
Thanks @AndrewAdarga, well found! The docs are generated from https://github.com/Jeffail/benthos/blob/master/lib/processor/branch.go#L111, so it'll need to be updated there followed by running `make docs` to update the generated bits. I'm happy to merge and update that part myself if you haven't got a build environment set up, just let me know.

AndrewAdarga:
Hi @Jeffail thanks for the guide, I've updated this change and it seems that the tests and lint have all passed in CI.

Jeffail:
Awesome, thanks @AndrewAdarga!

PR #1030, opened 2021-12-30
docs(kafka output): explicit min version for headers

cristiprg:
This PR explicitly mentions the min version 11 that supports kafka headers. For kafka input the version is already specified.

Background: this would've saved me a few hours of research, so I hope this does it to future developers :)

### DIFF
--- a/website/docs/components/outputs/kafka.md
+++ b/website/docs/components/outputs/kafka.md
@@ -111,7 +111,7 @@ The config field `ack_replicas` determines whether we wait for acknowledgement f
 
 Both the `key` and `topic` fields can be dynamically set using function interpolations described [here](/docs/configuration/interpolation#bloblang-queries).
 
-[Metadata](/docs/configuration/metadata) will be added to each message sent as headers, but can be restricted using the field [`metadata`](#metadata).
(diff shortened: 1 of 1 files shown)


Jeffail:
Hey @cristiprg, looks like a previous docs changed removed that warning. Sorry you got caught by this, and kudos for coming back and correcting it! The docs are generated so it'll need to be updated from here: https://github.com/Jeffail/benthos/blob/master/lib/output/kafka.go#L30, and then run `make docs`. If you don't have the build tools set up then let me know and I'll merge as is and translate, only takes me a second.

cristiprg:
@Jeffail Thanks! Yes, if you could do that, that'll great as I don't have the build tools set up. Glad to help!

Jeffail:
👍 also DING DING DING @cristiprg you just ticked us into 100 contributors https://github.com/Jeffail/benthos/graphs/contributors 🎉🎉🎉🎉

cristiprg:
Uhuu, that's so great! Let's pop the champagne!  🎉 🎉 🎉 🍾  🎉 🎉 🎉

PR #1035, opened 2022-01-02
Fix MongoDB processor logging

mihaitodor:
I think it's not desirable to log errors when the processor is unable to find any records, since it's up to the user to write the appropriate config to handle / log such cases.


Jeffail:
Thanks @mihaitodor, agree this makes more sense.

PR #1039, opened 2022-01-03
Support for multipart field in HTTP output client

YachikaRalhan:
With multipart support, we can divide a single batch of message into multiple HTTP request parts and send the POST request.
No need of having multiple messages to create a multipart request. Each part of the request will have its own headers and the fields of multipart array support interpolation
A sample benthos config for the same: set the parts in the bloblang pipeline and refer those in data of multipart 
```
pipeline:
  processors:
    - bloblang: |
        meta AttachmentName = filename
        root.part1 = "ABC"
        root.part2 = "XYZ"

output:
  http_client:
    headers:
      Content-Type: multipart/form-data
    url: ""
    verb: POST
    multipart:
      - content_disposition: form-data; name="bin"; filename='${! meta("AttachmentName")
        content_type: application/bin
        data: ${! json("part2") }
      - content_disposition: form-data; name="data"
        content_type: application/json; charset=utf-8
        data: ${! json("part1") }
```
Note: this is only valid for HTTP output client

### DIFF
--- a/internal/http/client.go
+++ b/internal/http/client.go
@@ -28,6 +28,12 @@ import (
 	"github.com/Jeffail/benthos/v3/lib/util/throttle"
 )
 
+type part struct {
+	contentDisposition *field.Expression
+	contentType        *field.Expression
+	data               *field.Expression
+}
+
 // Client is a component able to send and receive Benthos messages over HTTP.
 type Client struct {
 	client *http.Client
@@ -38,6 +44,7 @@ type Client struct {
 
 	url        *field.Expression
 	headers    map[string]*field.Expression
+	multipart  []part
 	host       *field.Expression
 	metaFilter *metadata.Filter
 
@@ -143,7 +150,6 @@ func NewClient(conf client.Config, opts ...func(*Client)) (*Client, error) {
 	if h.url, err = interop.NewBloblangField(h.mgr, conf.URL); err != nil {
 		return nil, fmt.Errorf("failed to parse URL expression: %v", err)
 	}
-
 	for k, v := range conf.Headers {
 		if strings.EqualFold(k, "host") {
 			if h.host, err = interop.NewBloblangField(h.mgr, v); err != nil {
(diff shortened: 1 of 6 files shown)


internal/http/client.go
```diff
@@ -210,6 +216,25 @@ func OptSetLogger(log log.Modular) func(*Client) {
 	}
 }
 
+// OptSetMultiPart sets the multipart to request.
+func OptSetMultiPart(multipart []client.Part) func(*Client) {
+	return func(t *Client) {
+		if len(multipart) > 0 {
+			for _, v := range multipart {
+				data, _ := interop.NewBloblangField(t.mgr, v.Data)
```
Jeffail, line 224: The error handling here really needs to be exposed somehow. Ideally the component that is calling `OptSetMultipart` should be calling `NewBloblangField` itself so that it can properly return errors, and then `client.Part` comes with the `*field.Expression` values.

Jeffail (commented):
Hey @YachikaRalhan, looks great, just one comment on error handling and we need some unit test coverage as well.

internal/http/client.go
YachikaRalhan: Ideally, this should happen in internal HTTP's NewClient method like we are handling for other fields, but as HTTP client doesn't have access to MultiPart because this is like an extra child to HTTP client as this belongs to only output processor only.
So NewClient's definition will change, somehow like this so that it got access to MultiParts and we do the error handling too.

Change purposing to an existing function definition -
 ```func NewClient(conf client.Config, parts []client.Part, opts ...func(*Client)) (*Client, error) {}```

original - ```func NewClient(conf client.Config, opts ...func(*Client)) (*Client, error) {}```
If this sounds okay to you, should I go ahead with the code change? I will cover unit test cases also.

Jeffail: Hey @YachikaRalhan, I've implemented what I was describing on this branch: https://github.com/Jeffail/benthos/tree/YachikaRalhan-http-multipart-request, I've also renamed the `data` field to `body`. I think with those changes I'm happy to merge, would you be able to rebase this PR from that branch and add some unit tests?

YachikaRalhan: Closing this PR, as I have created a new [PR](https://github.com/Jeffail/benthos/pull/1087) @Jeffail

YachikaRalhan:
Closing this PR, as I have created a new [PR](https://github.com/Jeffail/benthos/pull/1087)

PR #1044, opened 2022-01-05
Message ordering for `gcp_pubsub` output

torfjor:
Tested with and without ordering keys and seems to be working as intended.

Couldn't find any integration tests for `gcp_pubsub`. Where's a good place to start?


Jeffail:
Hey @torfjor, the gcp pubsub tests are in a legacy location: https://github.com/Jeffail/benthos/blob/master/lib/test/integration/gcp_pubsub_test.go

Jeffail:
Okay I just totally fumbled a rebase, merging and will correct it.

PR #1049, opened 2022-01-06
Draft: Feature json in env variables

r-franke:
As discussed in Discord. 

I've put together quick support for accessing values in JSON objects that are exposed through environment variables. Cloud Foundry is one example:

```
VCAP_SERVICES='{
      "p.rabbitmq": [
        {
          "label": "p.rabbitmq",
          "provider": null,
          "plan": "Exposed-plan",
          ...
        }]
}'
```

I've used [tidwall/gjson](https://github.com/tidwall/gjson) for parsing the object as its form may vary. This allows values from the JSON value to be used in the format that `gjson` describes in their docs. To use the label object one would do this:

```
input:
  amqp_0_9:
    urls: [${VCAP_SERVICES.p\.rabbitmq.0.label:NOTFOUND}]
```

Support for default values has been preserved as well. Tests have been updated and linters and fmt ran. Only linting error left has to do with import order I think.

As discussed in Discord, this might not be the best way to solve this as it touches core functionality, @Jeffail suggested templates might be another way to go. However I think the changes are quite small and harmless to support this.

### DIFF
--- a/lib/util/text/env_vars.go
+++ b/lib/util/text/env_vars.go
@@ -11,8 +11,8 @@ import (
 //------------------------------------------------------------------------------
 
 var (
-	envRegex        = regexp.MustCompile(`\${[0-9A-Za-z_.\[\]\\ ]+(:((\${[^}]+})|[^}])+)?}`)
-	escapedEnvRegex = regexp.MustCompile(`\${({[0-9A-Za-z_.\[\]\\ ]+(:((\${[^}]+})|[^}])+)?})}`)
+	envRegex        = regexp.MustCompile(`\${[0-9A-Za-z_.\[\]\\+\- ]+(:((\${[^}]+})|[^}])+)?}`)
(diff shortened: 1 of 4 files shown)


lib/util/text/env_vars.go
```diff
@@ -20,6 +22,12 @@ func ContainsEnvVariables(inBytes []byte) bool {
 	return envRegex.Find(inBytes) != nil || escapedEnvRegex.Find(inBytes) != nil
 }
 
+// TargetVarIsJSON returns true if inBytes contains either dots or square brackets
+// signalling that the targetVar is a json object
+func TargetVarIsJSON(inBytes []byte) bool {
+	return jsonPathRegex.Find(inBytes) != nil
+}
```
r-franke, line 29: This treats all env variables that have a `.` in it as JSON variables. This should not be causing problems as `.`s in environment variables are not valid notation normally.

r-franke:
Obsolete, this can be solved through templates as suggested.

PR #1053, opened 2022-01-10
Bump to latest deps and fix errors.

kennyp:
Bump to all of the latest deps and fix any reported errors.

Integration tests time out on my box. :-/


Jeffail:
Hey @kennyp, is there a specific feature or fix that this brings in? Sweeps of upgrades are quite a high risk change so I usually limit it to a few times a year unless there's a specific need.

kennyp:
I'm getting some push back from our security team since deps are out of date... So I figured I'd throw up a PR to bump the deps.

If there's a less massive way to break it up, I would be totally game.

kennyp:
The two specific issues they called out are:

- github.com/nats-io/jwt@v0.3.2: CVE-2020-26892 - Credential expiration is not working https://github.com/nats-io/jwt/security/advisories/GHSA-4w5x-x539-ppf5
- github.com/gogo/protobuf@v1.3.1: CVE-2021-3121 - potentially DoS if using Protobuf

Jeffail:
Reading through the version bumps and it all seems fairly low risk so I think we're good to merge, I'll probably end up putting a release candidate out this week for other bits so it'll have time in the wild as well hopefully. Will run the integration tests on my monster PC.

Loving those deletes in the parquet processors :heart:

PR #1063, opened 2022-01-14
Improve schema_registry processors URL support

mihaitodor:
This change allows users to configure schema registry URLs which have trailing slashes. It's a refinement for #963.


Jeffail:

PR #1064, opened 2022-01-16
Add support for BENTHOS_CONFIG_PATH to benthos-lambda distribution

cludden:
Adds support for customizing the file path where `benthos-lambda` looks for configuration when `BENTHOS_CONFIG` is not defined using a new `BENTHOS_CONFIG_PATH` environment variable


Jeffail:
Looks great @cludden, bonus thanks for updating the docs ❤️

PR #1067, opened 2022-01-18
Fix pulsar-output auth, allow out-of-order delivery on key_shared subscription

crtomirmajer:


Jeffail:
Thanks @crtomirmajer, looks good

PR #1068, opened 2022-01-18
DRAFT: Follow up at geoip support in benthos

gallypette:
Hi all,

Here is new PR to follow up on https://github.com/Jeffail/benthos/pull/866
There is no test yet as I don't see how make some without embedding maxmind licenced files (any ideas ?).

cheers.


Jeffail:
Thanks @gallypette, looks good, that's a bit of a doozy regarding testing. Are the files here: https://github.com/maxmind/MaxMind-DB/tree/5a0be1c0320490b8e4379dbd5295a18a648ff156/test-data something we can use?

gallypette:
Definitely, I will dig into it.

gallypette:
Some notes to keep track of my train of though:
- we can copy the relevant files
- [maxminddb-golang](github.com/oschwald/maxminddb-golang) and [geoip2-golang]( https://github.com/oschwald/geoip2-golang) integrates the test data through a git submodule
- we can copy the relevant files - as LFS - and add an .lfsconfig to excludes these files. We would fetch them in the makefile `git lfs pull`
- we can recreate some smaller mmdb test vectors from bits of the [source data](https://github.com/maxmind/MaxMind-DB/tree/main/source-data)
- we add some git refs and fetch the relevant files at runtime https://github.com/bradfitz/embiggen-disk/blob/master/integration_test.go#L44

Jeffail:
@gallypette, I'd strongly be against adding any git trickery. I think copying in some very small and trivial test data would be best as we're not testing the library itself, only that we're marshalling results into the right structure.

gallypette:
I tend to consider too many options :)

Where would you prefer to copy this data?
`internal/bloblang/query` ?  
(I think can shrink the relevant files to the bare minimum by regenerating from their source)

gallypette:
I added the tests and left maxmind test files untouched.
Let me know if it's alright for you.
cheers.

Jeffail:
This is awesome, thanks @jamesharr and @gallypette!

PR #1048, opened 2022-01-06
support raw output from jq processor

disintegrator:
Closes  #1045

Added an option to the `jq` processor to output raw strings. This emulates the behaviour of `jq --raw-output` which is described as:

> With  this option, if the filter's result is a string then it will be written directly to standard output rather than being formatted as a JSON string with quotes. This can be useful for making jq filters talk to non-JSON-based systems.

For example, this feature is quite useful for building up lines of CSV that can be conactenated (with `archive`) and output to a bucket.

The code in this change is adapted from the raw marshaller that is used by the gojq CLI: https://github.com/itchyny/gojq/blob/886515fe1b7e28bf5193778770619dce4787d85c/cli/marshaler.go#L17-L23

### Current behaviour

```yaml
input:
  label: sample_message
  generate:
    mapping: root = {"id":"1054fe28-a59e-4878-84f0-43b6fa41c2fc","msg":"sample \"log\" message","ts":1641393111}
    count: 1
pipeline:
  processors:
    - label: to_csv
      jq:
        query: '[.ts,.id,.msg] | @csv'
output:
  stdout:
    codec: lines
```

Outputs:

```
"1641393111,\"1054fe28-a59e-4878-84f0-43b6fa41c2fc\",\"sample \"\"log\"\" message\""
```

### New feature

```yaml
input:
  label: sample_message
  generate:
    mapping: root = {"id":"1054fe28-a59e-4878-84f0-43b6fa41c2fc","msg":"sample \"log\" message","ts":1641393111}
    count: 1
pipeline:
  processors:
    - label: to_csv
      jq:
        query: '[.ts,.id,.msg] | @csv'
        output_raw: true
output:
  stdout:
    codec: lines
```

Outputs:

```
1641393111,"1054fe28-a59e-4878-84f0-43b6fa41c2fc","sample ""log"" message"
```

PR #1075, opened 2022-01-19
pass on empty string result with jq raw output mode

disintegrator:
This changes fixes a bug in the `jq` processor when in raw output mode. Previously, if a query resulted in an empty string, it was treated as an empty message and dropped. Empty strings however are intentional and should not be coalesced into empty messages.

For example:

```yaml
input:
  generate:
    mapping: root = {"test":""}
    interval: ""
    count: 1

pipeline:
  processors:
    - jq:
        output_raw: true
        query: '.test'
    - log:
        message: 'message is `${! content() }`'

output:
  drop: {}
```

Should output a log message like this:

```
{"@timestamp":"2022-01-19T22:12:44Z","@service":"benthos","component":"benthos.pipeline.processor.1","level":"INFO","message":"message is ``"}
```

Currently, the message is dropped and no log message is printed

## Background

This was originally detected by @Jeffail in #1048 ([source](https://github.com/Jeffail/benthos/pull/1048#issuecomment-1013656418))

### DIFF
--- a/lib/processor/jq.go
+++ b/lib/processor/jq.go
@@ -240,7 +240,11 @@ func (j *JQ) ProcessMessage(msg types.Message) ([]types.Message, types.Response)
 				return false, err
 			}
 
-			if len(raw) == 0 {
+			// Sometimes the query result is an empty string. Example:
+			//    echo '{ "foo": "" }' | jq .foo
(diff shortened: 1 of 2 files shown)


Jeffail:
Thanks @disintegrator, looks good!

PR #1076, opened 2022-01-21
Support registering http endpoints in plugins

disintegrator:
This change adds a `RegisterEndpoint` method to the resource manager (`service.Resources`) that allows plugins to register http endpoints on the built-in http server.


disintegrator:
I don't need this anymore since #1077 solved the problem in a neater way!

PR #1077, opened 2022-01-22
Add cork input

### LINKED ISSUE #1069: `cork` higher order input
This is a proposal for a higher-order input called `cork` that controls reading from an underlying input.

```yaml
input:
  cork:
    initially_open: false
    signal_path: "/pubsub" 
    input:
      label: foo_input
      gcp_pub_sub: {}

output:
  drop: {}
```

In the example above, no messages are read from `foo_input`. A signal is required to "uncork" the input and start reading messages and likewise to "cork" and stop reads. In this example, the signal source is an http endpoint:

```
POST <benthos_prefix>/uncork/pubsub
```

… causes the pipeline to start reading messages from the input.

```
POST <benthos_prefix>/cork/pubsub
```

… causes the pipeline to stop reading messages from the input.

## Open questions

* I'm not sure if an http endpoint is the best fit for a signal source. Any other ideas?

## Example use-case

_This not real production use case so I'm quite interested if others have ones that benefit from this feature._

As part of a load test, I want to prepare a pubsub topic by loading it with messages with benthos instances deployed in kubernetes _on standby_. When I've completed loading the topic, I want to trigger the instances to start consuming.

disintegrator:
**This change represents an interesting idea I wanted to explore. While I don't have a need for it presently, I'm putting it forward for discussion and to see if it is interesting enough to include in Benthos.**

Closes #1069 

---

The `cork` input uses signals from one input source to control reads from a target input.

This utility input operates under two modes: corked and uncorked. When corked, reading from the target input is suspended. When uncorked, reading occurs as normal.

The signal input is responsible for produce messages that change the mode. The messages must be one of two string values: `"cork"`, `"uncork"`. Any other values are ignored and if the signal input is read to completion then the cork input will remain operating at the last mode it set.

## Example usage

```yaml
input:
  cork:
    input:
      generate:
        mapping: root = {"test":uuid_v4()}
        interval: 1s
        count: 10
    signal:
      broker:
        inputs:
          - http_server:
              path: /cork
              allowed_verbs: [POST]
            processors:
              - bloblang: root = "cork"
          - http_server:
              path: /uncork
              allowed_verbs: [POST]
            processors:
              - bloblang: root = "uncork"

output:
  stdout:
    codec: lines
```

In this config, we register two http endpoints that, when hit, get translated into the desired signals.

For example:

```
curl -XPOST http://localhost:4195/benthos/uncork
```

Will get converted by the bloblang processor into a message with contents `"uncork"`.

A more interesting use case perhaps is turning the cork input into a circuit breaker by tapping into health checks of downstream services to suspend/resume a Benthos pipeline:

```yaml
rate_limit_resources:
  - label: health_check_poll
    local:
      interval: 10s

input_resources:
  - label: cork_signal
    http_client:
      url: https://backend.blob/health_check
      rate_limit: "health_check_poll"
    processors:
      - bloblang: |
          root = if meta("http_status_code") == "200" { "uncork" } else { "cork" }

input:
  cork:
    input:
      gcp_pubsub:
        project: "${PROJECT}"
        subscription: "${SUBSCRIPTION}"
    signal:
      resource: cork_signal

# Rest of config…
```


website/docs/components/inputs/cork.md
```diff
@@ -0,0 +1,91 @@
...
+Use signals from one input to suspend or resume reads from another input
+
+```yaml
+# Config fields, showing default values
+input:
+  label: ""
+  cork:
+    initially_corked: false
+    input: null
+    signal: null
```
disintegrator, line 30: I'm guessing there's a bug in how the default (or zero value of NewInputField?) is rendered to markdown?

internal/impl/generic/cork.go
```diff
@@ -0,0 +1,287 @@
...
+			}),
+		)
+}
+
+func init() {
+	err := service.RegisterBatchInput(
+		"cork",
+		newCorkConfigSpec(),
+		func(conf *service.ParsedConfig, mgr *service.Resources) (service.BatchInput, error) {
+			return newCorkFromConfig(conf, mgr.Logger())
```
disintegrator, line 280: I decided not to add auto retry to this input as I wanted to delegate this concern to the underlying input.

disintegrator:
@Jeffail I'm so close to getting this done at this point so opening up for review. I hadn't realised my use of `atomic.Value` is incompatible with go1.16 😅 so I'm going to spend some time thinking of an alternative approach.

website/docs/components/inputs/cork.md
Jeffail: Yeah this is a docs thing, harmless for now but need to decide what to put in as a placeholder for docs examples.

internal/impl/generic/cork.go
Jeffail: Yeah that's the right call

Jeffail:
@disintegrator I'm going to struggle not to jump onto 1.18 as soon as it's out the oven, so I wouldn't stress too much about supporting 1.16. Although I think we're maybe only one more v3 release away from feature freeze so we could consider putting v4 onto 1.17 right now and merging this into the v4 branch.

disintegrator:
@Jeffail thanks for the heads up! I've just worked through an implementation that is compatible with Go pre-1.17 in https://github.com/Jeffail/benthos/pull/1077/commits/b16a10b3b28b8a38a216c049ba9ea394457ee385 

If you prefer the previous implementation then I'll revert the commit :)

Finally, I'm more than happy to wait out the feature freeze and launch of v4(!) before merging this change.

internal/impl/generic/cork_test.go
```diff
@@ -0,0 +1,247 @@
...
+	socketFile := createSignalSocket(t)
+	ctx := context.Background()
+	spec := newCorkConfigSpec()
+
+	cfg, err := spec.ParseYAML(`
+signal:
+  socket_server:
+    network: unix
+    address: `+socketFile+`
+    codec: lines
```
disintegrator, line 84: I couldn't think of a way to pass a mock input that would let me pass signals in my test code so i opted for this approach.

Jeffail:
@disintegrator same with this one.

PR #1078, opened 2022-01-24
update docs about dropping messages and acks

disintegrator:
This changes updates the documentation around dropping messages to clarify that this will result in message acknowledgement upstream. It is useful to call this out and avoid any misunderstanding about whether or not messages _simply vanish_ when dropped.


Jeffail:
Awesome, thanks @disintegrator!

PR #1081, opened 2022-01-26
aws_s3 output: add server_side_encryption field

alexthecarp:
This adds support a new field, `server_side_encryption`, to the `aws_s3` output. This simply exposes the `ServerSideEncryption` member of the `UploadInput` [structure](https://docs.aws.amazon.com/sdk-for-go/api/service/s3/s3manager/#UploadInput).

This is a bit tricky because a previously added flag, `kms_key_id` also implicitly sets `ServerSideEncryption` to "aws:kms", which is one of the possible values, but not the only one. To preserve backward compatibility, the new field will take precedence if specified, but otherwise will be ignored. Thus, existing configs that set only `kms_key_id` should continue to work.


Jeffail:
Looks good, thanks @alexthecarp!

PR #1082, opened 2022-01-27
Fix generate input panic on zero value interval

mihaitodor:
I think users expect `interval: 0s` to generate messages as fast as it can. Or maybe we should reject such values?

Without this fix, it panics for the following config:
```yaml
input:
  generate:
    mapping: "root = {}"
    interval: 0s
    count: 0

output:
  stdout: {}
```


Jeffail:
Good catch, thanks @mihaitodor

PR #1087, opened 2022-01-28
Support for multipart field in HTTP output client

YachikaRalhan:
With multipart support, we can divide a single batch of message into multiple HTTP request parts and send the POST request.
No need of having multiple messages to create a multipart request. Each part of the request will have its own headers and the fields of multipart array support interpolation
A sample benthos config for the same: set the parts in the bloblang pipeline and refer those in data of multipart 
```
pipeline:
  processors:
    - bloblang: |
        meta AttachmentName = filename
        root.part1 = "ABC"
        root.part2 = "XYZ"

output:
  http_client:
    headers:
      Content-Type: multipart/form-data
    url: ""
    verb: POST
    multipart:
      - content_disposition: form-data; name="bin"; filename='${! meta("AttachmentName")
        content_type: application/bin
        body: ${! json("part2") }
      - content_disposition: form-data; name="data"
        content_type: application/json; charset=utf-8
        body: ${! json("part1") }
```
Note: this is only valid for HTTP output client

### DIFF
--- a/internal/http/client.go
+++ b/internal/http/client.go
@@ -28,6 +28,16 @@ import (
 	"github.com/Jeffail/benthos/v3/lib/util/throttle"
 )
 
+// MultipartExpressions represents three dynamic expressions that define a
+// multipart message part in an HTTP request. Specifying one or more of these
+// can be used as a way of creating HTTP requests that overrides the default
+// behaviour.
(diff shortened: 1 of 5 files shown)


Jeffail:
Ace, thanks @YachikaRalhan this looks great.

PR #1091, opened 2022-01-31
Add `gcp_bigquery_select` input

disintegrator:
This change adds a new input, called `gcp_bigquery_select`, that can be used to run BigQuery jobs that generate rows of data. The rows are emitted as objects with column names as keys.

## Notice

This input is functionally similar to `sql_select`. BigQuery support could not be cleanly added to `sql_select` as it lacks an official or community-supported `database/sql` driver.

## Example

Given the following config:

```yaml
logger:
  level: DEBUG

input:
  gcp_bigquery_select:
    project: test-job-project
    table: bigquery-public-data.samples.shakespeare
    columns:
      - word
      - sum(word_count) as total_count
    where: length(word) >= ?
    suffix: |
      GROUP BY word
      ORDER BY total_count DESC
      LIMIT 10
    args_mapping: |
      root = [ 3 ]

output:
  stdout:
    codec: lines
```

… yields the following output:

```jsonl
{"@timestamp":"2022-01-31T18:40:16Z","@service":"benthos","component":"benthos.input","level":"DEBUG","message":"executing query: SELECT word, sum(word_count) as total_count FROM `bigquery-public-data.samples.shakespeare` WHERE length(word) >= ? GROUP BY word\nORDER BY total_count DESC\nLIMIT 10\n"}
{"@timestamp":"2022-01-31T18:40:16Z","@service":"benthos","component":"benthos","level":"INFO","message":"Launching a benthos instance, use CTRL+C to close."}
{"@timestamp":"2022-01-31T18:40:16Z","@service":"benthos","component":"benthos","level":"INFO","message":"Listening for HTTP requests at: http://0.0.0.0:4195"}
{"@timestamp":"2022-01-31T18:40:17Z","@service":"benthos","component":"benthos.input","job_id":"MfGJ7fVaUKssdSNUe8nKF7ElYK1","level":"DEBUG","message":"running bigquery job"}
{"@timestamp":"2022-01-31T18:40:19Z","@service":"benthos","component":"benthos.input","job_id":"MfGJ7fVaUKssdSNUe8nKF7ElYK1","level":"DEBUG","message":"bigquery job completed"}
{"total_count":25568,"word":"the"}
{"total_count":19649,"word":"and"}
{"total_count":12527,"word":"you"}
{"total_count":8561,"word":"that"}
{"total_count":8395,"word":"not"}
{"total_count":7780,"word":"And"}
{"total_count":7224,"word":"with"}
{"total_count":6811,"word":"his"}
{"total_count":6244,"word":"your"}
{"total_count":6154,"word":"for"}
{"@timestamp":"2022-01-31T18:40:20Z","@service":"benthos","component":"benthos.input","level":"DEBUG","message":"Waiting for pending acks to resolve before shutting down."}
{"@timestamp":"2022-01-31T18:40:20Z","@service":"benthos","component":"benthos.input","level":"DEBUG","message":"Pending acks resolved."}
{"@timestamp":"2022-01-31T18:40:20Z","@service":"benthos","component":"benthos","level":"INFO","message":"Pipeline has terminated. Shutting down the service."}
```

## TODO

- [x] Testing
- [x] Cleaning up commits


internal/impl/gcp/input_bigquery_select.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+
+	var row map[string]bigquery.Value
+	err := inp.iterator.Next(&row)
+	if err == iterator.Done {
+		return nil, nil, service.ErrEndOfInput
+	}
+	if err != nil {
+		return nil, nil, err
+	}
```
disintegrator, line 218: **Please note:** Based on conversations with @Jeffail, I am assuming `Read` is not called by multiple goroutines and therefore accessing the `*RowIterator` without syncronisation is safe. The `*RowIterator` type does not appear safe for concurrent access from my cursory read of the GCP Go SDK. If I got this wrong, then I can submit another commit to introduce a mutex around this block of code.

internal/impl/gcp/input_bigquery_select_test.go
```diff
@@ -0,0 +1,230 @@
...
+	server := httptest.NewServer(mux)
+
+	t.Cleanup(func() {
+		server.Close()
+	})
+
+	return server, mockAPI
+}
+
+func TestGCPBigQuerySelectInput_Connect(t *testing.T) {
```
disintegrator, line 199: This is the only test that needed elaborate mocking of the BigQuery API. The other tests that focus on `Read` were kept simple by passing a mock iterator to the input instance. Any future changes to the `Read` logic can continue to use this simpler testing path. Having said that, if we consider this particular test to be too much ceremony then I'm comfortable with deleting it and the associated mock server.

disintegrator:
I decided to extract the query building logic in https://github.com/Jeffail/benthos/pull/1091/commits/35baed265cb8cb66b53496e447c79aa58ffc2f8c to prepare for an upcoming _processor_ version of this input. It will share the same query building logic.

internal/impl/gcp/input_bigquery_select.go
```diff
@@ -0,0 +1,273 @@
...
+
+	bs, err := json.Marshal(row)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to marshal row to json: %w", err)
+	}
+
+	msg := service.NewMessage(bs)
```
disintegrator, line 241: I'm not sure why I need to manually marshal a `map[string]bigquery.Value` to json but passing it as is to `msg.SetStructured` resulted in an "unknown" types that could not be traversed in bloblang.

internal/impl/gcp/bigquery.go
```diff
@@ -0,0 +1,109 @@
+package gcp
```
disintegrator, line 1: The code in this file has the dual purpose of making testing simpler by mocking the bigquery client and enabling the development of a similar `gcp_bigquery_select` processor (upcoming PR).

disintegrator:
@Jeffail this is now ready for review and I'll be moving to build the processor equivalent. I do not expect to make any further changes to it beyond addressing review feedback you may have.

Jeffail:
Thanks @disintegrator, this looks great. Going the `json` marshal route should be fine for now, ideally we'd do something like `mapstructure.Decode` but I need to experiment a bit with the results before we can use it generally.

PR #1105, opened 2022-02-05
Batch calls to sendMsg in dispatchMessages

nibbleshift:
I am uncertain of the exact number this should be limited to, but this is to prevent hundreds of thousands of go routines being created all at once.

When ingesting massive amounts of data with lots of small records, I observed over 500,000 go routines running sendMsg.


lib/pipeline/processor.go
```diff
@@ -140,11 +140,15 @@ func (p *Processor) dispatchMessages(msgs []types.Message, ogResChan chan<- type
...
 
+	block := make(chan struct{}, 32)
+
 	for _, msg := range msgs {
+	    block <- struct{}{}
 		go func(m types.Message) {
 			sendMsg(m)
 			wg.Done()
 		}(msg)
+		<-block
```
Jeffail, line 151: Hey @nibbleshift, have you seen any notable changes when running this? I'm pretty sure this is a no-op as you'd need to put this lock release within the goroutine above for it to throttle them.

Jeffail (changes requested):

lib/pipeline/processor.go
nibbleshift: Yes. You're definitely correct.  It was late when I committed this, but I wanted to start a discussion.  I'd like the final solution to take a slightly different approach.  Instead of just throttling the number of go routines I think it might be more ideal to chunk the msgs slice into chunks.  The code I submitted above still results in starting len(msgs) goroutines, but the goal is to avoid spawning thousands and thousands of routines and instead using batching.

so if we did something like:
```
batchSize = 16 # Test to find an ideal number or possible make it configurable

batches = batchMessages(msg, batchSize)

for _, batch := range batches {
    go func(m []types.Message) {
         for _, m := range msgs {
             sendMsg(m)
          }
     }(batch)
}
```

nibbleshift: I was thinking we could use the number of cores on the system or the number of threads  set on the pipeline configuration.  Then from there we could come up with a multiplier to establish the number of concurrent batches.

PR #1107, opened 2022-02-06
Add AckWait config option to nats_jetstream input

### LINKED ISSUE #1106: Add AckWait config option to `nats_jetstream` input
For NATS JetStream input, it would be nice to be able to specify custom [AckWait](https://docs.nats.io/nats-concepts/jetstream/consumers#ackwait) configuration option.
It for configuring time that the NATS server will wait for an ack for any individual message. If an ack is not received in time, the message will be redelivered.

I found this useful for a pipeline that batches entries for extended time duration (above default consumer time limit of `30s`).

Also, I have already implemented this change [here](https://github.com/zbiljic/benthos/tree/feature/nats_jetstream_input_ack_wait), however following contributing guide I am creating this ticket (before possible PR).

Thank you for your consideration.

zbiljic:
The change adds `ack_wait` configuration option to `nats_jetstream` input.

Notes:
- unit tests are passing
- running linter returns 4 errors in `internal/impl/sql/integration_test.go`
- integration tests for NATS are passing (nothing changed there)
- no new integration tests were added (or current updated), but I could see about it (as far as I can see, it might require new kind of stream integration test)

closes #1106

### DIFF
--- a/internal/impl/nats/input_jetstream.go
+++ b/internal/impl/nats/input_jetstream.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/Jeffail/benthos/v3/internal/impl/nats/auth"
 	"github.com/Jeffail/benthos/v3/internal/shutdown"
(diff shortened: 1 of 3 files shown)


Jeffail:
Looks great, thanks @zbiljic!

PR #1109, opened 2022-02-06
Add `mongodb` input

egbuk:
Here we go. Example usage:

```yml
input:
  mongodb:
    url: "mongodb://localhost:27017"
    username: "user"
    password: "pass"
    database: "${MONGO_DATABASE}"
    collection: "${MONGO_COLLECTION}"
    query: |
      root.from = {"$lte": timestamp_unix()}
      root.to = {"$gte": timestamp_unix()}
```


Jeffail:
Hey @egbuk, this looks great, add it to the `Services` category. The `sql` input being in `Integration` is a mistake. Would also be good to match these fields of this input to the other `mongodb` components, so rename `dsn` to `url`.

egbuk:
Fixed

Jeffail:
@egbuk thanks, looks great. A test broke but I can fix that on merge.

PR #1112, opened 2022-02-07
XML: Add support for cast parameter

alexthecarp:
The library function we use, mxj.NewMapXml, exposes a `cast` parameter, which was left to the default value of `false`. In this PR I'm adding the support for setting that, both in the `xml` processor and in the `parse_xml()` string function.


alexthecarp:
This is a fairly small change in terms of code changes (~15 lines?). Most of the rest are new tests and examples and a couple of documentation lines.

PR #1119, opened 2022-02-09
fix: use float32 on cassandra float column

mfamador:
I have a table with a column `foo` of type `float`.

```
CREATE TABLE IF NOT EXISTS data.my_table
(
    id        text,
    foo       float,
    timestamp timestamp,
    PRIMARY key (id, timestamp)
) 
```

Using the “previous”  args parameter on cassandra output, it works fine:

```
    args:
      - ${! json("id") }
      - ${! json("foo") }
      - ${! json("timestamp") 
```

but with args_mapping:

```
    args_mapping: |
      root = [
        id,
        foo,
        timestamp
      ]
```

it fails with "Failed to send message to cassandra: can not marshal float64 into float"

Changing the table column to double instead of float, both args parameters work well.

As in https://github.com/gocql/gocql/issues/474, if we’d want to keep using float64 we’d need to change the column type to double or change our parameter to `float32`.

This PR sets the var as `float32` if the Cassandra column is `float`, and keeps `float64` if the column is `double`.

Signed-off-by: Marco Amador <marco.amador@anova.com>

### DIFF
--- a/internal/bloblang/query/type_helpers.go
+++ b/internal/bloblang/query/type_helpers.go
@@ -104,6 +104,27 @@ func IGetNumber(v interface{}) (float64, error) {
 	return 0, NewTypeError(v, ValueNumber)
 }
 
+// IGetFloat32 takes a boxed value and attempts to extract a number (float32)
+// from it.
+func IGetFloat32(v interface{}) (float32, error) {
+	switch t := v.(type) {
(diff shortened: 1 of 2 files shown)


Jeffail:
Nice! That's a pretty tidy solution, thanks @mfamador!

PR #1124, opened 2022-02-10
fix ID collision bug for bulk elasticsearch output

matt1484:
Solves this issue: https://github.com/Jeffail/benthos/issues/1120

### DIFF
--- a/lib/output/writer/elasticsearch.go
+++ b/lib/output/writer/elasticsearch.go
@@ -254,7 +254,7 @@ type pendingBulkIndex struct {
 	Routing  string
 	Type     string
 	Doc      interface{}
-	Id       string
+	ID       string
 }
 
 // WriteWithContext will attempt to write a message to Elasticsearch, wait for
@@ -287,7 +287,7 @@ func (e *Elasticsearch) Write(msg types.Message) error {
 			Routing:  e.routingStr.String(i, msg),
 			Type:     e.conf.Type,
 			Doc:      jObj,
-			Id:       e.idStr.String(i, msg),
+			ID:       e.idStr.String(i, msg),
 		}
 		return nil
 	}); err != nil {
@@ -342,7 +342,7 @@ func (e *Elasticsearch) Write(msg types.Message) error {
 							requests[errorCount] = requests[i]
 						}
 					}
-					errorCount += 1
+					errorCount++
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


lib/output/writer/elasticsearch.go
```diff
@@ -307,36 +309,49 @@ func (e *Elasticsearch) Write(msg types.Message) error {
...
+						reason := "no reason given"
+						if item.Error != nil {
+							if i == 0 {
+								returnedErr = item.Error
+							}
+							reason = item.Error.Reason
+						}
+						if !shouldRetry(item.Status) {
+							e.log.Errorf("Elasticsearch message '%v' rejected with code [%v]: %v\n", item.Id, item.Status, reason)
+							returnedErr = item.Error
```
matt1484, line 333: previously this was returning. but since we need the error count (since that was used previously) I just use `returnedErr` to effectively do the same thing after we have counted all the errors

lib/output/writer/elasticsearch.go
```diff
@@ -307,36 +309,49 @@ func (e *Elasticsearch) Write(msg types.Message) error {
 			return err
 		}
 
-		failed := result.Failed()
-		if len(failed) == 0 {
+		if !result.Errors {
+			requests = requests[:len(result.Items)]
```
matt1484, line 313: in case a subsequent loop has more requests to get through, I slice to skip them from being counted towards the offset using by `i` below

lib/output/writer/elasticsearch.go
```diff
@@ -307,36 +309,49 @@ func (e *Elasticsearch) Write(msg types.Message) error {
...
+							returnedErr = item.Error
+						} else {
+							e.log.Errorf("Elasticsearch message '%v' failed with code [%v]: %v\n", item.Id, item.Status, reason)
+							req := requests[i]
+							bulkReq, err := e.buildBulkableRequest(req)
+							if err != nil {
+								return err
+							}
+							b.Add(bulkReq)
+							requests[errorCount] = requests[i]
```
matt1484, line 342: reassigning allows the offset to be the same in subsequent requests

lib/output/writer/elasticsearch_integration_test.go
```diff
@@ -269,11 +282,11 @@ func testElasticErrorHandling(urls []string, client *elastic.Client, t *testing.
 	}()
 
 	if err = m.Write(message.New([][]byte{[]byte(`{"message":true}`)})); err == nil {
-		t.Error("Expected error")
+		t.Log("Expected error")
```
matt1484, line 285: not sure why this was using `t.Error` since it would cause the tests to fail, but feel free to change it back if you want

Jeffail: Hey @matt1484, if the index doesn't exist I'd expect the writes to fail, so this might indicate an issue

matt1484: Oh ok. I'll fix that and lint everything

Jeffail:
Thanks @matt1484, I'm going to take this an simplify it a bit as I don't think it's critical to have the total error count and I'm worried the logic right now is a challenge to follow.

PR #1126, opened 2022-02-11
add bloblang `assign` method for structured types

### LINKED ISSUE #1125: Bloblang `this.assign(that)` method for objects
This is a proposal to add an `assign` method to Benthos that overwrites values in a destination object with ones from a source object.

Example:

```
let x = { "foo": "I hate George", "bar": 123 }
let y = { "foo": "George is ok" }

let z = x.assign(y)

root = z
```

Output:

```json
{ "foo": "George is ok", "bar": 123 }
```

disintegrator:
Closes #1125 

This change adds a new bloblang method called assign which merges a source object into a destination object. When a collision occurs, as in a key exists in both source and destination object, then the value of that key is overwritten by that of the source object:

Given:
```
root = this.foo.assign(this.bar)
```

Input:
```
{"foo":{"first_name":"fooer","likes":"bars"},"bar":{"second_name":"barer","likes":"foos"}}
```

Output:
```
{"first_name":"fooer","likes":"foos","second_name":"barer"}
```

### DIFF
--- a/internal/bloblang/query/methods_structured.go
+++ b/internal/bloblang/query/methods_structured.go
@@ -922,6 +922,60 @@ func mergeMethod(target Function, args *ParsedParams) (Function, error) {
 
 //------------------------------------------------------------------------------
 
+var _ = registerMethod(
+	NewMethodSpec(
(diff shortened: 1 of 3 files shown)


Jeffail:

PR #1136, opened 2022-02-17
Support interpolation on `queue_name` field on `azure_queue_storage` input

mfamador:
Support interpolation on `queue_name` azure storage input (the output already supports it)

Signed-off-by: Marco Amador <amador.marco@gmail.com>

### DIFF
--- a/lib/input/azure_queue_storage.go
+++ b/lib/input/azure_queue_storage.go
@@ -9,6 +9,9 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/Jeffail/benthos/v3/internal/bloblang/field"
+	"github.com/Jeffail/benthos/v3/internal/interop"
+
 	"github.com/Azure/azure-storage-queue-go/azqueue"
 	"github.com/Jeffail/benthos/v3/internal/impl/azure"
 	"github.com/Jeffail/benthos/v3/lib/input/reader"
@@ -23,7 +26,8 @@ import (
 type azureQueueStorage struct {
 	conf AzureQueueStorageConfig
 
-	queueURL                 *azqueue.QueueURL
+	queueName                *field.Expression
+	serviceURL               *azqueue.ServiceURL
 	dequeueVisibilityTimeout time.Duration
 
 	log   log.Modular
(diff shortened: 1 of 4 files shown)


lib/input/azure_queue_storage.go
```diff
@@ -91,7 +99,7 @@ func (a *azureQueueStorage) ReadWithContext(ctx context.Context) (msg types.Mess
 			part := message.NewPart([]byte(queueMsg.Text))
 			meta := part.Metadata()
 			meta.Set("queue_storage_insertion_time", queueMsg.InsertionTime.Format(time.RFC3339))
-			meta.Set("queue_storage_queue_name", a.conf.QueueName)
+			meta.Set("queue_storage_queue_name", a.queueName.String(0, msg))
```
Jeffail, line 102: The value here might be different if the interpolation contains time based or random functions, we should capture the result of the first execution and use it in both cases.

lib/input/azure_queue_storage_config.go
```diff
@@ -51,8 +51,8 @@ Only one authentication method is required, ` + "`storage_connection_string`" +
 				"A storage account connection string. This field is required if `storage_account` and `storage_access_key` / `storage_sas_token` are not set.",
 			),
 			docs.FieldCommon(
-				"queue_name", "The name of the target Storage queue.",
-			),
+				"queue_name", "The name of the source Storage queue.",
+			).IsInterpolated(),
```
Jeffail, line 55: Might be worth adding an example here on how this might be used as it might be confusing given the absense of an input message.

mfamador: Added example

lib/input/azure_queue_storage.go
mfamador: You're right, changed

lib/input/azure_queue_storage_config.go
mfamador: Yes, I agree it's confusing because it'll never be interpolated from the `msg` since it was not yet produced, but this is actually a real use case where we want the queue name based on an env var, which value comes in PascalCase and Azure Storage Queue only supports queue names in lowercase.

Jeffail:
Looks good, thanks @mfamador!

PR #1139, opened 2022-02-18
Add support for protobuf Any fields

orktes:
This PR adds support for Any fields in protobuf definitions by creating a AnyResolver based on the protobuf import paths.

## Use case
Main use case for us is unified envelopes for all queues.

```proto
message Envelope {
  string id = 1;
  oogle.protobuf.Timestamp time = 2;
  google.protobuf.Any content = 3;
}
```

So for example

```go
event := Envelope{
  id: "someID",
  time: ...,
  content: SomeEvent{
     some_field: "foo"
     ...
  }
}
```

would after the change be parsed into

```js
{
  id: .., 
  time: .., 
  content: {
     some_field: "foo", 
     "@type": "type.googleapis.com/event.SomeEvent"
   }
}
```

instead of throwing an error for the missing `event.SomeEvent` when trying to resolve the Any field.


orktes:
@Jeffail seems like the lint failure is something not related to the change in the PR. Shall I fix those at the same time (import the new protobuf package) or leave it for later?

Issue #1143, opened 2022-02-22
MQTT shared consumers with hostnames (pod ids) as client-ids


rbroggi:
In MQTT clients should have different IDs and in a K8 environment, it is especially interesting to associate an MQTT consumer client-id to its hostname (which in k8 should resolve to a unique pod ID). 

Do you think it would be intesting supporting another option in the `dynamic_client_id_suffix` allowing for retrieval of the `os.Hostname()`? 

I could try to make a PR as this sounds very good as a newcomer-friendly contribution.

Jeffail:
Hey @rbroggi, we're actually close to entering the final phases of v4: https://github.com/Jeffail/benthos/issues/1071, and that means we can include some (relatively minor) breaking changes into the release. I think it might be worth changing the `client_id` field to support interpolation functions as it'd be more powerful and would let you specify things like `foo-${! nanoid() }-${! hostname() }`.

If you fancy trying that out then the changes would be similar to this PR: https://github.com/Jeffail/benthos/pull/1136, but done against the `v4` branch. However, I'd recommend waiting for maybe one more week or so until this is done: https://github.com/Jeffail/benthos/issues/1140 as it'd save you from needing to rebase and adapt your changes.

rbroggi:
I think that this is indeed more powerful. I will wait for a week or so and get back to you (hopefully with the PR). Thanks a lot for the fast reply, it is very much appreciated.

twmb:
This has been addressed — the MQTT component now has a dynamic_client_id_suffix field that supports nanoid and hostname suffixes, allowing unique client IDs per pod. Closing as resolved. If you need more flexible interpolation, feel free to open a new issue or PR.

PR #1144, opened 2022-02-22
issue #1143 - simple approach.

rbroggi:
to be reiterated with the support for generic client_id


Jeffail:
@rbroggi I'm an idiot, I deleted the `master` branch thinking these PRs would automatically change to target the default `main` branch but they were automatically closed instead. Can you reopen on `main`?

rbroggi:
Dear @Jeffail , no problem at all, this one was more of a marker than something else - as we have discussed, my goal (I hope to find the time for that) is to make a PR supporting the more generic solution you suggested for v4. BTW, just for me to have an idea of my deadlines: do you already have a target date when you plan to freeze new developments like the one above for v4?

Jeffail:
@rbroggi okay awesome. All new stuff goes into v4 now, there won't be a feature freeze for stuff like this as they small complementary changes rather than deep technical changes, so there's no current blockers and there won't be any so we can look at this whenever you're ready (time allowing).

PR #1147, opened 2022-02-23
Fix kafka_franz input and output docs

mihaitodor:


Jeffail:
❤️ ❤️ ❤️

Jeffail:
If I could merge it twice I would.

PR #1149, opened 2022-02-24
Update optum logo

soundarmoorthy:
Update optum logo to the new one. Use transparent PNG. Verified that the rendering is good by running the site locally.


Jeffail:
Thanks @soundarmoorthy!

Issue #658, opened 2021-02-08
Snowflake SQL Driver Option


nmbryant:
Is there any interest in having Snowflake as a driver option in the SQL processor/output? The driver can be found here: https://github.com/snowflakedb/gosnowflake

Jeffail:
Hey @nmbryant yes absolutely, any idea if there's a way to test it without using a snowflake account?

nmbryant:
Unfortunately it doesn't seem like testing can be without an account from what I have found. I think adding the driver option would be a small change, should I just make the change without tests?

ironchef:
Snowflake is pretty good with open source folks from what I've heard.  May be worth a shout out to someone like mpilman.  If not, I have access to snowflake and have used benthos in the past ... if I can swing some cycles I could potentially help.

mihaitodor:
I'm exploring the library right now and trying to build an output. @ironchef do you have experience with the Snowflake platform? I'd love to have a chat with someone who can provide some feedback on how people expect this to work in practice, especially from a data streaming perspective.

ironchef:
Yup yup .. mostly from a "hey .. execute this query and give me results" perspective on the golang side.  But I've integrated snowflake with AWS + S3, airflow, dbt, etc. (a bunch of other tooling).  I'll dm / email ya.

mihaitodor:
Update: The `snowflake_put` output was merged in #1160. I think it's worth adding a corresponding input and, potentially, some alternatives based on `SELECT` / `INSERT`. Leaving this open for feedback.

mihaitodor:
Closing this since, besides the existing `snowflake_put` output, Snowflake support was added to the various `sql_*` components in [v4.10.0](https://github.com/benthosdev/benthos/releases/tag/v4.10.0).

PR #1160, opened 2022-03-08
Add snowflake_put output

mihaitodor:
For anyone who'd like to test this out, Snowflake allows you to create a 30-day free trial account [here](https://signup.snowflake.com/).

I'd like to add a bunch of lints to the config, but wasn't sure what's the best way to do that currently. I know I did it [here](https://github.com/Jeffail/benthos/blob/3d36b990767f47449ae7fa3b0c664ce4fa590af3/internal/http/docs/client.go#L48), but not sure how to get to that when starting from `service.NewConfigSpec()`.

TODO: Add lints for:
- [x] upload_parallel_threads only allows values between 1 and 99 (see [here](https://docs.snowflake.com/en/sql-reference/sql/put.html#optional-parameters))
- [x] password / SSH private key are exclusive
- [x] don't allow snowpipe to be set when in user/password auth mode

Note: This is part of #658.


Jeffail:
@mihaitodor I've added blobl linting: https://github.com/benthosdev/benthos/commit/e7ee59a50a28899d131ec06359c18235ed19ebdc, worth trying when you get a chance

Jeffail:
Nice, extra props for fixing the batch docs.

PR #1162, opened 2022-03-09
Add file output option to Prometheus metrics

benschoch:
We're currently locked in some legacy windows infrastructure where the connection to a Prometheus push gateway is not possible without a lot of effort and also fetching from that server via HTTP is an issue due to security concerns.

This PR should mainly be a proof of concept to have an option to send the metrics to a file instead of a gateway or serving it via HTTP.

Unfortunately it is not possible to adapt or extend the metrics handling within our project as the metrics part is under `internal`. Maybe it could be an idea for an upcoming version to add the metrics to the public API, so we're able to implement our own metrics handler, similar to other components?

Would be nice if you could provide some feedback on whether you think this feature is useful and can be integrated without a major version increase.

### DIFF
--- a/internal/component/metrics/config_prometheus.go
+++ b/internal/component/metrics/config_prometheus.go
@@ -13,6 +13,7 @@ type PrometheusConfig struct {
 	PushBasicAuth      PrometheusPushBasicAuthConfig `json:"push_basic_auth" yaml:"push_basic_auth"`
 	PushInterval       string                        `json:"push_interval" yaml:"push_interval"`
 	PushJobName        string                        `json:"push_job_name" yaml:"push_job_name"`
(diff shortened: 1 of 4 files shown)


Jeffail:
Hey @benschoch, yeah I think this seems quite useful. I'll eventually have metrics plugin APIs that are more accessible but for now I'm happy to add this in. I'm prioritising preparation work for v4 so it might be a couple of weeks before I get around to merging this but I don't see a reason not to include it.

benschoch:
Thanks @Jeffail for your feedback. Looking forward to get this integrated ;)

Jeffail:
And it's in, thanks @benschoch!

Issue #1158, opened 2022-03-08
Expose connection sql settings as config fields


pcha:
Such as we speak in [slack](https://gophers.slack.com/archives/CLWCBK7FY/p1646757016622569) would be great if benthos support the [sql connection settings](https://github.com/go-sql-driver/mysql#dsn-data-source-name).

I thing that it could be setted in the keys `conns_max_lifetime`, `max_open_conns` and `max_iddle_conns`.

Full config example:

```yaml
# All config fields, showing default values
label: ""
sql_select:
  driver: ""
  dsn: ""
  table: ""
  columns: []
  where: ""
  args_mapping: ""
  prefix: ""
  suffix: ""
  conns_max_lifetime: 0
  max_open_conns: 0
  max_iddle_conns: 0
```

pcha:
PR: https://github.com/benthosdev/benthos/pull/1166

PR #1166, opened 2022-03-10
feat(input/sql_select): espose conn sql settings

pcha:
⚠️ Integration tests don't working on my machine

Refs: [#1158](https://github.com/benthosdev/benthos/issues/1158)

### DIFF
--- a/internal/impl/sql/input_sql_select.go
+++ b/internal/impl/sql/input_sql_select.go
@@ -243,7 +243,7 @@ func (s *sqlSelectInput) Connect(ctx context.Context) (err error) {
 	}()
 
 	db.SetConnMaxIdleTime(s.connMaxIdleTime)
-	db.SetConnMaxLifetime(s.connMaxIdleTime)
+	db.SetConnMaxLifetime(s.connMaxLifetime)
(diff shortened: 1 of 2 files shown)


Jeffail:
Hey @pcha this should be quick to add, I'm working through some remaining v4 work but will try and make time to get to this in the week. By default I'll be merging into `main` and it'll be tagged with the new v4 release, will you need this in v3 as well?

pcha:
He @Jeffail, than you. It depends of when the v4 will be released. I need this change incorporated for the next week, no pmather if it's in v3 or v4, do you think it would be possible?

Jeffail:
@pcha are you willing to run a release candidate? Otherwise we'd have to back-port it for v3 and do a minor bump.

pcha:
mmm I guess I can try with a rc for this case. If something breaks, I'll let you know and I'd ask for a minor bump in v3.

pcha:
Hi @Jeffail We just found another workaround and it's not more a priority for us, so you can keep the regular flow and merge it whe you consider it convenient. Thank you very much for the disposition to help!

Jeffail:
Looks good, thanks @pcha!

PR #1170, opened 2022-03-11
aws_sqs input - Add max_number_of_messages param in order to better integrate with fifo queues

ekeric13:
Currently when benthos's aws_sqs input module calls [receiveMessage](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html), it passes in the maximum number of messages allowed to be pulled at once: 10
This increases the efficiency of most queues but it can lead to some weird edge case behavior around fifo queues leading to a situation where the messages are not processed in order if the consumer of the queue responds with a nack. The only way to guarantee that is to only pull 1 message at a time.

This PR makes the number of messages pulled extendable by adding a param of "max_number_of_messages"

I noticed that there were a scattering of `10s` across the code and from my reading they seem to align with the number of messages inflight (number of messages pulled at once), so I modified those to match the new specified maxNumberOfMessages.

### DIFF
--- a/internal/old/input/aws_sqs.go
+++ b/internal/old/input/aws_sqs.go
@@ -63,6 +63,7 @@ You can access these metadata fields using
 			docs.FieldCommon("url", "The SQS URL to consume from."),
 			docs.FieldAdvanced("delete_message", "Whether to delete the consumed message once it is acked. Disabling allows you to handle the deletion using a different mechanism."),
 			docs.FieldAdvanced("reset_visibility", "Whether to set the visibility timeout of the consumed message to zero once it is nacked. Disabling honors the preset visibility timeout specified for the queue.").AtVersion("3.58.0"),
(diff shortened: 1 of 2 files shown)


ekeric13:
@Jeffail I have seen you have been working on the v4 roadmap. It definitely looks exciting! Is there a feature freeze until that is done?

Jeffail:
@ekeric13 not a freeze but its definitely gobbling up my time, I'm planning to put time aside for PRs this week. Most of V4 is done we're just wrapping things up so hopefully things will be moving again soon.

ekeric13:
Awesome! Appreciate the hard work.

ekeric13:
Any chance you will be able to get to this next week?

Jeffail:
On it, thanks @ekeric13! I'm planning to put a release candidate out for v4 this week.

PR #1183, opened 2022-03-22
Crdb changefeed input

danthegoodman1:
Remake see #971 for previous discussions


Jeffail:
@danthegoodman1 everything looks fine, you're registering the input correctly so I would expect that to be fine, I was planning on pulling this branch locally when its ready and properly digging in if the issue is still there.

danthegoodman1:
> @danthegoodman1 everything looks fine, you're registering the input correctly so I would expect that to be fine, I was planning on pulling this branch locally when its ready and properly digging in if the issue is still there.

@Jeffail Sorry for the late reply, was traveling all week - I would go ahead and test locally, it's ready and runs when I manually try changefeeds, but not sure why the integration test is failing. Would love some help on the final stretch here!

danthegoodman1:
Oh I think I found another issue (probably unrelated)... by default CRDB does not enable rangefeeds which is require for change data capture...

```
SET CLUSTER SETTING kv.rangefeed.enabled = true;
```

I can try using psql to manually connect and set this before running the test.

I don't think this is impacting benthos not selecting the correct input candidate, but I imagine this will be an issue once it does. I will add this to both the docs and make the tests enable this setting!

Jeffail:
Hey @danthegoodman1, I finally got around to looking at this, I've made a branch with a rework of this input so that it works similar to the input SQL components, the major difference is that here when the rows are exhausted we rerun the query with back off rather than close the input.

The work is on a branch: https://github.com/benthosdev/benthos/tree/crdb-changefeed, are you interested in rebasing off that and continuing? Otherwise I'll try and make some time to finish this up, I'm not entirely sure what we're meant to be doing regarding caching cursor timestamps so I'd need to do some reading up to finish it myself.

internal/impl/crdb/input_changefeed.go
```diff
@@ -0,0 +1,251 @@
...
+	"row": "{\"after\": {\"k\": \"1a7ff641-3e3b-47ee-94fe-a0cadb56cd8f\", \"v\": 2}, \"updated\": \"1637953249519902405.0000000000\"}", // stringified JSON object
+	"table": "strm_2"
+}`
+)
+
+func crdbChangefeedInputConfig() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Categories("Integration").
+		Summary(fmt.Sprintf("Listens to a CockroachDB Core Changefeed and creates a message for each row received. Each message is a json object looking like: \n```json\n%s\n```", sampleString)).
+		Description("Will continue to listen to the Changefeed until shutdown. If provided, will maintain the previous `timestamp` so that in the event of a restart the chanfeed will resume from where it last processed. This is at-least-once processing, as there is a chance that a timestamp is not successfully stored after being emitted and therefore a row may be sent again on restart.\n\nNote: You must have `SET CLUSTER SETTING kv.rangefeed.enabled = true;` on your CRDB cluster").
```
danthegoodman1, line 29:
```suggestion
		Description("Will continue to listen to the Changefeed until shutdown. If provided, will maintain the previous `timestamp` so that in the event of a restart the changefeed will resume from where it last processed. This is at-least-once processing, as there is a chance that a timestamp is not successfully stored after being emitted and therefore a row may be sent again on restart.\n\nNote: You must have `SET CLUSTER SETTING kv.rangefeed.enabled = true;` on your CRDB cluster").
```

danthegoodman1:
Hey @Jeffail ! Depends on how much work is remaning, as I don't have an immediate use case for this feature right now. If the branch you had worked on solved the issue in the original PR then I think the remaining work is quite minimal.

The changefeed cursor timestamp is used to track when to start feeding rows through the changefeed from. When the changefeed query is started, it normally starts by sending the entire table to catch the sink up to the DB. If you provide a timestamp, then the changefeed will only pick up from changes at or after that timestamp. This is important should benthos reboot or the connection be temporarily disconnected we do not want to send the whole table when restarting the changefeed query. All that should need to be done is persist the current timestamp somewhere durable every time a row comes through, and if one is proivided in the initial config set that as the first timestamp we have.

Is there a durable KV API within benthos that can be used to persist the cursor?

gedw99:
This looks great . Anyone know what version of cockroach this relies on. There are 2 types of change feeds in CRDB

danthegoodman1:
@gedw99 this is for core changefeeds, which are part of the open source distribution. Enterprise changefeeds handle this functionality already and would not need benthos (at least for most cases)

gedw99:
> @gedw99 this is for core changefeeds, which are part of the open source distribution. Enterprise changefeeds handle this functionality already and would not need benthos (at least for most cases)

Thank! Makes sense . 

Hoping to try this out when I get time

gedw99:
@danthegoodman1 
You could use nats kv I guess. It’s a plugin for benthos…

danthegoodman1:
> @danthegoodman1 You could use nats kv I guess. It’s a plugin for benthos…
> 
> it’s durable if you use jetstream It’s very easy to use nats kv .
> 
> https://docs.nats.io/using-nats/developer/develop_jetstream/kv

Would want to avoid additional dependencies

gedw99:
> > @danthegoodman1 You could use nats kv I guess. It’s a plugin for benthos…
> > it’s durable if you use jetstream It’s very easy to use nats kv .
> > https://docs.nats.io/using-nats/developer/develop_jetstream/kv
> 
> Would want to avoid additional dependencies

totally get you

https://github.com/benthosdev/benthos/pull/1704 is sort of interesting though.  Maybe its not such a dependency if its part of benthos anyway...

Jeffail:
Just as an update, I've had another look at this and I'm still stumped. I've implemented the ability to store and obtain a cursor from any cache resource: https://github.com/benthosdev/benthos/tree/crdb-changefeed, however the integration test that I've added is failing and I have no idea why.

I'm trying to:
- fill a table with some data
- consume the changefeed with the full state using a benthos stream
- stop the benthos stream
- create a new benthos stream that inherits the latest cursor from the previous data
- write more data into the table
- consume only the new data from the new benthos stream

Instead I'm getting nothing at all from the second stream. I assumed based on the docs in https://www.cockroachlabs.com/docs/stable/changefeed-examples?filters=core that I would be able to do this by setting `UPDATED`, storing the timestamp of the latest change, and then setting `CURSOR` to that same timestamp on the next run, but that's not working at all.

I'll try and come back to this again soon, but I'm hitting a wall and it's stealing time away from other things. If anyone has any tips please let me know.

Jeffail:
Done, merged via https://github.com/benthosdev/benthos/commit/ab0a2bc34486e97b5c913774f02d897c4a4cc60e

PR #1185, opened 2022-03-25
Add GCP Cloud Storage cache

kabukky:
Hi,

this is something we currently use as a custom benthos plugin, but would be great as a core feature.
Especially since AWS S3 is already available as a cache.

I could match the fields available in https://www.benthos.dev/docs/components/outputs/gcp_cloud_storage (thinking about `content_type`, `chunk_size`, `content_encoding`) but wanted to get a feel for what you think about it first.

Best
Kai

### DIFF
--- a/internal/impl/gcp/cache_cloud_storage.go
+++ b/internal/impl/gcp/cache_cloud_storage.go
@@ -2,7 +2,7 @@ package gcp
 
 import (
 	"context"
-	"io/ioutil"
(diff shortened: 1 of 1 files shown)


Jeffail:
Hey @kabukky, this looks great, happy to add it in. The docs probably ought to mention that `add` operations are non-atomic and therefore might not be 100% reliable.

PR #1186, opened 2022-03-27
added regex_pattern support to kafka_franz input

benmizrahi:
Added support of regex pattern with **kafka_franz** by adding regex_pattern indicator to configuration and set ConsumeRegex flag on consumer creation. consumer group will have in this case all topics that include add match topics.

This is usefully when  topics are based on some pattern and you don't need to statically supply all the list of topics needed
for example:

topic_a
topic_b
topic_c

If you need to consume them all into the same pipeline add the following pattern: [ topic_(.*) ]   
Tested locally


Jeffail:
Looks good, thanks @benmizrahi!

Jeffail:
FYI I'm planning to change the field name to `regexp_topics` in order to match up closer with similar fields.

PR #1188, opened 2022-03-30
add `mod_time` metadata to `file` input

benschoch:
Similar to i.e. the `aws_s3` input, the `file` input could also provide some more metadata provided via `os.FileInfo`.

In the first step, this includes `mod_time` and `mod_time_unix`. I could also add `size` but wasn't sure if that's desired or helpful as it just reflects `msg.Len()` or `len(part.Get())`.

Our project would heavily benefit from this as we filter files by modification time. We're currently using a custom input for that but we could get rid of it if that metadata would be available by default.

Let me know what you think 😉

### DIFF
--- a/internal/old/input/file.go
+++ b/internal/old/input/file.go
@@ -200,15 +200,17 @@ func (f *fileConsumer) ReadWithContext(ctx context.Context) (*message.Batch, rea
 
 		msg := message.QuickBatch(nil)
 		for _, part := range parts {
-			if len(part.Get()) > 0 {
-				part.MetaSet("path", currentPath)
-
-				modTimeUnix, modTime := f.getModTime(currentPath)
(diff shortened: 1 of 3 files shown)


Jeffail:
Hey @benschoch, looks good, I'm happy to add this in but I think ideally we'd need to pull out the mod time info during the `f.getReader(ctx)` so that it only happens once, otherwise we'd potentially be called fstat for each line of a file.

benschoch:
You're absolutely right! I did move it out of the `for _, part := range parts {` loop now but I didn't put it into `f.getReader` because I think it might get hard to read when 2 more return values were added to the method.
By putting it right before the loop, we should actually avoid repetitive calls now and also useless calls in error cases.
Let me know what you think.

Jeffail:
Thanks @benschoch, going to refactor this a little bit just in case we want to expand to add more stat stuff later.

PR #1190, opened 2022-03-31
fix(input/sql_select): take conn_max_lifetime correctly

pcha:


pcha:
@Jeffail Yes! I noticed that when I was trying to solve the conflicts. Thank you very much!

PR #1197, opened 2022-04-03
Fix: bloblang triple quote string parser

EmilLaursen:
During some fuzz testing of the bloblang parser, I believe a found a bug in the `TripleQuoteParser`, which can cause a panic.

This adds a test which provokes the panic, and a fix to the array bounds in the parser.


Jeffail:
Awesome, well found, thanks @EmilLaursen!

PR #1198, opened 2022-04-05
Fix schema_registry_decode processor

### LINKED ISSUE #1161: AVRO schema_registry_decode issue
I'm seeing some issues when decoding AVRO [logical types](https://avro.apache.org/docs/current/spec.html#Logical+Types) It looks like `codec.TextualFromNative` must be called [here](https://github.com/Jeffail/benthos/blob/3d36b990767f47449ae7fa3b0c664ce4fa590af3/internal/impl/confluent/processor_schema_registry_decode.go#L325), but even then there are some issues.

To reproduce, create the following `schema.avsc` file:
```json
{
  "type": "record",
  "name": "LongList2",
  "fields": [
    {
      "default": null,
      "name": "int_time_millis",
      "type": [
        "null",
        {
          "type": "int",
          "logicalType": "time-millis"
        }
      ]
    },
    {
      "default": null,
      "name": "long_time_micros",
      "type": [
        "null",
        {
          "type": "long",
          "logicalType": "time-micros"
        }
      ]
    },
    {
      "default": null,
      "name": "long_timestamp_micros",
      "type": [
        "null",
        {
          "type": "long",
          "logicalType": "timestamp-micros"
        }
      ]
    },
    {
      "default": null,
      "name": "pos_0_33333333",
      "type": [
        "null",
        {
          "logicalType": "decimal",
          "precision": 16,
          "scale": 2,
          "type": "bytes"
        }
      ]
    }
  ]
}
```

And the following `message.json` file:

```json
{"int_time_millis":{"int": 35245000},"long_time_micros":{"long":20192000000000},"long_timestamp_micros":{"long":62135596800000000},"pos_0_33333333":{"bytes":"!"}}
```

Start Redpanda in a Docker container:

```shell
> docker run --rm -p 8081:8081 -p 8082:8082 -p 9092:9092 --name redpanda docker.vectorized.io/vectorized/redpanda redpanda start --smp 1 --overprovisioned --kafka-addr 0.0.0.0:9092 --advertise-kafka-addr host.docker.internal:9092 --pandaproxy-addr 0.0.0.0:8082 --advertise-pandaproxy-addr host.docker.internal:8082
```

Start a Kafka tools Docker container and produce the above message on a topic called `test`:
```shell
> docker run --rm -it -v$(pwd):/workspace confluentinc/cp-schema-registry bash
>> cd /workspace
>> kafka-avro-console-producer --broker-list host.docker.internal:9092 --topic test --property schema.registry.url=http://host.docker.internal:8081 --property value.schema.file=schema.avsc < message.json
```

Then create the following Benthos `avro.yaml` config:

```yaml
input:
  kafka:
    addresses:
      - localhost:9092
    topics: [test]
    target_version: 1.0.0
    consumer_group: benthos_consumer_group
    client_id: benthos_kafka_input
    rack_id: ""
    start_from_oldest: true
    checkpoint_limit: 1
    commit_period: 1s
    max_processing_period: 100ms
    extract_tracing_map: ""
    group:
      session_timeout: 10s
      heartbeat_interval: 3s
      rebalance_timeout: 60s
    fetch_buffer_cap: 256

pipeline:
  threads: 1
  processors:
    - schema_registry_decode:
        url: "http://localhost:8081"

output:
  stdout:
    codec: lines
```

Running Benthos v3.65.0 produces this output:

```json
{"int_time_millis":{"int.time-millis":35245000000000},"long_time_micros":{"long.time-micros":20192000000000000},"long_timestamp_micros":{"long.timestamp-micros":"3939-01-01T00:00:00Z"},"pos_0_33333333":{"bytes.decimal":"33/100"}}
```

And changing the code [here](https://github.com/Jeffail/benthos/blob/3d36b990767f47449ae7fa3b0c664ce4fa590af3/internal/impl/confluent/processor_schema_registry_decode.go#L322-L332) to call `codec.TextualFromNative` produces this output:

```json
{"pos_0_33333333":{"bytes.decimal":"!"},"int_time_millis":{"int.time-millis":35245000},"long_time_micros":{"long.time-micros":1358741504},"long_timestamp_micros":{"long.timestamp-micros":62135596800000000}}
```

While the change does improve things, it looks like there's still an issue with `long_time_micros` which seems to require a bit of digging into goavro.

mihaitodor:
When dealing with AVRO [logical types](https://avro.apache.org/docs/current/spec.html#Logical+Types), we can't serialise the native struct emitted by goavro directly to JSON, since that will discard schema information that `codec.TextualFromNative` uses to produce the expected JSON.

Also, the tip version of goavro is required because the fix for this regression https://github.com/linkedin/goavro/issues/242 was merged, but there isn't a tagged release yet.

Fixes #1161.

For a bit of extra  detail, when converting this JSON to binary AVRO using the schema from the test I added (see #1161 for steps on how to do that):

```json
{"int_time_millis":{"int": 35245000},"long_time_micros":{"long":20192000000000},"long_timestamp_micros":{"long":62135596800000000},"pos_0_33333333":{"bytes":"!"}}
```

and then we skip the `codec.TextualFromNative` call when passing it through `schema_registry_decode`, we get the following serialised JSON from Benthos:

```json
{"int_time_millis":{"int.time-millis":35245000000000},"long_time_micros":{"long.time-micros":20192000000000000},"long_timestamp_micros":{"long.timestamp-micros":"3939-01-01T00:00:00Z"},"pos_0_33333333":{"bytes.decimal":"33/100"}}
```

which doesn't look like something that users will expect, but not sure. I have very limited experience with AVRO.

On the other hand, calling `codec.TextualFromNative` produces a JSON string which will then need to be unmarshalled back to a native struct if there are any subsequent processors. This is wasteful, but goavro doesn't seem to have any mechanism for crafting a native struct that is compatible with Benthos. I'm not sure right now what bloblang does under the hood if we call `Message.SetStructured()` with a native struct that contains types such as `time.Time`, `time.Duration`, `math/big.Rat` etc. Is that supported?


Jeffail:
Yeah I think this is the best we can do for now without making some custom component to walk the structure for custom types. In bloblang we can technically deal with any types for basic value getting, but really we need proper basic values in order for basic methods and arithmetic to work as expected.

It's not the end of the world if users make plugins that introduce odd types but our own processors definitely shouldn't as the user would have no way to work out what's happening.

mihaitodor:
Sounds good, thanks for the quick merge! :)

PR #1199, opened 2022-04-06
support CUE output format with `benthos list`

disintegrator:
This changes adds support for `benthos list --format cue` which will emit CUE types that describe the Benthos configuration file structure and the various Benthos components' configuration options.

With these types, users can build complex configurations using helpers to reduce the boilerplate, that would otherwise be present when working with YAML, while benefitting from strong type checking!

Indentation, Norway and other YAML nasties are a thing of the past!


Jeffail:
Loving it so far, one request would be to skip the cue tests if the `cue` binary isn't present, ideally I want it to be opt-in in the build process for now.

disintegrator:
> Loving it so far, one request would be to skip the cue tests if the `cue` binary isn't present, ideally I want it to be opt-in in the build process for now.

@Jeffail fixed.

cuetest/test.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
```
disintegrator, line 1: @Jeffail Did you want me to remove this altogether?

internal/cuegen/component.go
```diff
@@ -0,0 +1,92 @@
...
+		},
+	}
+
+	if len(addons) > 0 {
+		decls = append(decls, &ast.Field{
+			Label: opts.disjunctionIdent,
+			Value: ast.NewBinExpr(
+				token.AND,
+				opts.disjunctionIdent,
+				ast.NewStruct(
```
disintegrator, line 78: if you've got some `*ast.Field` (let's call it `myfield`) that you want to include in a Cue struct then you can go right ahead and do it with `ast.NewStruct(myfield)`. HOWEVER, if that field should be marked as optional then you need something a little more tricky like so: `ast.NewStruct(myfield.Label, token.Option, myField.Value)`. You will see this pattern in a few places in this PR.

internal/cuegen/component.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	decls := []ast.Decl{
+		&ast.Field{
+			Label: opts.collectionIdent,
+			Value: ast.NewStruct(fields...),
+		},
+		&ast.Field{
+			Label: opts.disjunctionIdent,
+			Value: ast.NewCall(ast.NewIdent("or"), ast.NewList(&ast.Comprehension{
```
disintegrator, line 56: So I wanted to build various disjunctions dynamically, based the available components for a given category (inputs, processors, outputs, ...). The cleanest way I've found to do this is to create a struct of all components of that category and then to you use a field comprehension and the Cue built-in function `or` to create the disjunction. You can try this following snippet in the Cue playground to see it in action:

```cue
#AllInputs: {
	broker: {
		copies?: int
		inputs?: [...null | #Input]
		batching?: {
			count?:     int
			byte_size?: int
			period?:    string
			check?:     string
			processors?: [...null | _]
		}
	}
	csv: {
		paths?: [...string]
		parse_header_row?: bool
		delimiter?:        string
		batch_count?:      int
	}
	discord: {
		channel_id:   string
		bot_token:    string
		poll_period?: string
		limit?:       int
		cache:        string
		cache_key?:   string
		rate_limit?:  string
	}
	dynamic: {
		inputs?: {
			[string]: null | #Input
		}
		prefix?: string
	}
	file: {
		paths?: [...string]
		codec?:            string
		max_buffer?:       int
		delete_on_finish?: bool
	}
}

#Input: or([for name, config in #AllInputs {(name): config}])

x: #Input

y: #Input & {
  	discord: {
		channel_id: 123  // FIXME: demoing that validation works. Changes to a string to verify.
		bot_token: "secret"
		cache: "whatever"
	}
}
```

website/docs/configuration/using_cue.md
```diff
@@ -0,0 +1,312 @@
...
+      - output:
+          retry:
+            max_retries: 5
+            output:
+              gcp_pubsub:
+                project: "sample-project"
+                topic: "sample-topic"
+```
+
+There are quite a few lines YAML here and we seem to be going sideways as we compose more functionality. We can try and make this more manageable with Cue!
```
mihaitodor, line 152: Small typo: missing `of`.

disintegrator: Thanks for catching it. Fixed!

website/docs/configuration/using_cue.md
```diff
@@ -0,0 +1,316 @@
...
+  // The desired output that will be wrapped with error handling mechanisms
+  #output: #Output
+
+  // The error text to emit if the output receives any messages which contained
+  // processing errors
+  #errorMessage: string
+
+  // The number of retries to attempt on the desired output (default is 3)
+  #maxRetries: uint | *3
+  // The error message to emit if the retry attempts are exhausted
```
mihaitodor, line 179: Missing newline

Jeffail:
This is awesome, thanks @disintegrator!

PR #1203, opened 2022-04-07
Update Changelog to mention the schema_registry_decode changes

mihaitodor:
I forgot to add this in the PR and I guess it's worth mentioning :)

LE: I also sneaked in here goavro v2.11.1, since they were nice enough to tag a release quickly.


Jeffail:
Nice, thanks @mihaitodor!

PR #1206, opened 2022-04-11
Add file_json_equals testing function

nenad:
This makes it possible to have JSON file fixtures that are big and yet human-readable. This makes it possible to have JSON file fixtures that are big and yet human-readable.

I'll submit a PR for v4 (main) as soon as this one gets in.


nenad:
I'm not exactly sure which files is the linter pointing to:
```
run golangci-lint
  Running [/home/runner/golangci-lint-1.45.2-linux-amd64/golangci-lint run --out-format=github-actions --timeout 10m cmd/... lib/... internal/... public/...] in [] ...
  Error: SA1019: strings.Title is deprecated: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead.  (staticcheck)
  Error: SA1019: strings.Title is deprecated: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead.  (staticcheck)
  Error: SA1019: bytes.Title is deprecated: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead.  (staticcheck)
  
  Error: issues found
  Ran golangci-lint in 413306ms
```

Jeffail:
Hey @nenad, ignore those linting errors, it's because the linters haven't been run against v3 for a while. Is this something you need specifically for v3? If so I wasn't planning to do a v3 release until we've built up some issue fixes that need a patch release. Ideally if you want to use this soon then I'd recommend targetting `main` and we'll include this in the first v4 release.

nenad:
Thanks @Jeffail. Closing this one as it's done in #1207.

PR #1207, opened 2022-04-11
Add file_json_equals testing function

nenad:
This makes it possible to have JSON file fixtures that are big and yet human-readable.


Jeffail:
Looks great, thanks @nenad!

PR #1216, opened 2022-04-16
Allow to specify headers in JWT tokens

mavimo:
Allow to add custom headers into JWT token.

When we try to access to google services using [JWT Auth](https://developers.google.com/identity/protocols/oauth2/service-account#jwt-auth) we need to specify the `kid` as header. At the moment is not possible to specify headers into JWT token.

Relevant informations
 - The JWT token from the original library [expose Headers property](https://github.com/dgrijalva/jwt-go/blob/v3.2.0/token.go#L23-L30)
 -  As suggested [here](https://github.com/dgrijalva/jwt-go/issues/198) to inject header the token heed to be manipulated (there aren't factory method or object methods that will allow to manipulate headers)


Jeffail:
Looks great, thanks @mavimo!

PR #1227, opened 2022-04-21
Improve the wording on the `snowflake_put` output examples

mihaitodor:
Small cosmetic changes, but I hope they clarify things a bit. In these examples, the files that end up in the Snowflake stage are not really valid JSON because they're not a proper JSON array, just a concatenation of multiple JSON documents. There is a paragraph in the Snowpipe docs [here](https://www.benthos.dev/docs/components/outputs/snowflake_put#snowpipe) where I explained how one can construct a valid JSON array if really needed, but it's not necessary and it requires extra configuration in the Snowpipe DDL.


Jeffail:
Great, thanks @mihaitodor!

PR #1231, opened 2022-04-26
Add LazyQuotes parameter to CSV input

### LINKED ISSUE #1038: [csv input] Make it possible to configure `csv.Reader`'s `LazyQuotes` field.
Under the hood, Benthos `csv input` uses the standard `encoding/csv` packages's [csv.Reader](https://pkg.go.dev/encoding/csv#Reader) struct.

The current implementation of csv input doesn't allow setting the `LazyQuotes` field.

We have a use case where we need to set the `LazyQuotes` field in order to make things work correctly.

AndrewSmith593:
Fixes #1038 

Maybe the file input [codec](https://www.benthos.dev/docs/components/inputs/file#codec) should be adjusted to take this parameter in a future PR, as well a codec option to set `parse_header_row` in order to read in CSV files with no header row.


Jeffail:
Thanks @AndrewSmith593! Looks great.

PR #1235, opened 2022-04-29
support setting job labels on `gcp_bigquery` output

disintegrator:
This change adds a config field on the `gcp_bigquery` output that can be used to attach labels to BigQuery load jobs created by this output.


Jeffail:
Looks great, thanks @disintegrator!

PR #1250, opened 2022-05-08
Add bloblang slug string method

mihaitodor:
I saw that gomplate [provides this feature](https://docs.gomplate.ca/functions/strings/#strings-slug) and I thought it would be nice to have it in Benthos too. It might come in handy for people who are offering blogging / eCommerce / etc sites as-a-service.


mihaitodor:
Hey @Jeffail, I had a look at the example you suggested and moved the implementation to `impl/lang/url` and it seems to work well.

One annoyance is that `query.MethodCategoryStrings` lives in `internal/bloblang/query/docs.go`, so not sure if you're OK to add dependencies to `internal` stuff inside the `impl` folder. I used `Category("String Manipulation")` for now, but let me know if I should change it.

Jeffail:
Very cool, thanks @mihaitodor!

Issue #1245, opened 2022-05-05
Feature Request: Add TLSTrustCertsFilePath option to Pulsar input and output


bemoore34:
**Request**: Add the Pulsar Client option to provide a CA certificate for trusting internally signed certificates.

I apologise if there is another way to support this in Benthos, but I couldn't find one.

When using TLS encrypted Broker connections ("pulsar+ssl://<URL/IP>:6651") the Pulsar client requires the signing CA's Cert in order to trust self-signed certificates. Without this, the client returns: `error="x509: certificate signed by unknown authority"`

Support for providing a CA cert can be added with the `TLSTrustCertsFilePath` option. This could be added to the pulsar.ClientOptions in the Pulsar input/output Connect methods along with the appropriate updates to the pulsarReader/Writer struct and config parsing:
```
opts := pulsar.ClientOptions{
		Logger:                 createDefaultLogger(p.log),
		ConnectionTimeout:      time.Second * 3,
		URL:               	p.url,
		TLSTrustCertsFilePath:  p.caCert
	}
```
Reference: https://pulsar.apache.org/docs/en/client-libraries-go/#tls-encryption-and-authentication

mihaitodor:
Hey @bemoore34, indeed, we are not setting that option in the input and output [here](https://github.com/benthosdev/benthos/tree/main/internal/impl/pulsar). Should be a straightforward addition. I think Ash will be happy to accept a PR if you'd like to contribute.

bemoore34:
@mihaitodor, that sounds good, I would be happy to contribute. I'll try to do that in the next couple days.

PR #1251, opened 2022-05-09
Pulsar updates

bemoore34:
Reference issue #1245.

Added the ability to configure a custom CA certificate to trust internally-signed server certificates when using TLS encryption on the broker service. Added the Pulsar GO Client's TLSTrustCertsFilePath property to the pulsar input and output files. This is useful when you are not using a trusted, public CA signed certificate for the service.

Successfully tested both the input and outputs with a Pulsar broker using TLS and not using TLS. Please let me know if you have any questions. Thanks, Brad.


Jeffail:
Hey @bemoore34 looks great, eventually it would be good to have proper support for a go `*tls.Config` but I imagine for now we're blocked by the client library.

One thing that would be good to change is to rename the field to [`tls.root_cas_file`](https://www.benthos.dev/docs/components/inputs/amqp_0_9#tlsroot_cas_file) just so that if they expand support later we can switch to the standard benthos tls config.

Jeffail:
Awesome, thanks @bemoore34!

PR #1256, opened 2022-05-11
Adding logging to file base on a "file_path" property

piclemx:
- Adding sending logs to a file
- Adding `lumberjack` to rotate logs


Jeffail:
Hey @piclemx, looks good but I think it might be worth expanding the config with a namespace and more fields before we release this, maybe something like this:

```yaml
logger:
  level: INFO
  file:
    path: ./foo/benthos.log
    rotate: true
    rotate_max_age: 1d
```

piclemx:
> Hey @piclemx, looks good but I think it might be worth expanding the config with a namespace and more fields before we release this, maybe something like this:
> 
> ```yaml
> logger:
>   level: INFO
>   file:
>     path: ./foo/benthos.log
>     rotate: true
>     rotate_max_age: 1d
> ```

I think the same as you. I was waiting on your feedback for that. I will try to get something today for that.

Jeffail:
Thanks @piclemx I'll take care of the rest on my end, I'm going to rework the logger docs to be generated so we get the full config spec listed.

PR #1268, opened 2022-05-21
Allow multiple docs in stream file

dingxiangfei2009:
This PR propose a new way to handle stream files.

It is natural to have multiple streams specified in one stream YAML file. For instance, scripts could generate several stream descriptions that may look alike or fall into the same category of functionalities. A straightforward approach by these scripts is to group those stream configurations and bundle them into one YAML file, instead of individual YAML files.

This PR will enable support for multiple documents in a stream file. Stream IDs are assigned in a way such that a suffix of form `@<document index>` is appended to the major stream ID inferred from the path of a stream file.

### DIFF
--- a/internal/cli/service.go
+++ b/internal/cli/service.go
@@ -98,9 +98,9 @@ func initStreamsMode(
 	}
 	logger.Infoln("Launching benthos in streams mode, use CTRL+C to close.")
 
-	if err := confReader.SubscribeStreamChanges(func(id string, newStreamConf stream.Config) bool {
-		if err = streamMgr.Update(id, newStreamConf, time.Second*30); err != nil && errors.Is(err, strmmgr.ErrStreamDoesNotExist) {
-			err = streamMgr.Create(id, newStreamConf)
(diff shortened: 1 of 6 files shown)


dingxiangfei2009:
Hold up. I need to add those tests. :rofl:

dingxiangfei2009:
There are a few unresolved design details pending discussion. For instance, if a stream is updated with a multi-doc, should `benthos` yank the the original _single_ stream?

Jeffail:
Hey  @dingxiangfei2009 very cool. I think the least surprising outcome would likely be removing the single stream in favour of the new identified streams. Needs a bit of thought, maybe an experimental flag in the docs to outline that multidocs are subject to being reworked.

dingxiangfei2009:
Just in case, I am currently working on putting this behind an experimental flag.

Jeffail:
Hey @dingxiangfei2009 are you still interested in this PR? I think realistically before we could consider supporting this we'd need to have a clear answer around linting and config unit tests for these multi document files. It's quite a large undertaking as there's lots of moving parts.

PR #1271, opened 2022-05-23
Fix: typo in input socket description

ahamidi:
Typo in codec description.


Jeffail:
Thanks @ahamidi, well spotted 👁️, the codec docs are generated so the root of the typo is here: https://github.com/benthosdev/benthos/blob/main/internal/codec/reader.go#L24, then afterwards run `make docs` and it'll update all the codec blurbs.

Jeffail:
Thanks @ahamidi!

PR #1277, opened 2022-05-31
Faker bloblang method

### LINKED ISSUE #1273: Bloblang method to generate fake data using Faker
## Overview
[Faker](https://github.com/bxcodec/faker) is a Golang library to generate fake data based on a struct. We want to generate a bunch of randomized test data and send those messages to a Kafka topic for performance testing and we currently use a simple Go app using the faker library. We would like to switch to using a Benthos pipeline to generate this data, so having a bloblang function that could do this would be helpful.

## Example
```
let x = {"foo":"bar"}
root = $x.fake()
```

Doing the above would replace "bar" with a randomly generate string. If instead of "bar" it was an integer, it would generate a random int value for foo. 
A lot of Faker's functionality comes from using tags in the struct, so I am not sure how we could leverage that, but there is fair amount that can be done without tags. Here's an [example](https://github.com/bxcodec/faker/blob/master/example_without_tag_test.go) of what can be done with Faker without using tags.

nmbryant:
Closes #1273

Adds a bloblang method for generating fake data using the [Faker](https://github.com/bxcodec/faker) library.
Call `fake()` on a structured object and it will go through the object replacing any Faker function keys (ex: `email`, `date`) with the result from the corresponding Faker function. Currently only supports those 2 faker functions, but will add support for all the [available functions](https://github.com/bxcodec/faker/blob/master/example_single_fake_data_test.go) assuming this initial code looks good.

Tests are currently failing for this branch due to the example I have for the `faker` method since it uses randomly generated strings. Not sure what the best way to handle that is. Also, I still need to update the docs for the `fake()` method after the initial review.


nmbryant:
Moved the `fake` function and method to the `lang` package. Still need to add more details to documentation after initial review.

Also @mihaitodor suggested improving the tests by extending [`NewExampleSpec`](https://github.com/nmbryant/benthos/blob/2ed21a30bf31270af25811271f476e14fb41d6fa/internal/bloblang/query/docs.go#L12) and adding `WithInitFunc` where the necessary `Set*` function in faker can be called so that each call to the corresponding faker method will return the same thing. I plan on making that change assuming these initial changes look good and we want to go forward with this function and method.

nmbryant:
Removed the method and updated the function to support all the faker functions. I included all the faker functions that are available, though there are some that don't seem particularly helpful. Considered removing them, but figured it would be best to just support all that faker provides.
Also, only have a single test for the faker function right now validating email. Should I add tests for each faker function?

Jeffail:
> Should I add tests for each faker function?

I think it's fine to test just a couple of them. The docs are looking pretty huge, I think maybe it's worth reducing down to just a couple of examples and adding an exhaustive list of types to the description rather than an example for each one. As long as the description has a link to [faker](https://github.com/bxcodec/faker) then users can dig in as much as they need.

nmbryant:
Makes sense, updated the docs to only have a few examples along with a list of the available functions.

Jeffail:
@nmbryant this looks good to me, happy for merging or do you want to keep this draft for now?

Jeffail:
Ace, looks great, thanks @nmbryant!

PR #1280, opened 2022-06-03
Add option for idempotent Kafka producer for Kafka output

### LINKED ISSUE #1223: Enable configuration of Kafka idempotent producer in output
Kafka's idempotent producer ensures that only one copy of a record is written to the Kafka broker, even in the face of network failures. Sarama supports this feature.

I'm not sure how it's done in Sarama, but usually you just need to configure a `transactional.id`, which uniquely identifies distinct client instances across restarts.

abhishekamte:
Resolves #1223 

Add a boolean option to kafka output to use idempotent producer

### DIFF
--- a/internal/component/output/config_kafka.go
+++ b/internal/component/output/config_kafka.go
@@ -10,27 +10,28 @@ import (
 
 // KafkaConfig contains configuration fields for the Kafka output type.
 type KafkaConfig struct {
-	Addresses        []string    `json:"addresses" yaml:"addresses"`
-	ClientID         string      `json:"client_id" yaml:"client_id"`
-	RackID           string      `json:"rack_id" yaml:"rack_id"`
-	Key              string      `json:"key" yaml:"key"`
-	Partitioner      string      `json:"partitioner" yaml:"partitioner"`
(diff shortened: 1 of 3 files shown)


Jeffail:
Thanks @abhishekamte, I think the description as it currently stands is a bit misleading, the guarantee of exactly-once delivery is only with respect to a single producer instance within the client library itself. In practice the majority of duplicates would be introduced during server crashes and catastrophic network failures as the services recover, in which case this field wouldn't help at all. I'm not 100% familiar with sarama internals but I'm assuming this based off https://stackoverflow.com/a/67909822 and the fact that we're not populating a transaction ID from the input.

I'd be happier with a more muted description such as: "Enable idempotent writes, this eliminates the possibility of message duplicates being written to a topic after recovering from network failures. Set this to true only if ack_replicas is true and Kafka version is above 0.11.0."

rupurt:
@Jeffail this is fairly important especially considering the default is true since [Kafka 3.0](https://blogs.apache.org/kafka/entry/what-s-new-in-apache6).

@abhishekamte are you still around to make the changes? Or if not can I fork this PR and re-submit with the modifications requested by @Jeffail?

PR #1283, opened 2022-06-07
Support expressions in redis output stream field

### LINKED ISSUE #1083: Add Interpolation/dynamic routing support for redis streams
For Kafka, benthos supports dynamic routing, e.g:
```
output:
  kafka:
    addresses: [ TODO ]
    topic: ${! meta("output_topic") }
    key: ${! meta("key") }
```

It would be great if similar support could be added for redis streams, if possible.

bmarinov:
PR fixes #1083 

This is a quick fix for the lack of expression support in the `stream` configuration on the redis stream output `output.redis_streams.stream` config field.

I tried to follow the integration test setup for redis, but in the end I had to add a new test that allows for additional metadata to be set on messages.

I have a couple of questions and some todo's - adding comments on the PR.


internal/integration/stream_test_definitions.go
```diff
@@ -873,3 +873,29 @@ func StreamTestOutputOnlyOverride(getFn GetMessageFunc) StreamTestDefinition {
...
+
+			messageContent := `{"foo": "bar"}`
+			metaFieldVal := []string{metaFieldName, metaFieldValue}
+
+			require.NoError(t, sendMessage(env.ctx, t, tranChan, messageContent, metaFieldVal...))
+			received := receiveMessage(env.ctx, t, input.TransactionChan(), nil)
+
+			messageMatch(t, received, messageContent, metaFieldVal...)
+
+			// todo: add proper assert?
```
bmarinov, line 898: Any good asserts here? My initial idea was to test the in/output configuration, but it is not available (for a good reason) and in theory it is enough to test that the message arrived. But then the test itself is kind of decoupled from what I am setting up on the other side in `integration_test.go` :/

internal/impl/redis/integration_test.go
```diff
@@ -100,6 +101,42 @@ input:
 		})
 	})
 
+	t.Run("stream-interpolated", func(t *testing.T) {
```
bmarinov, line 104: It feels wrong to have this test separate, but otherwise I would have to change / extend all the other redis streams tests. Is there a better way? Same problem with the `getFn` approach.

internal/impl/redis/output_streams.go
```diff
@@ -137,7 +142,7 @@ func (r *redisStreamsWriter) WriteWithContext(ctx context.Context, msg *message.
 	if msg.Len() == 1 {
 		if err := client.XAdd(&redis.XAddArgs{
 			ID:           "*",
-			Stream:       r.conf.Stream,
+			Stream:       r.interpolatedStreamName.String(0, msg),
```
bmarinov, line 145: This magic value (index `0`) is due to the fact that there is a single message in the batch. Is it obvious enough and have I understood the message loop for the expression handling correctly?

internal/impl/redis/integration_test.go
mihaitodor: If you'd like to avoid adding an extra test, I'd probably just make a minor modification to `TestIntegrationRedis` and do `stream: ${! "stream" }-$ID` instead of `stream: stream-$ID`. This way, there's no need to add `StreamTestWithMetadata`, since the existing tests will cover it.

internal/impl/redis/output_streams.go
```diff
@@ -72,14 +73,15 @@ func newRedisStreamsOutput(conf output.Config, mgr bundle.NewManagement, log log
 type redisStreamsWriter struct {
 	log log.Modular
 
-	conf       output.RedisStreamsConfig
-	metaFilter *metadata.ExcludeFilter
+	conf                   output.RedisStreamsConfig
+	interpolatedStreamName *field.Expression
```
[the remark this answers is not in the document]
mihaitodor: Its type does hint that it's an interpolated field, so I'd probably stick with `stream`, but up to you

internal/impl/redis/output_streams.go
```diff
@@ -90,6 +92,9 @@ func newRedisStreamsWriter(conf output.RedisStreamsConfig, log log.Modular) (*re
 	if r.metaFilter, err = conf.Metadata.Filter(); err != nil {
 		return nil, fmt.Errorf("failed to construct metadata filter: %w", err)
 	}
+	if r.interpolatedStreamName, err = mgr.BloblEnvironment().NewField(conf.Stream); err != nil {
```
mihaitodor, line 95: Maybe move this above the `metaFilter` part, so they're in the same order as the fields in the struct

internal/impl/redis/integration_test.go
bmarinov: Sounds like a good idea. Any suggestions on how to set custom metadata in messages through the test config? One option would be to add an additional `StreamTestOpt_` function and extend `StreamTestConfigVars`. Otherwise it seems that I would have to change the tests for the "streams" suite.

internal/impl/redis/output_streams.go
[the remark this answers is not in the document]
bmarinov: Renamed to `stream`.

internal/impl/redis/output_streams.go
bmarinov: Moved :)

internal/integration/stream_test_definitions.go
bmarinov: Now removed as no longer necessary to test this separately :)

internal/impl/redis/integration_test.go
bmarinov: Alright, fixed now. Removed the new test suite and used bloblang in the output processor to inject metadata in all streams tests. All are passing now.

Jeffail:
Thanks @bmarinov, looks great. I'll fix merge conflicts on my end as I've been refactoring constructor signatures.

PR #1284, opened 2022-06-08
Update `type()` method docs and examples

mihaitodor:
Following 1661dfa, we now have an extra type called `timestamp`.


Jeffail:
:heart: thanks @mihaitodor

PR #1294, opened 2022-06-15
Update kafka franz logging to be able to treat debug logs like trace logs

ekeric13:
This is kind of a crazy PR and not great code imo. Having one component have a different log level than the rest would be a much better alternative but I don't see a way to do that...


Jeffail:
Hey @ekeric13, I think I'd actually be in favour of converting debug to trace by default without introducing the extra configuration field. I think it's a fairly niche requirement and makes sense to limit the most verbose logs of client libraries to the tracing level.

PR #1295, opened 2022-06-15
Kafka Franz - add aws msk iam

ekeric13:
Still need to QA

Related:
https://github.com/benthosdev/benthos/pull/1292
https://github.com/benthosdev/benthos/pull/1293

### DIFF
--- a/website/docs/components/inputs/kafka_franz.md
+++ b/website/docs/components/inputs/kafka_franz.md
@@ -62,6 +62,14 @@ input:
       root_cas_file: ""
       client_certs: []
     sasl: []
+    credentials:
+      profile: ""
+      id: ""
+      secret: ""
+      token: ""
+      from_ec2_role: false
+      role: ""
+      role_external_id: ""
 ```
 
 </TabItem>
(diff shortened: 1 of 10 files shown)


website/docs/components/inputs/kafka_franz.md
```diff
@@ -62,6 +62,14 @@ input:
       root_cas_file: ""
       client_certs: []
     sasl: []
+    credentials:
+      profile: ""
+      id: ""
+      secret: ""
+      token: ""
```
ekeric13, line 69: instead of explicitly defining these in the kafka yaml, a user can simply just pass in `AWS_ACCESS_KEY` and `AWS_SECRET_KEY`. And it will be caught due to the behavior of this:
https://github.com/ekeric13/benthos/blob/9595631f6c0d1514ab9bc5006e9d38b7421cabc3/internal/impl/aws/session.go#L52

correct?

Jeffail: Yep, it's fine to leave the credentials fields empty or undefined.

Jeffail:
Thanks @ekeric13 looks good. I'm going to merge and then adapt these changes to make the AWS imports optional, similar to a recent change to elasticsearch: https://github.com/benthosdev/benthos/commit/e2772b5e5a57f57d32f37ee6596f262076179395, it'd be great if you could re-test after that change.

Jeffail:
I've moved a bit of stuff around: https://github.com/benthosdev/benthos/commit/0ed2244646ec735b5d7a29a99e86a80d6d11c765, hope you don't mind but the credentials fields are now within the mechanism object, which would potentially allow you to try two different sets of AWS credentials if needed. It looks like this:

```yaml
sasl:
  - mechanism: AWS_MSK_IAM
    aws:
      region: us-east-1
      credentials:
         profile: foo
```

ekeric13:
I don't mind moving stuff around at all. I do want to say that I wasn't able to fully test this one locally so I might have a follow up for it.

Issue #1276, opened 2022-05-31
Feature Request: add Pusher output / input


lorenz-qbees:
Some of our customers are using Pusher (https://pusher.com) to provide real time data to clients.
There is a simple GoLang integration, so it should be possible to consume / provide messages from / to their API.
We'll try to implement this feature on our own and keep this issue updated.

twmb:
Closing as stale due to inactivity. If a Pusher output is still needed, feel free to open a new issue or PR.

PR #1298, opened 2022-06-17
Add Pusher output

marcel-qbees:
This PR implements #1276

This allows to use [Pusher](pusher.com) Pubsub as an output.

Successfully tested, formatted and linted (make test, make fmt, make lint).

Also implements batching and allows to use an interpolation function for setting the Pusher channel.


Jeffail:
Very neat, thanks @marcel-qbees! I'm assuming there isn't any form of mock docker service we could do integration tests against? Doesn't block us from adding an experimental output but would be nice one day.

PR #1300, opened 2022-06-19
Add support for encrypted TLS private keys

chrisdoc:
**Context:**
We use TLS Certificates with an encrypted private key to connect to our Kafka brokers. Therefore I thought it would be a good idea to add support for encrypted keys to benthos.

**Solution:**
Decode the provided key and check if it is encrypted and try to decrypt it with the user provided password. I have added test for both certificates as file and string.


Jeffail:
Thanks @chrisdoc looks great, just need to add an additional entry to https://github.com/benthosdev/benthos/blob/main/internal/tls/docs.go#L44 in order for the documentation to match.

chrisdoc:
@Jeffail thank you for the fast feedback; I went ahead and added the missing documentation update :)

Jeffail:
Awesome, thanks @chrisdoc, extra :heart: for adding more tests

PR #1303, opened 2022-06-21
Allow multiple input batches in tests.

sapk:
It is a try to implement suggested enhancement: Implement https://github.com/benthosdev/benthos/issues/1289

I try to make as little changes as needed so let me know if this implementation is ok or if it need some adjustments.

I added a test to validate context across batches (count). ~Ideally, I would have liked to be able to use dedup to show dedup across batch or not depending on some configuration but I haven't found how to register cache resources in a test environment. 
Otherwise, I test this similar logic manually from a config file and it worked.~ I added a simple dedup use case to validate why I needed this feature.

### DIFF
--- a/config/test/deduplicate.yaml
+++ b/config/test/deduplicate.yaml
@@ -17,9 +17,16 @@ tests:
         - content: '2'
         - content: '3'
         - content: '4'
+        - content: '3'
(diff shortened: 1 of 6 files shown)


sapk:
It should be ok for review

PR #1304, opened 2022-06-21
Fix healthcheck bug

mihaitodor:
Don't write the header and message twice if both input and output aren't connected.

With the current code, the following will pop up in the logs:
```
2022/06/21 23:10:14 http: superfluous response.WriteHeader call from github.com/benthosdev/benthos/v4/internal/stream.New.func2 (type.go:57)
```


Jeffail:
Thanks @mihaitodor, if I remember right the dodgy code pathing here was because we wanted to print errors for both the input and output being unavailable at the same time when applicable. I might try and simplify it a little bit further.

PR #1308, opened 2022-06-27
Elasticsearch Upsert

DragFAQ:
https://github.com/benthosdev/benthos/issues/318
I tested it locally. It will create doc (if not exists) or update existance.


Jeffail:
Nice, thanks @DragFAQ!

PR #1309, opened 2022-06-30
Add remote ip and TLS information to http_server meta

### LINKED ISSUE #1052: Enhancement Request - http input - provide additional meta data
**Enhancement Request**

It would be beneficial for Benthos to provide additional network and tls meta data for the http input (and perhaps other network inputs).

This would be like Logstash provides for some sources.
For example - using Logtash's beats input, the user then can make use of the following provided variables:
@metadata.ip_address
@metadata.tls_peer.status
@metadata.tls_peer.protocol
@metadata.tls_peer.subject
@metadata.tls_peer.cipher_suite

References
https://www.elastic.co/guide/en/logstash/current/plugins-inputs-beats.html#plugins-inputs-beats-ecs_metadata

isae:
This fixes #1052

I couldn't come up with the solution how to test TLS fields population, so I just tested it locally (using self-signed certificate made with mkcert).

### DIFF
--- a/internal/impl/io/input_http_server.go
+++ b/internal/impl/io/input_http_server.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


Jeffail:
Looks great, thanks @isae!

PR #1310, opened 2022-06-30
Update catch.md

Myrotik:
updated the documentation for the correct word


Jeffail:
Thanks @Myrotik, well spotted, the docs are generated but the tidbit at the top of the file is wrong, the typo needs to get fixed here: https://github.com/benthosdev/benthos/blob/main/internal/impl/pure/processor_catch.go#L55, then run `make docs` and it'll generate the other bits.

Jeffail:
You don't really need Go experience but it does require a dev environment set up. No worries, it's only a quick change on my end so I'll merge and update.

PR #1316, opened 2022-07-07
Support for queue declare auto-delete bit in AMQP 0-9-1 input

initialed85:
**Context**

This PR addresses https://github.com/benthosdev/benthos/issues/1315.

**In this PR**

- Add `auto_delete` as an option for queue declaration in AMQP 0-9-1 inputs
- Extend integration test for AMQP 0-9-1 input to cover `auto_delete` omitted and `auto_delete` set

I ran the tests, linter and formatter and also ran the integration tests (just for AMQP 0-9-1 input) and it all seemed fine.

I was able to see the changed doco when I ran the website locally.

### DIFF
--- a/internal/impl/amqp09/input.go
+++ b/internal/impl/amqp09/input.go
@@ -85,6 +85,7 @@ then the declaration passively verifies that they match the target fields.`,
 			).WithChildren(
 				docs.FieldBool("enabled", "Whether to enable queue declaration.").HasDefault(false),
 				docs.FieldBool("durable", "Whether the declared queue is durable.").HasDefault(true),
+				docs.FieldBool("auto_delete", "Whether the declared queue will auto-delete.").HasDefault(false),
 			).Advanced(),
 			docs.FieldObject("bindings_declare",
 				"Allows you to passively declare bindings for the target queue.",
@@ -192,12 +193,12 @@ func (a *amqp09Reader) ConnectWithContext(ctx context.Context) (err error) {
 
 	if a.conf.QueueDeclare.Enabled {
 		if _, err = amqpChan.QueueDeclare(
-			a.conf.Queue,                // name of the queue
-			a.conf.QueueDeclare.Durable, // durable
-			false,                       // delete when unused
-			false,                       // exclusive
-			false,                       // noWait
-			nil,                         // arguments
+			a.conf.Queue,                   // name of the queue
+			a.conf.QueueDeclare.Durable,    // durable
+			a.conf.QueueDeclare.AutoDelete, // delete when unused
+			false,                          // exclusive
+			false,                          // noWait
+			nil,                            // arguments
 		); err != nil {
 			return fmt.Errorf("queue Declare: %s", err)
 		}
--- a/internal/impl/amqp09/integration_test.go
+++ b/internal/impl/amqp09/integration_test.go
@@ -1,7 +1,9 @@
 package amqp09
 
 import (
+	"encoding/json"
 	"fmt"
+	"net/http"
 	"testing"
 	"time"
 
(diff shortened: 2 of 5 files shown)


internal/impl/amqp09/input.go
```diff
@@ -111,7 +112,7 @@ then the declaration passively verifies that they match the target fields.`,
 	}
 }
 
-//------------------------------------------------------------------------------
+// ------------------------------------------------------------------------------
```
initialed85, line 115: My IDE (IntelliJ, has formal Go support) basically forced these comment formatting changes (I think it runs `gofmt` any time I save).

Is that okay?

internal/impl/amqp09/integration_test.go
```diff
@@ -14,30 +16,43 @@ import (
 )
 
 func TestIntegrationAMQP09(t *testing.T) {
-	integration.CheckSkip(t)
-	t.Parallel()
+	assertQueueStateFromRabbitMQManagementAPI := func(resource *dockertest.Resource) {
```
initialed85, line 19: This closure here runs as a test cleanup (because I needed to be sure the suites had completed) but it does do some some assertion.

Is there a way that's more in keeping with the pattern to do this?

internal/impl/amqp09/integration_test.go
```diff
@@ -64,27 +79,82 @@ input:
...
+	commonTests := integration.StreamTests(
 		integration.StreamTestOpenClose(),
 		integration.StreamTestMetadata(),
 		integration.StreamTestMetadataFilter(),
 		integration.StreamTestSendBatch(10),
 		integration.StreamTestStreamSequential(1000),
 		integration.StreamTestStreamParallel(1000),
+	)
+
+	lossyTests := integration.StreamTests(
```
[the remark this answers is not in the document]
initialed85: Hmm- on thinking about it (and testing in my real environment), the fact that the lossy tests fail in this configuration is actually problematic.

Lossy connections are a fact of life and we want Benthos to keep consuming as per config in that event.

I'll change the tests so that all suites run for both config variants and look what what I can do with the AMQP lifecycle to ensure we re-establish the queues and bindings on reconnect when auto-delete is set.

internal/impl/amqp09/integration_test.go
```diff
@@ -80,11 +126,29 @@ input:
...
+	t.Cleanup(func() {
+		assertQueueStateFromRabbitMQManagementAPI(resource)
+	})
+}
+
+func TestIntegrationAMQP09WithoutQueueDeclareAutoDelete(t *testing.T) {
+	doSetupAndAssertions(false, t)
+}
+
+func TestIntegrationAMQP09WithQueueDeclareAutoDelete(t *testing.T) {
```
initialed85, line 152: Unfortunately this test is failing (hitting the timeout) and I can't work out why.

We end up at https://github.com/benthosdev/benthos/blob/3c94c306b5cd16562f642fd26a5a4887d24c04a9/internal/integration/stream_test_helpers.go#L598

initialed85: I'm running this fork for the purpose mentioned in the issue (siphon off a cloned stream of prod stats for testing) and I am intermittently seeing on one of the instances (always the most swamped one, might be unrelated) that Benthos eventually (silently) stops doing as required.

Config looks like this:

```json
http:
  address: 0.0.0.0:4196

input:
  label: "api_stat_import_stream_source"
  amqp_0_9:
    urls:
      - "amqp://${STAT_IMPORT_CLONE_WMT_MQ_USERNAME}:${STAT_IMPORT_CLONE_WMT_MQ_PASSWORD}@${STAT_IMPORT_CLONE_WMT_MQ_IP}:5672/"
    queue_declare:
      enabled: true
      durable: false
      auto_delete: true
    bindings_declare:
      - exchange: "APIStatImportServer_main_fanout"
    auto_ack: true

output:
  label: "api_stat_import_stream_destination"
  amqp_0_9:
    urls:
      - "amqp://${WMT_MQ_USERNAME}:${WMT_MQ_PASSWORD}@${WMT_MQ_IP}:5672/"
    exchange: "APIStatImportServer_main_fanout"
    exchange_declare:
      enabled: true
      type: fanout
      durable: true
```

If I look at the management UI for the RabbitMQ servers, the source (prod) still has a happy consumer and the queue isn't growing, so some part of Benthos is still draining the queue as expected- just not getting anything in the destination RabbitMQ instance (also seems to have happy connections).

No errors in the Benthos CLI, but I haven't yet tried cranking up logging.

Jeffail: That test closes an existing consumer and starts a new one up, so you're probably seeing the result of the old consumer being torn down along with all the data, then the consumer starts up and the queue gets remade. I would just disable that test when `auto_delete: true` as the two are incompatible anyway and I wouldn't expect it to work.

internal/impl/amqp09/input.go
Jeffail: I doubt it's `gofmt` as I've never seen it happen before, I'd be happier just deleting the lines if you can't prevent your IDE from changing it.

internal/impl/amqp09/integration_test.go
```diff
@@ -77,14 +123,41 @@ input:
...
-		integration.StreamTestStreamParallelLossy(1000),
-		integration.StreamTestStreamParallelLossyThroughReconnect(1000),
 	)
-	suite.Run(
-		t, template,
-		integration.StreamTestOptSleepAfterInput(500*time.Millisecond),
-		integration.StreamTestOptSleepAfterOutput(500*time.Millisecond),
+
+	// we can't run these tests when auto-delete is not set because the disconnect / reconnect cycle cleans up the queues under test
+	if !setQueueDeclareAutoDelete {
```
initialed85, line 129: This has changed since last revision- we don't run the lossy tests while exercising the auto-delete functionality.

initialed85:
@Jeffail I believe this is ready for re-review now, apologies for taking so long to update it.

Integration tests pass locally (w/ lossy tests not run for the auto-delete mode).

Jeffail:
Awesome, looks great, thanks @initialed85!

PR #1329, opened 2022-07-15
Add note about http_server not being at least once delivery.

kennyp:
Just a documentation update.


Jeffail:
Thanks @kennyp, looks good.

PR #1333, opened 2022-07-20
Upgrade github.com/microcosm-cc/bluemonday

mihaitodor:
The version that's currently used by Benthos was retracted:

> warning: github.com/microcosm-cc/bluemonday@v1.0.17: retracted by module author: Retract older versions as only latest is to be depended upon


Jeffail:

PR #1334, opened 2022-07-20
log: expose naming of message field via logrus

sondermax:
hey there! just piggy backing on top of d5da5fd40fce575c4bd6ed9b6a2511a729acb76c , we saw that the logger `message` field changed to `msg` with logrus. this just adds configuration for that field, so that tooling can optionally make it whatever it wants!

let me know if this is stepping on toes, just trying to get back to the old behavior and follow the timestamp changes. :) 

thanks for the look!


Jeffail:
Hey @sondermax, makes sense and looks good, thanks!

PR #1337, opened 2022-07-21
add avro-ocf input codec

loicalleyne:
Hi, I'm not trying to short circuit the contribution guidelines, I wasn't able to run any tests of this code on my machine because I'm having issues building Benthos but hoping you don't mind taking a quick look to see if I'm on the right path.

### DIFF
--- a/internal/codec/reader.go
+++ b/internal/codec/reader.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/benthosdev/benthos/v4/internal/docs"
 	"github.com/benthosdev/benthos/v4/internal/message"
+	"github.com/linkedin/goavro"
 )
 
 // ReaderDocs is a static field documentation for input codecs.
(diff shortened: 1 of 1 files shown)


Jeffail:
Hey @loicalleyne yeah looks like you've pretty much got it done, I'd probably err on the side of caution and add an "EXPERIMENTAL: " prefix to the docs field description.

loicalleyne:
I thought the import defaults to v2, in the docs it says if you want v1 you need to specify it.
I just saw the tests failed, I think downgrading my go to v1. 16 might fix my build problem, will try that and  try to get some actual testing done with some of my avro files.

mihaitodor:
Good job! Your import definitely needs to be `"github.com/linkedin/goavro/v2"` just like [here](https://github.com/benthosdev/benthos/blob/a58fac0e2647d673928af570081c81894b36ab77/internal/impl/confluent/processor_schema_registry_decode.go#L18) and it should build just fine with Go 1.18

loicalleyne:
This is embarassing... my builds were failing because go in WSL was still at v1.13.
anywho, the output of a simple file input(avro-ocf) to file output(lines) looks has the data but not in JSON, think the codec.TextualFromNative() is the wrong method to use, perhaps NativeFromBinary() followed by a marshall to JSON? Now that my builds are working I'll test some more and resubmit the PR.

PR #1339, opened 2022-07-22
chore: upgrade aztables module to v0.6.0

davidandradeduarte:
Upgrading [aztables](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/data/aztables@v0.6.0) to ~~v1.0.1~~ v0.6.0 (the latest version compatible with go 1.17).

Small changes were required in the azure_table_storage output.

Tested against a real Azure storage account and an [Azurite container](https://hub.docker.com/_/microsoft-azure-storage-azurite) with the following [client](https://github.com/davidandradeduarte/benthos-storage-aztables).

Context:
We have custom application code using a different aztables module (v0.6.0), which isn't retro compatible with the one Benthos uses internally (v0.5.0). Ideally, the Azure SDK for Go would not add breaking changes in a minor version, but not everyone respects SemVer. 🤷

Cheers @Jeffail


mihaitodor:
Another good reason to thumbs up https://github.com/benthosdev/benthos/issues/1307

Jeffail:
Thanks @davidandradeduarte, I think a sensible plan is to use this version for now, wait for 1.19 to come out before we lean fully into 1.18, and then upgrade to v1 as soon as that happens. I'll add a note to the 1.18 issue so we (hopefully) remember to revisit.

PR #1349, opened 2022-07-28
Add method to parse url encoded data

mariosant:
Adds method `parse_url_encoded` to bloblang strings so it can transform form url encoded data to something benthos can work with.

Example:
```
pipeline:
  processors:
    - bloblang: |
        root = content().parse_url_encoded()
```


Jeffail:
Looks great, thanks @mariosant!

PR #1354, opened 2022-07-30
Fix: azure_table_storage output not writing the first message when the table does not exist

eduardodbr:
Hello!

I have found a bug with azure_table_storage output. Whenever it tries to write a message to a table that doesn't exist, it creates the table and exits, without writing the message. This PR fixes that issue and I also introduced a performance improvement by only trying to create the table if the transaction fails with a "TableNotFound" error. The current output is always trying to create the table. 

NOTE: in case you are wondering why I'm searching for the string "TableNotFound" instead of searching for a 404 code, the azure sdk accepts the request (returning 202) and only the multipart data will have 4XX responses for the batch request that failed. Might be useful to read on azure error handling, if you want: https://docs.microsoft.com/en-us/rest/api/storageservices/performing-entity-group-transactions#sample-error-response


Jeffail:
Looks good, thanks @eduardodbr!

PR #1355, opened 2022-07-31
fix(cue): avoid cycles in higher-order components

disintegrator:
Higher-order components such as the `retry` output can cause structural cycles within CUE's type system. To break these cycles, we need to declare their child component fields as optional.

Additionally, the test I had initially written was not detecting issues in the generated CUE schema. This has now been fixed and I was able to get it to fail, as expected, on the structural cycle error before fixing it in this PR.


disintegrator:
I've asked the CUE community on Slack for better solutions and might follow up with a refactor PR if I find one.

disintegrator:
According to CUE maintainer, there might be a bug in CUE so I've opened an issue with them to get to the bottom of it. I think it's safe for us to proceed with this fix in the interim. Equally, I'm happy to wait for a response on the issue.

https://github.com/cue-lang/cue/issues/1827

PR #1360, opened 2022-08-03
Add "content_type" field to gcs cache

kabukky:
This is something we actively use, since the official Google Cloud Storage library is kind of flaky when automatically inferring the Content-Type.
The "content_type" field should be optional to still use the automatic inferring of the library if so desired.

### DIFF
--- a/internal/impl/gcp/cache_cloud_storage.go
+++ b/internal/impl/gcp/cache_cloud_storage.go
@@ -17,7 +17,7 @@ func gcpCloudStorageCacheConfig() *service.ConfigSpec {
 		Field(service.NewStringField("bucket").
 			Description("The Google Cloud Storage bucket to store items in.")).
 		Field(service.NewStringField("content_type").
-			Description("Optional field to explicitly set the Content-Type.").Default("").Optional())
+			Description("Optional field to explicitly set the Content-Type.").Optional())
 
 	return spec
 }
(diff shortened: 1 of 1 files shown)


internal/impl/gcp/cache_cloud_storage.go
```diff
@@ -15,7 +15,9 @@ func gcpCloudStorageCacheConfig() *service.ConfigSpec {
 		Summary(`Use a Google Cloud Storage bucket as a cache.`).
 		Description(`It is not possible to atomically upload cloud storage objects exclusively when the target does not already exist, therefore this cache is not suitable for deduplication.`).
 		Field(service.NewStringField("bucket").
-			Description("The Google Cloud Storage bucket to store items in."))
+			Description("The Google Cloud Storage bucket to store items in.")).
+		Field(service.NewStringField("content_type").
+			Description("Optional field to explicitly set the Content-Type.").Default("").Optional())
```
kabukky, line 20: I guess I can get rid of `.Default("")` when also using `.Optional()` but I wanted to make sure. What do you say?

Jeffail: Hey @kabukky, yeah that's correct, all you need is optional.

kabukky: Thanks! I've changed it now.

Jeffail:
Sorry, merged this on stream so forgot to comment, thanks @kabukky!

PR #1362, opened 2022-08-04
fix: OTEL propagation for the HTTP server input

robertmarsal:
At the moment when an HTTP request is made to Benthos on the HTTP input with a `traceparent` header the propagation is not happening because Go normalises headers and changes the case (uppercase first). The headers are then converted to a `TextMap` carrier which is case-sensitive so when extraction happens no values are found since the OTEL libraries are looking for `traceparent` in lowercase.

This PR ensures the headers are lowercased before passed to the extractor for the HTTP input where we deal with headers.

Propagation now works as expected and spans are being linked.


Jeffail:
Sorry, merged this on stream so forgot to comment, thanks @robertmarsal!

PR #1363, opened 2022-08-04
feat: allow delaying shutdown of benthos process

disintegrator:
This change adds a top-level `shutdown_delay` config option that sets an optional shutdown delay duration. This option can be used to delay the start of the shutdown procedure. This is useful for pipelines that need a short grace period to have their metrics and traces scraped. While the shutdown delay is in effect, the HTTP metrics endpoint continues to be available for scraping and any active tracers are free to flush remaining traces.

The shutdown delay can be interrupted by sending the Benthos process a second OS interrupt or termination signal.

**Why not use a [Prometheus Pushgateway](https://prometheus.io/docs/practices/pushing/) instead?**

Push gateways are designed to be a workaround for systems that cannot be scraped, either because they don't expose metrics data over an endpoint or they are behind a firewall that prevents scrapers from accessing metrics endpoints. Additionally, deploying a Pushgateway instance increases operational burden and adds a single point of failure in the chain (one more thing to keep alive). According to the official Prometheus docs, Pushgateway is only recommended for such limited use cases and the preference should be to have service metrics scraped from an endpoint.

For users that share these views, they will tend to delay the termination of there short-lived pipelines and batch jobs using certain hacks. For example in Benthos this can setup can be used to create a synthetic delay:

```yaml

input:
  sequence:
      # The input that has a finite amount of data
    - gcp_bigquery_select: {}
    - generate:
        mapping: "root = {}"
        interval: 30s
        # the count is 2 since it'll fire immediately the first time
        # and then wait one interval
        count: 2
      processors:
        - bloblang: root = deleted()
```

This is a cumbersome solution that convolutes an otherwise simple pipeline configuration compared to:

```yaml
shutdown_delay: 30s

input:
  gcp_bigquery_select: {}
```

---

Sample config:

```yaml
shutdown_delay: 5s

input:
  generate:
    mapping: |-
      root = { "id": uuid_v4(), "message": "Hello, World!" }
    count: 5
    interval: 1s

output:
  drop: {}

```

Sample output:

```
INFO Running main config from specified file       @service=benthos path=scratch/sample.yml
INFO Dropping messages.                            @service=benthos label="" path=root.output
INFO Shutdown delay is in effect for 5s            @service=benthos
```

### DIFF
--- a/internal/cli/service.go
+++ b/internal/cli/service.go
@@ -383,6 +383,15 @@ func cmdService(
 		close(httpServerClosedChan)
 	}()
 
+	var exitDelay time.Duration
+	if td := conf.SystemCloseDelay; len(td) > 0 {
+		var err error
+		if exitDelay, err = time.ParseDuration(td); err != nil {
+			logger.Errorf("Failed to parse shutdown delay period string: %v\n", err)
+			return 1
+		}
+	}
+
 	var exitTimeout time.Duration
 	if tout := conf.SystemCloseTimeout; len(tout) > 0 {
 		var err error
@@ -394,6 +403,18 @@ func cmdService(
 
 	// Defer clean up.
 	defer func() {
+		if exitDelay > 0 {
+			select {
+			// We should only cause a shutdown delay if the pipeline has terminated.
+			// Other termination conditions are considered intentional and require
+			// immediate shutdown.
+			case <-dataStreamClosedChan:
+				logger.Infof("Shutdown delay is in effect for %s\n", exitDelay)
+				<-time.After(exitDelay)
+			default:
+			}
+		}
+
 		go func() {
 			_ = httpServer.Shutdown(context.Background())
 			select {
--- a/internal/cli/service.go
+++ b/internal/cli/service.go
@@ -404,14 +404,9 @@ func cmdService(
 	// Defer clean up.
 	defer func() {
 		if exitDelay > 0 {
-			select {
-			// We should only cause a shutdown delay if the pipeline has terminated.
-			// Other termination conditions are considered intentional and require
-			// immediate shutdown.
-			case <-dataStreamClosedChan:
-				logger.Infof("Shutdown delay is in effect for %s\n", exitDelay)
-				<-time.After(exitDelay)
-			default:
+			logger.Infof("Shutdown delay is in effect for %s\n", exitDelay)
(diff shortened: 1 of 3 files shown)


internal/cli/service.go
```diff
@@ -394,6 +403,18 @@ func cmdService(
 
 	// Defer clean up.
 	defer func() {
+		if exitDelay > 0 {
+			select {
+			// We should only cause a shutdown delay if the pipeline is has terminated.
```
mihaitodor, line 408: Small typo :)

website/docs/configuration/about.md
```diff
@@ -279,10 +279,25 @@ benthos -c ./your-config.yaml echo
 
 You can check the output of the above command to see if certain sections are missing or fields are incorrect, which allows you to pinpoint typos in the config.
 
+## Shutting down
+
+Under normal operating conditions, the Benthos process will shut down when there are no more messages produced by inputs and the final message has been processed. The shutdown procedure can also be initiated by sending the process a interrupt (`SIGINT`) or termination (`SIGTERM`) signal. There are two top-level configuration options that control the shutdown behaviour: `shutdown_timeout` and `shutdown_delay`.
+### Shutdown delay
+
+The `shutdown_delay` option can be used to delay the start of the shutdown procedure. This is useful for pipelines that will process a finite number of messages and terminate but need a short grace period to have their metrics and traces scraped (or pushed). While the shutdown delay is in effect, the HTTP metrics endpoint continues to be available for scraping and any active tracers are free to flush remaining traces.
```
mihaitodor, line 287: > to have their metrics and traces scraped (or pushed)

Doesn't Benthos attempt to push metrics just before exiting? https://www.benthos.dev/docs/components/metrics/prometheus#push-gateway

mihaitodor (approved):

internal/cli/service.go
disintegrator: Thanks for catching it! It's fixed now.

website/docs/configuration/about.md
disintegrator: Ah nice catch! I remove the `(or pushed)` wording but you've also touched on a subject that I should at least discuss in the PR description.

Jeffail (approved):

Jeffail:

PR #1365, opened 2022-08-07
Add "json_path" bloblang method

kabukky:
This is something we are actively using.
JSONPath is very useful when handling unstructured/schemaless data.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,8 +22,8 @@ require (
 	github.com/Jeffail/grok v1.1.0
 	github.com/Masterminds/squirrel v1.5.2
 	github.com/OneOfOne/xxhash v1.2.8
-	github.com/PaesslerAG/gval v1.2.0 // indirect
-	github.com/PaesslerAG/jsonpath v0.1.1 // indirect
+	github.com/PaesslerAG/gval v1.2.0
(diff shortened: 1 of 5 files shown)


Jeffail:
Hey @kabukky this looks good but could you rework it to use the public plugin APIs? And place it within its own package similar to https://github.com/benthosdev/benthos/tree/main/internal/impl/maxmind so that we can isolate the dependencies.

Also I think it'd be a good idea to mark this as experimental in the docs like here: https://github.com/benthosdev/benthos/blob/main/internal/impl/maxmind/bloblang_geoip.go#L18 since this is going to be a really powerful feature I'd like to get user feedback before marking it stable.

kabukky:
Hey @Jeffail, thank you so much for your comment on this.
I've reworked this PR to use the public plugin API. Is this okay? Please tell me if anything is missing.

mihaitodor:
Small nit: Would you mind running `make deps` and committing the changes so the tests will run?

kabukky:
Hey @mihaitodor, thanks for your comment. I've done that now, hope it helps.

Jeffail (approved):
Looks good, thanks @kabukky, I'll update the docs on my side.

PR #1374, opened 2022-08-10
Fix schema registry encode's field "avro_raw_json"

ekeric13:
### Manual test description

If the schema is:
```
{
  "type": "record",
  "name": "heartbeat_v3",
  "namespace": "outbox_events",
  "fields": [
      {
          "name": "heartbeat",
          "type": [
              "null",
              "int"
          ],
          "default": null
      }
  ]
}
```

When the benthos config is:
```
schema_registry_encode:
  url: ${SCHEMA_REGISTRY_ENDPOINT}
  subject: ${! meta("EventName")}
  avro_raw_json: false
```

If the input is:
```
'{"heartbeat": {"int": 1}}'
```

The logs show a success:
Screen Shot 2022-08-10 at 5 35 41 AM

If the input is:
```
'{"heartbeat": 1}'
```

The logs show a failure:
Screen Shot 2022-08-10 at 5 35 53 AM

When the benthos config is:
```
schema_registry_encode:
  url: ${SCHEMA_REGISTRY_ENDPOINT}
  subject: ${! meta("EventName")}
  avro_raw_json: true
```

If the input is:
```
'{"heartbeat": {"int": 1}}'
```

The logs show a failure:
Screen Shot 2022-08-10 at 5 36 48 AM

If the input is:
```
'{"heartbeat": 1}'
```

The logs show  a success:
Screen Shot 2022-08-10 at 5 37 13 AM

### Thoughts on default value

So this PR does keep `avro_raw_json` defaulted to false.

Though this PR doesn't change the default value of `avro_raw_json` (it is already false), it does changes the default codec mechanism from [`NewCodecForStandardJSON`](https://pkg.go.dev/github.com/linkedin/goavro/v2#NewCodecForStandardJSON) to [`NewCodec`](https://pkg.go.dev/github.com/linkedin/goavro/v2#NewCodec), so previously though we stated the default mechanism was avro json the code was expecting standard json.
So there is an argument to change `avro_raw_json` to default to `true`, as this is a breaking change given that the code now expects you to have your json formatted differently when `avro_raw_json` is false.
The argument to set the default to `true` is bolstered by the fact that most new time avro users will expect to be able to pass normal json, and there are some slight differences in standard json and the spec: https://avro.apache.org/docs/current/specification/_print/#json-encoding

But I think this will cause more confusion in the long run as:
1. The standard java implementation of avro doesn't support raw json
People have been asking avro to support normal json for as far back as 2014
https://issues.apache.org/jira/browse/AVRO-1582
And as recently in 2021:
https://issues.apache.org/jira/browse/AVRO-3210
But from the responses you can tell that avro maintainers are open to supporting raw json but in no rush.
So many users, I assume most are coming from java, will be caught off guard by defaulting to raw json.

2. No matter how the avro is encdoded (whether `avro_raw_json` is true or false), it is always decoded to avro json.
For example if your input to avro encoder is this:
```
`{"Address":{"City":"foo","State":"bar"},"Name":"foo","MaybeHobby":"dancing"}`
```
then after you decode the binary data your output is this:
```
{"Address":{"my.namespace.com.address":{"City":"foo","State":"bar"}},"MaybeHobby":{"string":"dancing"},"Name":"foo"}
```
Given this, I think it would be unnatural for the kafka producers to be handling a different json format than the kafka consumers by default.

3. Changing the default value to `true` does mess anyone up who didn't set the value of `avro_raw_json` and kept the default of false, thinking they are using avro json this whole time.
For example they may not have used any union types, they upgrade benthos to latest version, and then at a much later date they go to add an avro json value that involves a union, and then they are unexpectedly given an error because the default value of `avro_raw_json` has switched to true and wants raw json. So I would argue that whether we keep the value as `false` or set it to `true`, we are creating a breaking change. And `avro_raw_json` defaulting to `false` will cause less confusion in the long run.

### DIFF
--- a/internal/impl/confluent/processor_schema_registry_encode.go
+++ b/internal/impl/confluent/processor_schema_registry_encode.go
@@ -353,9 +353,7 @@ func (s *schemaRegistryEncoder) getLatestEncoder(subject string) (schemaEncoder,
 		return nil, 0, err
 	}
 
-	s.logger.Infof("codec 1 is: %#v", codec)
-
-	s.logger.Infof("codec 2 is: %v", codec)
+	s.logger.Tracef("codec is: %#v", codec)
 
 	return func(m *service.Message) error {
 		var datum interface{}
@@ -372,7 +370,7 @@ func (s *schemaRegistryEncoder) getLatestEncoder(subject string) (schemaEncoder,
 			return err
 		}
 
-		s.logger.Infof("datum is: %#v", datum)
+		s.logger.Tracef("datum is: %#v", datum)
 
 		binary, err := codec.BinaryFromNative(nil, datum)
 		if err != nil {
--- a/internal/impl/confluent/processor_schema_registry_encode.go
+++ b/internal/impl/confluent/processor_schema_registry_encode.go
@@ -353,6 +353,10 @@ func (s *schemaRegistryEncoder) getLatestEncoder(subject string) (schemaEncoder,
 		return nil, 0, err
 	}
 
+	s.logger.Infof("codec 1 is: %#v", codec)
+
+	s.logger.Infof("codec 2 is: %v", codec)
+
 	return func(m *service.Message) error {
 		var datum interface{}
 		if s.avroRawJSON {
@@ -367,6 +371,8 @@ func (s *schemaRegistryEncoder) getLatestEncoder(subject string) (schemaEncoder,
 		} else if datum, err = m.AsStructured(); err != nil {
 			return err
 		}
+
+		s.logger.Infof("datum is: %#v", datum)
 
 		binary, err := codec.BinaryFromNative(nil, datum)
 		if err != nil {
--- a/internal/impl/confluent/processor_schema_registry_encode.go
+++ b/internal/impl/confluent/processor_schema_registry_encode.go
@@ -35,7 +35,7 @@ Currently only Avro schemas are supported.
 
 ### Avro JSON Format
 
-By default this processor expects documents formatted as [Avro JSON](https://avro.apache.org/docs/current/spec.html#json_encoding) when encoding Avro schemas. In this format the value of a union is encoded in JSON as follows:
+By default this processor expects documents formatted as [Avro JSON](https://avro.apache.org/docs/current/specification/_print/#json-encoding) when encoding Avro schemas. In this format the value of a union is encoded in JSON as follows:
 
 - if its type is ` + "`null`, then it is encoded as a JSON `null`" + `;
 - otherwise it is encoded as a JSON object with one name/value pair whose name is the type's name and whose value is the recursively encoded value. For Avro's named types (record, fixed or enum) the user-specified name is used, for other types the type name is used.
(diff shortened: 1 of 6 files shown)


internal/impl/confluent/processor_schema_registry_encode_test.go
```diff
@@ -210,7 +210,7 @@ func TestSchemaRegistryEncodeAvro(t *testing.T) {
 		{
 			name:        "message doesnt match schema",
 			input:       `{"Address":{"my.namespace.com.address":"not this","Name":"foo"}}`,
-			errContains: "schema does not specify default value",
+			errContains: "cannot decode textual union: cannot decode textual record",
```
ekeric13, line 213: Changed as was getting this new error:

```
--- FAIL: TestSchemaRegistryEncodeAvro (0.00s)
    --- FAIL: TestSchemaRegistryEncodeAvro/message_doesnt_match_schema (0.00s)
        processor_schema_registry_encode_test.go:231:
            	Error Trace:	processor_schema_registry_encode_test.go:231
            	Error:      	"cannot decode textual record \"foo.namespace.com.identity\": cannot decode textual union: cannot decode textual record \"my.namespace.com.address\": expected: '{'; actual: '\"' for key: \"my.namespace.com.address\" for key: \"Address\"" does not contain "schema does not specify default value"
            	Test:       	TestSchemaRegistryEncodeAvro/message_doesnt_match_schema
```

I am assuming because a different codec parser is being used the error message is different.

internal/impl/confluent/processor_schema_registry_encode.go
```diff
@@ -348,26 +348,33 @@ func (s *schemaRegistryEncoder) getLatestEncoder(subject string) (schemaEncoder,
...
+			return nil, 0, err
+		}
+	} else {
+		if codec, err = goavro.NewCodec(resPayload.Schema); err != nil {
+			s.logger.Errorf("failed to parse response for schema subject '%v': %v", subject, err)
+			return nil, 0, err
+		}
 	}
 
+	s.logger.Tracef("codec is: %#v", codec)
```
ekeric13, line 363: please let me know a good format for trace logs. I think it would be nice to expose the user what codec has been parsed if trace log level is enabled.

internal/impl/confluent/processor_schema_registry_encode.go
```diff
@@ -348,26 +348,33 @@ func (s *schemaRegistryEncoder) getLatestEncoder(subject string) (schemaEncoder,
...
+		}
 
-			if datum, _, err = codec.NativeFromTextual(b); err != nil {
-				return err
-			}
-		} else if datum, err = m.AsStructured(); err != nil {
+		datum, _, err := codec.NativeFromTextual(b)
+		if err != nil {
 			return err
 		}
```
mihaitodor, line 374: Hey @ekeric13, when I looked at this [on Discord](https://discord.com/channels/746368194196799589/754099840354615296/1006945611578351667), I got the impression that we still need this if / else logic here, but we should probably come up with some relevant test cases for both cases. Feel free to hit me up on Discord if you'd like to brainstorm through the corner cases.

mihaitodor (approved):
Thank you for your patience with this and for adding the tests. LGTM!

ekeric13:
I strongly appreciated the feedback. Feel a lot better about it after talking it over and putting in a bit more work.

Jeffail (approved):
Thanks @ekeric13, this stuff is becoming a bit of a nightmare to navigate to appreciate all the info. I'll try and come up with a meaningful changelog entry but would appreciate a proof read of it.

Jeffail:
Also these processors are still marked as experimental so if there's a solid argument for changing defaults then I think we're okay to do so.

PR #1379, opened 2022-08-12
Add `Snowflake ID` to `Bloblang`

### LINKED ISSUE #1377: Feature suggestion: Add snowflake ID to boblang
Hey, Benthos Team.

Recently we come is a situation that we could use `Snowflake ID` in Benthos. 
If it sounds like a good idea, I'd be more than happy to create a PR and add this feature.

ehcaning:
Here is a tiny PR to add `snowflake_id` support to Benthos.
Resolves #1377

### DIFF
--- a/internal/bloblang/query/functions.go
+++ b/internal/bloblang/query/functions.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/Jeffail/gabs/v2"
-	"github.com/bwmarrin/snowflake"
 	"github.com/gofrs/uuid"
 	gonanoid "github.com/matoous/go-nanoid/v2"
 	"github.com/segmentio/ksuid"
@@ -697,38 +696,6 @@ var _ = registerSimpleFunction(
 		return u4.String(), nil
 	},
 )
-
-//------------------------------------------------------------------------------
-
-var _ = registerFunction(
-	NewFunctionSpec(
-		FunctionCategoryGeneral, "snowflake_id",
-		"Generate a new snowflake ID each time it is invoked and prints a string representation.",
-		NewExampleSpec("", `root.id = snowflake_id()`),
-		NewExampleSpec("It is possible to specify the node_id.", `root.id = snowflake_id(1)`),
-	).
-		Param(ParamInt64("node_id", "An optional node_id.").Optional()),
-	snowflakeidFunction,
-)
-
-func snowflakeidFunction(args *ParsedParams) (Function, error) {
-	nodeArg, err := args.FieldOptionalInt64("node_id")
-	if err != nil {
-		return nil, err
-	}
-
-	return ClosureFunction("function snowflake_id", func(ctx FunctionContext) (interface{}, error) {
-		nodeID := int64(1)
-		if nodeArg != nil {
-			nodeID = *nodeArg
-		}
-		node, err := snowflake.NewNode(nodeID)
-		if err != nil {
-			return nil, err
-		}
-		return node.Generate().String(), nil
-	}, nil), nil
-}
 
 //------------------------------------------------------------------------------
 
--- a/internal/bloblang/query/functions.go
+++ b/internal/bloblang/query/functions.go
@@ -703,7 +703,7 @@ var _ = registerSimpleFunction(
 var _ = registerSimpleFunction(
 	NewFunctionSpec(
 		FunctionCategoryGeneral, "snowflake_id",
-		"Generates a snowflake ID and prints a string representation.",
+		"Generate a new snowflake ID each time it is invoked and prints a string representation.",
 		NewExampleSpec("", `root.id = snowflake_id()`),
 	),
 	func(_ FunctionContext) (interface{}, error) {
--- a/internal/bloblang/query/functions.go
+++ b/internal/bloblang/query/functions.go
@@ -700,21 +700,35 @@ var _ = registerSimpleFunction(
 
 //------------------------------------------------------------------------------
 
-var _ = registerSimpleFunction(
+var _ = registerFunction(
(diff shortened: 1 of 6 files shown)


internal/bloblang/query/functions.go
```diff
@@ -699,6 +700,24 @@ var _ = registerSimpleFunction(
...
 //------------------------------------------------------------------------------
 
+var _ = registerSimpleFunction(
+	NewFunctionSpec(
+		FunctionCategoryGeneral, "snowflake_id",
+		"Generate a new snowflake ID each time it is invoked and prints a string representation.",
+		NewExampleSpec("", `root.id = snowflake_id()`),
+	),
+	func(_ FunctionContext) (interface{}, error) {
+		node, err := snowflake.NewNode(1)
```
mihaitodor, line 710: I guess some users might want to pass in a custom node number, but not sure. Maybe consider having it as an optional parameter with `1` as the default value

internal/bloblang/query/functions.go
```diff
@@ -699,6 +700,24 @@ var _ = registerSimpleFunction(
...
+		if err != nil {
+			panic(err)
```
mihaitodor, line 712: We'd rather avoid panicking here if possible. Please have a look at other functions for error handling patterns.

mihaitodor (changes requested):
Thanks for this addition! Since the underlying library is doing some small computations [on init](https://github.com/bwmarrin/snowflake/blob/3cf6d89d1fe90688e862739ec26be74ebef77a57/snowflake.go#L57-L76), it would be better to have this added [here](https://github.com/benthosdev/benthos/blob/main/internal/impl/lang/bloblang.go) instead so people can opt out from importing it when creating a custom binary.

internal/bloblang/query/functions.go
ehcaning: Thanks for the suggestion, it's done ✅

internal/bloblang/query/functions.go
ehcaning: Thanks for the suggestion, it's done ✅

mihaitodor:
Thanks for the changes! Don't forget to move the new code to https://github.com/benthosdev/benthos/blob/main/internal/impl/lang/bloblang.go as mentioned above

mihaitodor:
> > Thanks for the changes! Don't forget to move the new code to https://github.com/benthosdev/benthos/blob/main/internal/impl/lang/bloblang.go as mentioned above
> 
> Sorry, I'm a bit lost here, can you explain more please? I don't know what should I do 😄

No prob! You added the functionality to `internal/bloblang/query/functions.go`, but these are core Benthos functions which should come by default for anyone who imports `github.com/benthosdev/benthos/v4/public/components/pure` into their code. However, there's also `github.com/benthosdev/benthos/v4/public/components/pure/extended` for such extras that might not get used frequently and add bloat. In Go, all [`init()` functions](https://go.dev/ref/spec#Package_initialization) get run on startup before `main()`, so, ideally, we'd want to make it possible for people to skip importing them. So, instead of importing `github.com/benthosdev/benthos/v4/public/components/all`, they might cherry-pick only the imports they need from https://github.com/benthosdev/benthos/tree/main/public/components. Given this, you'll want to move your code to https://github.com/benthosdev/benthos/blob/main/internal/impl/lang/bloblang.go and https://github.com/benthosdev/benthos/blob/main/internal/impl/lang/bloblang_test.go so it only gets picked up when people import `github.com/benthosdev/benthos/v4/public/components/pure/extended` (or `github.com/benthosdev/benthos/v4/public/components/all`). Hope that makes sense.

ehcaning:
> > > Thanks for the changes! Don't forget to move the new code to https://github.com/benthosdev/benthos/blob/main/internal/impl/lang/bloblang.go as mentioned above
> > 
> > 
> > Sorry, I'm a bit lost here, can you explain more please? I don't know what should I do 😄
> 
> No prob! You added the functionality to `internal/bloblang/query/functions.go`, but these are core Benthos functions which should come by default for anyone who imports `github.com/benthosdev/benthos/v4/public/components/pure` into their code. However, there's also `github.com/benthosdev/benthos/v4/public/components/pure/extended` for such extras that might not get used frequently and add bloat. In Go, all [`init()` functions](https://go.dev/ref/spec#Package_initialization) get run on startup before `main()`, so, ideally, we'd want to make it possible for people to skip importing them. So, instead of importing `github.com/benthosdev/benthos/v4/public/components/all`, they might cherry-pick only the imports they need from https://github.com/benthosdev/benthos/tree/main/public/components. Given this, you'll want to move your code to https://github.com/benthosdev/benthos/blob/main/internal/impl/lang/bloblang.go and https://github.com/benthosdev/benthos/blob/main/internal/impl/lang/bloblang_test.go so it only gets picked up when people import `github.com/benthosdev/benthos/v4/public/components/pure/extended` (or `github.com/benthosdev/benthos/v4/public/components/all`). Hope that makes sense.

Thanks for the detailed comment. I've updated my PR.
I'll appreciate a review. 🤝

mihaitodor (approved):
LGTM. Might be nice to place a sample Snowflake ID in the description, so people will know what they're getting when looking at the docs

ehcaning:
@mihaitodor Thanks, can we go ahead and merge it?

mihaitodor:
> @mihaitodor Thanks, can we go ahead and merge it?

Unfortunately, it will have to wait until @Jeffail has a bit of time to look at it, since he's the only maintainer. If it's urgent for you, the best I can suggest is to use a custom build.

Jeffail (approved):
Looks good, thanks @ehcaning!

PR #1380, opened 2022-08-14
Convert YAML config to JSON for HandleList

nibbleshift:
The dynamic CRUD API is marshalling a YAML object into json.RawMessage object.   The HandleList endpoints are returning an internal server error because it cannot marshal the YAML into JSON.     This PR converts the YAML into valid JSON before it is returned to the caller.


nibbleshift:
Fixes #1398

Jeffail:
Fixed slightly further upstream: https://github.com/benthosdev/benthos/commit/c151d57b32a51011feca9114ad633ddccd0dbfb0

PR #1386, opened 2022-08-16
Fix redis processor config bug

mihaitodor:
I believe this is a regression introduced in 530547a.


Jeffail (approved):
Nice catch, thanks @mihaitodor!

PR #1388, opened 2022-08-17
New Opensearch sink 

arunx2:
Added new OpenSearch output component. I just opened a PR for initial review and planning to make some more changes for metrics and use public config API


internal/impl/opensearch/writer_integration_test.go
```diff
@@ -0,0 +1,667 @@
...
+	if err = m.Write(message.QuickBatch([][]byte{
+		[]byte(`{"message":"hello world","user":"2"}`),
+		[]byte(`{"message":"hello world","user":"3"}`),
+	})); err != nil {
+		t.Error(err)
+	}
+
+	for i := 0; i < 3; i++ {
+		id := fmt.Sprintf("foo-%v", i+1)
+		// nolint:staticcheck // Ignore SA1019 Type is deprecated warning for .Index()
```
mihaitodor, line 172: Nit: `.Index()`?

go.mod
```diff
@@ -86,6 +86,7 @@ require (
 	github.com/nsqio/go-nsq v1.1.0
 	github.com/olivere/elastic/v7 v7.0.31
 	github.com/opencontainers/runc v1.0.3 // indirect
+	github.com/opensearch-project/opensearch-go v1.1.0
```
mihaitodor, line 89: Shouldn't we be using https://pkg.go.dev/github.com/opensearch-project/opensearch-go/v2 ?

internal/impl/opensearch/output.go
```diff
@@ -0,0 +1,454 @@
...
+It's possible to enable AWS connectivity with this output using the `+"`aws`"+`
+fields. However, you may need to set `+"`sniff` and `healthcheck`"+` to
+false for connections to succeed.`),
+		Config: docs.FieldComponent().WithChildren(
+			docs.FieldString("urls", "A list of URLs to connect to. If an item of the list contains commas it will be expanded into multiple URLs.", []string{"http://localhost:9200"}).Array(),
+			docs.FieldString("index", "The index to place messages.").IsInterpolated(),
+			docs.FieldString("action", "The action to take on the document. This field must resolve to one of the following action types: `create`, `index`, `update`, `upsert` or `delete`.").IsInterpolated().Advanced(),
+			docs.FieldString("pipeline", "An optional pipeline id to preprocess incoming documents.").IsInterpolated().Advanced(),
+			docs.FieldString("id", "The ID for indexed messages. Interpolation should be used in order to create a unique ID for each message.").IsInterpolated(),
+			docs.FieldString("type", "The document mapping type. This field is required for versions of elasticsearch earlier than 6.0.0, but are invalid for versions 7.0.0 or later.").Optional().IsInterpolated(),
```
mihaitodor, line 87: Should this still reference elasticsearch?

internal/impl/opensearch/output.go
```diff
@@ -0,0 +1,454 @@
...
+func (e *OpenSearch) Write(msg message.Batch) error {
+	if e.client == nil {
+		return component.ErrNotConnected
+	}
+
+	requests := make([]*pendingBulkIndex, msg.Len())
+	if err := msg.Iter(func(i int, part *message.Part) error {
+		jObj, ierr := part.AsStructured()
+		if ierr != nil {
+			e.log.Errorf("Failed to marshal message into JSON document: %v\n", ierr)
```
mihaitodor, line 287: This seems a bit redundant, but I see it's [the same](https://github.com/benthosdev/benthos/blob/ef8fb69ad85a3b7311bacaa78565b8a6b5127ead/internal/impl/elasticsearch/output.go#L275) in the ES output. Might be a leftover that we should clean up (cc @Jeffail)

internal/impl/opensearch/output.go
```diff
@@ -0,0 +1,454 @@
...
+			ID:       e.idStr.String(i, msg),
+		}
+		return nil
+	}); err != nil {
+		return err
+	}
+	start := time.Now()
+	b, _ := opensearchutil.NewBulkIndexer(opensearchutil.BulkIndexerConfig{
+		FlushInterval: time.Second * 1,
+		FlushBytes:    0,
```
mihaitodor, line 306: I see these two have different defaults in [the docs](https://pkg.go.dev/github.com/opensearch-project/opensearch-go/v2/opensearchutil#BulkIndexerConfig). Why are we adjusting them?

internal/impl/opensearch/output.go
```diff
@@ -0,0 +1,454 @@
...
+		Client:        e.client,
+	})
+	ctx := context.Background()
```
mihaitodor, line 309: I think we'll want to propagate the `ctx` parameter from `WriteBatch` here. Otherwise, `b.Add` and `b.Close` could block forever...

internal/impl/opensearch/output.go
```diff
@@ -0,0 +1,454 @@
...
+	for _, v := range requests {
+		bulkReq, err := e.buildBulkableRequest(v)
+		if err != nil {
+			return err
+		}
+		_ = b.Add(ctx, *bulkReq)
```
mihaitodor, line 315: Should we not check the error here?

internal/impl/opensearch/output.go
```diff
@@ -0,0 +1,454 @@
...
+	}
+
+	_ = b.Close(ctx)
```
mihaitodor, line 318: Should we not check the error here?

internal/impl/opensearch/output.go
```diff
@@ -0,0 +1,454 @@
...
+	if biStats.NumFailed > 0 {
+		e.log.Errorf(
+			"Indexed [%s] documents with [%s] errors in %s (%s docs/sec)",
+			humanize.Comma(int64(biStats.NumFlushed)),
+			humanize.Comma(int64(biStats.NumFailed)),
+			dur.Truncate(time.Millisecond),
+			humanize.Comma(int64(1000.0/float64(dur/time.Millisecond)*float64(biStats.NumFlushed))),
+		)
+	} else {
+		e.log.Infof(
```
mihaitodor, line 332: This might get noisy, I think. Might be better to use `Debugf`.

internal/impl/opensearch/output.go
```diff
@@ -0,0 +1,454 @@
...
+func (e *OpenSearch) Close(context.Context) error {
+	return nil
+}
+
+// Build a bulkable request for a given pending bulk index item.
+func (e *OpenSearch) buildBulkableRequest(p *pendingBulkIndex) (*opensearchutil.BulkIndexerItem, error) {
+	switch p.Action {
+	case "update":
+		// opensearch util needs the body root element should be "doc"
+		doc, _ := sjson.SetOptions("", "doc", p.Doc, &sjson.Options{ReplaceInPlace: true})
```
mihaitodor, line 353: Should we not check the error here?

internal/impl/opensearch/output.go
```diff
@@ -0,0 +1,454 @@
...
+			Index:      p.Index,
+			DocumentID: p.ID,
+			Action:     "delete",
+		}
+		if p.Routing != "" {
+			r.Routing = &p.Routing
+		}
+		return r, nil
+	case "index":
+		jsonData, _ := json.Marshal(p.Doc)
```
mihaitodor, line 378: Should we not check the error here?

internal/impl/opensearch/output.go
```diff
@@ -0,0 +1,454 @@
...
+		}
+		if p.ID != "" {
+			r.DocumentID = p.ID
+		}
+		if p.Routing != "" {
+			r.Routing = &p.Routing
+		}
+		return r, nil
+	case "create":
+		jsonData, _ := json.Marshal(p.Doc)
```
mihaitodor, line 392: Should we not check the error here?

internal/impl/opensearch/output.go
```diff
@@ -0,0 +1,454 @@
...
+	}
+	return st.Transport.RoundTrip(req)
+}
+
+// Read method to read the content from io.reader to string
+func Read(r io.Reader) string {
+	var b bytes.Buffer
+	_, _ = b.ReadFrom(r)
+	return b.String()
+}
```
mihaitodor, line 454: I believe this one's only used inside the unit test code and I think swallowing the error returned by `b.ReadFrom(r)` isn't ideal. Would it be possible to remove it from here?

mihaitodor (changes requested):
Hey @arunx2, thanks for this contribution! Looks good overall. I haven't tested the code yet, but I did do an initial read through it and left some comments

internal/impl/opensearch/output.go
arunx2: This must use the config value. Tried out something quickly forgot to clean

internal/impl/opensearch/writer_integration_test.go
arunx2: I'll remove it.

internal/impl/opensearch/output.go
arunx2: I'll test this out. Because OpenSearch 1.x has default type as _doc and in 2.x, it is fully removed.

internal/impl/opensearch/output.go
arunx2: We already had it checked before calling this method. I thought it would be redundant. Let me see if I can improve

internal/impl/opensearch/output.go
arunx2: Done! I love these entire review feedback of nitty-gritty details

Jeffail:
Hey @arunx2, any chance you could shift this over to the new public plugin APIs? A good example would be something like the jetstream input: https://github.com/benthosdev/benthos/blob/main/internal/impl/nats/input_jetstream.go. Otherwise I can do the migration myself but might need to wait a little while.

arunx2:
I have that in my to-do. I'll do it in a day or two

Thanks
Arun

internal/impl/opensearch/output.go
arunx2: I moved this method back to tests and handled err as well.

go.mod
arunx2: Thanks for the suggestion. I updated to v2

mihaitodor: Awesome, glad that worked! I'll have to look at the changes in more detail, but, on a quick note, I think you might need to add an import here as well: https://github.com/benthosdev/benthos/blob/main/public/components/all/package.go

internal/impl/opensearch/output.go
arunx2: Totally Agreed!

arunx2:
> 

I tried to migrate to new public API, and it looks like we might need to implement helper methods for handling basic authentication, optional AWS configuration like we did for TLS configuration and back off configurations.

internal/impl/opensearch/output.go
arunx2: Done!

internal/impl/opensearch/output.go
arunx2: This might be required for opensearch go client as the update requires interface{} object.

arunx2:
@Jeffail @mihaitodor Can you review when you get time? I've some cycles to work on this week and the next. Thanks!

arunx2:
@Jeffail I'm happy to work on next steps if known :) Thanks.

internal/impl/opensearch/output.go
arunx2: done

internal/impl/opensearch/output.go
arunx2: Done

arunx2:
> Hey @arunx2, thanks for this contribution! Looks good overall. I haven't tested the code yet, but I did do an initial read through it and left some comments

I pushed new changes and adopted the new API. Please review when you have time.

arunx2:
@Jeffail @mihaitodor Can one of you kindly review the PR? Kindly let me know what are the next steps. I'm just keeping the PR up to date with the main branch as much as I can. As I'm working with many customers in OpenSearch, I hope I can show a great demo to them with Benthos and OpenSearch.

arunx2 (commented):
I modified the code as you suggested. Please review when you have time.

is3ka1:
Hi @arunx2 ,

I think it need to make a minor adjustment of adding 
`_ "github.com/benthosdev/benthos/v4/public/components/opensearch"` import to the [public/components/all/package.go](https://github.com/arunx2/benthos/blob/opensearch/public/components/all/package.go) file, so that `cmd/benthos/main.go` cli can include the opensearch output.

Also appreciate the efforts to keep the opensearch output feature usable. As someone who also needs this feature, I really hope this pr will be merged soon.

Thanks for all the work

arunx2:
> Hi @arunx2 ,
> 
> I think it need to make a minor adjustment of adding `_ "github.com/benthosdev/benthos/v4/public/components/opensearch"` import to the [public/components/all/package.go](https://github.com/arunx2/benthos/blob/opensearch/public/components/all/package.go) file, so that `cmd/benthos/main.go` cli can include the opensearch output.
> 
> Also appreciate the efforts to keep the opensearch output feature usable. As someone who also needs this feature, I really hope this pr will be merged soon.
> 
> Thanks for all the work

Thanks @is3ka1, for your review and hope. I added the OpenSearch package to all/package.go

arunx2:
@Jeffail @mihaitodor, Can you comment on the status/plan, please?

Jeffail:
Hey @arunx2 it looks as though you're still using the older plugin APIs and config structs to register the plugin so before merging this we'd need to manually migrate it ourselves. New component plugins shouldn't generally require any imports from the internal package (with the exception being the inner `aws` package in this case). This is something we can potentially do for you but it's not a priority over other things going on right now.

piclemx:
Do we know when this will be retaken?

Jeffail:
The existing elasticsearch output is now using the new APIs so you can mostly just copy/paste it: https://github.com/benthosdev/benthos/blob/main/internal/impl/elasticsearch/output.go

Jeffail:
@piclemx take a look at my follow up commit: https://github.com/benthosdev/benthos/commit/0b25d042b73ec1bd1bfe0d25db0d00eef45c76b5

piclemx:
yeah, I saw it after commenting sorry 😂

PR #1393, opened 2022-08-17
Refactor dispatchMessages to remove the need for mutex

nibbleshift:
Slight refactoring of dispatch messages to remove the need for the mutex and allocate the array for pending up front since we know the size in advance.


nibbleshift:
I will look into the failure on this one.

PR #1394, opened 2022-08-18
Minor improvement to DeserializeBytes

nibbleshift:
made a modification to pre-allocate the part slice upfront since we know the number of parts in advance and avoid the append.


Jeffail (approved):
LGTM, thanks @nibbleshift

PR #1397, opened 2022-08-18
Update Clickhouse driver

### LINKED ISSUE #1396: Update Clickhouse driver, so it will support insert of experimental semi-structured JSON feature + insert DateTime from string
For now, to implement such pipeline, I had to add temporary table with raw JSON and insert from Benthos to it + add materialized view with transformation.

mihaitodor:
Fixes #1396


Jeffail (approved):
Looks good, thanks @mihaitodor

PR #1400, opened 2022-08-19
feat: implement header propagation for the kafka input

robertmarsal:
This PR allows propagation of parent spans when using the `kafka` input. For this we are using the `traceparent` header that can be attached to a Kafka message.

Example: 
image

### DIFF
--- a/internal/impl/io/input_http_server.go
+++ b/internal/impl/io/input_http_server.go
@@ -330,12 +330,9 @@ func (h *httpServerInput) extractMessageFromRequest(r *http.Request) (message.Ba
 	})
 
 	textMapGeneric := map[string]interface{}{}
-	// Go normalises headers changing the way they are capitalised, here we are converting from normalised headers to
-	// a TextMap format which is case-sensitive. To ensure propagation still happens we need to convert the headers to
-	// lowercase as that is the format expected by the OTEL libraries.
 	for k, vals := range r.Header {
 		for _, v := range vals {
-			textMapGeneric[strings.ToLower(k)] = v
+			textMapGeneric[k] = v
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


Jeffail:
Hey @robertmarsal I'm not too savvy with tracing stuff but it's possible that this is already possible with the https://www.benthos.dev/docs/components/inputs/kafka#extract_tracing_map field, in your case you're using headers which are delivered as metadata so the value `extract_tracing_map: root = meta()` would potentially work.

robertmarsal:
Hi @Jeffail thank you for having a look, that indeed works! I should have read the manual before looking into this sorry 🙈 

There is an issue however, similar to the one I fixed here: https://github.com/benthosdev/benthos/pull/1362/files If the Kafka header is in uppercase it will not be picked up. Would you be open to moving that fix here: https://github.com/benthosdev/benthos/blob/main/internal/tracing/otel.go#L162 so that we don't have to fix it for every single input? According to the spec we should be accepting both uppercase and lowercase headers.

Should I also update the tracing docs with a mention to this `extract_tracing_map` config? I think it can be easily missed at the moment. Thanks

Jeffail:
@robertmarsal no worries, yeah a docs update would be great, also yes definitely open to a fix. A temporary work around would be `root = meta().map_each_key(key -> key.lowercase())` to get you unblocked for now.

robertmarsal:
Thank you. I've moved the lowercase fix, added a test for it and update the docs. Also tested locally and both the http and kafka inputs can make us of the traceparent header now 🎉 

Thanks for the help.

Jeffail (approved):
Awesome, looks great, thanks @robertmarsal!

PR #1404, opened 2022-08-21
feat: add lint to require labels on components

disintegrator:
This change adds a new lint option that checks that all relevant components have labels.

## Why

Labelled components will have stable metric names that can improve discoverability and prevent clobbering past time series.

Currently, an unlabelled processor has the following time series:

```
processor_batch_sent{label="",path="root.pipeline.processors.0"} 9
```

This is problematic because, if the config changes, then the processor at `root.pipeline.processors.0` can change to something else entirely and disrupt the time-series.

Labelling it will result in:

```
processor_batch_sent{label="format_msg",path="root.pipeline.processors.0"} 9
```

This is better because it can tolerate changes to the structure of the config if users query the metric by the `label` label in their dashboards.

## Example

Given:

```yaml
input:
  broker:
    inputs:
      - generate:
          mapping: |
            root = {
              "id": uuid_v4(),
              "msg": "hello world"
            }
    batching:
      count: 5

pipeline:
  processors:
    - label: format_msg
      bloblang: |
        root = this
        root.msg = this.msg.uppercase()

output:
  drop: {}
```

Output:

```
❯ ./target/bin/benthos lint --labels ./scratch/sample.yml
./scratch/sample.yml: line 4: label is required for generate
./scratch/sample.yml: line 2: label is required for broker
./scratch/sample.yml: line 21: label is required for drop
```

### DIFF
--- a/internal/docs/yaml.go
+++ b/internal/docs/yaml.go
@@ -465,18 +465,22 @@ func LintYAML(ctx LintContext, cType Type, node *yaml.Node) []Lint {
 	}
 
 	reservedFields := ReservedFieldsByType(cType)
+	_, canLabel := reservedFields["label"]
+	hasLabel := false
 	for i := 0; i < len(node.Content)-1; i += 2 {
-		if node.Content[i].Value == name || node.Content[i].Value == "type" {
+		key := node.Content[i].Value
+		if key == name || key == "type" {
 			continue
 		}
-		if node.Content[i].Value == "plugin" {
+		if key == "plugin" {
 			if nameFound || !cSpec.Plugin {
 				lints = append(lints, NewLintError(node.Content[i].Line, "plugin object is ineffective"))
 			} else {
 				lints = append(lints, cSpec.Config.LintYAML(ctx, node.Content[i+1])...)
 			}
 		}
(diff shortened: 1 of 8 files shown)


internal/docs/yaml.go
```diff
@@ -465,18 +465,22 @@ func LintYAML(ctx LintContext, cType Type, node *yaml.Node) []Lint {
...
+		if key == "plugin" {
 			if nameFound || !cSpec.Plugin {
 				lints = append(lints, NewLintError(node.Content[i].Line, "plugin object is ineffective"))
 			} else {
 				lints = append(lints, cSpec.Config.LintYAML(ctx, node.Content[i+1])...)
 			}
 		}
-		spec, exists := reservedFields[node.Content[i].Value]
+		spec, exists := reservedFields[key]
+		hasLabel = key == "label"
```
Jeffail, line 483:
```suggestion
		hasLabel = hasLabel || (key == "label")
```

Jeffail (changes requested):
Awesome, this looks great, one comment that looks like it might be a bug but other than that we're good to go.

Jeffail (approved):
LGTM, thanks @disintegrator!

PR #1410, opened 2022-08-23
Add Jaeger tracer default tags

mihaitodor:
Inspired from this blog article: http://www.inanzzz.com/index.php/post/4qes/implementing-opentelemetry-and-jaeger-tracing-in-golang-http-api

This code adds 2 default tags to the Jaeger tracer: `service.name` and `service.version` (which can still be overridden via the `jaeger.tags` configuration field.


Jeffail (approved):
LGTM, thanks @mihaitodor!

Issue #1385, opened 2022-08-15
Allow import of individual components from public package


tprestegard:
Love love love the ability to build-your-own custom Benthos, but what do you think of making all of the provided components available publicly on an individual basis? I am running a simple Benthos configuration with a `nats` input, custom Bloblang processing, and `http_client` output, so I don't need to import most of the components.  I see that I could import stuff from `public/components/io` separately, which would pull in `http_client`, but AFAIK, I can't use NATS without doing the full "import all".

It's not a huge deal, but it would definitely clean up my dependency tree, reduce the executable size, and speed up the compile time if I could separately import `nats` without needing all the other dependencies. I think it would be fine to keep the "standard library" components as a package deal since those don't pull in a ton of dependencies.

mihaitodor:
Hey @tprestegard, glad to hear you're enjoying Benthos! If you have a look [here](https://github.com/benthosdev/benthos/blob/6e100eaf5c1fedeae956c53e772cd967c031bfcc/public/components/all/package.go#L37-L43), the plan is, indeed, to offer what you're asking for. It will require a small PR to add a `nats` folder [here](https://github.com/benthosdev/benthos/tree/main/public/components) and an update to `public/components/all/package.go`. Feel free to submit a PR if you'd like.

tprestegard:
Great, I will definitely take a look at that. Thanks!

PR #1411, opened 2022-08-23
Add individual public packages for components

tprestegard:
Following the discussion on #1385, this PR adds public packages for all of the individual components. I have added each package in a separate commit to make changes easy during the review process, but I will plan to squash them once approved.


Jeffail (approved):
Ace, thanks @tprestegard! I'm going to merge and then modify a few of these in order to reduce the granularity slightly, but thanks for putting in the legwork.

PR #1413, opened 2022-08-23
feat: add file_json_contains

### LINKED ISSUE #1409: feature: implement file_json_contains similar to json_contains
It would be nice to be able to externalize the JSON content from a test file and still have `json_contains` behavior.

This makes test file(s) cleaner/simpler/shorter. It's also a nice feature when `file_json_equals` is not quite what I am looking for.

gena01:
closes #1409


Jeffail (approved):
Thanks @gena01, looks great!

PR #1418, opened 2022-08-25
Bump franz-go deps and refactor code to handle breaking changes

ekeric13:
Take 2 of: https://github.com/benthosdev/benthos/pull/1361

Main purpose for upgrading is to handle a bug I was having in an older version of franz-go that I have at least yet to reproduce on the newer version. The maintainer, Travis, was very responsive and even upgraded franz-go to handle an edge case that we were hitting with committing empty offsets:
https://github.com/twmb/franz-go/issues/186

Main breaking change was the fact that [`PollFetches` now returns a deadlineExceeded error if the context is cancelled](https://github.com/twmb/franz-go/blob/master/CHANGELOG.md#v151), which it is after 1 second.

I do want to add some test this go around but please let me know @Jeffail if this is the right direction.


Jeffail:
Thanks @ekeric13, merged as https://github.com/benthosdev/benthos/commit/3a94c7cec01a2c1d96870bd4f0d71af65e797735 (bumped a version ahead).

PR #1421, opened 2022-08-29
add endpoint to serve pprof goroutine profile

disintegrator:
Registers an endpoint at `/debug/pprof/goroutine` to serve the goroutine profile.


Jeffail (approved):
LGTM, thanks @disintegrator!

Issue #1415, opened 2022-08-24
Tracing: end spans do not emit the correct traceparent


robertmarsal:
At the moment Benthos propagates the `traceparent` from the input to the output however the output span should update the traceparent to point to itself or else chained calls do not nest properly. In my example below `validated-routing-benthos` consumes from the topic to which `ingestion-service-benthos` produces however they are both at the same level instead of being nested.

![image](https://user-images.githubusercontent.com/507871/186447840-688a7dfe-c9f3-4e36-886c-65708239f4f8.png)

Would you be open to a contribution that would make that end span generate its own `traceparent`?

robertmarsal:
For Kafka I managed to make it work by using the [inject_tracing_map](https://www.benthos.dev/docs/components/outputs/kafka#inject_tracing_map) functionality. I did encounter some issues following the docs for that as `meta()` always returns empty (which sounds like a bug on it's own maybe?) so I had to do the following (I also wanted to delete the original `Traceparent` meta):

```
    inject_tracing_map: |
      meta Traceparent = deleted()
      meta traceparent = this.traceparent
```

For HTTP there is no similar option so I created a PR that will correctly propagate the HTTP headers here: https://github.com/benthosdev/benthos/pull/1424/checks

robertmarsal:
Hi @Jeffail when you get a chance can you please have a look at the PR I sent to fix this issue? If there's any changes you'd like me to make please let me know. Thanks 🙏

PR #1424, opened 2022-08-30
feat: use end span as parent instead of propagating input parent

robertmarsal:
Solves #1415 by setting the `traceparent` on http request to the value from the last span rather than propagating the input value

### DIFF
--- a/internal/http/client.go
+++ b/internal/http/client.go
@@ -495,6 +495,7 @@ func (h *Client) checkStatus(code int) (succeeded bool, retStrat retryStrategy) 
 // to be consumed.
 func (h *Client) SendToResponse(ctx context.Context, sendMsg, refMsg message.Batch) (res *http.Response, err error) {
 	var spans []*tracing.Span
+	var currentTraceparent string
 	if sendMsg != nil {
 		spans = tracing.CreateChildSpans(h.mgr.Tracer(), "http_request", sendMsg)
 		defer func() {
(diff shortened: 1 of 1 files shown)


cristiprg (approved):

Jeffail:
Hey @robertmarsal, looks good to me but I'm wondering whether we should make this configurable, and if so maybe instead of adding an explicit field to the HTTP components we could actually add a function to bloblang that extracts the span as a text map, that'd allow us to do something like this:

```
http:
  url: example.com
  verb: POST
  headers:
    traceparent: ${! span_text_map().traceparent }
```

And it'd be reusable elsewhere. Would that work for you? If so let me know if you're happy to do it otherwise I can take a look at adding it myself.

robertmarsal:
Hi @Jeffail thank you for taking the time to look at this. I agree that a more generic solution would be better and solve more use cases. I am not that comfortable around that area of the code to introduce new bloblang functionality so I would appreciate your help if you have the time. Happy to help with testing/documentation :) Thanks again

Jeffail:
Hey @robertmarsal got it implemented here: https://github.com/benthosdev/benthos/commit/177387332c27da3353ed031332dcabb5cc61edc7, any chance you can give this a try? Just want to be sure it solves your use case before closing this.

robertmarsal:
That is awesome @Jeffail thank you so much for taking the time for this.

This works however at the moment it seems to return the main span rather than the currently active span 🤔 . If you look at my test below I was expecting the `audit_summary` span to be nested under the `http_request` rather than at the same level. The `traceparent` it receives is the one from `input_kafka` rather than `http_request`.

This is the config I am using:
```yaml
    - http_client:
        url: http://localhost:8080
        retries: 3
        headers:
          traceparent: ${! tracing_span().traceparent }
```

image

Jeffail:
I think I know why this is, it's a general issue across a few processor implementations, will take a look this evening.

Jeffail:
Think I got it: https://github.com/benthosdev/benthos/commit/455040b8060161432fa8baec0e40252db94ca04d, there's actually a bit more work needed to improve processor traces for proper hierarchies but I'll need to tackle that separately.

robertmarsal:
Works as expected now thank you! 

image

I believe this and the associated issue can be closed now. Thank you again for the help with this.

Jeffail:
Awesome, thanks for testing @robertmarsal!

PR #1432, opened 2022-09-02
Fix max_in_flight docs for various batched outputs

mihaitodor:
I believe this should cover all the batched outputs. However, I didn't adjust the default value, which, in some cases, is probably wrong. See [MongoDB](https://www.benthos.dev/docs/components/outputs/mongodb/#max_in_flight) (64). Should I go and set all those to 1? It would be a breaking change...

I'm a bit confused by [`azure_blob_storage`](https://www.benthos.dev/docs/components/outputs/azure_blob_storage/) because it's missing the batching section. I see that it's [calling](https://github.com/benthosdev/benthos/blob/ff756c1785ca2d34b987c0e1611e409d344f38f8/internal/impl/azure/output_blob_storage.go#L95) `output.OnlySinglePayloads()`, but it's missing `batcher.NewFromConfig()`, unlike `aws_s3` and `gcp_cloud_storage`.

Also, I saw that "output.OnlySinglePayloads expands message batches into individual payloads, respecting the max in flight of the wrapped output", so I'm a bit unsure if `max_in_flight` is still referring to batches in this case. If not, then I'll go back and revert the changes for the outputs which call this utility.


mihaitodor:
Uh-oh... I hit the once in a blue moon `TestTypeAPIBasicOperations` failure in `./internal/stream/manager`

Jeffail (approved):
LGTM, the `OnlySinglePayloads` bit is just to prevent needing to iterate batches when the output implementation sends single payloads. That doesn't necessarily mean the output isn't suitable for batching but it's expected that you archive the batch before it reaches the output, which is where the confusion is.

It's kind of similar in the new plugin APIs as you can define a BatchOutput that expects more than one message for each call, but a non-batched Output might still document that it's sensible to batch your messages and archive them.

Issue #718, opened 2021-03-31
beanstalkd input and out plugin


ea3hsp:
beanstalkd plugins input and output will be awesome !!

Jeffail:
Hey @ea3hsp, sounds like a good fit, I had a quick look for Go client libraries and there are one or two but they don't seem well used, happy to take suggestions if anyone has one they would recommend.

Looks like there's a docker image we can use for testing: https://hub.docker.com/r/schickling/beanstalkd

djmetzle:
I was hoping to be able to be able to use Connect, but the `beanstalkd` input does not currently support using anything but the default tube.
https://github.com/redpanda-data/connect/blob/21a9d9745ea09f7ae74941730339f153ff177a2b/internal/impl/beanstalkd/input.go#L57

The client library seems to support this. 
https://github.com/beanstalkd/go-beanstalk/blob/52e4a01c453b3de42105389beb2374f256751064/conn.go#L18-L28
and the current implementation even uses `Reserve`:
https://github.com/redpanda-data/connect/blob/21a9d9745ea09f7ae74941730339f153ff177a2b/internal/impl/beanstalkd/input.go#L102
which appears to be a function on a tubeset:
https://github.com/beanstalkd/go-beanstalk/blob/52e4a01c453b3de42105389beb2374f256751064/tubeset.go#L29

It could make the beanstalkd input much more useful if we are able to read an arbitrary tube, or tubeset.

twmb:
Closing as stale. The beanstalkd input exists but is missing tube selection — PRs welcome to add this. Feel free to open a new issue or PR.

PR #1434, opened 2022-09-06
Beanstalkd

mkvolkov:
Added beanstalkd input/output. (Issue #718 )


Jeffail:
Hey @mkvolkov, this looks great. Ideally we'd want to introduce these components using the newer public plugin APIs, there's an example here: https://github.com/benthosdev/benthos/blob/main/internal/impl/nats/input_jetstream.go, basically the way that config is defined is slightly different but other than that it's mostly the same.

mkvolkov:
Hello @Jeffail. Thanks. You mean impl/beanstalkd/input_jetstream.go and impl/beanstalkd/output_jetstream.go, am I right?

Jeffail:
> Hello @Jeffail. Thanks. You mean impl/beanstalkd/input_jetstream.go and impl/beanstalkd/output_jetstream.go, am I right?

Just using those as examples, I mean using the newer APIs for registering the plugins like https://pkg.go.dev/github.com/benthosdev/benthos/v4@v4.6.0/public/service#RegisterInput and so on.

mkvolkov:
@Jeffail I've changed the configuration. It is really better.

mihaitodor:
Nice job! I know it's a big ask, but would you consider looking at adding integration tests for it? There's https://hub.docker.com/r/websmurf/beanstalkd that's generated from https://github.com/websmurf/beanstalkd-docker/blob/main/Dockerfile, which is a trivial container. In the future, if the owner stops maintaining this container, we could run an alpine container directly and install this package on startup and then run it.

Here are some example integration tests: https://github.com/benthosdev/benthos/blob/main/internal/impl/nats/integration_jetstream_test.go or https://github.com/benthosdev/benthos/blob/main/internal/impl/nats/integration_nats_test.go. There are various `integration.StreamTests` that you can plug in there, but even the first basic ones would be great to have.

mkvolkov:
@mihaitodor Thanks. Of course, integration tests should be added.

mkvolkov:
Well, we have StreamTestOpenClose and SendBatch. They do not work being launched in the same suite (I think it is due to the one message topic inside the Beanstalkd), but separately tests pass.

mkvolkov:
@mihaitodor I've added four tests.

mihaitodor:
Cool, thank you very much! ❤️

Jeffail (approved):
Thanks @mkvolkov! LGTM but I'm going to make some minor adjustments after merging, one thing I'm going to change is a rename of the `tcp_address` fields to just `address`. If we want to expand support for other protocols later we can add an explicit field with `tcp` default.

mkvolkov:
@Jeffail thank you very much! You are right about ### address field.

Issue #1430, opened 2022-09-02
[influxdb] Mismatched Metric Label and Values


newlandk:
I'm seeing a situation where metric labels and values seem to get swapped. I'm uncertain of how to reliably reproduce this, but I'm seeing it on a simple InfluxDB output where I often see the `path` label swapped with many other label value like `mechanism` on batching and in the example below `path` is swapped with `operation`. 

This is being seen on release 4.4.1.

cache_success,**path=add**,host=hostname,label=my_redis_cache,**operation=root.cache_resources** count=0i <timestamp>
cache_success,**operation=get**,**path=root.cache_resources**,host=hostname,label=my_leveled_cache count=0i <timestamp>

Any thoughts or recommendations on where to look for where this may be occurring? Thank you!

Jeffail:
Hey @newlandk thanks for raising this, looks like we need to do a bit of digging.

newlandk:
I created a docker compose that seems to reliably reproduce this for the cache scenario. When I look a the tags on the /metrics HTTP endpoint I do not see the tags swapping, but almost immediately with the Influx output.

`docker-compose.yaml`
```yaml
version: "3.9"
services:
  benthos:
    image: jeffail/benthos:4.6.0
    volumes:
      - ./benthos.yaml:/benthos.yaml
    depends_on:
      - telegraf
  telegraf:
    image: telegraf:1.23.4
    volumes:
      - ./telegraf.conf:/etc/telegraf/telegraf.conf

```

`benthos.yaml`
```yaml
metrics:
  influxdb:
    url: http://telegraf:8086
    interval: 1s

logger:
  level: INFO
  format: json
  add_timestamp: true
  static_fields:
    '@service': benthos

pipeline:
  threads: 1
  processors:
    - branch:
        request_map: root = this.type
        processors:
          - label: cache_lookup
            cache:
              resource: foo_cache
              operator: get
              key: ${! content() }
        result_map: root.lookup = content().string()


cache_resources:
  - label: foo_cache
    memory:
      init_values:
        foo: bar
        bar: foo

input:
  resource: generator_input

input_resources:
  - label: generator_input
    generate:
      interval: 1s
      mapping: |
        root = if random_int() % 2 == 0 {
          {
            "type": "foo",
            "foo": "is yummy"
          }
        } else {
          {
            "type": "bar",
            "bar": "is gross"
          }
        }

output:
  label: output_drop
  drop: {}
```

`telegraf.conf`
```toml
[...]

PR #1435, opened 2022-09-06
Fix InfluxDB Key/Value Mismatches

newlandk:
Sort tags from copy in order to retain correct order of original tagNames to stop tagName/tagValue mismatches.

#1430


Jeffail (approved):
Awesome, thanks @newlandk! Looks like the logger based metrics exporter has the same issue so I'll port your fix over to that as well.

Issue #1408, opened 2022-08-22
add crc32 hash function


siviae:
Use case: I build a benthos -> clickhouse pipeline and I want to add deduplication id to the events (so, the key of an event will be the (timestamp, deduplicationId) tuple). It is perfectly fine for deduplication id it to be an integer, to save some memory. Thus, I want to be able to calculate integer hash from arbitrary byte array. 

I suggest to add crc32 algorithm to bloblang `hash` function.

twmb:
This has been implemented — the hash method now supports crc32 with IEEE, Castagnoli, and Koopman polynomials. Example: this.value.hash(algorithm: "crc32"). Closing as resolved.

PR #1436, opened 2022-09-08
Add support to string hash crc32

peczenyj:
This Pull Request fix issue #1408

Since there is no indication of what variation of crc32 I decide to use the key argument to receive 3 values and use IEEE by default. If anyone need a more specific table they can submit another pull request :)

please check https://pkg.go.dev/hash/crc32@go1.19.1#pkg-constants

### DIFF
--- a/internal/bloblang/query/methods_strings.go
+++ b/internal/bloblang/query/methods_strings.go
@@ -16,6 +16,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


Jeffail (approved):
Thanks @peczenyj looks great. I might change it slightly to add a new optional field `polynomial` rather than overload the `key` field, but the implementation will stay mostly the same.

peczenyj:
Thanks

Issue #1437, opened 2022-09-08
add support to sqlite driver


peczenyj:
Hello

I have to deal with some sqlite files and benthos seems the right tool for this.

I don't know what is the criteria to add a new driver, I made a small POC and it works fine (I guess)

Enjoy

srenatus:
Isn't this done? 🤔

Jeffail:
@srenatus yes 😅

PR #1438, opened 2022-09-08
Add support to driver sqlite

peczenyj:
This fixes issue #1437


mihaitodor (commented):
Hey @peczenyj, thanks for this contribution! The main distribution of Benthos is CGo-free and I could be wrong, but I believe `github.com/mattn/go-sqlite3` requires CGo.

There's an alternative pure-Go implementation of the SQLite driver [here](https://gitlab.com/cznic/sqlite) which has some benchmarks [here](https://datastation.multiprocess.io/blog/2022-05-12-sqlite-in-go-with-and-without-cgo.html). If you feel that the performance impact is acceptable, then I think this would be a good compromise (assuming it doesn't blow up the binary size by a lot).

Ideally, I think Benthos should make it easier to inject a new SQL driver via custom plugins. Maybe @Jeffail has some thoughts on this aspect.

peczenyj:
thanks @mihaitodor for the feedback, I was no aware of such rule of CGO-free

if I understand well, when you talk about plugins it is about write code to run with a custom build of benthos, right?

why not use the golang plugin system (even if it is only available on Linux) https://pkg.go.dev/plugin ?

For instance, this can be a minimal code like 

```golang
package main

import (
	_ "github.com/mattn/go-sqlite3"
)
```

then we can run benthos with `-p path/to/plugin-to-load-sqlite3-or-other-thing.so`

I think we can build a minimal interface over such plugins like

```golang
package main

import (
  "context"

)

// this function is a service.RegisterCallback (suggested name)

func Register(ctx context.Context, serviceRegister service.ServiceRegister) error {
 // TODO
}
```

then each plugin added to the -p parameter will Lookup for Register and execute it.

I think we can use an OO approach on this case, where ServiceRegister is an interface with methods like RegisterCache, etc

sounds good?

Jeffail:
Yeah if we can swap it out for modernc.org/sqlite then we're good to go, if I remember right they added ARM support so all builds will work.

peczenyj:
I change the dependency and I note some integration tests may fail on:

```
            integration_test.go:226: 
                	Error Trace:	integration_test.go:226
                	            				asm_amd64.s:1571
                	            	database is locked (5) (SQLITE_BUSY)

```

seems a limitation on the driver itself when running some parallel tests. Since I don't have experience with this code, I have no idea if it is expected or not...

mihaitodor:
> if I understand well, when you talk about plugins it is about write code to run with a custom build of benthos, right?

Yep, but right now it's not really possible to inject a new driver to the existing `sql_*` components and I think that might be handy in this case as well as for DB2 an such.

> why not use the golang plugin system (even if it is only available on Linux) https://pkg.go.dev/plugin ?

I think the initial goal was to provide a mechanism that works across all platforms that are supported by `go build`, but I don't know the history. Since users still need to run `go build` to create their plugin and then they need to figure out how to install the produced binary somewhere on their system (or in a Benthos Docker container) where Benthos can access it, is there any specific benefit to enabling this alternate plugin system over the existing one? Like, right now, one can import Benthos as a library and, if [this variable](https://github.com/benthosdev/benthos/blob/649fee1887ea9d4280af8cf3f424acf4a70034e9/internal/impl/sql/conn_fields.go#L12) is populated at runtime (maybe via [`sql.Drivers()`](https://pkg.go.dev/database/sql#Drivers), then one could write the following:

```go
package main

import (
	"context"

	// Import all standard Benthos components
	_ "github.com/benthosdev/benthos/v4/public/components/all"

	// Import extra SQL drivers
       _ "github.com/mattn/go-sqlite3"
)

func main() {
	service.RunCLI(context.Background())
}
```

peczenyj:
ok it seems to fail in some combinations of  os / architecture, I can try to fix it properly soon

mihaitodor:
I think the failure might be due to missing support for the `freebsd_arm64` architecture. Not sure if `modernc.org/sqlite` supports it. Maybe related to [this](https://gitlab.com/cznic/sqlite/-/issues/88)?

Jeffail (approved):
Awesome, thanks for all the work @peczenyj, this is going to be extremely useful!

PR #1441, opened 2022-09-12
Add ability to authenticate with the schema registry with basic auth credentials

mathewdgardner:
This addresses https://github.com/benthosdev/benthos/issues/1425.

One question I had was how we best validate for the username and password to go together in their config. If we are setting username, then we should expect a password as well as vice versa.


Jeffail:
Hey @mathewdgardner, looks good but ideally I'd like to follow the same format as the current HTTP components (https://www.benthos.dev/docs/components/processors/http#basic_auth), so we have a config section `basic_auth` with an explicit `enabled` field.

The old HTTP implementation is within https://github.com/benthosdev/benthos/tree/main/internal/http/docs/auth, eventually I want to consolidate into a single implementation that offers `public/service` config specs but for now I'm happy to stick with a separate implementation if that's easier.

mathewdgardner:
@Jeffail I changed the configs to follow that pattern though it is not exact. Please let me know what you think.

Jeffail (approved):
LGTM

Jeffail:
Awesome, thanks @mathewdgardner!

PR #1443, opened 2022-09-15
Log shutdown signal name correctly

mihaitodor:
After this change, the logs will look like this:
- `kill -2 benthos_pid`: "INFO Received SIGINT, the service is closing       @service=benthos"
- `kill -15 benthos_pid`: "INFO Received SIGTERM, the service is closing      @service=benthos"

### DIFF
--- a/internal/cli/service.go
+++ b/internal/cli/service.go
@@ -462,8 +462,17 @@ func cmdService(
 
 	// Wait for termination signal
 	select {
-	case <-sigChan:
-		logger.Infoln("Received SIGTERM, the service is closing")
+	case sig := <-sigChan:
+		var sigName string
+		switch sig {
+		case os.Interrupt:
+			sigName = "SIGINT"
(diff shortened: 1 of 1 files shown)


internal/cli/service.go
```diff
@@ -462,8 +462,17 @@ func cmdService(
...
-		logger.Infoln("Received SIGTERM, the service is closing")
+	case sig := <-sigChan:
+		var sigName string
+		switch sig {
+		case os.Interrupt:
+			sigName = "SIGINT"
+		case syscall.SIGTERM:
+			sigName = "SIGTERM"
+		default:
+			sigName = sig.String()
```
mihaitodor, line 473: I could also just use this for all cases, but it prints `interrupt` instead of `SIGINT` and `terminated` instead of `SIGTERM`, which might be confusing for some people.

Jeffail (approved):
LGTM, thanks @mihaitodor

PR #1450, opened 2022-09-19
increase backoff intervals for aws sqs input

### LINKED ISSUE #1449: Add configurable backoff to aws_sqs input
The problem: right now aws_sqs input generates one error log per second if queue does not exists or the application has insufficient permissions to receive messages from it. This generates huge amount of logs very fast, especially if there are multiple sqs queues which could be added and removed dynamically (my case). It would be nice if it had backoff configuration, so I could add exponential backoff behaviour and drastically shorten the amount of logs.

siviae:
This fixes #1449


Jeffail (approved):
LGTM, thanks @siviae, I suspect we'll need to continue tuning this over time but I'm happy with these intervals.

PR #1459, opened 2022-09-24
Ensure output gcp cloud storage will remove temp files when the collision mode is append

### LINKED ISSUE #1458: benthos output gcp cloud storage does not clean several temporary files properly
Hello

using output gcp cloud storage we not that several temporary files from different days are present on the gcp bucket when using collision mode **append**.

I find no clue in the logs. By checking the code I think the delete process may be skip on certain conditions (such as error handling).

peczenyj:
using my previous experience with gcp cloud storage, I manage to change the error handling to ensure we close the temporari files properly when the collision mode is append

this ticket may close #1458


peczenyj:
I decided to update the dependency cloud.google.com/go/storage on this PR

since the previous version was from 2021-10-18, I think we can include almost 1 year of improvements and bugfixes

PR #1460, opened 2022-09-24
use errors.Is instead err == XXX

peczenyj:
using `errors.Is` is a better approach than `err == something`

the code is cleaner 

if possible I will change some code to use errors.As too


Jeffail:
Sorry @peczenyj looks like my recent refactor broke this and your other PR, they both look good at a glance so I'll try and get them merged before I continue.

Jeffail (approved):
LGTM, thanks @peczenyj!

PR #1462, opened 2022-09-25
Add Oracle DB support

### LINKED ISSUE #821: Support oracle db in sql components
Hi, how long will it take for benthos to support oracle db?

mihaitodor:
Fixes #821.

Co-authored-by: @Solverj

TODO:
- [x] Fix tests... Oracle capitalises column names and might have a few more issues...
- [ ] Send a PR to the Oracle image repo: https://github.com/gvenzl/oci-oracle-xe#users-of-these-images

Here are some relevant blog posts for posterity:
- https://geraldonit.com/2021/08/15/oracle-xe-docker-images/
- https://fullstackcode.dev/2022/07/09/run-oracle-database-with-docker/

### DIFF
--- a/internal/impl/sql/integration_test.go
+++ b/internal/impl/sql/integration_test.go
@@ -6,8 +6,8 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"path/filepath"
 	"regexp"
-	"runtime"
 	"strings"
 	"sync"
 	"testing"
@@ -22,6 +22,7 @@ import (
 	"github.com/benthosdev/benthos/v4/public/service"
 
 	_ "github.com/benthosdev/benthos/v4/public/components/pure"
+	_ "github.com/benthosdev/benthos/v4/public/components/sql"
 )
 
 type testFn func(t *testing.T, driver, dsn, table string)
@@ -30,19 +31,19 @@ func testProcessors(name string, fn func(t *testing.T, insertProc, selectProc se
 	return func(t *testing.T, driver, dsn, table string) {
 		t.Run(name, func(t *testing.T) {
 			insertConf := fmt.Sprintf(`
-driver: %v
-dsn: %v
-table: %v
-columns: [ foo, bar, baz ]
+driver: %s
+dsn: %s
+table: %s
+columns: [ "\"foo\"", "\"bar\"", "\"baz\"" ]
 args_mapping: 'root = [ this.foo, this.bar.floor(), this.baz ]'
 `, driver, dsn, table)
 
 			queryConf := fmt.Sprintf(`
-driver: %v
-dsn: %v
-table: %v
-columns: [ foo, bar, baz ]
-where: foo = ?
+driver: %s
+dsn: %s
+table: %s
+columns: [ "\"foo\"", "\"bar\"", "\"baz\"" ]
+where: '"foo" = ?'
 args_mapping: 'root = [ this.id ]'
 `, driver, dsn, table)
 
@@ -73,11 +74,13 @@ func testRawProcessors(name string, fn func(t *testing.T, insertProc, selectProc
 			valuesStr := `(?, ?, ?)`
 			if driver == "postgres" || driver == "clickhouse" {
 				valuesStr = `($1, $2, $3)`
+			} else if driver == "oracle" {
+				valuesStr = `(:1, :2, :3)`
 			}
 			insertConf := fmt.Sprintf(`
-driver: %v
-dsn: %v
-query: insert into %v ( foo, bar, baz ) values `+valuesStr+`
+driver: %s
+dsn: %s
+query: insert into %s ( "foo", "bar", "baz" ) values `+valuesStr+`
 args_mapping: 'root = [ this.foo, this.bar.floor(), this.baz ]'
 exec_only: true
 `, driver, dsn, table)
@@ -85,11 +88,13 @@ exec_only: true
 			placeholderStr := "?"
 			if driver == "postgres" || driver == "clickhouse" {
 				placeholderStr = "$1"
+			} else if driver == "oracle" {
+				placeholderStr = ":1"
 			}
 			queryConf := fmt.Sprintf(`
-driver: %v
-dsn: %v
-query: select foo, bar, baz from %v where foo = `+placeholderStr+`
+driver: %s
+dsn: %s
+query: select "foo", "bar", "baz" from %s where "foo" = `+placeholderStr+`
 args_mapping: 'root = [ this.id ]'
 `, driver, dsn, table)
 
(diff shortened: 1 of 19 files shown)


internal/impl/sql/output_sql_insert.go
```diff
@@ -212,7 +219,7 @@ func (s *sqlInsertOutput) WriteBatch(ctx context.Context, batch service.MessageB
 		if tx, err = s.db.Begin(); err != nil {
 			return err
 		}
-		sqlStr, _, err := insertBuilder.Values().ToSql()
+		sqlStr, _, err := insertBuilder.ToSql()
```
mihaitodor, line 222: Any idea if this breaks Clickhouse? The Clickhouse tests do pass after this change, so I think it's fine, but not sure... What happens is that the current implementation produces a query like so: `INSERT INTO FOOTABLE (C1) VALUES ()` which Oracle doesn't like. After this change, the statement becomes `INSERT INTO FOOTABLE (foo) VALUES (:1)`.

internal/impl/sql/integration_test.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/benthosdev/benthos/v4/public/service"
 
 	_ "github.com/benthosdev/benthos/v4/public/components/pure"
+	_ "github.com/benthosdev/benthos/v4/public/components/sql"
```
mihaitodor, line 25: Guess these tests haven't been run in a long time

website/docs/components/outputs/sql_raw.md
```diff
@@ -152,6 +155,7 @@ The query to execute. The style of placeholder to use depends on the driver, som
 | `postgres` | Dollar sign |
 | `mssql` | Question mark |
 | `sqlite` | Question mark |
+| `oracle` | Colon |
```
Solverj, line 158: There seems to be a lot of copy paste overhead between sql files, refering to one example of: output/sql.md:106 output/sql_raw.md:158

Is it possible in markdown to somehow refer to other document's content? It's going to be, in the future, really hard to maintain the documentation.

Solverj (commented):
DRY and maintanance

internal/impl/sql/integration_test.go
peczenyj: I guess too... I had some troubles with this when I was adding the sqlite but I had no prior experience with this

website/docs/components/outputs/sql_raw.md
mihaitodor: I agree, there is some duplication here that should be eliminated. PRs are welcome :)

However, these .md files are generated automatically via `make docs`, so that should be fine. This bit in particular comes from here: https://github.com/benthosdev/benthos/blob/f0e07b3b3bdb677ba105289c2a4e91753ebf0d38/internal/impl/sql/conn_fields.go#L22-L28

internal/impl/sql/output_sql_insert.go
```diff
@@ -148,6 +149,16 @@ func newSQLInsertOutputFromConfig(conf *service.ParsedConfig, logger *service.Lo
 	s.builder = squirrel.Insert(tableStr).Columns(columns...)
 	if s.driver == "postgres" || s.driver == "clickhouse" {
 		s.builder = s.builder.PlaceholderFormat(squirrel.Dollar)
+	} else if s.driver == "oracle" {
+		s.builder = s.builder.PlaceholderFormat(squirrel.Colon)
+	}
+
+	if s.useTxStmt {
+		values := make([]any, 0, len(columns))
```
mihaitodor, line 157: Kludge: `Values` requires `...interface{}` and `columns` is `[]string` so I can't just use `columns...`

internal/impl/sql/integration_test.go
```diff
@@ -550,7 +562,7 @@ func clickhouseIntegration(t *testing.T) {
 	if err != nil {
 		t.Skipf("Could not connect to docker: %s", err)
 	}
-	pool.MaxWait = 30 * time.Second
+	pool.MaxWait = 3 * time.Minute
```
mihaitodor, line 565: I bumped this for all the SQL tests and it seems fine on my really old Macbook

internal/impl/sql/integration_test.go
```diff
@@ -396,13 +405,13 @@ processors:
...
-	"foo": "doc-%v",
-	"bar": %v,
+	"foo": "doc-%d",
+	"bar": %d,
 	"baz": "and this"
 }`, i, i))))
 		}
 		require.NoError(t, inFn(context.Background(), insertBatch))
-		require.NoError(t, streamIn.StopWithin(time.Second))
+		require.NoError(t, streamIn.StopWithin(15*time.Second))
```
mihaitodor, line 414: I hope this is reasonable. With one second, I get timeouts for MySQL and Oracle.

internal/impl/sql/integration_test.go
```diff
@@ -840,31 +847,33 @@ func msSQLIntegration(t *testing.T) {
 }
 
 func sqliteIntegration(t *testing.T) {
+	t.Skip("Broken for now...")
```
mihaitodor, line 850: The SQLite test is broken for now...

internal/impl/sql/integration_test.go
```diff
@@ -30,19 +31,19 @@ func testProcessors(name string, fn func(t *testing.T, insertProc, selectProc se
...
 		t.Run(name, func(t *testing.T) {
 			insertConf := fmt.Sprintf(`
-driver: %v
-dsn: %v
-table: %v
-columns: [ foo, bar, baz ]
+driver: %s
+dsn: %s
+table: %s
+columns: [ "\"foo\"", "\"bar\"", "\"baz\"" ]
```
mihaitodor, line 37: Unlike other drivers, Oracle returns the columns as upper case if they're not explicitly created as lower case in the `create table` statement. If I change all these columns to upper case, then it breaks the other providers...

internal/impl/sql/integration_test.go
```diff
@@ -712,21 +724,20 @@ func postgresIntegration(t *testing.T) {
...
 		t.Skipf("Could not connect to docker: %s", err)
 	}
-	pool.MaxWait = 30 * time.Second
+	pool.MaxWait = 3 * time.Minute
 
 	resource, err := pool.RunWithOptions(&dockertest.RunOptions{
 		Repository:   "mysql",
 		ExposedPorts: []string{"3306/tcp"},
+		Cmd: []string{
+			"--sql_mode=ANSI_QUOTES",
```
mihaitodor, line 739: This forces MySQL to allow double quotes around column names instead of Backticks. Backticks are not compatible with all the other DB engines.

internal/impl/sql/output_sql_insert.go
Jeffail: If the tests pass then I think we're fine, the original was likely me just trying stuff until it worked anyway.

Jeffail (approved):
This is awesome, thanks @Solverj and @mihaitodor!

PR #1466, opened 2022-09-26
add option async_batch on cassandra output

### LINKED ISSUE #1465: add option on output cassandra to write using async batches for performance reasons
Hello

currently the component output cassandra handle a batch of messages using a `unlogged batch`. However, this may not be efficient, as I can read from our cassandra logs:

> "Unlogged batch covering 52 partitions detected against table [xxx.yyy]. You should use a logged batch for atomicity, or asynchronous writes for performance."

From https://foundev.medium.com/cassandra-batch-loading-without-the-batch-keyword-40f00e35e23e

> Unlogged batches require the coordinator to do all the work of managing these inserts, and will make a single node do more work. Worse if the partition keys are owned by other nodes then the coordinator node has an extra network hop to manage as well. The data is not delivered in the most efficient path.

The suggestion is to use async writes instead unlogged batches if there are no correlation in terms of partition keys between the messages.

Here is an example from the go driver

https://pkg.go.dev/github.com/gocql/gocql#hdr-Executing_multiple_queries_concurrently

I think we can add an option like `async_batch` that does not use unlogger batches but write on individual goroutines.

peczenyj:
This PR fix #1465

using an advanced configuration `async_batch` (by default is false, to be backward compatible)

we can perform async queries by running each query in a dedicated goroutine.

the batch will wait for all goroutine finish and check the amount of errors.

if more than one, we will return the last one, with a message giving a number of errors over total.

each individual error will also trigger a log with ERROR level

I am also update the go driver to use the latest stable version available, to be possible use the latest fixes and improvements

BTW since one batch may result in several errors, I am not sure what should we do.

One option is abort the entire batch in the first error, another is count the errors and return one (first or last). I think use some kind of multierr is overkill since I will log each error. what do you think?

Also, I am not 100% sure about run each querie in a goroutine. I tryed to be explicit on the documentation that a batch of N messages will generate N goroutines.

I can use another approach like create X workers and consume the messages via channels, however I think this approach is a little bit complex.


Jeffail (commented):
Hey @peczenyj, looks good, can you add an integration test?

peczenyj:
> Hey @peczenyj, looks good, can you add an integration test?

I will try :)

peczenyj:
> Hey @peczenyj, looks good, can you add an integration test?

hey @Jeffail 

I am not sure about his syntax on the integration tests, perhaps this is **deprecated** and should be **removed**?

the cassandra docs mention the field `args`

```yaml
$ cat foo.yaml 
output:
  cassandra:
    addresses:
      - localhost:$PORT
    query: 'INSERT INTO testspace.table$ID (id, content, created_at) VALUES (?, ?, ?)'
    args:
      - ${! json("id") }
      - ${! json("content") }
      - ${! timestamp("2006-01-02T15:04:05.999Z07:00") }
$ benthos lint foo.yaml
foo.yaml: line 6: field args not recognised
```

Thanks

Jeffail:
Hey @peczenyj, added a comment on the original issue: https://github.com/benthosdev/benthos/issues/1465#issuecomment-1264700062, basically I think we're actually able to just use logged batches. Closing but feel free to reopen if I've misunderstood and we need to proceed.

PR #1470, opened 2022-09-28
add checkpoint_limit to redis_list input

### LINKED ISSUE #1469: Allow controlling of the prefetch count on redis_list input
As discussed in https://github.com/benthosdev/benthos/discussions/1451, it seems that the greedy behavior of the `redis_list` input prefetching, cannot be controlled in a meaningful way. I would argue that it could be useful to be able to control this given the following reasons (note, motivated by personal needs, not necessarily representative):

1. If you have a distributed setup (for example in k8s), where you have multiple Benthos instances essentially acting as a workers, it would make sense that this count can be adjusted, to avoid worker starvation. This is particularly true if the mean task execution time is more than just a couple of seconds (which is my case), while at times having a slow stream of new items being put on the redis list.
    
    I don't really have a use-case for optimizing fetching items from the redis list, as the actual pipeline work is of magnitudes higher in execution time. So in this case I am fine with fetching a single item a time, with the potential risk of loosing some I/O throughput speed (as I imagine this greedyness perhaps helps with).
3. For the purpose of observability, specifically, I would like to know which items are currently being worked on (not just having been fetched and awaiting to be put through the pipeline) versus queued. I guess I could achieve it by having some initial step in the processor, which should expose this information, but it seems quite cumbersome to be able to do a simple distinction between queued/processing. Because not only do I have to count the items on the redis_list I also have to count those that have been popped, but are awaiting processing inside Benthos.

Perhaps I am misusing Benthos, and there are better ways to achieve this ! Should the change to control this however be reasonably easy to implement, I do not necessarily an issue in exposing it for configuration. But I leave the judgement of that up to you :). Thanks !

sarnikowski:
# Changes

* Adds a `checkpoint_limit` to the `redis_list`. This sets a limit on the combined number of messages that are either in a prefetched or in-processing state. Has been tested using the setup described in https://github.com/benthosdev/benthos/discussions/1451. As this is quite a specific use-case, I've gone ahead with a default of `0`, which implies no limit.

Notice I have gone with the name `checkpoint_limit` as discussed in #1469. We can change this if needed (I leave this decision to you). I have also gone ahead and added a disclaimer in the description, so that people do not run into a stalling setup (hopefully it is clear from the description).

Closes #1469

### DIFF
--- a/internal/component/input/config_redis_list.go
+++ b/internal/component/input/config_redis_list.go
@@ -6,16 +6,18 @@ import (
 
 // RedisListConfig contains configuration fields for the RedisList input type.
 type RedisListConfig struct {
-	bredis.Config `json:",inline" yaml:",inline"`
-	Key           string `json:"key" yaml:"key"`
-	Timeout       string `json:"timeout" yaml:"timeout"`
+	bredis.Config   `json:",inline" yaml:",inline"`
+	Key             string `json:"key" yaml:"key"`
+	Timeout         string `json:"timeout" yaml:"timeout"`
+	CheckpointLimit int    `json:"checkpoint_limit" yaml:"checkpoint_limit"`
 }
 
 // NewRedisListConfig creates a new RedisListConfig with default values.
 func NewRedisListConfig() RedisListConfig {
 	return RedisListConfig{
-		Config:  bredis.NewConfig(),
-		Key:     "",
-		Timeout: "5s",
+		Config:          bredis.NewConfig(),
+		Key:             "",
+		Timeout:         "5s",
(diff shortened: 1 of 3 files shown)


Jeffail:
Thanks @sarnikowski, implementation looks good I'm just going to take a day or two to think over the field name. I reckon it's likely we'll reproduce this field in other inputs and I'm having second thoughts about `checkpoint_limit`.

Jeffail (approved):
LGTM, I'm going to name the field `max_in_flight` in the end because of very important reasons such as:
- eh
- hmmph
- mmm?
- yeah f it

PR #1473, opened 2022-09-28
beanstalkd version fixed

mkvolkov:
I specified the wrong version 3.46.0 instead of 4.7.0.


Jeffail (approved):
Whoops! Well spotted @mkvolkov, thanks!

Issue #1257, opened 2022-05-12
Watcher in streams mode: watch for new file or deleted file


norbjd:
Hello! :wave:

## Issue

I can't make Benthos work in Streams mode with watcher option when **adding new files or removing files**:

- adding new files does not trigger a loading of the new config
- removing an existing file won't remove the stream

I don't know if this is by design or not.

## How to reproduce?

I am using Docker to run Benthos, with the latest Benthos version at the moment (4.1.0):

```shell
docker run --name benthos --rm -it -p 4195:4195 \
    -v $PWD/configs:/configs \
    jeffail/benthos:4.1.0 --watcher streams /configs/*.yaml
```

Benthos is running, I can see the following logs:

```
Config file not specified, reading from /benthos.yaml
INFO[0000] Launching benthos in streams mode, use CTRL+C to close.  @service=benthos
INFO[0000] Listening for HTTP requests at: http://0.0.0.0:4195/  @service=benthos
```

When starting the container, my `/configs` directory **is empty**, so `curl http://localhost:4195/streams` returns `{}`, as expected.

But if I create a file in the `/configs` directory, either by copying a `yaml` or directly by `touch`ing a file and then editing it, logs does not show that Benthos have loaded my new config file. I can confirm that no streams are added because `curl http://localhost:4195/streams` still returns `{}`. I have also waited 5 minutes (in case there was some delay), but without success.

For info, my config (simple NATS -> HTTP) is valid, because if I restart my container after adding the new file in my `configs` directory, the logs show:

```
INFO[0000] Sending messages via HTTP requests to: https://domain.com/  @service=benthos label="" path=root.output stream=hello
INFO[0000] Receiving NATS messages from JetStream subject: sub1  @service=benthos label="" path=root.input stream=hello
```

I know that if I start my container with one config (like `hello.yaml`) and if I change **this particular** config, it will be reloaded, as the docs states.

## Question

So I have experienced the two following behaviors:

- when adding a new file, the new config is not loaded
- when removing an existing file, the associated stream won't disappear (`curl http://localhost:4195/streams` will still return the stream corresponding to the deleted file)

Are these behaviors normal? If so, can those features (loading new files and deleting streams if files are deleted) be added? It will be really great for dynamic configuration with files of Benthos!

I'm also thinking: maybe if **all streams were defined in a single file**, the reload will work since the file name won't change, and reloading works great. But I'm wondering if having multiple streams in one file is possible. It could be one way to mitigate my issue temporarily.

[...]

PR #1476, opened 2022-09-29
add failing test for creating stream configs in watch mode

DaniTulp:
Took a crack at #1257, but wasn't sure what the correct implementation was, I saw a todo about refreshing these paths occasionally but wasn't able to get that working cleanly. So I wrote a test that fails ❌.

Another idea might be instead of refreshing it occasionally is to use https://github.com/fsnotify/fsnotify feature to watch a folder whenever the path contains a glob pattern and update it that way.

I think it would be really sweet if additions/deletions of files within a stream path that matches would be caught, it would also match the way the streaming rest api functions.

Also please feel free to reject this, I got as far as writing the test, so thought I'd might send it ;).

### DIFF
--- a/internal/config/reader_test.go
+++ b/internal/config/reader_test.go
@@ -63,6 +63,59 @@ output:
 	assert.Equal(t, "drop", updatedConf.Output.Type)
 }
 
+func TestReaderStreamWatching(t *testing.T) {
+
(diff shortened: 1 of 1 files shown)


DaniTulp:
I'll close this for now and maybe revisit it at a later point.

Issue #1478, opened 2022-09-29
Kafka Inputs lose headers if multiple have the same key


andimiller:
This is similar to #1389 but happens in both kafka inputs.

Both inputs loop over the headers and will overwrite previous headers if they have the same key.

PR #1479, opened 2022-09-29
Add `multi_header` support for both kafka inputs

andimiller:
# Problem

Today both kafka inputs lose headers if multiple headers have the same key (#1478)

# Suggested Solution

This exposes a new `multi_header` config for both kafka inputs

By default it is set to `false` to preserve old behaviour

When set to `true`, it will decode kafka header values into lists stored in `meta`

For example:
```
Header(key="word", value="hello")
Header(key="word", value="world")
```
Would just decode to `meta("word") == "world"` today

This change encodes the values as JSON, and allows you to write bloblang like this:

```
meta("word").contains("hello") # also evaluates to true
```

# Verification

I've verified this code myself against a kafka topic with messages with multiple headers, this solution works for me on both linux and an ARM mac


andimiller:
An alternate way to do this would be to populate `foo` and `foo_multi` together, where `foo` contains the value of the last `foo` header seen, and `foo_multi` contains the list, let me know if that would be better

andimiller:
> An alternate way to do this would be to populate `foo` and `foo_multi` together, where `foo` contains the value of the last `foo` header seen, and `foo_multi` contains the list, let me know if that would be better

After sleeping on it, I'm amending this PR so that `multi_header` enables decoding into extra keys with the suffix `_multi` so that people can have both sets active at once for ease of u se

Jeffail (commented):
Thanks @andimiller, I've done a bit of work on the internals and we now support arbitrary metadata values, messages now have `MetaSetMut` and `MetaGetMut` methods which include the public APIs: https://github.com/benthosdev/benthos/commit/3729601fadefdf1884dd78bbbd796dd8e49d553b, so you don't need to do any manual JSON serialisation for either of the implementations.

I'm also thinking the behaviour ought to simply replace all header values with a []any slice when `multi_header` is enabled rather than have both single and multiple value variants.

If you're happy to make those adjustments let me know, otherwise I'll probably merge as-is and make the changes myself tomorrow.

andimiller:
> Thanks @andimiller, I've done a bit of work on the internals and we now support arbitrary metadata values, messages now have `MetaSetMut` and `MetaGetMut` methods which include the public APIs: [3729601](https://github.com/benthosdev/benthos/commit/3729601fadefdf1884dd78bbbd796dd8e49d553b), so you don't need to do any manual JSON serialisation for either of the implementations.

That sounds great, I can pull and update this PR with that!

> I'm also thinking the behaviour ought to simply replace all header values with a []any slice when `multi_header` is enabled rather than have both single and multiple value variants.

I was originally thinking that, then I tried to use my fork for real and ended up having to do a lot of `.0` for the headers where I expected only one value, I would be happy with either though

andimiller:
I've now updated this PR
* I've pulled from main
* Updated to use `MetaSetMut`
* Swapped to just decoding everything as lists when `multi_header` is enabled, because the other behaviour was probably overcomplicating it

Jeffail (approved):
Awesome, thanks @andimiller!

PR #1481, opened 2022-09-30
Change `INSERT` operation back to previous v3 behaviour on azure table storage

mfamador:
Changing the `INSERT` operation back to the previous behaviour and making it again compliant to the docs.

https://www.benthos.dev/docs/components/outputs/azure_table_storage#insert_type

The INSERT should fail if the partition key and row key already exist, INSERT_MERGE should merge the values, and INSERT_REPLACE completely replace the entire row.

I've also removed the other operations (UPDATE and DELETE) since the field is called `insert_type` and there's no doc about them. I can add them back if you think they might be helpful for any use case.

CC: @davidandradeduarte

### DIFF
--- a/internal/impl/azure/output_table_storage.go
+++ b/internal/impl/azure/output_table_storage.go
@@ -293,18 +293,12 @@ func (a *azureTableStorageWriter) addToBatch(batch []aztables.TransactionAction,
 	}
 	var err error
 	switch strings.ToUpper(insertType) {
-	case "ADD":
+	case "INSERT":
 		batch, err = appendFunc(batch, aztables.TransactionTypeAdd, entity)
-	case "INSERT", "INSERT_MERGE", "INSERTMERGE":
+	case "INSERT_MERGE":
 		batch, err = appendFunc(batch, aztables.TransactionTypeInsertMerge, entity)
-	case "INSERT_REPLACE", "INSERTREPLACE":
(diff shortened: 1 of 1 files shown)


Jeffail:
@mfamador we definitely can't remove stuff if it's possibly being used, I would keep the extra options even though they don't fit the field name. In future we could deprecate the `insert_type` field and add something more generic like `transaction_type`.

mfamador:
Sure, then I'll also add UPDATE_REPLACE and DELETE to the docs as well.

What about the operations with no underscore (not in the docs, I believe no one is using them (?))

mfamador:
Actually, if the field is still
`docs.FieldString("insert_type", "Type of insert operation").HasOptions(
				"INSERT", "INSERT_MERGE", "INSERT_REPLACE",
			).IsInterpolated().Advanced(),`
it means that no one if suing any of the operations if the switch case, right? Because it will fail in this validation.

mfamador:
I've just tried with `insert_type: 'ADD'` and it fails with
``` 
ERRO Config lint error . . value ADD is not a valid option for this field"
```
@Jeffail, in that case, I think it's pretty safe to keep the PR as it is because no one is really using for sure the `ADD`, `UPDATE` or `DELETE` operations.

Jeffail (approved):
Ah fair enough, thanks @mfamador!

PR #1487, opened 2022-10-01
Cassandra input

mkvolkov:
Added Cassandra/Scylla input.


Jeffail (approved):
Looks great, thanks @mkvolkov! Having both an input and output means we could potentially set up round-trip integration tests, I might look into that if I have time this evening.

PR #1489, opened 2022-10-01
Small cleanup in the snowflake_put and kafka_franz outputs

mihaitodor:
I opened this one to start a discussion on what's the minimal synchronisation plumbing that's needed for output components, since I'm struggling to figure out which edge cases need to be accounted for.


internal/impl/kafka/output_kafka_franz.go
```diff
@@ -111,14 +110,12 @@ type franzKafkaWriter struct {
 
 	client *kgo.Client
 
-	log     *service.Logger
-	shutSig *shutdown.Signaller
```
mihaitodor, line 115: It's not used in this component, but should it? I see that inputs do leverage it in the disconnect / shutdown logic

internal/impl/snowflake/output_snowflake_put.go
```diff
@@ -670,6 +672,12 @@ func (s *snowflakeWriter) callSnowpipe(ctx context.Context, snowpipe, requestID,
 }
 
 func (s *snowflakeWriter) WriteBatch(ctx context.Context, batch service.MessageBatch) error {
+	s.connMut.Lock()
+	defer s.connMut.Unlock()
```
mihaitodor, line 676: I'm not sure how to tell if / when this mutex is necessary. For example, the `kafka_franz` output doesn't have it, but the `nats_jetstream` one [does](https://github.com/benthosdev/benthos/blob/63559aa93c1e00632ba5e2448c3b9b97f552ad64/internal/impl/nats/output_jetstream.go#L74).

internal/impl/kafka/output_kafka_franz.go
Jeffail: We definitely don't need a shutdown signaller, that's probably a copy/paste artifact.

internal/impl/snowflake/output_snowflake_put.go
Jeffail: It's probably not needed, a mutex around connections is just an extra blanket of safety as it's the role of the channel based output wrapper to only call Close on a single thread. It would however be a problem if under certain circumstances we were to disconnect the client from the write call, as write can be called in parallel (as can connect).

Overall though it's never a problem to add a mutex just in case.

Jeffail (approved):

Jeffail:
LGTM, I think this is fine to merge.

mihaitodor:
OK, cool! That does clear up a few things, thanks @Jeffail! I took the liberty to clean up some more of these.

Jeffail:
Thanks @mihaitodor, precious deletes.

mihaitodor:
work-out-guillotine

PR #1491, opened 2022-10-03
add jetstream metadata

cortopy:
This solves https://github.com/benthosdev/benthos/issues/1486

NATS Jetstream is different from normal NATS in that metadata like timestamp and sequence is not send as part of headers, but in the reply subject. The NATS client library has convenient methods to extract this information, which after this PR is added to the message and made available to processors

This is tested locally and it works. I didn't know how to add unit tests and documentation, but happy to do so if you think is relevant

### DIFF
--- a/internal/impl/nats/input_jetstream.go
+++ b/internal/impl/nats/input_jetstream.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


Jeffail (approved):
LGTM, thanks @cortopy, I'm going to make one small tweak which is to rename `nats_timestamp` to `nats_timestamp_unix_nano`. Soon we'll be able to have proper timestamps as metadata at which point we can add `nats_timestamp` with an explicit `time.Time`.

PR #1477, opened 2022-09-29
feat: add `sql_outbox` input

disintegrator:
Adds a `sql_outbox` input that consumes messages from a database table implementing the transaction outbox pattern. 

- Outbox items that are processed successfully are deleted from the table.
- Outbox items that fail to be processed will have a retry counter incremented on their corresponding database row.

This current implementations makes hard assumptions about the implementation of this pattern. Namely:

- The outbox table must have an "id" column: We need to target outbox items for deletions and updates based on their processing outcomes
- The outbox table must have a "retry count" column: For failed items we are opting to increment a counter in the database table for bookkeeping. This counter can be used to skip over outbox items that will terminally fail.
- The processing strategy is consume-delete-update.

---

### Local testing
<details>
<summary>PostgreSQL</summary>

You can use the docker postgres image to setup a database with sample data very quickly for the purpose of testing this input.

Create a file called `docker-entrypoint-initdb.d/create-outbox.sql`:

```sql
ALTER DATABASE postgres
  SET log_statement = 'all';

CREATE TABLE outbox_items (
  id BIGINT GENERATED BY DEFAULT AS IDENTITY,
  created_at TIMESTAMP DEFAULT NOW(),
  deliverable_at TIMESTAMP DEFAULT NOW(),
  payload JSON DEFAULT '{"message": "Hello, World!"}'::json,
  attempt_count INT DEFAULT 0,
  PRIMARY KEY (id)
);

INSERT INTO outbox_items SELECT id FROM generate_series(1, 1000) AS g (id);
```

Then run:

```
docker run \
  --name benthospg \
  -p 5432:5432 \
  -e POSTGRES_HOST_AUTH_METHOD=trust \
  -v "${PWD}/docker-entrypoint-initdb.d:/docker-entrypoint-initdb.d" \
  --rm \
  postgres:14
```

The database server should be up and running and bootstrapped with an outbox table containing a mock entries.

In a different terminal, run a Benthos pipeline with this config:

```yaml
rate_limit_resources:
  - label: outbox_poll_rate
    local:
      count: 1
      interval: 10s

input:
  sql_outbox:
    driver: postgres
    dsn: postgres://postgres@localhost:5432/postgres?sslmode=disable
    table: outbox_items
    columns:
      - payload
      - created_at
      - id
    where: deliverable_at <= now()
    id_column: id
    retry_count_column: attempt_count
    max_retries: 3
    item_count: 10
    rate_limit: outbox_poll_rate

output:
  stdout: {}
```

[...]

PR #1493, opened 2022-10-03
Input sql raw

akai-katto:
Any possibility on getting this added? Tested with 

```
input:
  label: ""
  sql_raw:
    driver: "postgres"
    dsn: postgres://username:password@localhost:5432/default_database?sslmode=disable
    query: "select * from dataclass_hits where dc_name = $1;"
    args_mapping: '[ "ssn" ]'
```
 
```
tyler@Tylers-MacBook-Pro bin % ./benthos -c config.yaml
INFO Running main config from specified file       @service=benthos path=config.yaml
{"count":10,"dc_name":"ssn","timestamp":"2022-08-31T00:00:00Z"}
{"count":10,"dc_name":"ssn","timestamp":"2022-07-31T00:00:00Z"}
{"count":15,"dc_name":"ssn","timestamp":"2022-07-28T00:00:00Z"}
```

If revisions are needed please point me in the right direction. Would love to have this added for our project.


akai-katto:
The reason for this PR is we (we as in my team) have to preform aggregates over large databases at our company, which require some more complex SQL statements at times, and `input_sql_select.go` is a bit limiting at times with the (what I define to be) more rigid structure at times of what's querable. We were looking to have the same flexibility that processor_sql_raw offers, but for the input part of benthos. 

For example, we we needed to preform this aggregate, which isn't possible (from what I poked around in the benthos discord / read the source code for) from input_sql_select: 

```
"
SELECT sum(counted_uniques) 
FROM (
  SELECT smalltable, count(*) as counted_uniques
    FROM (SELECT jsonb_array_elements(some_json) AS smalltable FROM jsonbtesting
  ) alias1 GROUP BY smalltable
) alias2;"
```

We need benthos to be able to aggregate streams of this, which outputs a table containing rows like:

```
"string1"  :  5
"string2" :  3
```


---

We were able to hack around this by simply creating a one line csv file, and using processor_sql_raw to hydrate the data, but it was super hacky:

```
input:
  label: ""
  csv:
    paths: ["onelinecsv.csv"]  # This is a csv file with one line
    parse_header_row: true
    delimiter: ','
    lazy_quotes: false

pipeline:
  processors:
    - branch:
        processors:
          - sql_raw:
              driver: postgres
              dsn: postgres://username:password@localhost:5432/default_database?sslmode=disable
              query: "SELECT sum(counted_uniques) 
                      FROM (
                        SELECT smalltable, count(*) as counted_uniques
                          FROM (SELECT jsonb_array_elements(some_json) AS smalltable FROM jsonbtesting
                        ) alias1 GROUP BY smalltable
                      ) alias2;"

        result_map: 'root.aggregate = this'
```

My PR solves the problem we needed by having a lose and dangerous input_sql_raw input that processor_sql_raw  provides, but for the input stage.

disintegrator:
Hey @akai-katto, the PR looks fairly straightforward and easy to read. Also thanks for including the motivation as a follow-up! I'm curious why you think this is worth including as a plugin in Benthos versus keeping it as part of your custom Benthos binary that you may have today. It's not a loaded question intended to gatekeep or anything like that, just passing curiosity.
I recently tried to contribute `sql_outbox` in #1477 but had to walk it back because I didn't really put it to the test before contributing it. The API needs a slight edit to support different strategies for example. I'm taking my time with it now.

mihaitodor:
I'll let others chime in on this addition, but I thought I'd point out that the `generate` input with `count: 1` and `interval: 0s` and `mapping: root = {}` can be used instead of the `csv` one as a trigger for the `sql_raw` processor.

akai-katto:
> Hey @akai-katto, the PR looks fairly straightforward and easy to read. Also thanks for including the motivation as a follow-up! I'm curious why you think this is worth including as a plugin in Benthos versus keeping it as part of your custom Benthos binary that you may have today. It's not a loaded question intended to gatekeep or anything like that, just passing curiosity. I recently tried to contribute `sql_outbox` in #1477 but had to walk it back because I didn't really put it to the test before contributing it. The API needs a slight edit to support different strategies for example. I'm taking my time with it now.

Answering some of the questions laid out, 

As for adding this to the main repo:

I strongly believe that the use case here of running benthos over a set of arbitrary queries for SQL is not a one off request use case. Most of my answers include "well theres a way to do it in benthos already", but having access to raw queries would just simplify some pipelines.

Some examples of why I think it should be included (including the one I listed above) would be:

- Performing joins on tables where a given message should contain rows from multiple tables. From my knowledge of the documentation, this is not possible with `sql_select`. I don't think there's a way around this using benthos currently, at least in the input stage. 

- Performing aggregates / counts over databases. This one is useful for using all the tools that benthos has already implemented (hydrating data [i.e, hash and salting, http endpoint, etc],  and then posting results to a new table or uploading a new file). Our use case for this is redacting and hashing sensitive information from aggregates, adding necessary hydrated fields where necessary (converting mappings between uuids and plaintext strings they represent).

- Casting jsonb columns to check whether or not a given row should be included or not should be processed as a message. While this can be done in the processing stage of benthos, this circles back to just reducing the amount of work on the processing stage overall by having the SQL application do some of the heavy lifting. Companies tend to rely on jsonb columns in postgres (from what I'm aware) quite often due to their flexibility. 


As for running our own binary, there are a few tangential issues for us, 

1) A custom plugin couldn't get made due to the heavy reliance on the sql code that benthos has written implemented, so the plugin architecture wouldn't allow me to touch internal code without essentially copy and pasting every missing code bit and renaming functions to avoid function name overlap. Unless I'm missing something here this would be a preventative piece for me creating my own binary using benthos using the plugin approach given what the use case provided. 

2) Our company relies on benthos very heavily and migrating upstream changes is not a long term plan for us, there's not much buy in from a team standpoint to maintain a custom benthos binary and pulling upstream changes each release.

akai-katto:
> Hey @akai-katto, the PR looks fairly straightforward and easy to read. Also thanks for including the motivation as a follow-up! I'm curious why you think this is worth including as a plugin in Benthos versus keeping it as part of your custom Benthos binary that you may have today. It's not a loaded question intended to gatekeep or anything like that, just passing curiosity. I recently tried to contribute `sql_outbox` in #1477 but had to walk it back because I didn't really put it to the test before contributing it. The API needs a slight edit to support different strategies for example. I'm taking my time with it now.

by the way, how do you make the auto-gen docs file?

disintegrator:
Running `make doc` will add/update markdown files which you need to add to your PR. If you continue to run into the same error then rebase your branch because there may have been docs edits in main that are not in your branch - the lint compares your branch to main.

akai-katto:
> Running `make doc` will add/update markdown files which you need to add to your PR. If you continue to run into the same error then rebase your branch because there may have been docs edits in main that are not in your branch - the lint compares your branch to main.

btw `make doc` generated the file, despite it still raising an error. We'll see how the pipeline feels about that,

internal/impl/sql/input_sql_raw.go
```diff
@@ -0,0 +1,213 @@
...
+input:
+  sql_raw:
+    driver: postgres
+    dsn: postgres://foouser:foopass@localhost:5432/testdb?sslmode=disable
+    query: "SELECT name, count(*) FROM person WHERE last_updated < $1 GROUP BY name;"
+    args_mapping: |
+      root = [
+        now().ts_unix() - 3600
+      ]
+]
```
Jeffail, line 48: Looks like this extra square bracket is causing the docs to fail linting (it's a check that verifies documentation examples are valid YAML).

internal/impl/sql/input_sql_raw.go
```diff
@@ -0,0 +1,213 @@
...
+			Description("A [Bloblang mapping](/docs/guides/bloblang/about) which should evaluate to an array of values matching in size to the number of columns specified.").
+			Example("root = [ this.cat.meow, this.doc.woofs[0] ]").
+			Example(`root = [ meta("user.id") ]`).
+			Optional())
+	for _, f := range connFields() {
+		spec = spec.Field(f)
+	}
+
+	spec = spec.
+		Version("3.59.0").
```
Jeffail, line 34: Needs to be set for the next release (4.10.0)

internal/impl/sql/input_sql_raw.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	rows, err := s.db.QueryContext(ctx, s.queryStatic, args...)
+	if err != nil {
+		s.logger.Warnf("Failed to run query: %v", err)
+	} else {
+		// this is only re-assigned so linting will be happy : \. Will throw error if s.rows
+		// is immediately assigned straight out of QueryContext
+		s.rows = rows
+	}
```
Jeffail, line 164: I'm suggesting we match the behaviour of the `sql_select` input:

```go
	if s.rows, err = s.db.QueryContext(ctx, s.queryStatic, args...); err != nil {
		return fmt.Errorf("failed to run query: %w", err)
	}
```

Where a query error is the equivalent of a connection error (and gets retried loudly). Otherwise this issue might be easy to slip under the radar.

Jeffail (commented):
Hey @akai-katto, looks good, I think it's a reasonable addition given the improved memory efficient versus using a processor. Left a couple of comments but they're only minor,  I'd like to get this merged this week so I'm happy to make those changes after merging if you're busy, just let me know.

akai-katto:
resolved comments and updated the docs!

Jeffail (approved):
LGTM, going to make a few minor adjustments, don't worry about the lint that looks like a misfire.

PR #1494, opened 2022-10-04
Deprecate `insert_type` field and add new `transaction_type` to azure_table_storage output

mfamador:
Deprecated `insert_type` field to add a new field `transaction_type` supporting 3 new non insert operations: `UPDATE_MERGE`, `UPDATE_RE PLACE` and ` DELETE`.

Despite `insert_type` is supposed to be interpolated it only supported 3 static values. This was fixed on this field and on the new one.
It's now possible to do something like this:
 
```
       insert_type: '${! json("operation") }'
```
or
```
        transaction_type: '${! json("operation") }'
```


Jeffail (approved):
LGTM, thanks @mfamador

PR #1496, opened 2022-10-06
fix cassandra docs

peczenyj:
removing mention to `args` field

mention that uses logged batches

use latest stable gocql driver


Jeffail (approved):
LGTM, thanks @peczenyj!

PR #1498, opened 2022-10-06
feat: make batch type configurable on cassandra output

eduardodbr:
Hi!

The version 4.9.0 changed the default behavior of the cassandra output to insert logged batches of data instead of unlogged batches. However, [AWS keyspaces](https://docs.aws.amazon.com/keyspaces/latest/devguide/functional-differences.html#functional-differences.batch) and [Azure CosmoDB](https://learn.microsoft.com/en-us/azure/cosmos-db/cassandra/cassandra-support#cql-commands) don't support logged batches. 

This PR makes the batch type configurable by adding a new boolean field `logged_batch` that is true by default.


Jeffail (approved):
LGTM, thanks @eduardodbr!

PR #1502, opened 2022-10-07
Azure table storage input

mfamador:
I've created this input to query an Azure Account Table Storage.

We can set a filter that if not set returns all rows.
And a `select` field on which we can limit the columns to be retrieved. If not set all cols are returned

Examples:

```yml 
input:
  label: input_table_storage
  azure_table_storage:
    storage_connection_string: <REDACTED>
    table_name: Foo
    filter: "PartitionKey eq 'foo' and RowKey gt '1'"
    select: "PartitionKey,RowKey,Bar,Timestamp"
```

```yml 
input:
  label: input_table_storage
  azure_table_storage:
    storage_connection_string: <REDACTED>
    table_name: Foo
```


Jeffail (approved):
LGTM, thanks @mfamador!

PR #1503, opened 2022-10-08
Add `trace_id` function

mfamador:
We need to log the messages' `trace_id` so that we can see the traces directly on Grafana Tempo from Grafana Loki logs.

I added a `trace_id` func and we can use it like this:

```yml
  processors:
  - log:
      level: INFO
      message: "message"
      fields:
        payload: ${! content() }
        metadata: ${! meta() }
        traceID: ${! trace_id() }
```

Another option would be changing the already existent `tracing_span` function and add to the resulting map the `trace_id` key, what do you prefer?


Jeffail (commented):
Hey @mfamador, can we rename this `tracing_id` so it's consistent with `tracing_span`? Also it's minor but it'd be good to abstract the actual ID extraction to https://github.com/benthosdev/benthos/blob/main/internal/tracing/otel.go, the main reasoning is so that it's less of a headache to track all the places otel gets imported and used.

mfamador:
Sorry, will create another PR with the suggestions. Messed up again rebasing upstream. Need to go to git school again :(

PR #1507, opened 2022-10-11
Add Snowflake SQL driver to SQL components

mihaitodor:
When inserting large amounts of data, the `snowflake_put` output should outperform `sql_insert` because it uses Snowpipe directly. However, the Snowflake SQL driver also attempts to send the data to a temporary stage when the number of messages in a batch exceeds an [undocumented](https://pkg.go.dev/github.com/snowflakedb/gosnowflake#hdr-Batch_Inserts_and_Binding_Parameters) threshold, but it's not very consistent. Details [here](https://github.com/snowflakedb/gosnowflake/issues/540).

This implementation should satisfy the basic use cases, but it might be possible to optimise inserts later by binding array variables to parameters. Details [here](https://pkg.go.dev/github.com/snowflakedb/gosnowflake#hdr-Binding_Parameters_to_Array_Variables).

Unfortunately, I was unable to add integration tests based on MySQL as described [here](https://github.com/snowflakedb/gosnowflake/issues/279) because [this code](https://github.com/snowflakedb/gosnowflake/blob/74e351e5e110c5b4c409730b47d5fa4058ab1c6f/auth.go#L45-L91) does not allow me to set the `authenticator` parameter to `tokenaccessor`, so I can't get the driver to bypass the authentication step. There might be other blockers too, not sure.

### DIFF
--- a/internal/impl/sql/conn_fields.go
+++ b/internal/impl/sql/conn_fields.go
@@ -9,7 +9,7 @@ import (
 	"github.com/benthosdev/benthos/v4/public/service"
 )
 
-var driverField = service.NewStringEnumField("driver", "mysql", "postgres", "clickhouse", "mssql", "sqlite", "oracle").
+var driverField = service.NewStringEnumField("driver", "mysql", "postgres", "clickhouse", "mssql", "sqlite", "oracle", "snowflake").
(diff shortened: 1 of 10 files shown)


Jeffail (approved):
LGTM, epic docs :heart:

PR #1508, opened 2022-10-11
Cleanup SQLite include logic and integration tests

mihaitodor:
- Use in-memory SQLite database to avoid file locking issues during integration tests as suggested here: https://github.com/mattn/go-sqlite3/issues/677#issuecomment-450203752
- The modernc.org/sqlite FreeBSD build was fixed: https://gitlab.com/cznic/sqlite/-/commit/fd7b2cc5b3b1be8d962638485629a39431cac462#bb99b1baec3b0c8f02dc4e87b04926bc377fd8db
- github.com/ClickHouse/clickhouse-go now supports ARM: https://github.com/ClickHouse/clickhouse-go/issues/519
- Run `go fix ./...` to switch to `//go:build` across the codebase
since we're on Go 1.18+: https://stackoverflow.com/questions/68360688/whats-the-difference-between-gobuild-and-build-directives


Jeffail (approved):
Awesome, thanks @mihaitodor, I'll fix the lint after merging.

PR #1510, opened 2022-10-11
Add missing field descriptions

mihaitodor:
I noticed one of these when I raised #1387, but I must've rushed when I raised that one, since the change I made there was utter nonsense.


Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #1513, opened 2022-10-12
Add govulncheck linter

mihaitodor:
This is temporary while https://github.com/golangci/golangci-lint/issues/3094 is being worked on here: https://github.com/golangci/golangci-lint/pull/3199

Also, `actions/setup-go` added a `check-latest` flag that can be set to `true` to ensure that the latest matching release is pulled in if it's newer than the cached one. I enabled this because otherwise this linter will complain about vulnerabilities in the standard lib whenever there's a new patch version released until the cached version expires.

There's https://github.com/Templum/govulncheck-action which can also upload a vulnerability report to GitHub, but I don't think it's worth the hassle of maintaining that.


Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #1520, opened 2022-10-15
Add base64_encode and base64_decode functions to awk processor

### LINKED ISSUE #1512: awk processor decode and assign
I want to decode the string into base64 in awk processor and assign it to a variable，just like this:
```
#! /bin/bash

astr=MjAyMi0xMC0xMiAxMDoyMTowNg==
bstr=`echo -n $astr|base64 -d`
echo "$bstr"

```
Decode astr with base64 and assign it to bstr
How to implement the above function in awk processor？

mihaitodor:
Fixes #1512.

### DIFF
--- a/internal/impl/awk/processor.go
+++ b/internal/impl/awk/processor.go
@@ -3,6 +3,7 @@ package awk
 import (
 	"bytes"
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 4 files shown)


internal/impl/awk/processor.go
```diff
@@ -590,6 +604,14 @@ var awkFunctionsMap = map[string]any{
 	"print_log": func(value, level string) {
 		// Do nothing, this is a placeholder for compilation.
 	},
+	"base64_encode": func(data string) string {
+		// Do nothing, this is a placeholder for compilation.
```
Jeffail, line 608: We only need placeholders for functions that interact directly with messages, since this is processing the argument directly we can just add the implementations here and ignore the customFuncs lower down.

mihaitodor: Spot on! I was trying to figure out how these needs to work and then forgot to check if I still need the boilerplate. Thanks, fixed! 😽

Jeffail (approved):

Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #1522, opened 2022-10-18
Bump Go version in Readme

mihaitodor:
This is going to be annoying to keep track of...


Jeffail:
I say we replace it with something more vague like `Build with Go (any [currently supported version](https://go.dev/dl/))`, since we always intend to support the lowest supported Go version anyway.

mihaitodor:
That makes a lot of sense, thanks!

PR #1526, opened 2022-10-19
Update docs to reference the mapping processor

mihaitodor:
The bloblang processor will be [deprecated](https://www.benthos.dev/docs/components/processors/bloblang#component-rename) in the future.

Not sure if some of these should be `mutation` instead of `mapping`. Guess not. Worst case, people who copy / paste won't get the best perf they can get 😈

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ input:
 
 pipeline:
   processors:
-    - bloblang: |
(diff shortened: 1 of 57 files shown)


Jeffail (approved):
🙌 doing the blobs work

PR #1532, opened 2022-10-20
Fix outter typo

mihaitodor:
Guess this should cover the rename to `outer` including the deprecation notice. Also feel free to close it if you think it's over the top and unnecessary :) Just thought it would be good to match the SQL `OUTER JOIN`.


Jeffail (approved):
I know you're right, I googled it and everything, it just doesn't _look_ right. But I guess I'm willing to concede to <every dictionary everywhere and the whole internet>.

Issue #1511, opened 2022-10-11
Who uses arm6+7?


Jeffail:
We currently support pretty much just every hecking platform we can, because why not? However, there is a steadily increasing gap of enthusiasm between support for ARM64 and 32-bit ARM architectures (6/7) in the library ecosystem. For us this mostly just means disabling some components for the 32-bit build artifacts with build flags, which is easy enough and only means a minor amount of effort even in some of the worst cases.

The question though is whether that effort is worth it at all, from download analytics is appears as though only bots download the ARM 6/7 artifacts. If we were to base a decision on support going forwards that's really the only information we have currently, so please let us know with a thumbs up or a comment whether you use these build artifacts, would be interested in learning what for if you're willing to share.

Professionals, hobbyists, historical reenactors, your thoughts are all equally valid here so please get them in.

robinhuiser:
I do - I work in the financial industry and I have build a "bank in a box" concept (mobile/web-app, digital microservices and core banking) on a RPI 7 cluster (6 nodes). Shortly we are looking into Benthos as a streaming platform for ingesting 3rd party core data. I am also the maintainer of Concourse for RPI (see: https://github.com/robinhuiser/concourse-arm64)

Jeffail:
Oh that's awesome, thanks for sharing @robinhuiser!

Jeffail:
Looks like we're keeping it going for now 🎊

PR #1534, opened 2022-10-21
Update kafka-franz and re-enable old ARM builds

mihaitodor:
Some bug fixes [went in](https://github.com/twmb/franz-go/blob/master/CHANGELOG.md#v190) following the PR I raised here: https://github.com/twmb/franz-go/pull/223.

Since Travis was kind enough to also fix the old ARM builds, I thought it should be fine to re-enable them by reverting 414cf1f.

Related to #1511.


Jeffail (approved):
Awesome, thanks @mihaitodor

PR #1535, opened 2022-10-21
fix: inject_tracing_map only working for the first message in a batch

### LINKED ISSUE #1492: inject_tracing_map only works for the first message in a batch
Hi! Whenever we try to write a batch with multiple messages, only the first one is written with the traceparent ID. Here is a simple pipeline to test it:

```
input:
  generate: 
    # uncomment next line to work
    # mapping: '{"foo":"bar","samples":[{"baz":1}]}'
    mapping: '{"foo":"bar","samples":[{"baz":1},{"baz":2},{"baz":3}]}'
    count: 1

pipeline:
  processors:
    - bloblang: | 
        root = this.explode("samples")
    - unarchive:
         format: json_array

output:
  kafka:
    addresses:
    - localhost:9092
    topic: "test"
    inject_tracing_map: "meta = this"
```

with a single message everything works, if we have multiple messages in a batch it throws a warning for every extra message :

`WARN Failed to inject span: failed assignment (line 1): unable to reference message as structured (with 'this.traceparent'): message is empty`

The AsyncWriter injectSpans() func does the folowing (with some comments):

```
for i := 0; i < msg.Len(); i++ {
  spanMapGeneric, err := spans[i].TextMap()
  if err != nil {
	  w.log.Warnf("Failed to inject span: %v", err)
	  continue
  }
  
  spanPart := message.NewPart(nil)
  spanPart.SetStructuredMut(spanMapGeneric)
  
  spanMsg := message.Batch{spanPart} // this is always a batch with a single part
  // MapOnto() receives an index ("i" in this case) to retrieve that index from the reference ("spanMsg" in this case)
  // however, spanMsg is always a batch with a single part so getting an index > 0 is always nil
  if tmpMsg, err := w.injectTracingMap.MapOnto(msg.Get(i), i, spanMsg); err != nil { 
	  w.log.Warnf("Failed to inject span: %v", err)
  } else {
	  msg[i] = tmpMsg
  }
}
```
w.injectTracingMap.MapOnto() is always called with a message.Batch with a single Part. However, the function MapOnto() keeps trying to retrieve the current index from the batch, although the batch is always a single message, so it fails if the batch has more than one message.

I'd like to know if this is really a bug or if I am doing something wrong.

eduardodbr:
This PR fixes #1492

### DIFF
--- a/internal/component/output/async_writer.go
+++ b/internal/component/output/async_writer.go
@@ -106,7 +106,7 @@ func (w *AsyncWriter) injectSpans(msg message.Batch, spans []*tracing.Span) {
 		spanPart.SetStructuredMut(spanMapGeneric)
 
 		spanMsg := message.Batch{spanPart}
-		if tmpMsg, err := w.injectTracingMap.MapOnto(msg.Get(i), i, spanMsg); err != nil {
+		if tmpMsg, err := w.injectTracingMap.MapOnto(msg.Get(i), 0, spanMsg); err != nil {
 			w.log.Warnf("Failed to inject span: %v", err)
 		} else {
 			msg[i] = tmpMsg


Jeffail (approved):
LGTM, thanks @eduardodbr!

PR #1537, opened 2022-10-24
Fix typo in filtering cookbook

tragiclifestories:
Extremely minor thing, but I'm such a pedant I can't help myself!


Jeffail (approved):
LGTM, thanks @tragiclifestories, pedants are very welcome here just try not to have fun while doing it.

PR #1538, opened 2022-10-24
Set backoff.MaxElapsedTime to 0 in aws_sqs input

siviae:
This is a follow-up for https://github.com/benthosdev/benthos/pull/1450

Unfortunately, I didn't noticed that ExponentialBackOff implementation has MaxElapsedTime set to 15 minutes by default.
After that, NextBackOff() starts to return -1, effectively turning polling loop into hot one and we are back to the initial problem :(.

Thus, I propose to set MaxElapsedTime to 0, because I think this is the expected behaviour here.


Jeffail (approved):
Ahhhh that makes sense.

PR #1542, opened 2022-10-24
Add extra array methods

mihaitodor:
I saw this [come up on Slack](https://gophers.slack.com/archives/CLWCBK7FY/p1665005208025259) where someone was looking smth similar to [`index_of`](https://www.benthos.dev/docs/guides/bloblang/methods#index_of) for arrays and thought it might be a good-to-have utility, although after I implemented these methods, I realised that [`enumerated`](https://www.benthos.dev/docs/guides/bloblang/methods#enumerated) can be leveraged for the use case I had in mind. I picked `find` because there's [`std::find`](https://en.cppreference.com/w/cpp/algorithm/find) in C++ and C# has [`Find` for arrays](https://learn.microsoft.com/en-us/dotnet/api/system.array.find). I'm OK to rename them to smth else or discard them entirely if they don't really seem useful.

I also did a bit of cleanup in 6865938.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ All notable changes to this project will be documented in this file.
 - New `base64_encode` and `base64_decode` functions for the awk processor.
 - Param `use_number` added to the `parse_json` bloblang method.
 - Fields `init_statement` and `init_files` added to all sql components.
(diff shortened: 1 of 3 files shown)


Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #1545, opened 2022-10-26
Add small note to the mapping processor

mihaitodor:
A colleague pointed out that people who jump right into the `mapping` processor docs might not realise immediately that it's a drop-in replacement for the `bloblang` one.


Jeffail (approved):
Yeah that makes sense, thanks @mihaitodor!

PR #1551, opened 2022-10-30
doc: fix typo error in pipeline interface comments

proost:
remove duplicated word `based` from `Pipeline` interface comments.


Jeffail (approved):
Thanks @proost!

PR #1557, opened 2022-11-03
Fix snowflake_put output token expiry bug

### LINKED ISSUE #1556: Snowflake_put output: Auth token expires after 4 hours of inactivity
My setup (kafka input, snowflake_put output) gets its data sporadically, usually in a burst once per day.

I get the following error:
snowflake_auth_error

Apparently a snowflake session is only valid for 4 hours. My current, temporary, solution is to restart the benthos client on a schedule. 

The gosnowflake [config ](https://pkg.go.dev/github.com/snowflakedb/gosnowflake@v1.6.6#Config) has a `KeepSessionAlive` field that I think would remedy this problem. 

Wouldn't it be nice to have this as a param in the snowflake_put config block, and add it [here](https://github.com/benthosdev/benthos/blob/547723e3ab6b49881d7f52a1e350d418080a261e/internal/impl/snowflake/output_snowflake_put.go#L563
)?

Or perhaps I'm thinking about this incorrectly and there's another way of solving my issue?

Thanks,
Niclas

mihaitodor:
This is an attempt to prevent the Snowflake connection from becoming invalid after 4 hours as described [here](https://github.com/snowflakedb/gosnowflake/issues/556).

The change is inspired from [this example](https://github.com/snowflakedb/gosnowflake/blob/82aecfae87c571aa7dfe1c7542e1e48517e79ec8/cmd/keepalive/keepalive.go#L37-L53) in the github.com/snowflakedb/gosnowflake code.

Fixes #1556.

### DIFF
--- a/internal/impl/snowflake/output_snowflake_put.go
+++ b/internal/impl/snowflake/output_snowflake_put.go
@@ -560,6 +560,12 @@ func newSnowflakeWriterFromConfig(conf *service.ParsedConfig, mgr *service.Resou
 		authenticator = gosnowflake.AuthTypeSnowflake
 	}
 
+	params := make(map[string]*string)
+	valueTrue := "true"
+	// This parameter must be set to prevent the auth token from expiring after 4 hours.
+	// Details here: https://github.com/snowflakedb/gosnowflake/issues/556
+	params["client_session_keep_alive"] = &valueTrue
+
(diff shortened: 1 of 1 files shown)


Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #1559, opened 2022-11-04
Expose the client_session_keep_alive param for snowflake_put

mihaitodor:
This is a refinement to the fix in #1557 for issue #1556. I started to worry that there might be some reason (permissions, costs, etc) for which the `gosnowflake` authors left this as an opt-in thing.


Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #1562, opened 2022-11-07
Fix memory cache docs bug

mihaitodor:
Not sure if that `either` was used to suggest some alternative mechanism in the past, but I guess setting `compaction_interval: ""` is the only mechanism to disable compaction now.


Jeffail (approved):
Good catch, thanks @mihaitodor!

PR #1564, opened 2022-11-07
feat: add find and update and mongo client options

lucasoares:
This PR will add a new operation for MongoDB (find-and-update) and also new client options:

Default values:
```
  password_set: false
  auth_source: ""
  auth_mechanism: ""
  app_name: ""
  connect_timeout: 10s
  socket_timeout: 30s
  server_selection_timeout: 30s
  min_pool_size: "0"
  max_pool_size: "100"
```

It will also add two processor option:

```
  sort_map: ""
  comment: ""
```

### DIFF
--- a/internal/impl/mongodb/client/config.go
+++ b/internal/impl/mongodb/client/config.go
@@ -26,6 +26,8 @@ const (
 	OperationUpdateOne Operation = "update-one"
 	// OperationFindOne Find one operation.
 	OperationFindOne Operation = "find-one"
+	// OperationFindAndUpdate Find and update operation.
+	OperationFindAndUpdate Operation = "find-and-update"
 	// OperationInvalid Invalid operation.
 	OperationInvalid Operation = "invalid"
 )
@@ -45,6 +47,8 @@ func NewOperation(op string) Operation {
 		return OperationUpdateOne
 	case "find-one":
 		return OperationFindOne
+	case "find-and-update":
+		return OperationFindAndUpdate
 	default:
 		return OperationInvalid
 	}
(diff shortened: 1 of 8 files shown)


internal/impl/mongodb/client/config.go
```diff
@@ -121,5 +232,21 @@ func ConfigDocs() docs.FieldSpecs {
...
+		docs.FieldString("auth_mechanism", `The mechanism to use for authentication. This can also be set through the "authMechanism" URI option. (e.g. "authMechanism=PLAIN"). For more information, see [MongoDB docs](https://docs.mongodb.com/manual/core/authentication-mechanisms/).`).Optional().Advanced().AtVersion("4.11.0"),
+		docs.FieldObject("auth_mechanism_properties", `Used to specify additional configuration options for certain mechanisms. For more information, see [MongoDB docs](https://docs.mongodb.com/manual/core/authentication-mechanisms/).`).Optional().Advanced().AtVersion("4.11.0").WithChildren(
+			docs.FieldString("service_name", "The service name to use for GSSAPI authentication.").Optional(),
+			docs.FieldBool("canonicalize_host_name", "Whether to canonicalize the host name for GSSAPI authentication.").Optional(),
+			docs.FieldString("service_realm", "The service realm to use for GSSAPI authentication.").Optional(),
+			docs.FieldString("service_host", "The service host to use for GSSAPI authentication. The service_host and canonicalize_host_name properties must not be used at the same time on Linux and Darwin systems.").Optional(),
```
mihaitodor, line 242: > The service_host and canonicalize_host_name properties must not be used at the same time on Linux and Darwin systems.

Interesting! Do they have to be used at the same time on Windows? I think we could add a lint rule for this where `runtime.GOOS` is concatenated into the bloblang code.

internal/impl/mongodb/processor.go
```diff
@@ -315,9 +386,16 @@ func (m *Processor) ProcessBatch(ctx context.Context, spans []*tracing.Span, bat
...
-		case client.OperationFindOne:
+		case client.OperationFindOne, client.OperationFindAndUpdate:
 			var decoded any
-			err := collection.FindOne(context.Background(), filterJSON, findOptions).Decode(&decoded)
+			var err error
+
+			if m.operation == client.OperationFindOne {
+				err = collection.FindOne(context.Background(), filterJSON, findOptions).Decode(&decoded)
+			} else {
+				err = collection.FindOneAndUpdate(context.Background(), filterJSON, docJSON, findAndUpdateOptions).Decode(&decoded)
```
mihaitodor, line 396: Hmm... Looking at this code makes me wonder if we should be using the `ctx` parameter from `ProcessBatch()` here instead of `context.Background()`. Same below for `collection.BulkWrite()`. This might be a bug...

mihaitodor (commented):
Nice one & thanks for bearing with me! It took me a while to go over these changes... I haven't tried to test this new code manually, but I think it should be fine, since you added some tests too. I left a few small comments.

internal/impl/mongodb/processor.go
lucasoares: Exactly mt thoughts at first, but since other commands were using the background too, I used the same...

Best way to test this is by executing a long-execution command and then signal a shutdown?

mihaitodor: I think it might be hard to repro since Benthos calls `os.Exit` eventually. But yeah, it’s rude to let it terminate the operation that way.

internal/impl/mongodb/client/config.go
lucasoares: Hey @mihaitodor how can I do a lint rule in this API? I saw a LinterFunc but I don't know how to access other fields in it..

internal/impl/mongodb/processor.go
lucasoares: Fixed.

internal/impl/mongodb/client/config.go
mihaitodor: There are a few examples [here](https://github.com/benthosdev/benthos/blob/3310065cf97944a8545563187997cc47e409fa92/internal/impl/snowflake/output_snowflake_put.go#L197-L209). Not sure if that helps you.

PR #1567, opened 2022-11-09
feat: add redis rate_limit

### LINKED ISSUE #337: using a cache for Rate Limiting
@Jeffail,

We have created our own custom processor which can use a redis cluster for rate limiting purposes and would like to contribute the code as new functionality in benthos. Before getting started though I am wondering where and how it should be added in terms of structure. I see a number of different approaches:

- Create a new `cache` or `redis` rate limits resource effectively duplicating the existing `cache` resource and define new args for this functionality under the existing `rate_limit` processor
- Allow the `rate_limit` processor to use cache resources and define new args for this functionality
- Create a new `rate_limit_cache` or `rate_limit_redis`  processor and disregard the `rate_limits` and `cache` resource types

Any preferences between these or other ways of structuring this new functionality??

lucasoares:
Closes #337 

@Jeffail there is 1 thing I want to be fixed before the PR is accepted and another thing I need your opinion..

1. (want to be fixed before this PR) - Currently the implementation always returns the wait time. What will happen if I have more accumulated messages waiting to be processed other than my rate limit supports? Example:

```
count: 10
interval: 1s
```

With 30 processor threads, then all 30 messages will access the rate_limit simultaneously. 10 will be processed because it is inside my rate limit window. The other 20 will receive a sleep (for example 1s) and will follow after this 1s, which skip my rate_limit window and process 20 messages in the 1s interval.

Should we change the implementation so the rate_limit `Access` block itself and only return when a free slot is available? I would be able to re-access the cache until a slot is free for that message.

I can change the implementation to always return `0` in the `Access` for this redis implementation, and deal with the sleep inside the component, but I don't know if this is acceptable, please let me know.

2. (to think later, not needed for this PR but want your opinion) - The rate limit implementation doesn't let the component to have an interpolated string field because the `Access` method doesn't receive the message, only the `context.Context`. Do you it's possible to change this in the future? I want the rate limit `key` for redis to be an interpolated string field to have dynamic keys.

Thanks.


Jeffail:
Hey @lucasoares, the interval returned by a rate limit isn't intended to match the exact amount of time for the caller to wait before moving on, it's meant as a guestimate as to how long it should wait _before attempting to access again_, docs here: https://pkg.go.dev/github.com/benthosdev/benthos/v4@v4.10.0/public/service#RateLimit. You can see an example of it in action here: https://github.com/benthosdev/benthos/blob/main/internal/impl/pure/processor_rate_limit.go#L64

That behaviour is important to maintain as it's common for rate limits to be used in synchronous requests, where the rate limited resonse is expected to be non-blocking, but the client needs an indicator as to how severely it's being rate limited.

As for 2, we could potentially expand rate limits with an optional key activated method but this wouldn't be backwards compatible so we couldn't enforce it generally. I'm not sure how viable that would be going forwards as it would make rate limit implementations much more complicated if they were all expected to behave dynamically by key.

lucasoares:
> Hey @lucasoares, the interval returned by a rate limit isn't intended to match the exact amount of time for the caller to wait before moving on, it's meant as a guestimate as to how long it should wait _before attempting to access again_, docs here: https://pkg.go.dev/github.com/benthosdev/benthos/v4@v4.10.0/public/service#RateLimit. You can see an example of it in action here: https://github.com/benthosdev/benthos/blob/main/internal/impl/pure/processor_rate_limit.go#L64
> 
> That behaviour is important to maintain as it's common for rate limits to be used in synchronous requests, where the rate limited resonse is expected to be non-blocking, but the client needs an indicator as to how severely it's being rate limited.
> 
> As for 2, we could potentially expand rate limits with an optional key activated method but this wouldn't be backwards compatible so we couldn't enforce it generally. I'm not sure how viable that would be going forwards as it would make rate limit implementations much more complicated if they were all expected to behave dynamically by key.

Nice! If this is the expected behavior this PR is ready to be merged.. May you take a look @Jeffail ?

Also for 2, I will create an issue because the dynamic key will be very important for few projects I'm dealing with and maybe we can think something good for this :D

PR #1569, opened 2022-11-14
Add support for PKCS8 encrypted keys for TLS client certs

### LINKED ISSUE #1565: Kafka TLS: Thoughts on supporting PKCS #8 PEM private key?
### Discussed in https://github.com/benthosdev/benthos/discussions/1519

<div type='discussions-op-text'>

<sup>Originally posted by **dtanner** October 14, 2022</sup>
Reading [the docs](https://www.benthos.dev/docs/components/outputs/kafka#tlsclient_certspassword) for `tls.client_certs[].password`, it looks like there's a single supported encryption format of PKCS #1.  Is that correct?  If so, any plans to support other formats?  

Background:
I got a `tls: failed to parse private key` error when trying to use a password-protected, PEM-formatted PKCS8 private key as my `key_file` for a kafka output. The same file works with other tools like `kcat` and the java kafka client.  The java client doesn't support PKCS #1 😢 but does support PKCS #8 (only with passwords though!), and I'm trying to find a single compatible format for all my kafka tooling.  
I resolved the issue by using an unencrypted PKCS #8 PEM file, which does work in Benthos.  So hopefully I'm missing something, but if not I'm curious if there's anything on the roadmap for encryption support.

</div>

mihaitodor:
Fixes #1565.


Jeffail:
How much do we trust github.com/youmark/pkcs8 to get everything right? I'm a bit shy of adding new crypto stuff.

mihaitodor:
> How much do we trust github.com/youmark/pkcs8 to get everything right? I'm a bit shy of adding new crypto stuff.

It's a messy situation, where the standard library authors deprecated the obsolete algorithm, but didn't put in the effort to implement the newer ones. See here: https://github.com/golang/go/issues/8860

However, we're [using it](https://github.com/benthosdev/benthos/blob/56c6a589b0fddef52a7e505db919fa24f54678cf/go.mod#L99) already since I needed to add it for the `snowflake_put` output [here](https://github.com/benthosdev/benthos/blob/a6e4783ca3ea76e2b049e46310c5df832b5d52eb/internal/impl/snowflake/output_snowflake_put.go#L345). I wasn't too happy to have to do that, but that's what the Snowflake Terraform provider does [here](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/e1227159be50bf26841acead8730dad516a96ebc/pkg/provider/provider.go#L453).

The MongoDB folks seem to trust it enough to add it to their driver: https://github.com/mongodb/mongo-go-driver/blob/766f8281c63d88369371f5b079e0381a75fc7953/mongo/options/clientoptions.go#L23. See some background discussion [here](https://jira.mongodb.org/browse/GODRIVER-364).

TL; DR: It looks like a standard algo and the 3rd party lib doesn't do anything super low-level, but we can also wait until there's official support. I just saw it as low-hanging fruit and felt bored 😅

Jeffail (approved):
Awesome, LGTM then, thanks for the investigation @mihaitodor!

PR #1571, opened 2022-11-15
add application properties map to amqp_1 output

DaniTulp:
Should populate the following in amqp_1 https://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-application-properties.

Azure Service Bus or Event Hub use fields in this to provide deduplication. Internally we use it for routing messages to proper handlers as well without having to run multiple queues/hubs.

### DIFF
--- a/internal/impl/amqp1/output.go
+++ b/internal/impl/amqp1/output.go
@@ -197,7 +197,6 @@ func (a *amqp1Writer) WriteBatch(ctx context.Context, msg message.Batch) error {
 
 	return output.IterateBatchedSend(msg, func(i int, p *message.Part) error {
 		m := amqp.NewMessage(p.AsBytes())
-		var err error
 
 		if a.applicationPropertiesMap != nil {
 			value, err := a.applicationPropertiesMap.Query(p)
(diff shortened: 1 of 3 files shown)


internal/impl/amqp1/output.go
```diff
@@ -184,14 +197,28 @@ func (a *amqp1Writer) WriteBatch(ctx context.Context, msg message.Batch) error {
 
 	return output.IterateBatchedSend(msg, func(i int, p *message.Part) error {
 		m := amqp.NewMessage(p.AsBytes())
+		var err error
```
Jeffail, line 200: Looks like this isn't needed

internal/impl/amqp1/output.go
```diff
@@ -184,14 +197,28 @@ func (a *amqp1Writer) WriteBatch(ctx context.Context, msg message.Batch) error {
...
+
+		if a.applicationPropertiesMap != nil {
+			value, err := a.applicationPropertiesMap.Query(p)
+			if err != nil {
+				return err
+			}
+			applicationProperties, ok := value.(map[string]interface{})
+			if !ok {
+				return errors.New("")
```
Jeffail, line 209: Don't you hold out on me, daddy wants the error text

internal/impl/amqp1/output.go
DaniTulp: should be fixed

PR #1575, opened 2022-11-17
feat: add support for primitive.ObjectId type

### LINKED ISSUE #1574: Add support to MongoDB ObjectId type as a bloblang method
ObjectId is an internal type used by MongoDB and currently, it is impossible to query fields using this type.

lucasoares:
Closes #1574

### DIFF
--- a/internal/bloblang/query/type_helpers.go
+++ b/internal/bloblang/query/type_helpers.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/Jeffail/gabs/v2"
+	"go.mongodb.org/mongo-driver/bson/primitive"
 )
 
 // SliceToDotPath returns a valid dot path from a slice of path segments.
@@ -230,6 +231,8 @@ func IGetTimestamp(v any) (time.Time, error) {
 		return time.Parse(time.RFC3339Nano, string(t))
 	case string:
 		return time.Parse(time.RFC3339Nano, t)
+	case primitive.ObjectID:
+		return t.Timestamp(), nil
 	}
 	return time.Time{}, NewTypeError(v, ValueNumber, ValueString)
 }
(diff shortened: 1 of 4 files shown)


internal/impl/mongodb/bloblang.go
```diff
@@ -0,0 +1,38 @@
...
+
+func init() {
+	objectIdParseSpec := bloblang.NewPluginSpec().
+		Beta().
+		Category("Parsing").
+		Description("Decodes a string or a date into a [MongoDB's ObjectId](https://www.mongodb.com/docs/manual/reference/method/ObjectId/).").
+		Example("", `root._id = content().parse_object_id()`)
+
+	if err := bloblang.RegisterMethodV2(
+		"parse_object_id", objectIdParseSpec,
```
Jeffail, line 19: The naming here should probably be prefixed with something like `mongodb_` in order to make it clear that this is very specific to mongodb use cases

internal/bloblang/query/type_helpers.go
```diff
@@ -230,6 +231,8 @@ func IGetTimestamp(v any) (time.Time, error) {
 		return time.Parse(time.RFC3339Nano, string(t))
 	case string:
 		return time.Parse(time.RFC3339Nano, t)
+	case primitive.ObjectID:
```
Jeffail, line 234: This is specifically needed for the use case? We shouldn't really be importing third party client libraries into `./internal/bloblang` as it'll then bring them into every single possible custom build.

Sebmaster:
Instead of treating ObjectId specifically, would it be possible to expose [UnmarshalExtJSON](https://github.com/mongodb/mongo-go-driver/blob/v1.11.1/bson/unmarshal.go#L59) instead? That'd allow creating object ids via `UnmarshalExtJSON({$oid: "<hexstr>"})`, but also enable all other mongo-specific types (e.g. `Timestamp`).

lucasoares:
I'm closing since I don't think benthos should support primitive types from third party tools. Probably the best way to go is using @Sebmaster suggestion in https://github.com/benthosdev/benthos/pull/1642

PR #1578, opened 2022-11-18
Add optional range parameters to `random_int` function

mfamador:
I've added optional range parameters to the `random_int` function that allows us to specify a range for our resulting random numbers.

We can use it in a mapping like this: 

```yml
input:
  label: generate
  generate:
    mapping: |
      root = {
        "random": random_int(), 
        "random_seed": random_int(1), 
        "random_ge_0_le_20": random_int(timestamp_unix_nano(), 0, 20),
        "random_le_20": random_int(max:20),
        "random_ge_0_le_5_with_seed": random_int(seed:timestamp_unix_nano(), min:0, max:5)
      }
```

### DIFF
--- a/internal/bloblang/query/functions.go
+++ b/internal/bloblang/query/functions.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"math/rand"
 	"sync"
 	"time"
@@ -631,20 +632,28 @@ var _ = registerFunction(
 var _ = registerFunction(
 	NewFunctionSpec(
 		FunctionCategoryGeneral, "random_int",
-		"Generates a non-negative pseudo-random 64-bit integer. An optional integer argument can be provided in order to seed the random number generator.",
+		"Generates a non-negative pseudo-random 64-bit integer. An optional integer argument can be provided in order to seed the random number generator. Optional `min` and `max` arguments can be provided to make the generated numbers within a range.",
 		NewExampleSpec("",
 			`root.first = random_int()
-root.second = random_int(1)`,
+root.second = random_int(1)
+root.third = random_int(max:20)
+root.forth = random_int(min:10, max:20)
+root.fifth = random_int(timestamp_unix_nano(), 5, 20)
+root.sixth = random_int(seed:timestamp_unix_nano(), max:20)
+`,
 		),
 		NewExampleSpec("It is possible to specify a dynamic seed argument, in which case the argument will only be resolved once during the lifetime of the mapping.",
 			`root.first = random_int(timestamp_unix_nano())`,
+			`root.second = random_int(timestamp_unix_nano(), 5, 20)`,
 		),
 	).
 		Param(ParamQuery(
 			"seed",
 			"A seed to use, if a query is provided it will only be resolved once during the lifetime of the mapping.",
 			true,
-		).Default(NewLiteralFunction("", 0))),
+		).Default(NewLiteralFunction("", 0))).
+		Param(ParamInt64("min", "The minimum value the random generated number will have. The default value is 0.").Default(0)).
+		Param(ParamInt64("max", fmt.Sprintf("The maximum value the random generated number will have. The default value is %d (math.MaxInt64 - 1).", uint64(math.MaxInt64-1))).Default(int64(math.MaxInt64-1))),
 	randomIntFunction,
 )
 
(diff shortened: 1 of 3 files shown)


mfamador:
Or we could add the range `min` and `max` parameters to the existent `random_int`, after `seed`to keep it backwards compatible.

mfamador:
I guess I prefer the latter, you tell me.

mihaitodor:
Whichever you like more. For me, the latter means less to remember :)

mfamador:
Ok, I'll add the range parameters to the existent `random_int`. Who said we couldn't do function overloading in golang? :D

internal/bloblang/query/functions.go
```diff
@@ -653,6 +662,14 @@ func randomIntFunction(args *ParsedParams) (Function, error) {
...
 		return nil, err
 	}
+	min, err := args.FieldInt64("min")
+	if err != nil {
+		return nil, err
+	}
+	max, err := args.FieldInt64("max")
+	if err != nil {
+		return nil, err
+	}
```
mihaitodor, line 672: Might be worth checking somewhere here that `max - min >= 0`. Otherwise, `Int63n` will panic for such cases.

internal/bloblang/query/functions.go
```diff
@@ -675,7 +692,11 @@ func randomIntFunction(args *ParsedParams) (Function, error) {
 			r = rand.New(rand.NewSource(seed))
 		}
 
-		v := int64(r.Int())
+		n := max - min
+		if max < int64(math.MaxInt64) {
+			n++
+		}
```
mihaitodor, line 698: I'm sure there must be a simple explanation for this, but I'm struggling to get it... If `Int63n(n)` returns values from `[0,n)`, then I guess the goal is to obtain values in `[0,n]`, but please confirm and maybe leave a comment in the code if you don't mind :)

internal/bloblang/query/functions.go
```diff
@@ -631,20 +632,28 @@ var _ = registerFunction(
...
 		),
 	).
 		Param(ParamQuery(
 			"seed",
 			"A seed to use, if a query is provided it will only be resolved once during the lifetime of the mapping.",
 			true,
-		).Default(NewLiteralFunction("", 0))),
+		).Default(NewLiteralFunction("", 0))).
+		Param(ParamInt64("min", "The minimum value the random generated number will have. The default value is 0").Default(0)).
+		Param(ParamInt64("max", "The maximum value the random generated number will have. The default value is 9223372036854775807").Default(int64(math.MaxInt64))),
```
mihaitodor, line 656: Nit: Instead of hardcoding that number, consider using `+ fmt.Sprint(math.MaxInt64)` instead

internal/bloblang/query/functions.go
mfamador: Yes, the interval is half open so we need to increase the input parameter with 1 if we want to get random numbers equal to `max`. This check was to prevent it'd overflowing if max == maxInt64.
I've changed the doc and validation to make sure max is always smaller than maxInt64 so we can remove this check.

internal/bloblang/query/functions.go
mfamador: Added validations to make sure min is positive and max if equal or bigger than min

internal/bloblang/query/functions.go
mfamador: Fixed it

internal/bloblang/query/functions.go
mfamador: This was a problem if we'd want to set min = max, when Int64n would panic with n = 0.
Maybe we want that max is always bigger than min?

mfamador: Even though I guess we'd expect a closed interval on `random_in` when specifying `max`, so we need to add 1 to `n` on `Int63n`

mihaitodor (approved):
LGTM, thanks! I did some quick search for prior art in other languages and there are diverging opinions on how the max value should work (inclusive vs exclusive) and if min == max is allowed or not and if min and max should be automagically swapped if min > max, but there doesn't seem to be any consensus, as expected https://xkcd.com/927/ :) The current approach makes sense to me. The inclusive max has the small downside that users might be tempted to use `math.MaxInt64` for max, but that's grand. They'll pay more attention if they get an error 👹  Would you mind changing `The default value is %d` to `The default value is %d ([math.MaxInt64](https://pkg.go.dev/math#pkg-constants))`?

mfamador:
Changed, @mihaitodor. Thanks

mihaitodor (approved):

Jeffail (approved):
LGTM, thanks @mfamador!

PR #1582, opened 2022-11-21
feat: add support for subscription creation on gcp_pubsub input

### LINKED ISSUE #1579: Add support for subscription creation on gcp_pubsub input
Benthos have no support for automatically creation of subscriptions on gcp_pubsub input. 

It adds an effort because every new subscription should be mannually created on GCP. 

Is it a good feature for the project?

manoelstilpen:
If the user configures it, Benthos can submit a new subscription to PubSub;

I just created a new object property on gcp_pubsub input:

```yaml
create_subscription:
  enabled: true
  topic: "topic-id"
```

Closes #1579

### DIFF
--- a/internal/component/input/config_gcp_pubsub.go
+++ b/internal/component/input/config_gcp_pubsub.go
@@ -2,13 +2,17 @@ package input
 
 // GCPPubSubConfig contains configuration values for the input type.
 type GCPPubSubConfig struct {
-	ProjectID              string `json:"project" yaml:"project"`
-	SubscriptionID         string `json:"subscription" yaml:"subscription"`
-	TopicID                string `json:"topic" yaml:"topic"`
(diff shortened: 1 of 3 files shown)


Jeffail:
Nice, I think we ought to structure this within its own configuration section as the topic is only relevant for sub creation, and I can imagine maybe we'd need to add more creation specific fields in the future (like timeouts), so change:

```yaml
topic: foo
sync: true
create_subscription: true
```

To look like this:

```yaml
sync: true
create_subscription:
    enabled: true
    topic: foo
```

And also we should immately return an error in construction if `enabled` is `true` but the topic is not set as that'd be a configuration error.

manoelstilpen:
@Jeffail done

Jeffail (approved):
Awesome, thanks @manoelstilpen!

PR #1584, opened 2022-11-24
Add format_xml bloblang method

mihaitodor:
Also add `no_indent` parameter to the `format_json()` bloblang method. Unfortunately, we can't currently distinguish between `format_json()` and `format_json("    ")`, so the easiest thing to do is to add an extra parameter to be consistent with `format_xml()`. Otherwise, people can't force these methods to not insert any `\n` characters, because setting the indent to `""` still emits them and, then again, diverging from the Go API behaviour is probably a bad idea...

Unfortunately, I couldn't use `encoding/xml` for this task because it has some limitations such as https://github.com/golang/go/issues/48624. It's not clear to me why it doesn't support the same kinds of inputs as `encoding/json`, but, alas...


Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #1590, opened 2022-11-28
Add service.AutoRetryNacks test for read after close

mihaitodor:
I believe this test should succeed, but I could be wrong... 

I'm trying to fix this other test https://github.com/benthosdev/benthos/blob/main/internal/impl/mongodb/input_test.go and it looks like there's smth wonky with the way `public/service/input_auto_retry.go` works. Basically, I'm expecting `selectInput.Read(ctx)` here https://github.com/benthosdev/benthos/blob/b195d9eb87f6b4a5637330f4e276b06ea344ff17/internal/impl/mongodb/input_test.go#L235 to unblock when `service.ErrEndOfInput` is returned by the input `Read()` method, but that doesn't seem to be the case. Right now, it's just blocking forever, because the test author of the test didn't use `context.WithTimeout()`, but that's another issue... There are a few other issues in this MongoDB test that I'll address once this issue is clarified.

### DIFF
--- a/public/service/input_auto_retry_test.go
+++ b/public/service/input_auto_retry_test.go
@@ -339,3 +339,66 @@ func TestAutoRetryBuffer(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, exp3, string(b))
 }
+
+func TestAutoRetryReadAfterClose(t *testing.T) {
+	t.Parallel()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+
+	readerImpl := newMockInput()
+	readerImpl.msgsToSnd = append(readerImpl.msgsToSnd, NewMessage([]byte("foo")))
+
+	pres := AutoRetryNacks(readerImpl)
+
+	wg := sync.WaitGroup{}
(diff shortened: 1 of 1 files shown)


public/service/input_auto_retry_test.go
```diff
@@ -339,3 +339,51 @@ func TestAutoRetryBuffer(t *testing.T) {
...
+			t.Error("Timed out")
+		}
+
+		// Force mockInput.Read() to return ErrEndOfInput
+		close(readerImpl.readChan)
+	}()
+
+	require.NoError(t, pres.Connect(ctx))
+
+	msg, _, err := pres.Read(ctx)
```
Jeffail, line 375: It's because this read is never acked, so as far as the auto retryer is concerned we're not finished, as if this message gets nacked we need to ensure the input is still being read from.

mihaitodor: Ahhhhh! That makes sense. I forgot about the ack function... Indeed, it has to be called to unblock the workflow. Thanks a million! 😘

Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #1593, opened 2022-12-01
Set `Content-Type` header when returning lint errors

danmcgee-soda:
I noticed today that although lint errors return a JSON body and a 400 response code, the `Content-Type` header isn't explicitly set. This causes the http module to assume `text/plain`.

We're using a library that only parses the body as JSON if the content type is properly set, which caused us to trip up when errors were encountered in bloblang syntax.

This adds an explicit `Content-Type` header to any lint error return path, as well as tests for it to be set.

I removed a `sort.Strings()` call as well on one of the four places lint errors were returned. It seemed odd that we didn't do it in all cases, and thinking through it, it is best to leave lint errors in line number order. Doing a sort causes a line 11 error to be sorted before a line 9 error, which isn't intuitive.

### DIFF
--- a/internal/stream/manager/api.go
+++ b/internal/stream/manager/api.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"sort"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


Jeffail (approved):
LGTM, thanks @danmcgee-soda! I have a vague memory that the sorting may have been necessary for tests as some lints were randomly ordered, but that's probably no longer needed. If we see tests occasionally failing it's easy enough to add back in later (or fix the tests instead).

PR #1604, opened 2022-12-07
Document bug fix in sql_insert output from c6a71e9

mihaitodor:


Jeffail (approved):
Nice catch, thanks @mihaitodor!

PR #1607, opened 2022-12-08
Add open_message_type flag for websocket input

mihaitodor:
According to [the docs](https://pkg.go.dev/github.com/gorilla/websocket?utm_source=godoc#hdr-Data_Messages), the type is recorded in the data frame. It looks like in some cases, it should be set to `TextMessage` instead of `BinaryMessage`. For example this config doesn't work without setting `open_message_type: binary`:

```yaml
input:
  websocket:
    url: "wss://ws.blockchain.info/inv"
    open_message: '{"op":"ping_block"}'
  processors:
    - log:
        message: ${! content() }

output:
  stdout: {}
```

### DIFF
--- a/internal/component/input/config_websocket.go
+++ b/internal/component/input/config_websocket.go
@@ -5,10 +5,21 @@ import (
 	btls "github.com/benthosdev/benthos/v4/internal/tls"
 )
 
+// OpenMsgType represents the type of the open_message field.
+type OpenMsgType string
+
+const (
+	// OpenMsgTypeBinary sets the type of open_message to binary.
+	OpenMsgTypeBinary OpenMsgType = "binary"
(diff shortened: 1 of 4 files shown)


Jeffail:
LGTM but I'm maybe tempted to go with an enum of `text` and `binary` just for the extremely unlikely case that they expand it, although I have no idea what they'd expand it with.

mihaitodor:
Indeed, adding that bool field was sloppy. Should look better now :)

Jeffail (approved):
LGTM, thanks @mihaitodor!

Edit: Although it looks like that test error isn't a flakey one.

mihaitodor:
Should be fixed now. The websocket handler was getting stuck when one of the tests wasn't sending an OpenMsg.

PR #1608, opened 2022-12-08
Preallocate a few slices in public/service/processor.go

mihaitodor:
TODO: Test if this has any impact on #1605 (guess it won't be much if any).

I am seeing a negligible improvement (3s) on the overall duration of `go test -run=XXX -bench=StreamPipelines ./public/service`, so I guess either the current tests don't really get these slices to grow much or the optimiser is smart enough nowadays.

[optimised.txt](https://github.com/benthosdev/benthos/files/10181744/optimised.txt)
[orig.txt](https://github.com/benthosdev/benthos/files/10181745/orig.txt)

### DIFF
--- a/public/service/processor.go
+++ b/public/service/processor.go
@@ -100,9 +100,9 @@ func newAirGapBatchProcessor(typeStr string, p BatchProcessor, mgr bundle.NewMan
 }
 
 func (a *airGapBatchProcessor) ProcessBatch(ctx context.Context, spans []*tracing.Span, batch message.Batch) ([]message.Batch, error) {
-	inputBatch := make([]*Message, batch.Len())
+	inputBatch := make([]*Message, 0, batch.Len())
 	_ = batch.Iter(func(i int, p *message.Part) error {
-		inputBatch[i] = newMessageFromPart(p)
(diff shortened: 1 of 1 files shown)


public/service/processor.go
```diff
@@ -145,7 +145,11 @@ func (o *OwnedProcessor) Process(ctx context.Context, msg *Message) (MessageBatc
 		return nil, res
 	}
 
-	var b MessageBatch
+	outputBatchLength := 0
+	for _, iMsg := range iMsgs {
```
mihaitodor, line 149: This might be overkill...

Jeffail:
Yass I imagine this is the right place to look, I did consider options for airgap optimisations but it's kind of edge caseish based on large batches, worth doing though. Allocating the right length will get us somewhere but also we can consider pooling some of this stuff as it's just wrapper types and slices.

filippog:
Thank you @mihaitodor !

> TODO: Test if this has any impact on #1605 (guess it won't be much if any).

I did another test with this commit and can confirm I can't see any significant improvement :(

mihaitodor:
I'll close this for now, since it needs more detailed analysis and proper benchmarking. I'll submit a new PR when I'll get a chance to revisit this in detail.

PR #1613, opened 2022-12-12
Fix broken link

mihaitodor:
`/docs/configuration/interpolation#metadata` -> `/docs/configuration/interpolation#bloblang-queries`

Also regenerate Pulsar docs.

### DIFF
--- a/website/docs/components/inputs/pulsar.md
+++ b/website/docs/components/inputs/pulsar.md
@@ -43,6 +43,8 @@ input:
     topics: []
     subscription_name: ""
     subscription_type: shared
+    tls:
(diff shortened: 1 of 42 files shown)


website/docs/components/inputs/pulsar.md
```diff
@@ -43,6 +43,8 @@ input:
     topics: []
     subscription_name: ""
     subscription_type: shared
+    tls:
```
mihaitodor, line 46: Looks like the Pulsar docs haven't been refreshed in a while. They don't get picked up by automatically by the docs generator since they're not included in `public/components/all/package.go`.

Issue #1568, opened 2022-11-11
`asdf` as an installer?


rolandtritsch:
Hi All, 

any plans to add support for `asdf`? 

If people think that it is a good idea (with that you can have multiple versions of `benthos` installed and can switch between them easily), I am happy to contribute a PR.

Thoughts?

mihaitodor:
Hey @rolandtritsch, nice to see you here! 😄 That's a great idea! Would love to see Benthos among the official `asdf` [plugins](https://github.com/asdf-vm/asdf-plugins). I had a quick look over [this doc](https://github.com/asdf-vm/asdf/blob/master/docs/plugins/create.md) and I guess a separate repo is required to maintain a bunch of scripts which fetch the version list via the GitHub API. Guess it's just a fork of [this template](https://github.com/asdf-vm/asdf-plugin-template) with a new GitHub personal token. Is that what you were thinking to do?

rolandtritsch:
Hi @mihaitodor. LOL ... good to "see" you too. Small world. Yes. That's sounds about right. Would look to "host" the repo on `benthosdev` (or I can create/host it on `rolandtritsch` or `community`). Sounds good?

mihaitodor:
Small world indeed! ❤️  OK, I'll check with @Jeffail about getting that set up under `benthosdev`. Should be sorted early next week. I'll keep you posted.

mihaitodor:
I created the plugin repo [here](https://github.com/benthosdev/asdf-benthos) from the template and I ran the initial `bash setup.bash` setup. The next step is to look at the TODOs and see why the build failed. Also, I have to check what GitHub personal access token needs to be created and where it needs to be set up.

rolandtritsch:
Hi Mihai. Excellent. Happy to help/take over whenever you want/need it.

rolandtritsch:
Hi @mihaitodor. Thanks for setting this up. I have forked the repo and have started to work on the implementation. - https://github.com/rolandtritsch/asdf-benthos

rolandtritsch:
@mihaitodor One quick suggestion/ask: Can we/Should we rename the repo to `benthos-asdf`? Most of the other repos start with `benthos-`. Just for symmetry/consistency. Thoughts?

mihaitodor:
Hey @rolandtritsch, thanks for jumping in! I was meaning to get to it myself, but happy to let you jump into it if you know what must be changed. I think we need to set up a token somehow, just not sure what the process is.

[...]

PR #1614, opened 2022-12-12
Make install take bensos_install_path as a second parameter

rolandtritsch:
We are currently working to add `asdf` as an additional way to install `benthos` (because this gives us an elegant way to have multiple versions of benthos installed at the same time). #1568

The implementation is [here](https://github.com/benthosdev/benthos-asdf/pull/1).

This PR suggests a small backward compatible change/enhancement so that `install` can be used to implement the `asdf` plugin.

The change is to add the ability to also pass the `install_path` as an optional second parameter.

### DIFF
--- a/website/static/sh/install
+++ b/website/static/sh/install
@@ -36,7 +36,25 @@ check_tools() {
 install_benthos()
 {
 	trap 'echo -e "Aborted, error $? in command: $BASH_COMMAND"; trap ERR; exit 1' ERR
-	install_path="/usr/local/bin"
+
+  # Process the command line
+	if [[ "$#" -eq 2 ]]; then
+		benthos_tag="v$1"
+		benthos_version="$1"
+    benthos_install_path="$2"
+	elif [[ "$#" -eq 1 ]]; then
+		benthos_tag="v$1"
+		benthos_version=$1
+    benthos_install_path="/usr/local/bin"
+	elif [[ "$#" -eq 0 ]]; then
+		benthos_tag=$(curl -s https://api.github.com/repos/benthosdev/benthos/releases/latest | grep 'tag_name' | cut -d\" -f4)
+		benthos_version=$(echo ${benthos_tag} | cut -c2-)
+    benthos_install_path="/usr/local/bin"
+	else
+		echo "Too many arguments."
+		exit 1
+	fi
+
 	benthos_os="unsupported"
 	benthos_arch="unknown"
 	benthos_arm=""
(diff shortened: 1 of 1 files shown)


website/static/sh/install
```diff
@@ -45,12 +63,12 @@ install_benthos()
 	check_tools
 
 	if [[ -n "$PREFIX" ]]; then
-		install_path="$PREFIX/bin"
+		benthos_install_path="$PREFIX/bin"
```
rolandtritsch, line 66: Just renaming `install_path` to `benthos_install_path` (for consistency).

website/static/sh/install
```diff
@@ -111,19 +129,6 @@ install_benthos()
 	echo "Downloading Benthos for ${benthos_os}/${benthos_arch}${benthos_arm}..."
 	benthos_file="benthos_${benthos_os}_${benthos_arch}${benthos_arm}${benthos_dl_ext}"
 
-	if [[ "$#" -eq 0 ]]; then
```
rolandtritsch, line 114: This piece was moved to the top of the file.

website/static/sh/install
```diff
@@ -137,12 +142,15 @@ install_benthos()
...
 
-	echo "Putting benthos in $install_path (may require password)"
-	$sudo_cmd mv "$PREFIX/tmp/$benthos_bin" "$install_path/$benthos_bin"
+	echo "Putting benthos in $benthos_install_path (may require password)"
+	$sudo_cmd mv "$PREFIX/tmp/$benthos_bin" "$benthos_install_path/$benthos_bin"
 	$sudo_cmd rm -- "$dl"
 
 	# check installation
-	$benthos_bin -version
+	$benthos_install_path/$benthos_bin -version
```
rolandtritsch, line 150: When installing with `asdf` benthos is not on the path yet (this happens when you run `asdf global ...`.

Means we need to distinguish between we can execute the bin and it's presence on the path.

Jeffail:
👍 thanks @rolandtritsch, the whitespace needs fixing as it's all over the place but I'm happy with the changes.

rolandtritsch:
> +1 thanks @rolandtritsch, the whitespace needs fixing as it's all over the place but I'm happy with the changes.

Uuupppsss. Sorry. Fixed.

Jeffail:
> > +1 thanks @rolandtritsch, the whitespace needs fixing as it's all over the place but I'm happy with the changes.
> 
> Uuupppsss. Sorry. Fixed.

It's consistent but can I be a total pain and ask for tabs instead of spaces? 😅 I don't have a strong preference either way but it makes for a smaller diff since we had tabs before.

rolandtritsch:
> It's consistent but can I be a total pain and ask for tabs instead of spaces? sweat_smile I don't have a strong preference either way but it makes for a smaller diff since we had tabs before.

Good point. Fixed.

Jeffail (approved):
LGTM, thanks @rolandtritsch!

rolandtritsch:
Hi @Jeffail. I now want to cleanup the implementation of the [asdf][] plugin, but when I run `curl -Lsf https://sh.benthos.dev` I am still seeing the old/previous `install.sh`. When will it be available on `sh.benthos.dev`? (cc: @mihaitodor)

[asdf]: https://github.com/benthosdev/benthos-asdf/pull/1

mihaitodor:
Hey @rolandtritsch, Happy New Year! I think that gets updated when a new release is published, but I'll have to defer to @Jeffail for that.

PR #1618, opened 2022-12-13
Allow FS to be swapped by service users

codegangsta:
There are use cases where I'd like to override the underlying ifs.FS for a particular stream/environment. Mainly to be able to source credentials from memory rather than store it on disk (think keystores etc).

I know not all components use `FS()` (I think I need to fix that in our NATS components), but this should hopefully be a start.

### DIFF
--- a/public/service/environment.go
+++ b/public/service/environment.go
@@ -21,6 +21,7 @@ import (
 	"github.com/benthosdev/benthos/v4/internal/component/tracer"
 	"github.com/benthosdev/benthos/v4/internal/config"
 	"github.com/benthosdev/benthos/v4/internal/docs"
+	"github.com/benthosdev/benthos/v4/internal/filepath/ifs"
 	"github.com/benthosdev/benthos/v4/public/bloblang"
 )
 
@@ -32,11 +33,13 @@ import (
 type Environment struct {
 	internal    *bundle.Environment
 	bloblangEnv *bloblang.Environment
+	fs          ifs.FS
 }
 
 var globalEnvironment = &Environment{
 	internal:    bundle.GlobalEnvironment,
 	bloblangEnv: bloblang.GlobalEnvironment(),
+	fs:          ifs.OS(),
 }
 
 // GlobalEnvironment returns a reference to the global environment, adding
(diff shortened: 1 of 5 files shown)


public/service/environment.go
```diff
@@ -67,6 +71,12 @@ func (e *Environment) UseBloblangEnvironment(bEnv *bloblang.Environment) {
 	e.bloblangEnv = bEnv
 }
 
+// UseFS configures the service environment to use an implementation of ifs.FS
+// as its filesystem.
+func (e *Environment) UseFS(fs ifs.FS) {
```
Jeffail, line 76: The interface we accept here will need to be public, a copy/paste jobby that matches `ifs.FS` is fine and then if in future we ever expand `ifs.FS` we'll be forced to do the right thing and bridge the gap between this unchanged public interface and the internal one.

Jeffail (commented):
Thanks @codegangsta, just one comment and the rest LGTM.

public/service/environment.go
codegangsta: @Jeffail Good point! I opted to replace the `service.FS` struct with the interface, I don't believe this would create a breaking change for anyone since the struct didn't export anything and there would be no point to dereference it, but lmk if that's too heavy handed and I can just include the new interface as something like `service.IFS`

Jeffail: Argh yeah I forgot about the FS struct we export. It's a bit of an awkwardness in trying to cover my butt for all angles of backwards compatibility complaints as it allows me to add more capabilities as methods without breaking anything, and avoids annoying type assertions on the user side.

Unfortunately we can't get rid of the struct in either case as we'd be swapping `*service.FS` for `service.FS`. One option that might help prevent duplicating the API is to add a public constructor for `*service.FS` that accepts a https://pkg.go.dev/io/fs#FS, and if the underlying implementation also supports `ifs.FS` then we ref it directly, otherwise we wrap it with the other bits we need.

If you modify this PR to accept a `*service.FS` as it currently is in `UseFS` then I'm happy to add the constructors for `*service.FS` to accomodate your unit tests while I merge it.

We could alternatively accept a `fs.FS` in `UseFS` with the exact same process but my thinking is we might want to add some more helper constructors in the service package in future, or mark the first pass constructor as deprecated if we decide that `fs.FS` is too limited and we want to force the full signature.

codegangsta: Ah, true dat. I like the constructor idea. I'll throw that in today

codegangsta: @Jeffail Thanks for the feedback, I brought back the struct and added a constructor and wrapper, but feel free to change it to your liking.

Jeffail (approved):
Awesome, LGTM, thanks @codegangsta!

PR #1622, opened 2022-12-14
add csv:x delimiter to unarchive processor

sapk:
This PR add the ability to use a custom delimiter similar to file input for csv format.


Jeffail (approved):
LGTM, thanks @sapk. The default linting rule for enum fields tries to validate that the value exactly matches one of the options listed, so pattern based options will be rejected by default. I'll fix that separately though.

PR #1633, opened 2022-12-20
add opt 'dump_request_log' on http_client (i/o) and http (processor)

yusufsyaifudin:
It might be useful if we can print the HTTP request and response via logger with the level of this HTTP log is defined by user. For example, user wants to set the level to `INFO` but their main logger config is set to `ERROR` then the HTTP request-response log (this PR) won't show on their log (because we use the same logger instance).

Also, we can let the user to opt-in/opt-out this HTTP logging (by default off) if they think that the request-response body may contain sensitive data.

When this feature is useful:
* When user want to track and make sure that their `bloblang` or `mapping` processor is really send in the expected format (currently we have to create new `log` processor each time we modify our data)
* When user want to see the actual HTTP response (body and header) on their server log. Currently, it seems hard to see the actual HTTP response returned by the HTTP service unless we're using `propagate_response` which only propagate back to the input (in most cases, we only want to know the log, not propagate back the response)

### DIFF
--- a/internal/httpclient/client.go
+++ b/internal/httpclient/client.go
@@ -77,6 +77,11 @@ func NewClientFromOldConfig(conf oldconfig.OldConfig, mgr bundle.NewManagement, 
 	h.clientCtx, h.clientCancel = context.WithCancel(context.Background())
 	h.client = conf.OAuth2.Client(h.clientCtx)
 
+	h.client.Transport, err = newRequestLog(h.client.Transport, h.log, conf.DumpRequestLog)
+	if err != nil {
+		return nil, fmt.Errorf("failed to config logger for request dump: %v", err)
+	}
+
 	if tout := conf.Timeout; len(tout) > 0 {
 		var err error
 		if h.client.Timeout, err = time.ParseDuration(tout); err != nil {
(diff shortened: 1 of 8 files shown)


yusufsyaifudin:
just seeing the email that the workflow failed on golangci linter and documentation is not updated.. I have fixed that.. please review again @Jeffail

Jeffail:
Hey @yusufsyaifudin, this looks great and is definitely useful.

I'm considering whether we ought to simplify the configuration as it's probably unlikely someone will need to customise the actual log message. Maybe a single field alternative would be to add a field `dump_request_log_level` which is unset by default, but if you set it to a level like `DEBUG` then you get the logs at that level. Does that sound reasonble? If you're okay with it I'm happy to make the change myself after merging.

yusufsyaifudin:
> Hey @yusufsyaifudin, this looks great and is definitely useful.
> 
> I'm considering whether we ought to simplify the configuration as it's probably unlikely someone will need to customise the actual log message. Maybe a single field alternative would be to add a field `dump_request_log_level` which is unset by default, but if you set it to a level like `DEBUG` then you get the logs at that level. Does that sound reasonble? If you're okay with it I'm happy to make the change myself after merging.

@Jeffail are you want to remove the `dump_request_log_level.enable` field? if so, yes, I'm definitely agree with that..

Jeffail:
Okay I've just refactored the HTTP components, it breaks the merge of this PR but that's fine, I'm going to rework this PR as I merge it so no need for any action on your side.

Jeffail:
Merged: https://github.com/benthosdev/benthos/commit/95923f66cde3e835a27a8abc7be659b2ed59f0bc

Issue #1516, opened 2022-10-14
Add some sort of `format_bson` bloblang method with a format specifier parameter


sytgj7896321:
![image](https://user-images.githubusercontent.com/70554029/195746361-9a693b00-2d53-4290-b569-cd3fe1736277.png)
When insert document to mongo, the all number field type is double, be actually I want it to be int32 or int64

mihaitodor:
Alternatively, the `mongodb` processor and output could have something like a `type_annotations` field which contains a `path -> type` mapping. For example, something like this:

```coffee
root.foo = "int32"
root.bar.baz = "int64"
```

Dunno if this makes more sense or it will be more difficult to use.

sytgj7896321:
> Alternatively, the `mongodb` processor and output could have something like a `type_annotations` field which contains a `path -> type` mapping. For example, something like this:
> 
> ```coffee
> root.foo = "int32"
> root.bar.baz = "int64"
> ```
> 
> Dunno if this makes more sense or it will be more difficult to use.

I think it will be better by this way

PR #1642, opened 2022-12-25
Allow ExtJSON in mongo queries

Sebmaster:
We now purposefully unmarshal the Mongo query via the bson library. This adds implicit support for ObjectIds ($oid), Timestamps ($ts), Regexes ($regex), etc. in queries.

This _may_ offer a workaround for #1516 as well (from what I've understood).

First golang contribution, feel free to nitpick 😅


Jeffail:
Hey @Sebmaster, this looks great. I'm not 100% familiar with the implications of this change but it looks as though it's a good idea for usability, if someone else with more experience with MongoDB can chime in I'm mostly concerned about backwards compatibility, but if that's not an issue I'm happy to merge.

Sebmaster:
The biggest (only?) problem I see is that if someone uses the same keys as the ExtJSON schema but wants them to be literal objects, that'd break. But they're all [prefixed with $](https://www.mongodb.com/docs/manual/reference/mongodb-extended-json/#type-representations) so it seems like fairly small breakage (if at all).

lucasoares:
I think all modifications are OK and it will not bring any breaking change since MongoDB doesn't allow fields to be prefixed with `$` in the root of the document: 
image

And since mongodb components are experimental, minor breaking changes would be OK...

@Sebmaster I'm okay to proceed with the merge but I think you should add an integration test case for the success path (where the query does return an element)

torbjornvatn:
Any news on this one @lucasoares or @Jeffail ?

I trying to use Benthos for querying Mongodb with on either on Date fields or the ObjectId `_id`, and this PR would make my life 💯% better!

lucasoares:
> Any news on this one @lucasoares or @Jeffail ?
> 
> I trying to use Benthos for querying Mongodb with on either on Date fields or the ObjectId `_id`, and this PR would make my life 💯% better!
> 
> I can help with creating an integration test case if that's helpful for you guys 😉

Feel free to create the integration test!

Jeffail:
Thanks all, finally got around to reviewing this stuff. The components were massively refactored a while back so the conflicts were too much to work through. I've added this functionality over the refactored common file which is nice and neat: https://github.com/benthosdev/benthos/commit/9f692589be99109dfa32c5566bba536bed8613ab

@Sebmaster the remaining item that isn't merged yet is unmarshalling in the input query as extjson, I figured I'd leave that so that you can rework the PR to only include that part and it'll dramatically reduce the conflicts. This way you'll keep the attribution. However, I do need the DCO check to pass so you'll need to commit with `-s` before I can merge: https://github.com/apps/dco

@torbjornvatn if you're happy to raise a PR with your test case I'd love to take it.

twmb:
Checking in on if you'd like to continue this PR.

PR #1648, opened 2022-12-30
Set the default value of conn_max_idle for sql_* components to 2

mihaitodor:
Currently, this is implicitly set to 0, which deviates from the default value (2) that is specified in the standard library docs [here](https://pkg.go.dev/database/sql#DB.SetMaxIdleConns).

Guess some people who are currently using `conn_max_idle: 0` implicitly might bump into performance issues, not sure...


Jeffail (approved):
LGTM, thanks @mihaitodor.

PR #1649, opened 2022-12-30
Add Couchbase cache implementation

sapk:
This is a simple cache with ttl over the same couchbase client as the processor.


Jeffail:
Awesome, looks great, thanks @sapk!

Jeffail (approved):

PR #1650, opened 2023-01-01
Fix MongoDB input integration test

mihaitodor:
Guess this one hasn't been run in a while...

LE: The tests should finish quickly once the container is running, so we shouldn't need to use `testing.Short()` unless we want to do that for all Docker-based integration tests.


Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #1657, opened 2023-01-05
added input_bigquery_raw

loicalleyne:
Implemented as a new gcp input - essentially the same as input_bigquery_select but without using squirrel to build the query.


Jeffail:
Hey @loicalleyne, looks good, any chance you can add some tests?

mihaitodor:
There's an emulator for BigQuery [here](https://github.com/goccy/bigquery-emulator) which we can use, but I wonder if it might be better to consider adding [go-gorm/bigquery](https://github.com/go-gorm/bigquery) as a SQL driver, so we can leverage the existing `sql_*` components. The library is still young, but I think it should work OK now that [this PR](https://github.com/go-gorm/bigquery/pull/13) got merged. Might be worth waiting until it gets a new release, though.

loicalleyne:
tbh My validation tests were done directly with BigQuery in a production environment. I'm not that experienced with writing tests, I might need a bit of guidance.

 Is there any functionality lost by using gorm instead of Google's own BigQuery library?  
Looking at the difference between sql_select and gcp_bigquery_select, the same logic for keeping those separate and not just adding a BigQuery driver to sql_select might be applied to the adding of gcp_bigquery_rawselect or adding a BigQuery driver to sql_rawselect.

If the BigQuery API features exposed in Benthos differ enough from the APIs of other DBs supported in sql_*select then it makes sense to keep them separate, if not then gcp_bigquery_select could be merged with sql_select and BigQuery raw queries implemented in sql_rawselect. 

In my job I use BigQuery all day every day so my biased personal preference would be an implementation that is BigQuery-specific.

PR #1659, opened 2023-01-06
Enable nacos service registry

intfish123:
feat: Enable to register self to nacos service registry center. In order to better collect service metrics with running in docker mode, or for more functions in futures.


Jeffail:
Hey @intfish123, I'm not sure I understand what the value of this feature is. Could this potentially be something that custom builds implement rather than a feature within the main distribution of benthos?

gedw99:
It’s this ? https://apisix.apache.org/blog/2022/02/21/nacos-api-gateway/

intfish123:
> Hey @intfish123, I'm not sure I understand what the value of this feature is. Could this potentially be something that custom builds implement rather than a feature within the main distribution of benthos?

In more cases, we need to use service discover to collect every pod's business metrics. So that I just read these infomations from service register center, do not care about prometheus gateway. 

But now I want collect every pod's besiness metrcis, and I must to use  prometheus gateway, I feel it's inconvenient.

intfish123:
> It’s this ? https://apisix.apache.org/blog/2022/02/21/nacos-api-gateway/

almost like this

PR #1668, opened 2023-01-10
Add descriptions to cue structs and fields

codegangsta:
This PR adds comments to Cue structs and fields when exporting via cue. This is turn also outputs `description` fields when exporting the cue to and open api spec, which can be useful for editor tools to drive an LSP like experience.

This does, however, greatly increase the size of the output when running `benthos list --format cue`, so I'd be open to moving this to another format directive: `benthos list --formate cue-full` or something like that.

Would love feedback and if y'all think this is a good idea.

### DIFF
--- a/internal/cuegen/cue.go
+++ b/internal/cuegen/cue.go
@@ -2,6 +2,7 @@ package cuegen
 
 import (
 	"fmt"
+	"strings"
 
 	"cuelang.org/go/cue/ast"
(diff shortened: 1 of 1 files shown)


disintegrator (approved):
LGTM

Jeffail (approved):
LGTM, if there's demand for a smaller export we can add `cue-scrubbed` or something.

PR #1669, opened 2023-01-11
Fix aws_sqs backoff reset behavior

cludden:
The current implementation results in artificial delays when reading from a queue that has an average receive message frequency lower than the configured receive wait timeout. For example, if a queue receives, on average, a burst of messages every 15-30m, and this input is configured with long polling enabled (using the max receive wait timeout of 20s), most messages will experience an artificial delay of 5m. This change removes the conditional clause around receiving > 0 messages for resetting the backoff, and instead resets it on every successful poll.

### DIFF
--- a/internal/impl/aws/input_sqs.go
+++ b/internal/impl/aws/input_sqs.go
@@ -215,8 +215,8 @@ func (a *awsSQSReader) readLoop(wg *sync.WaitGroup) {
 	}()
 
 	backoff := backoff.NewExponentialBackOff()
-	backoff.InitialInterval = 100 * time.Millisecond
-	backoff.MaxInterval = 5 * time.Minute
+	backoff.InitialInterval = 10 * time.Millisecond
+	backoff.MaxInterval = 30 * time.Second
 	backoff.MaxElapsedTime = 0
 
 	getMsgs := func() {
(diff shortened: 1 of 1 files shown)


Jeffail:
Hey @cludden, it's probably worth tuning the backoff parameters instead so that the upper and lower intervals are smaller. Polling the queue in a busy loop each time we get zero messages back was costing some users quite a lot relative to the frequency of messages coming through.

I'd be happy to tune the parameters so that the initial interval is maybe 10ms, and the max interval is something more like 30s. Eventually I'll expose them as config parameters for even more control but I'd like to find sensible defaults first.

cludden:
What do you think about guarding the reset with a conditional clause that checks for either >0 messages received or long polling enabled? Something like:

```go
if len(res.Messages) > 0 || a.conf.WaitTimeoutSeconds > 0 {
    backoff.Reset()
}
```

While I'm not opposed to your suggestion of tuning the backoff settings, tt seems counterintuitive to me to both utilize long polling and also introduce exponentially longer artificial delays in between _successful_ (empty or not) responses from long polling queries. I could be misunderstanding the long polling feature, but it seems its intended to minimize the overhead of consumer queries, which seems to overlap with the intention of the backoff functionality here. Thoughts?

Jeffail (approved):
Ahh of course, I forgot about long polling. Yeah that works, thanks @cludden!

PR #1670, opened 2023-01-11
Fix Oracle timestamp marshalling in sql components

### LINKED ISSUE #1640: Oracle SQL driver doesn't return timestamps correctly
This issue is meant to track https://github.com/sijms/go-ora/issues/287 which has all the details. In essence, one can't currently `SELECT` columns of type `TIMESTAMP` directly, since they won't be deserialised correctly in [`sqlRowToMap`](https://github.com/benthosdev/benthos/blob/48c7bb7a0718bdcef453cc156f60f2ffafb48730/internal/impl/sql/util.go#LL48C6-L48C17) and they'll end up containing an empty object (`{}`) instead of the actual timestamp value. I guess a workaround is to call the `TO_CHAR()` function on the `TIMESTAMP` column in the `SELECT` statement until this is addressed somehow.

mihaitodor:
Fixes #1640.

It would be nice to enhance the integration tests to have one of the columns contain a timestamp...


Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #1671, opened 2023-01-11
Add support for NATS credentials loaded via service.FS

codegangsta:
This PR adds support for loading NATS user credentials via `service.FS`, giving the services API another way to bootstrap itself with NATS credentials without having to put it on the filesystem.

I wanted to get this out of the way before I started a more broad rework of the current NATS components to use the services API.

### DIFF
--- a/internal/impl/nats/auth.go
+++ b/internal/impl/nats/auth.go
@@ -1,13 +1,22 @@
 package nats
 
 import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+
 	"github.com/nats-io/nats.go"
+	"github.com/nats-io/nkeys"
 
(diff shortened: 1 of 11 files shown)


internal/impl/nats/auth.go
```diff
@@ -38,3 +56,96 @@ func AuthFromParsedConfig(p *service.ParsedConfig) (c auth.Config, err error) {
...
+		return sig, nil
+	}
+}
+
+// Just wipe slice with 'x', for clearing contents of creds or nkey seed file.
+func wipeSlice(buf []byte) {
+	for i := range buf {
+		buf[i] = 'x'
+	}
+}
```
Jeffail, line 96: Thorough :heart:

Jeffail (approved):
LGTM, thanks @codegangsta!

PR #1673, opened 2023-01-13
Update NATS input and NATS output to use the v4 public services api

codegangsta:
This PR updates the `nats` input and output components to use the new services API. That's about it


Jeffail (approved):
Nice, looks great, thanks @codegangsta! Unfortunately I missed the stream but looks like it went well ❤️

PR #1675, opened 2023-01-14
Fix dump_request_log_level in http components

mihaitodor:
Prevent the logger `roundTripper` from getting overwritten when applying TLS and proxy settings.

This came up on Discord. For both TLS and proxy settings we perform a type assertion for `*http.Transport`, which isn't going to work if the underlying type is `roundTripper`.


Jeffail (approved):
Thanks @mihaitodor! I'll make a note to eventually add some sort of test matrix of all these different settings as they all interact quite violently.

PR #1681, opened 2023-01-17
Add JWT signing support for EdDSA

alexthecarp:
This PR is adding another algorithm to the list of JWT signing algorithms, EdDSA. Slight refactor to use the more generic key type `crypto.PrivateKey`.


Jeffail (approved):
LGTM, thanks @alexthecarp!

PR #1682, opened 2023-01-17
Enhance snowflake_put output

mihaitodor:
- Add fields `file_name`, `file_extension` and `request_id`
- Add interpolation support to the `path` field
- Add ZSTD compression support to the `compression` field
- Add extra examples
- Update github.com/snowflakedb/gosnowflake


Jeffail (approved):
Awesome, thanks @mihaitodor!

PR #1685, opened 2023-01-18
Add POP3 input

mfamador:
(sorry, messed up the previous branch)

This input connects to a POP3 server and reads the emails indefinitely at an interval (by default) or a finite number of times (to be used on a job). After delivering the messages the emails are deleted from the server.

The messages have the email body in their content and multiple metadata fields relative to the email, if present, i.e.:
- Subject
- From
- To
- Cc
- Bcc
- Date
- Received
- Content-Type
- Return-Path
- Delivered-To

Example:
```yml
input:
  label: "pop3"
  pop3:
    host: ${HOST:pop.gmail.com}
    port: ${PORT:995}
    username: ${USERNAME}
    password: ${PASSWORD}
    tls:
      enabled: ${TLS_ENABLED:true}
      skip_cert_verify: ${TLS_SKIP_VERIFY:true}
    pull:
      count: ${PULL_COUNT:0} # loop forever
      interval: ${PULL_INTERVAL:5m} # pull the messages periodically every 5 min
  processors:
  - label: "log"
    log:
      message: "msg"
      fields:
        count: ${! count("msg") }
        content: ${! content() }
        subject: ${! meta("Subject") }
        from: ${! meta("From") }
        to: ${! meta("To") }
        cc: ${! meta("Cc") }
        bcc: ${! meta("Bcc") }
        date: ${! meta("Date") }
        received: ${! meta("Received") }
        contentType: ${! meta("Content-Type") }
        returnPath: ${! meta("Return-Path") }
        deliveredTo: ${! meta("Delivered-To") }
```


internal/impl/mail/input_pop3.go
```diff
@@ -0,0 +1,300 @@
...
+		messageChan:  make(chan *envelope, maxInFlight),
+		done:         make(chan struct{}),
+		ack:          make(chan int, maxInFlight),
+		log:          log,
+	}
+
+	return p, nil
+}
+
+var onlyOnce sync.Once
```
Jeffail, line 200: Looks like this needs to be owned by an individual `*pop`, otherwise multiple instances of this input would be ineffective.

mfamador: Fixed.

internal/impl/mail/input_pop3.go
```diff
@@ -0,0 +1,299 @@
...
+		Field(service.NewStringField(passwordParam).
+			Description("The password to login on the POP3 server.")).
+		Field(service.NewObjectField(tlsParam,
+			service.NewBoolField(tlsEnabledParam).
+				Description(`Whether TLS is enabled.`),
+			service.NewBoolField(tlsSkipCertVerifyParam).
+				Description(`Whether to skip server side certificate verification.`)).Advanced()).
+		Field(service.NewObjectField(pullParam,
+			service.NewIntField(countParam).
+				Description(`An optional number of pulls to fetch messages, if set above 0 the specified number of pulls is executed and then the input will shut down.`).
```
Jeffail, line 73: This is an interesting one, we could achieve the same thing by placing this input within a `read_until` input, not sure it's worth having an explicit field here.

internal/impl/mail/input_pop3.go
```diff
@@ -0,0 +1,299 @@
...
+			Description("The timeout to establish the connection with the POP3 server.").
+			Default("30s").Advanced()).
+		Field(service.NewIntField(maxInFlightParam).
+			Description("The maximum number of messages to have in flight at a given time. Increase this to improve throughput.").
+			Default(1).Advanced()).
+		Field(service.NewIntField(maxBodySizeParam).
+			Description("The max number of bytes when reading the message body.").
+			Default(4096).Advanced()).
+		Field(service.NewIntField(listPageSizeParam).
+			Description("The max number of emails returned by the POP3 server. This limit may be configured on the server or may be imposed by the server software.").
```
Jeffail, line 88: Does this mean the value must match what the server has configured? Or just that it's optional as the server could be specifying it instead?

internal/impl/mail/input_pop3.go
```diff
@@ -0,0 +1,299 @@
...
+			p.log.Errorf("error retrieving email %d: %v", m.ID, err)
+		}
+		body := make([]byte, p.maxBodySize)
+		n, err := entity.Body.Read(body)
+		if err != nil {
+			p.log.Debugf("error reading email body %d: %v", m.ID, err)
+		}
+		message := service.NewMessage(body[:n])
+		for k, v := range entity.Header.Map() {
+			message.MetaSet(k, v[0])
```
Jeffail, line 259: Rather than use the keys exactly as they are I'm considering whether we should match the style of keys from other inputs, so it'd be `pop3_return_path` rather than `Return-Path`.

internal/impl/mail/input_pop3.go
```diff
@@ -0,0 +1,299 @@
...
+					p.log.Errorf("error acknowledging email %d: %v", ID, err)
+				}
+				wg.Done()
+			case <-acked:
+				return
+			}
+		}
+	}()
+	wg.Wait()
+	acked <- true
```
Jeffail, line 279: The stuff above is kind of difficult to get my head around, lots of channels doing stuff to keep track of.

Jeffail (commented):
Pretty sure we're all good in terms of behaviour but left a few comments around config options. Also not sure I like the number of channels being used here, it's making it difficult to understand how everythings working together which is a bit risky, since that also seems to be an important factor due to the connections lifetime being tied to a single flush of data.

I'd like to have a go at reworking this to maybe use more synchronous calls within the Read method. However, we really need some form of testing otherwise I'll have no confidence that I'm getting it right.

PR #1687, opened 2023-01-19
Add a NATS error handler option when connecting to NATS

albinou:
When a NATS error occurs, it was not logged using the logrus formatter. This commits fixes it.

The `nats_stream` input and output component implementation is still using the old `Modular` logger, but I guess it will be fixed later when applying a change like #1673.


Jeffail (approved):
Nice and clean, thanks @albinou!

PR #1692, opened 2023-01-23
Add support for `x-death` AMQP headers

ehcaning:
This is a quick solution for handling `x-death` headers.

[Official doc](https://www.rabbitmq.com/dlx.html#effects):
image

It will be available in meta: `meta("x_death_0_count")`

Please let me know if I can add any more details.


Jeffail (approved):
LGTM, thanks @ehcaning!

PR #1697, opened 2023-01-25
Add `timestamp_unix_milli` function

mfamador:
I've added the function `timestamp_unix_milli` to allow us to have the current unix timestamp in milliseconds.


Jeffail (approved):
LGTM, thanks @mfamador!

PR #1698, opened 2023-01-25
Add `ts_unix_milli` method

mfamador:
I've added the `ts_unix_milli` method to allow us to get the unix timestamp with millisecond precision without needing to perform any additional operation.

### DIFF
--- a/internal/impl/pure/bloblang_time.go
+++ b/internal/impl/pure/bloblang_time.go
@@ -451,6 +451,37 @@ The output format is defined by showing how the reference time, defined to be Mo
 		panic(err)
 	}
 
+	formatTSUnixMilliSpec := bloblang.NewPluginSpec().
+		Category(query.MethodCategoryTime).
+		Beta().
+		Static().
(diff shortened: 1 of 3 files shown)


internal/impl/pure/bloblang_time.go
```diff
@@ -451,6 +451,37 @@ The output format is defined by showing how the reference time, defined to be Mo
...
+		return bloblang.TimestampMethod(func(target time.Time) (any, error) {
+			return target.UnixMilli(), nil
+		}), nil
+	}
+
+	if err := bloblang.RegisterMethodV2("ts_unix_milli", formatTSUnixMilliSpec, formatTSUnixMilliCtor); err != nil {
+		panic(err)
+	}
+
+	if err := bloblang.RegisterMethodV2("format_timestamp_unix_milli", formatTSUnixMilliSpecDep, formatTSUnixMilliCtor); err != nil {
```
mfamador, line 481: It's deprecated but some could still want to use this naming

Jeffail (approved):
LGTM, it looks odd but I agree it's fair to add the deprecated variant as well.

PR #1700, opened 2023-01-25
feat: support configuring flow control for pubsub

disintegrator:
The PubSub publisher client maintains an internal message buffer for each topic whose capacity is controlled by flow control settings. This change exposes these as Benthos config fields on the gcp_pubsub output.

### DIFF
--- a/internal/component/output/config_gcp_pubsub.go
+++ b/internal/component/output/config_gcp_pubsub.go
@@ -4,6 +4,25 @@ import (
 	"github.com/benthosdev/benthos/v4/internal/metadata"
 )
 
+// GCPPubSubFlowControlConfig configures a flow control policy of the PubSub
+// client. This affects the internal buffering mechanism that the client manages
+// for each topic when publishing messages.
+type GCPPubSubFlowControlConfig struct {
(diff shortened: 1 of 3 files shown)


disintegrator:
Context: we've encountered intermittent `context.DeadlineExceeded` errors in a pipeline publishing about 1500 messages/s with default settings on the pubsub client. We think the underlying issue is that pubsub client is not blocking publish calls when its internal buffers are full. Instead it continues expanding the buffer and scheduling future work. If enough of a backlog develops, it may translate into context timeouts - each publish call returns a "future" value that is bound to a context with finite lifetime.

This is all conjecture at this stage and we are running more tests to isolate our issue. Nevertheless, it's worthwhile expose the config options presented in this PR.

This issue on the client's github repo may also be relevant: https://github.com/googleapis/google-cloud-go/issues/2831

Jeffail (approved):
LGTM, if I understood correctly then `HasDefault("ignore")` doesn't mean by default we lose data, just that by default memory go boom or something. I'll tweak the docs a little bit to make that clearer.

disintegrator:
@Jeffail unfortunately the package docs aren't very helpful. "ignore" appears to disable buffer size limits.

https://pkg.go.dev/cloud.google.com/go/pubsub#LimitExceededBehavior

PR #1701, opened 2023-01-26
Optimize cache hot-path in Ristretto.

nibbleshift:
Each time cache operations are invoked we pool.Get() a backoff.BackOff and use a defer func to reset the
BackOff and then place it back in the pool using a deferred func.

Currently the BackOff is only necessary if a cache operation fails and the backoff needs to wait and attempt a retry.  This commit moves the Pool.Get() and the defer to a conditional in the failure path.

This completely eliminates the need for the pool.Get, the defer that does backoff.Reset and Pool.Put in success conditions.


Jeffail:
Nice, do you have any performance numbers for the networked components? Unless there's a significant boost then I'd be tempted to reduce these changes down just to the ristretto cache in favour of keeping the code a bit tidier, I'd wager that as soon as networked calls are involved the cost of obtaining a backoff becomes less noticable.

nibbleshift:
I don't have the tests handy, but the ristretto was the biggest improvement, but I did see an improvement in redis access.

Not having the deferred function and the mutex accesses in the hot-path did improve the overall throughput.  If you'd like, I can spend some time getting some benchmarks together for memcache and redis.

nibbleshift:
Left main branch, right remove mutexs/pools from cache hot path.  Test uses redis for the cache.  Performing only Get operations.
image

nibbleshift:
Updated code to only optimize hot-path in the ristretto cache.

Jeffail (approved):
LGTM, thanks @nibbleshift!

PR #1703, opened 2023-01-26
Update cache propagation between levels and optimize.

nibbleshift:
I've noticed some cache inconsistency between levels that causes keys to not get propagated between the levels.  This aims to approve the cache consistency and also defer some of the cache updates to go routines to improve the performance of Get, Set, Add operations.

### DIFF
--- a/internal/impl/pure/cache_multilevel.go
+++ b/internal/impl/pure/cache_multilevel.go
@@ -88,10 +88,11 @@ func newMultilevelCache(levels []string, mgr cacheProvider, log *service.Logger)
 
 //------------------------------------------------------------------------------
 
-func (l *multilevelCache) setUpToLevelPassive(ctx context.Context, i int, key string, value []byte) {
-	for j, name := range l.caches {
-		if j == i {
-			break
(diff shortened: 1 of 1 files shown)


Jeffail:
Hey @nibbleshift, can you give a summary of what the behaviour currently is versus what you've changed it to? I'm not entirely sure whether this is an implementation bug or maybe a behaviour we should document. For example I'm not convinced that it's useful given caches a, b, c for us to guarantee c is hydrated with data obtained from b during a get request, as the main use for a multilevel cache is in cases where you'd assume b is a subset of c and a is a subset of b (and each level is increasingly expensive performance wise).

Also we can't run the hydration on a separate goroutine as a hot config could easily in this case flood the service with unbounded dangling set requests. Having the lower tier caches populated with acquired values during a Get is a chosen performance hit with a multilevel cache.

nibbleshift:
I ended up going a different route and implemented a front/back cache pattern that supports non-blocking writes \w the ability to apply back pressure.  it ends up performing significantly better for my use case.

The multi-level cache doesn't quite fit the problem I was attempting to solve.  We can close this out.

PR #1706, opened 2023-01-28
Add OTLP tracer default tags

### LINKED ISSUE #1695: No way to set service.name on the trace provider when using the open_telemetry_collector
Hi, Currently is not possible to set the service.name to change the name from the default **OTLPResourceNoServiceName** when using the open_telemetry_collector.

Thanks!

mihaitodor:
Fixes #1695.

This adds 2 default tags to the `open_telemetry_collector` tracer: `service.name` and `service.version` (which can still be overridden via the open_telemetry_collector.tags configuration field.

Based on `jaeger` tracer fix from #1410.

### DIFF
--- a/internal/impl/otlp/tracer_otlp.go
+++ b/internal/impl/otlp/tracer_otlp.go
@@ -15,6 +15,7 @@ import (
 	tracesdk "go.opentelemetry.io/otel/sdk/trace"
 	"go.opentelemetry.io/otel/trace"
 
+	"github.com/benthosdev/benthos/v4/internal/cli"
 	"github.com/benthosdev/benthos/v4/public/service"
 )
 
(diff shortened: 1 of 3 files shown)


internal/impl/otlp/tracer_otlp.go
```diff
@@ -122,6 +123,14 @@ func newOtlp(config *otlp) (trace.TracerProvider, error) {
 		attrs = append(attrs, attribute.String(k, v))
 	}
 
+	if _, ok := config.tags[string(semconv.ServiceNameKey)]; !ok {
+		attrs = append(attrs, semconv.ServiceNameKey.String("benthos"))
+	}
+
+	if _, ok := config.tags[string(semconv.ServiceVersionKey)]; !ok {
+		attrs = append(attrs, semconv.ServiceVersionKey.String(cli.Version))
+	}
```
disintegrator, line 132: I think the default version should only be set if the default service name is being set. This would cover situations where user's only set service name.

Also, does it work to pass these as default tags so they're reflected in the docs? 👇 

```
		Field(service.NewStringMapField("tags").
			Description("A map of tags to add to all tracing spans.").
			Default(map[string]string{
				string(semconv.ServiceNameKey): "benthos",
				string(semconv.ServiceVersionKey): cli.Version,
            }).
Advanced())
```

Jeffail: Having them as an explicit config default gets a bit weird, it'd mean the docs change after every tagged release, also means setting any tag would wipe away the defaults which we might not want.

mihaitodor: @disintegrator Given Ash's remark, I addressed this part: "they should accidentally get the benthos version as an additional tag". Guess you meant "shouldn't".

I also changed the logic for the Jaeger tracer, which I guess is a bit of a breaking change, but I doubt anyone will get tripped up by that...

disintegrator: @mihaitodor yep that’s a typo 🤦 
Fixed…

Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #1713, opened 2023-02-02
Enable nacos service registry 3

intfish123:
feat: Enable to register self to nacos service registry center. In order to better collect service metrics with running in docker mode, or for more functions in futures.

In more cases, we need to use service discover to collect every pod's business metrics. So that I just read these infomations from service register center, do not care about prometheus gateway.

But now I want collect every pod's besiness metrcis, and I must to use prometheus gateway, I feel it's inconvenient.


intfish123:
I fix some lint problems, hope you review and merge it, or edit it if necessary.

PR #1715, opened 2023-02-02
add a new metadata entry to kafka input

fouadkada:
# Context

In a Kafka topic that is configured with a cleanup policy of `compact`, a producer can produce message with a `null` value to flag that on the next topic compaction, this message will be deleted from the topic.

# Description

By adding this new metadata, one does not have to check in the pipeline if the `content().length() == 0` to be able to determine if a message is a tombstone message or not. This tidies thing a tad bit

### DIFF
--- a/internal/impl/kafka/input_kafka_franz.go
+++ b/internal/impl/kafka/input_kafka_franz.go
@@ -44,6 +44,7 @@ This input adds the following metadata fields to each message:
 - kafka_partition
 - kafka_offset
 - kafka_timestamp_unix
+- kafka_tombstone_message
 - All record headers
 ` + "```" + `
 `).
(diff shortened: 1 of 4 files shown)


Jeffail:
Hey @fouadkada, looks good, I'm wondering if we should maybe only set this field when the message is a tombstone since it's redundant to have the majority of messages carry `tombstone: false`.

fouadkada:
hey @Jeffail , i'm fine with both approaches...The first school of thoughts would be that it's always available with either a value of `true` or `false` and the other school of thoughts would be that it'll be available when needed with the value `true`.

Let me know, if for consistency's sake, you prefer one approach over the other.

Jeffail (approved):
LGTM, thanks @fouadkada!

PR #1720, opened 2023-02-03
bloblang: add `compare_bcrypt` string method

disintegrator:
This method is useful for checking input strings against bcrypt-hashed secrets. Equivalent methods for scrypt and argon2 may be useful in the future as well.


Jeffail (approved):
LGTM, especially the bit where you drop a soul crushing truth, thanks @disintegrator!

PR #1721, opened 2023-02-03
Add new s3 input sqs config to drop sqs message on specific errors

reflection:
We have gzipped objects in our S3 bucket where some may be corrupted (`unexpected EOF` error).  Benthos will requeue these, but in our use case these objects will not be "uncorrupted" later, so we need a configuration to drop rather than requeue these sqs messages.


Jeffail:
Hey @reflection, I like the concept and I think this is useful. I'm still thinking through alternatives to having an explicit list of errors for cases where they might be many but similar, maybe a bloblang mapping that resolves to a boolean but I'm not sure yet (sorry George), leave this one with me for now.

Jeffail:
Hey @reflection, I'm actually considering implementing this at the output level, where maybe we can expand https://www.benthos.dev/docs/components/outputs/drop_on to have an `error_matches` field, where the error is ignored and messages acked when the error matches one of a given list of patterns.

Jeffail:
Finally got around to this: https://github.com/benthosdev/benthos/commit/5dd8b9d06b0268a5e3bf60f375a6120e359e6054

PR #1723, opened 2023-02-04
bloblang: add `compare_argon2` string method

disintegrator:
This method is useful for checking input strings against argon2i/argon2id-hashed secrets.


Jeffail (approved):
LGTM, thanks @disintegrator

PR #1725, opened 2023-02-05
Add `ts_unix_micro` method

mfamador:
Added the missing `ts_unix_micro` method to allow us to get the epoch unix timestamp with microsecond precision without needing to perform any additional operation.


Jeffail (approved):
LGTM, thanks @mfamador

PR #1726, opened 2023-02-05
Add `timestamp_unix_micro` function

mfamador:
Added the missing function `timestamp_unix_micro` to allow us to have the current epoch unix timestamp with microsecond precision.


Jeffail (approved):
LGTM, thanks @mfamador!

PR #1735, opened 2023-02-09
bloblang: add hmac-sha jwt signing methods

disintegrator:
This change adds three bloblang methods for signing objects that represent JWT claims using HMAC-SHA. These methods are:

- `sign_jwt_hs256(signing_secret)`
- `sign_jwt_hs384(signing_secret)`
- `sign_jwt_hs512(signing_secret)`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ All notable changes to this project will be documented in this file.
 - JWT based HTTP authentication now supports `EdDSA`.
 - New `flow_control` fields added to the `gcp_pubsub` output.
 - Added bloblang methods `sign_jwt_hs256`, `sign_jwt_hs384` and `sign_jwt_hs512`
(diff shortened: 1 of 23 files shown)


Jeffail (approved):
LGTM, one thing we could consider is whether to make the algorithm a parameter similar to the hash method, but given this is crypto stuff maybe it's better to have isolated methods.

Jeffail (approved):

disintegrator:
@Jeffail I had a similar thought process and opted for the explicit method names. It seemed harder to fumble these both from usage and implementation (didn't need to consider if user entered arbitrary algorithm in the code). Documentation was also simpler compared to golang-jwt docs below:
image

If those aren't compelling reasons then I'd be happy to move to a signature like so:

```
sign_jwt(algorithm, signing_key) 
```

PR #1736, opened 2023-02-09
[cassandra output] Mitigate "gocql: no hosts available in the pool" cannot recover without restart 

mfamador:
To mitigate an issue (https://github.com/gocql/gocql/issues/915 and https://github.com/gocql/gocql/issues/831) with `gocql` when Cassandra's IP address resolves differently or on some particular network glitches, the session does not recover and there's the need to kill the container to be able to write again, an Observer was introduced to monitor and recreate the session when needed.

We're hitting this issue for a while now, and since we're using a custom benthos image with this "band-aid" it never happened again, it's being able to recover by itself.

### DIFF
--- a/internal/impl/cassandra/output.go
+++ b/internal/impl/cassandra/output.go
@@ -146,6 +146,7 @@ type cassandraWriter struct {
 	backoffMax time.Duration
 
 	session  *gocql.Session
+	conn     *gocql.ClusterConfig
 	connLock sync.RWMutex
 
 	argsMapping *mapping.Executor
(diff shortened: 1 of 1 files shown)


mfamador:
Converted it do draft, there's still a discussion going on in what would be the best fix, maybe on `gocql`.

mfamador:
This "band-aid" is the only thing that's been avoiding the issue to happen so far. I'll keep an eye on `gocql`discussions but both open PRs didn't solve it.

mfamador:
After a week it's getting stuck again, not in the same place, though.

PR #1740, opened 2023-02-11
Enhance integer parsing methods

mihaitodor:
Allow the `int64()`, `int32()`, `uint64()` and `uint32()` bloblang methods to infer the number base when parsing strings.

### DIFF
--- a/internal/bloblang/query/type_helpers.go
+++ b/internal/bloblang/query/type_helpers.go
@@ -422,9 +422,9 @@ func IToInt(v any) (int64, error) {
 	case json.Number:
 		return t.Int64()
 	case []byte:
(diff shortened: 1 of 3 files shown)


nibbleshift:
What are your thoughts on allowing the caller to set the base explicitly?  this.field.int64(10) or this.field.int64(2)

Jeffail (approved):
LGTM, thanks @mihaitodor. I think I agree with @nibbleshift that it might be nice to allow an explicit base but we could add that in a subsequent PR with an optional parameter.

mihaitodor:
I thought of that, but it would be for somewhat niche use cases. Feel free to send a PR @nibbleshift

nibbleshift:
Will do.

PR #1741, opened 2023-02-11
Use optimized compression module (klauspost/compress)

nibbleshift:
Use klauspost/compress drop-in replacements for:

- snappy
- gzip
- flate
- zlib

Benchmarks for gzip comparing compress/gzip to implementation in klauspost/compress:
image


Jeffail (approved):
LGTM, thanks @nibbleshift!

PR #1742, opened 2023-02-11
Adding the Socket_Syslog input component

xenbyte:
After talking to @Jeffail on Discord about adding the datagram unix domain socket input, it was decided this type would be a **completely new component** and it won't interfere with the already existing `server_socket`. 
Since in most 3rd party services (nginx, PowerDNS) they all send their logs in the syslog format to a datagram unix domain socket, I thought adding a new component that does exactly that can be beneficial to a lot of people (including me).
There won't be any further need to parse incoming syslog messages, this input will handle that.


internal/impl/io/input_socket_syslog.go
```diff
@@ -0,0 +1,252 @@
...
+		if err := os.Remove(p.address); err != nil {
+			p.log.Errorf("Couldn't create socket: %v", err)
+		}
+	}
+	conn, err := net.Listen(p.socketType, p.address)
+	if err != nil {
+		return nil, err
+	}
+
+	c := make(chan os.Signal, 1)
```
Jeffail, line 113: We can't really allow this within a component implementation since in set ups such as streams mode we need to be able to dynamically spin up and tear down components potentially multiple times. Ideally we want a mechanism for `Close` to trigger this listener clean up logic.

internal/impl/io/input_socket_syslog.go
```diff
@@ -0,0 +1,252 @@
...
+			return "", err
+		}
+		message := fmt.Sprintf("%v", m.Dump()["content"])
+		return message, nil
+	case "rfc3164":
+		m := rfc3164.NewParser(buf)
+		if err := m.Parse(); err != nil {
+			return "", err
+		}
+		message := fmt.Sprintf("%v", m.Dump()["content"])
```
Jeffail, line 201: Would it make sense to pull out other bits of data and add them as metadata fields? In which case this method will need to return a message type instead.

Jeffail (commented):
Hey @xenbyte, mostly looks good but one comment that needs addressing and one minor nit, extracting metadata fields is something we can do at a later time if needed, just thought it'd be cool. The shutdown stuff definitely needs changing though.

xenbyte:
Unfortunately I couldn't find a solution to [this](https://github.com/benthosdev/benthos/pull/1742#discussion_r1155309095) issue the first couple weeks working on this and totally forgot about. I'm closing this issue. Will try an alternate solution if I can.

PR #1745, opened 2023-02-12
Add go 1.20.x to matrix

nibbleshift:
Add 1.20.x to matrix in CI/CD test.


peczenyj:
We can consider deprecate go 1.18.x ?

PR #1748, opened 2023-02-13
Narrow locking in ResultStore.Add() and drop defer.

nibbleshift:
It's not necessary to acquire the lock while assembling the new batch.  The lock is only needed when updating r.payloads.


Jeffail (approved):
LGTM, thanks @nibbleshift!

PR #1750, opened 2023-02-13
wip: Optionally fail when env vars are unset

wpcarro:
Opt-into failing at the boundaries of the application by setting the `--strict-env` flag.


wpcarro:
Currently blocked on not knowing the best way to pass around the CLI configuration and apply the `c.Bool("strict-env")` value to the `ReplaceEnvVariables` call. The good news is that in some local tests, `ReplaceEnvVariables` works as expected (when the condition is hard-coded).

Also left an open TODO as a comment for discussion.

wpcarro:
Also not sure what the recommended way to throw an error in go-land was, so I did `panic`, but I don't think that's what we want. Forgive the noob

Jeffail:
Hey @wpcarro I decided to rework this into a linting error: https://github.com/benthosdev/benthos/commit/15e107cb34f4328179916182deced9acfcbf4970 and https://github.com/benthosdev/benthos/commit/82f72713b2e2dbce648524902a1eb30b3137de52, thanks!

wpcarro:
oh cool - thanks :)

PR #1753, opened 2023-02-14
number_type() bloblang method

loicalleyne:
"Returns the type of a number value as a string, providing one of the following values: `int`, `uint`, `int8`, `uint8`, `int16`, `uint16`, `uint32`, `uint64`, `float32`, `float64` or `null`.",


loicalleyne:
The tests that failed were due to me not having figured out how to do the PR directly from my local repo, omiting an import in my github fork, and some embarrassing manual editing in github to avoid losing my work on my local machine.

Jeffail:
Hey @loicalleyne I think rather than a separate method it might make more sense to add customisation fields to the `type` method so that you can do something like `root.foo_type = this.foo.type(numerical_types: true)`. However, there's also the option to do `root.type = "%T".format(this.foo)` which I think makes more sense here as you're specifically aiming to access the raw Go type under the hood.

PR #1758, opened 2023-02-19
Add bloblang `trim_prefix` and `trim_suffix` string methods

mihaitodor:


internal/bloblang/query/methods_strings.go
```diff
@@ -1940,3 +1940,63 @@ root.description = this.description.trim()`,
...
+root.description = this.description.trim_prefix("foobar_")`,
+			`{"description":"unchanged","name":"foobar_blobton"}`,
+			`{"description":"unchanged","name":"blobton"}`,
+		),
+	).Param(ParamString("prefix", "The leading prefix substring to trim from the string.")),
+	func(args *ParsedParams) (simpleMethod, error) {
+		prefix, err := args.FieldString("prefix")
+		if err != nil {
+			return nil, err
+		}
```
nibbleshift, line 1961: You could add byte conversion here and avoid the conversion each execution:
```
bytesPrefix = []byte(prefix)
```

internal/bloblang/query/methods_strings.go
```diff
@@ -1940,3 +1940,63 @@ root.description = this.description.trim()`,
...
+root.description = this.description.trim_suffix("_foobar")`,
+			`{"description":"unchanged","name":"blobton_foobar"}`,
+			`{"description":"unchanged","name":"blobton"}`,
+		),
+	).Param(ParamString("suffix", "The trailing suffix substring to trim from the string.")),
+	func(args *ParsedParams) (simpleMethod, error) {
+		suffix, err := args.FieldString("suffix")
+		if err != nil {
+			return nil, err
+		}
```
nibbleshift, line 1991: You could add byte conversion here and avoid the conversion each execution:
```
bytesSuffix := []byte(suffix)
```

internal/bloblang/query/methods_strings.go
mihaitodor: Good point, thanks! ❤️

Jeffail (approved):
LGTM, thanks @mihaitodor!

Issue #1768, opened 2023-02-23
benthos panics when flooded by nats messages


pcosquer:
Hello,

I'm running benthos with a nats input:

```yaml
input:
  nats:
    urls:
      - nats://localhost:4222
    subject: ">"
pipeline:
  processors:
    - bloblang: |
        root.event = this.catch(deleted())
        root.subject = meta("nats_subject")
output:
  label: ""
  stdout:
    codec: lines
```

Nats is running in a nats container

```
docker run -p 4222:4222 -p 8222:8222 -p 6222:6222 --rm -ti nats:2.9.0 -js 
```

When running this go program, benthos panics

```go
package main

import "github.com/nats-io/nats.go"

func main() {
	nc, _ := nats.Connect(nats.DefaultURL)

	for i := 0; i < 500; i++ {
		nc.Publish("foo", []byte("Hello World"))
	}

	nc.Drain()
	nc.Close()
}
```

```
 ./target/bin/benthos -c config.yaml
INFO Running main config from specified file       @service=benthos path=config.yaml
INFO Receiving NATS messages from subject: >       @service=benthos label="" path=root.input
INFO Launching a benthos instance, use CTRL+C to close  @service=benthos
INFO Listening for HTTP requests at: http://0.0.0.0:4195  @service=benthos
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x20a3523]

goroutine 76 [running]:
github.com/nats-io/nats%2ego.(*Subscription).ConsumerInfo(0xc000baa4f0?)
        /home/pcosquer/go/pkg/mod/github.com/nats-io/nats.go@v1.15.0/js.go:2386 +0x43
github.com/benthosdev/benthos/v4/internal/impl/nats.errorHandlerOption.func1(0x0?, 0xc0012a56c0, {0x43ce2a0?, 0xc0006f3a40})
        /home/pcosquer/dev/benthos/internal/impl/nats/errors.go:17 +0x174
github.com/nats-io/nats%2ego.(*Conn).processMsg.func2()
        /home/pcosquer/go/pkg/mod/github.com/nats-io/nats.go@v1.15.0/nats.go:2974 +0x36
github.com/nats-io/nats%2ego.(*asyncCallbacksHandler).asyncCBDispatcher(0xc0015b2400)
[...]

PR #1704, opened 2023-01-28
Add support for NATS KV Processor/Input/Output. Add default NATS urls

codegangsta:
This PR goes out to all the blobs out there who believed there would come a day where Benthos would have more NATS support:

- 🎉 **NATS KV Input** - it watches for updates on keys, isn't that neat?!
- 🎉 **NATS KV Output** - It puts your keys **and** values into a bucket!
- 🎉 **NATS KV Processor** - This thing `puts`, `gets`, `deletes`, `creates`, `updates`, `deletes`, slices, dices and even folds your laundry (if you are so inclined)

We got **docs!** 😮

We got **integration tests!** 😄 

We even got **default urls for NATS components!** 🤷 (for real though, We like using the default URL in all our NATS tool, since NATS always kicks itself off on `localhost:4222`, I think this makes it a wee bit easier to get started with NATS and Benthos)

Go ahead, tear it apart.... I dare you.

Bring it el jeffe! (aka @Jeffail )

And if you are really nice I'll write a Cache next

PR #1769, opened 2023-02-23
Update NATS client libs

mihaitodor:
This fixes the panic reported in #1768 which is caused by a regression introduced in #1687 (45f785a).

#1704 also brings in this updated version of the NATS client libs, so if that can be merged, feel free to ignore this one. I haven't run the integration tests, so I'm not sure if there are any other regressions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


Jeffail (approved):
LGTM, I'll get a patch release out today. Thanks @mihaitodor and @pcosquer for the quick investigation :heart:

PR #1770, opened 2023-02-23
Add asdf to the Changelog

mihaitodor:
Looks like we forgot to mention this for v4.12.0... Does it make sense to update the notes postmortem or would it be better to mention this for the next release?


Jeffail:
I'd typically not add something like this to the CHANGELOG as it wasn't a change made to Benthos itself and as far as I'm aware it's not specific to a version of Benthos, other than this happens to be the first version that it's coincided with.

mihaitodor:
Good point, people who use asdf will find it anyway

PR #1773, opened 2023-02-23
feat: add csv-safe codec

ryanhallcs:
Our use case requires a lot of error handling and tracking of file intake - so we needed a bit more resiliency for parsing failures in the csv codec. While this may or may not be desirable upstream, I still wanted to share our use case and solicit feedback. If it is helpful, even better!

- Adds 'row_number' metadata (starting at 2 for 
- Adds 'row_parse_error' metadata on any error returned from `scanner.Read()`
- Adds 'row_empty' metadata

Errors will not short-circuit this codec which is our intent, the payload in the message part will be an empty `map[string]any` with the relevant metadata so downstream configurations can handle as appropriate


Jeffail (approved):
LGTM, thanks @ryanhallcs!

PR #1774, opened 2023-02-23
feat: add ability to skip a UTF BOM in files

ryanhallcs:
BOMs are difficult to track down and annoying to deal with, especially when dealing with structured data.

Instead of doing this in our benthos configurations:
```
...
          - label: strip_bom
            bloblang: |
              # Some CSVs we get have a UTF-8 byte order mark, which screws up one column header.
              # This will strip byte order marks, and should have no effect if there wasn't one.
              let bom = "EFBBBF".decode("hex")
              root = this.map_each_key(key -> key.replace($bom, ""))
...
```

We preferred to have it handled at the codec level and happily process our files without any ticking time bombs lurking in our bytes

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,6 +34,7 @@ require (
 	github.com/couchbase/gocb/v2 v2.6.0
 	github.com/denisenkom/go-mssqldb v0.12.3
 	github.com/dgraph-io/ristretto v0.1.1
+	github.com/dimchansky/utfbom v1.1.1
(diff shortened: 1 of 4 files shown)


Jeffail:
Merged via: https://github.com/benthosdev/benthos/commit/31049847779559e08b5bded6a77145bf64c9c2ad, I rewrote it to use a new component that does the skipping as we might want to reuse it for other stuff.

PR #1776, opened 2023-02-27
Fix the type of the messages returned by the sequence input

### LINKED ISSUE #1775: Sequence input component gives messages of type "unknown"
If I use the example provided in documentation:

```yml
input:
  sequence:
    sharded_join:
      type: full-outer
      id_path: uuid
      merge_strategy: array
    inputs:
      - csv:
          paths:
            - ./hobbies.csv
            - ./main.csv
```

The output result is as expected giving me the 3 messages as indicated:

```json
{"age":"28","name":"Emma","uuid":"BBB"}
{"age":"45","hobby":"pokemon go","name":"Geri","uuid":"CCC"}
{"age":"34","hobby":["rowing","golf"],"name":"Melanie","uuid":"AAA"}
```

Unfortunately these messages are not usable in a pipeline, and are of type `"unknown"`.

For example if I add this mapping:

```yml
input:
  sequence:
    sharded_join:
      type: full-outer
      id_path: uuid
      merge_strategy: array
    inputs:
      - csv:
          paths:
            - ./hobbies.csv
            - ./main.csv

pipeline:
  processors:
    - mapping: |
        root = {}
        root.original = this
        root.original_type = this.type()
        root.name_value = this.name
        root.name_type = this.name.type()
```

I get the following output:

```json
{"name_type":"null","name_value":null,"original":{"age":"28","name":"Emma","uuid":"BBB"},"original_type":"unknown"}
{"name_type":"null","name_value":null,"original":{"age":"34","hobby":["rowing","golf"],"name":"Melanie","uuid":"AAA"},"original_type":"unknown"}
{"name_type":"null","name_value":null,"original":{"age":"45","hobby":"pokemon go","name":"Geri","uuid":"CCC"},"original_type":"unknown"}
```

Basically `this` is accessible, with an `unknown` type but none of it's properties are.

Also, any attempt to add a property to `this`will result in an error: 
```yml
    - mapping: |
        root = this
        root.test = 5
```

Gives:
```
failed to assign result (line 2): unable to set target path original as the value of the root was a non-object type (unknown)
```

mihaitodor:
The code needs to return the underlying Gabs container object.

Fixes #1775.

### DIFF
--- a/internal/impl/pure/input_sequence.go
+++ b/internal/impl/pure/input_sequence.go
@@ -177,7 +177,7 @@ type joinedMessage struct {
 
 func (j *joinedMessage) ToMsg() message.Batch {
 	part := message.NewPart(nil)
-	part.SetStructuredMut(message.CopyJSON(j.fields))
+	part.SetStructuredMut(message.CopyJSON(j.fields.Data()))
 	for k, v := range j.metadata {
 		part.MetaSetMut(k, v)
 	}
--- a/internal/impl/pure/input_sequence_test.go
+++ b/internal/impl/pure/input_sequence_test.go
@@ -255,7 +255,11 @@ func TestSequenceJoinsMergeStrategies(t *testing.T) {
 						break consumeLoop
 					}
 					assert.Equal(t, 1, tran.Payload.Len())
-					act = append(act, string(tran.Payload.Get(0).AsBytes()))
+					m := tran.Payload.Get(0)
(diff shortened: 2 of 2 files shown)


Jeffail (approved):
Ahhhh, nice find. This will be a regression caused by changes to CopyJSON, as it went from doing a really dodgy deep clone on unrecognised types to just naively carrying them over so the old implementation would have masked this problem.

PR #1778, opened 2023-02-28
fix vulnerability GO-2023-1571

peczenyj:
By running the **govulncheck** I found this issue in the benthos vendors and it is easy to fix:

```
Vulnerability #1: GO-2023-1571
  A maliciously crafted HTTP/2 stream could cause excessive CPU
  consumption in the HPACK decoder, sufficient to cause a denial
  of service from a small number of small requests.

  More info: https://pkg.go.dev/vuln/GO-2023-1571

  Module: golang.org/x/net
    Found in: golang.org/x/net@v0.6.0
    Fixed in: golang.org/x/net@v0.7.0
```


Jeffail:
Looks like we've already upgraded the deps in a previous PR but I'll merge for the CHANGELOG update.

PR #1784, opened 2023-03-01
Add support for inline NATS user creds in the configuration

albinou:
This PR allows for the NATS user credentials to be given in the Benthos stream configuration (User JWT + User NKey Seed).

Example:

```yaml
input:
  nats_jetstream:
    urls:
      - "nats://nats.mycloud.tld"
    stream: "input"
    subject: "input"
    auth:
      user_jwt: ""
      user_nkey_seed: ""
output:
  nats_jetstream:
    urls:
      - "nats://nats.mycloud.tld"
    subject: "output"
    auth:
      user_jwt: ""
      user_nkey_seed: ""
```

This is very convenient not to rely on local storage. Please comment if the code needs improvements :smiley:

### DIFF
--- a/internal/impl/nats/auth.go
+++ b/internal/impl/nats/auth.go
@@ -38,6 +38,12 @@ func authConfToOptions(auth auth.Config, fs *service.FS) []nats.Option {
 		))
 	}
 
+	if auth.UserJWT != "" && auth.UserNkeySeed != "" {
+		opts = append(opts, nats.UserJWTAndSeed(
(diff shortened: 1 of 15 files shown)


codegangsta:
Hey @albinou, are you using Benthos directly or the service api?

If using the service API, the NATS components all now support the `FS` interface so you could essentially overlay a filesystem with the proper creds. That may be a bit more secure than this approach

Jeffail (approved):
LGTM, thanks @albinou, I think it's useful to have both options as they'll suit different deployment styles.

PR #1785, opened 2023-03-02
Update Changelog for PR #1776

mihaitodor:
Just so it doesn't fall through the cracks 😅 since I forgot to do it in #1776...


Jeffail (approved):
LGTM

PR #1787, opened 2023-03-04
Fix the return type of the find_all bloblang method

mihaitodor:
I keep messing this stuff up... Maybe it's worth making the tests a bit stricter to assert the type too


Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #1790, opened 2023-03-06
Use AMQP message value if data is empty

tomwphillips:
I couldn't get the integration test to run locally. I tried editing the test to run against my local ActiveMQ instance but had no luck. Hopefully it'll run in CI.

As I mentioned on Discord, sometimes there's data in the `Message.value` attribute. This change pulls that out into message metadata.

I have limited experience with Go so I hope this is ok 😅

### DIFF
--- a/internal/impl/amqp1/input.go
+++ b/internal/impl/amqp1/input.go
@@ -239,6 +239,7 @@ func (a *amqp1Reader) ReadBatch(ctx context.Context) (message.Batch, input.Async
 	}
 
 	part := message.NewPart(amqpMsg.GetData())
+	amqpSetMetadata(part, "amqp_value", amqpMsg.Value)
 	if amqpMsg.Properties != nil {
 		amqpSetMetadata(part, "amqp_content_type", amqpMsg.Properties.ContentType)
 		amqpSetMetadata(part, "amqp_content_encoding", amqpMsg.Properties.ContentEncoding)
--- a/internal/impl/amqp1/input.go
+++ b/internal/impl/amqp1/input.go
@@ -238,7 +238,16 @@ func (a *amqp1Reader) ReadBatch(ctx context.Context) (message.Batch, input.Async
 		return nil, nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


andreas-habel:
Hi @tomwphillips,
we have a similar issue with messages received by Benthos being empty, because the whole data of the message is stored in the `Message.Value` property and `Message.GetData()` returns nothing.
According to the OASIS specification of a message body ...

> 3.2 Message Format - The body consists of one of the following three choices: one or more [data](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-data) sections, one or more [amqp-sequence](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-amqp-sequence) sections, or a single [amqp-value](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-amqp-value) section.

I'd suggest to not put the Value into a separte Meta field, but replace the actual content

```golang
var part *message.Part

if data := amqpMsg.GetData(); data != nil {
   part = message.NewPart(data)
}
if value, ok := amqpMsg.Value.(string); ok && len(value) > 0 {
   part = message.NewPart([]byte(value))
}
```

tomwphillips:
Thanks for bringing this to my attention, @andreas-habel. It looks like a better solution. I will update the PR.

tomwphillips:
Like I [said above](https://github.com/benthosdev/benthos/pull/1790#issue-1611510146), I can't get the tests running locally. I will build and manually test with my local ActiveMQ installation instead.

tomwphillips:
I did some manual testing and it works as expected.

Looks like a maintainer needs to approve the workflow before the integration tests will run in CI.

Jeffail (approved):
Thanks @tomwphillips, couple linting errors that I can fix on my end, I might also get rid of the `amqp_value` metadata key add as it's duplicate data.

PR #1792, opened 2023-03-07
otlp: open_telemetry_collector add insecure options for http and grpc

vishen:
There is some use-cases where you have secure otel collector. To allow for this a new 'insecure' option was added to both the 'http' and 'grpc' configs in 'open_telemetry_collector'.

### DIFF
--- a/internal/impl/otlp/tracer_otlp.go
+++ b/internal/impl/otlp/tracer_otlp.go
@@ -26,11 +26,17 @@ func init() {
 			service.NewURLField("url").
 				Description("The URL of a collector to send tracing events to.").
 				Default("localhost:4318"),
+			service.NewBoolField("insecure").
+				Description("Use HTTP instead of HTTPS.").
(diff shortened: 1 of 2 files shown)


Jeffail:
Hey @vishen, looks like we're already setting `Insecure` on both collector types so for backwards compatibility we likely need to retain `insecure: false` as a default.

Also, I'm a bit confused as to how this interacts with the HTTP collectors, if we're URL based does it not make sense to use the URL to determine TLS, so `http://example.com` versus `https://example.com` rather than using a boolean config flag? I might be missing something related to otel but if that makes sense I'd rather avoid adding the field to the http collectors.

Jeffail:
Closing, looks like another PR added a `secure` field: https://www.benthos.dev/docs/components/tracers/open_telemetry_collector#httpsecure, I'm pretty sure we should be renaming these `url` fields to `address`, will make a note.

PR #1795, opened 2023-03-08
allow --env-file flag multiple times

disintegrator:
This change updates to the --env-file flag definition so that it can be repeated multiple times. This is useful in settings where benthos is supplied with multiple files containing environment variables. For example, a container orchestration framework can inject one file containing application secrets, another file containing database access credentials and there may be a file included in the repository of a benthos project. These can be loaded like so:

```
benthos \
  --env-file ./.env.production \
  --env-file /tmpfs/secrets.env \
  --env-file /tmpfs/db.env \
  --config config.yml
```

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -126,10 +126,10 @@ func Run() {
 			Value:   false,
 			Usage:   "display version info, then exit",
 		},
-		&cli.StringFlag{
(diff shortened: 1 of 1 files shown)


Jeffail (approved):
LGTM, thanks @disintegrator!

PR #1807, opened 2023-03-18
Disable gocritic and re-enable wastedassign

mihaitodor:
- gogritic is very slow (golangci-lint v1.52.0)
- wastedassign received support for generics: https://github.com/golangci/golangci-lint/pull/3689

LE: Looks like the PR liniting time went down from 7 to 2 minutes

### DIFF
--- a/internal/httpclient/client.go
+++ b/internal/httpclient/client.go
@@ -225,8 +225,8 @@ func (h *Client) waitForAccess(ctx context.Context) bool {
 }
 
 // ResponseToBatch attempts to parse an HTTP response into a 2D slice of bytes.
-func (h *Client) ResponseToBatch(res *http.Response) (resMsg message.Batch, err error) {
-	resMsg = message.QuickBatch(nil)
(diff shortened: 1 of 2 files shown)


internal/httpclient/client.go
```diff
@@ -244,25 +244,25 @@ func (h *Client) ResponseToBatch(res *http.Response) (resMsg message.Batch, err
...
 	defer res.Body.Close()
 
 	var mediaType string
 	var params map[string]string
+	var err error
 	if contentType := res.Header.Get("Content-Type"); len(contentType) > 0 {
 		if mediaType, params, err = mime.ParseMediaType(contentType); err != nil {
 			h.log.Warnf("Failed to parse media type from Content-Type header: %v\n", err)
 		}
-		err = nil
```
mihaitodor, line 257: I think it makes more sense for this method to not use named return parameters

Jeffail (approved):
Niccceeeee, thanks for looking into this, that's a massive win ❤️

PR #1814, opened 2023-03-22
treat path config as prefix in http_server input

### LINKED ISSUE #1813: http_server input path matching changed in 4.13.0
The switch from `http.ServeMux` to gorilla's `mux.Router` changed the path matching behaviour in the `http_server` input. Given the following config:

```yaml
input:
  http_server:
    address: 0.0.0.0:3000
    allowed_verbs: [GET]
    path: /

output:
  sync_response: {}
  processors:
    - mapping: |
        root = {
          "message": "ok",
          "path": @http_server_request_path
        }
```

With Benthos 4.12.1:

```
❯ curl localhost:3000/hello/world
{"message":"ok","path":"/hello/world"}
```

With Benthos 4.13.0
```
❯ curl localhost:3000/hello/world
404 page not found
```

I expected the behaviour under 4.12.1 and prior versions to be preserved.

disintegrator:
This change restores the path matching behaviour of the http_server input from Benthos 4.12.1 and prior where the path config was treated as a prefix.

Closes #1813

### DIFF
--- a/internal/impl/io/input_http_server.go
+++ b/internal/impl/io/input_http_server.go
@@ -222,10 +222,10 @@ func newHTTPServerInput(conf input.Config, mgr bundle.NewManagement) (input.Stre
 	wsHdlr := gzipHandler(h.wsHandler)
 	if gMux != nil {
 		if len(h.conf.Path) > 0 {
-			gMux.HandleFunc(h.conf.Path, postHdlr)
+			gMux.PathPrefix(h.conf.Path).Handler(postHdlr)
 		}
 		if len(h.conf.WSPath) > 0 {
-			gMux.HandleFunc(h.conf.WSPath, wsHdlr)
+			gMux.PathPrefix(h.conf.Path).Handler(wsHdlr)
 		}
 	} else {
 		if len(h.conf.Path) > 0 {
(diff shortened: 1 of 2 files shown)


internal/impl/io/input_http_server.go
```diff
@@ -222,10 +222,10 @@ func newHTTPServerInput(conf input.Config, mgr bundle.NewManagement) (input.Stre
 	wsHdlr := gzipHandler(h.wsHandler)
 	if gMux != nil {
 		if len(h.conf.Path) > 0 {
-			gMux.HandleFunc(h.conf.Path, postHdlr)
+			gMux.PathPrefix(h.conf.Path).Handler(postHdlr)
 		}
 		if len(h.conf.WSPath) > 0 {
-			gMux.HandleFunc(h.conf.WSPath, wsHdlr)
+			gMux.PathPrefix(h.conf.Path).Handler(wsHdlr)
```
Jeffail, line 228:
```suggestion
			gMux.PathPrefix(h.conf.WSPath).Handler(wsHdlr)
```

Something odd going on here as the websocket tests should be failing.

Jeffail: Nvm that's because all of the WS tests run through the global server. Still digging into the test failure.

Jeffail:
Got it, the test failure is because rather than spin a custom server that specific test uses `apiRegGorillaMutWrapper`. I've added a test for the custom server and merged with comments above addressed. I've also reflected the same change in the general HTTP server, it's a mild change in behaviour that fixes a previous change in behaviour (back when path params were added to that as well).

Jeffail:
Merged as: https://github.com/benthosdev/benthos/commit/61aa1d860051ced3a6540a30cc3f08b88621fb99

PR #1816, opened 2023-03-23
deps: github.com/bxcodec/faker/v3 -> github.com/go-faker/faker/v4

srenatus:
The bxcodec repo has been archived.


mihaitodor:
Might be worth noting that it will increase the binary size a tiny bit, since it embeds a [155k JSON data file](https://github.com/go-faker/faker/blob/main/misc/addresses-us-1000.min.json) over [here](https://github.com/go-faker/faker/blob/5effebba7fe9e25350305432ffdca5e19151edad/address.go#L12). However, since it adds this address functionality, it would be nice to also expose [`real_address`](https://github.com/go-faker/faker/blob/5effebba7fe9e25350305432ffdca5e19151edad/example_single_fake_data_test.go#L15) (and any other extras?)

Jeffail (approved):
LGTM, thanks @srenatus! The binary size increase is not insignifant but I think it's the lesser evil of using stale deps or having a fork so let's go with this for now.

PR #1817, opened 2023-03-24
add `ulid(encoding, random_src)` bloblang function

disintegrator:
This change adds a custom `ulid(encoding, random_src)` function to bloblang used to generate Universally Unique Lexicographically Sortable Identifiers (ULIDs). This ID format is useful when storing data in databases and is implemented in many languages unlike several other competing ID formats.

More info here:
https://supabase.com/blog/choosing-a-postgres-primary-key
https://blog.daveallie.com/ulid-primary-keys


internal/impl/lang/bloblang.go
```diff
@@ -1,12 +1,17 @@
...
 	"fmt"
+	"io"
 	"strings"
+	"time"
 
 	"github.com/bwmarrin/snowflake"
 	"github.com/bxcodec/faker/v3"
 	"github.com/gosimple/slug"
+	"github.com/oklog/ulid"
+	frand "golang.org/x/exp/rand"
```
disintegrator, line 14: `math/rand.Source` is not safe for use from multiple goroutine. The alternative is to use this package which implements a [LockedSource](https://pkg.go.dev/golang.org/x/exp/rand#LockedSource)

Jeffail (approved):
LGTM, thanks @disintegrator!

PR #1819, opened 2023-03-27
Allow users to omit the snowpipe field of the snowflake_put output

mihaitodor:
~~Is there a reason to not have a default value (even if it's just the datatype default) for optional fields? If I don't use it, then I need to remember to do `if conf.Contains("snowpipe")`, which feels a bit too ceremonious.~~  Ignore that... It makes sense in some cases where we want to have optional flags and such. There's no point to have a "default value" for those.


Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #1821, opened 2023-03-30
add skip_on config to cached processor

### LINKED ISSUE #1804: cached processor with deleted()
I noticed that when using a `cached` processor if one of the `cached` child processors errors and deletes (drops) the message, benthos will error with a message saying: `Child processors failed: child processors resulted in zero messages`. For example, use this config and input:
```json
{ "question": "what is the best stream processing pipeline?", "answer": "benthos" }
```

```yaml
cache_resources:
  - label: dedupe
    memory:
      init_values: {}
input:
  stdin:
    codec: lines
pipeline:
  threads: 1
  processors:
    - branch:
        processors:
          - cached:
              key: 'dedupe::${! this.question }'
              cache: dedupe
              processors:
                # mocking root = if errored() { deleted() }
                - bloblang: root = deleted()
output:
  stdout:
    codec: lines
```

Will result in this output:
```
INFO Running main config from specified file       @service=benthos path=config.yaml
ERRO Child processors failed: child processors resulted in zero messages  @service=benthos label="" path=root.pipeline.processors.0
{ "question": "what is the best stream processing pipeline?", "answer": "benthos" }
```

What's the right way to drop a message under the cached processor (like on an error)? Is the `child processors resulted in zero messages` error expected in this case? Thanks!

disintegrator:
This change adds an optional `skip_on` config field to the `cached` processor. This field takes a bloblang mapping that must evaluate to a boolean value. If it evaluate to `true` then the resulting batch from is not cached. This can be useful when the result is an error and should not be cached:

```yaml
pipeline:
  processors:
    - cached:
        key: "${! this.id }"
        # If the sql lookup fails then do not cache the error.
        skip_on: "errored()"
        processors:
          - sql_select:
              driver: "sqlite"
              dsn: "file:/sample.db"
              table: "cats"
              columns: "name,breed"
              where: "coat = 'tuxedo'"
```

Closes #1804

### DIFF
--- a/internal/impl/pure/processor_cached.go
+++ b/internal/impl/pure/processor_cached.go
@@ -10,6 +10,7 @@ import (
 
 	"golang.org/x/sync/errgroup"
 
+	"github.com/benthosdev/benthos/v4/public/bloblang"
 	"github.com/benthosdev/benthos/v4/public/service"
 )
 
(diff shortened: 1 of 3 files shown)


internal/impl/pure/processor_cached.go
```diff
@@ -159,29 +174,39 @@ func (proc *cachedProcessor) Process(ctx context.Context, msg *service.Message)
 		return nil, err
 	}
 
+	shouldCache := true
 	var collapsedBatch service.MessageBatch
 	for _, b := range resultBatch {
+		skip, err := shouldSkip(b, proc.skipOn)
+		if err != nil {
```
Jeffail, line 181: I'm going to change this so that it's a soft-error where we log and set `shouldCache = false`, as this seems similar to the actual cache write errors where we treat them as annoyances where we ought to carry on.

Jeffail (approved):
LGTM, thanks @disintegrator, Just going to make a minor change to error handling.

PR #1826, opened 2023-04-01
improve error message in methods and functions

disintegrator:
This change improves error messages in methods and functions particularly when invoked with dynamic arguments. Previously the error message did not highlight the affect method/function name and the argument name didn't quote stand out in the message.

Before:

```
failed to extract input arg path:
  wrong argument type, expected string, got null
```

After:

```
method 'get':
  failed to extract input arg 'path':
    wrong argument type, expected string, got null
```


Jeffail (approved):
LGTM, thanks @disintegrator!

PR #1831, opened 2023-04-03
Add support for configuring GCP Credentials as json (from env variable)

pradeeppattamatta:
Changes to add this feature  - https://github.com/benthosdev/benthos/issues/1761


pradeeppattamatta:
Hi @Jeffail Can you please take a look at this PR and merge if it looks good. Let me know your thoughts, if any changes are needed. Thank you!

internal/impl/gcp/cache_cloud_storage.go
```diff
@@ -19,7 +22,9 @@ func gcpCloudStorageCacheConfig() *service.ConfigSpec {
 		Field(service.NewStringField("bucket").
 			Description("The Google Cloud Storage bucket to store items in.")).
 		Field(service.NewStringField("content_type").
-			Description("Optional field to explicitly set the Content-Type.").Optional())
+			Description("Optional field to explicitly set the Content-Type.").Optional()).
+		Field(service.NewStringField("credentials_json").
+			Description("An optional field to set Google Service Account Credentials json as base64 encoded string.").Optional().Secret().Default(""))
```
Jeffail, line 27: Is base64 encoding these credentials a standard practice for gcp creds? If so is there any documentation explaining it? It seems like an odd extra step and maybe a bit confusing when the field is called `_json`.

internal/impl/gcp/cache_cloud_storage.go
```diff
@@ -60,6 +70,25 @@ func newGcpCloudStorageCacheFromConfig(parsedConf *service.ParsedConfig) (*gcpCl
 	}, nil
 }
 
+func getClientOptionsForCloudStorageCache(parsedConf *service.ParsedConfig) ([]option.ClientOption, error) {
```
Jeffail, line 73: Having a unique `getClientOptionsForXXX` seems a bit error prone as they all have to reproduce the credentials logic, could we collapse this into a single `getAuthClientOptions(parsedConf *service.ParsedConfig)` that they all share?

Jeffail (changes requested):
Thanks @pradeeppattamatta, mostly good, just left a couple of comments.

internal/impl/gcp/cache_cloud_storage.go
pradeeppattamatta: @Jeffail  since the credential is a multi-line json and has all type of characters, using the Base64 encoding of it would avoid the need to remove line breaks and use escape characters. Added in the documentation at all places to make this expectation clear. If you prefer, I can change the attribute from `credentials_json` to `credentials_json_enc` or something similar to make it more clear.

mihaitodor: You can also use the yaml raw string features to avoid those pitfalls:
```yaml
output:
  gcp_cloud_storage:
    bucket: foo
    credentials_json: |
      { ...creds json... }
    path: benthos
    ...
```

pradeeppattamatta: That's a good option. Thanks @mihaitodor. Since it's a secret, we have to consider the option of setting such multi-line string through environment variable or cli-flags. I think that shouldn't be a problem. I will check that out from my side. Please let me know your thoughts as well.

mihaitodor: @pradeeppattamatta There's precedent in having 2 mutually exclusive fields, like, for example [tlsroot_cas](https://www.benthos.dev/docs/components/inputs/kafka_franz#tlsroot_cas) and [tlsroot_cas_file](https://www.benthos.dev/docs/components/inputs/kafka_franz#tlsroot_cas_file) for the `kafka_franz` components. I'd consider doing the same here. If you want to make it extra user-friendly, you can even add linter validation rules via bloblang. See for example [here](https://github.com/benthosdev/benthos/blob/ef48b45c8866d92023f7df4074a7b69c98a7d701/internal/impl/kafka/output_kafka_franz.go#L74).

pradeeppattamatta: Thanks @mihaitodor The yaml raw string feature worked, but setting that format through environment variables was having below problem. 

-  I tried by removing the Base64 encoding and setting the entire json as env variable. It was working for a certain type of service account, but it was not working for the Google Service Account that has private key in json (with `\n` chars in private key field of json). It was not being resolved correctly by the driver when set through env variable. I tried to replace the `\n` chars to newline, but that didn't help either. 
     
    image

- The option of the seperate file path field wouldn't solve my issue as well. GCP driver already inherently supports reading the Service Account Credentials from file path based env variable (GOOGLE_APPLICATION_CREDENTIAL as also documented in https://www.benthos.dev/docs/guides/cloud/gcp). But in some situations (like our environment) where we don't have an option of securely making the file available in container/app , adding support of setting it using environment variable would help. 

- Considering these, I think using the base64 encoded format is the workable solution. If the field name was misleading, I can name the field more clearly like `credentials_json_enc` or `credentials_json_encoded` and have the documentation to clearly mention that the field expects Base64 encoded value.

I will revert back to using Base64 encoding in my next commit.

mihaitodor: > It was not being resolved correctly by the driver when set through env variable. 
It's probably because of `"` not getting escaped properly

My suggestion was to support injecting the creds JSON inline in the yaml, but I didn't really think too much about env vars. One hack would be to enable interpolation for this field and then do something like `${! env("GCP_CRED_JSON").decode("base64") }`, but I think that's probably overkill...

What I don't get now is why do you need to set a custom env var and inject it in this field instead of just setting `GOOGLE_APPLICATION_CREDENTIAL`. Are you looking to be able to customise the name of the env var?

pradeeppattamatta: In that case, are we good with the decode option? Can I go ahead and commit back with those changes?

mihaitodor: I'll defer to @Jeffail, since he's the maintainer. Thanks for looking into this!

internal/impl/gcp/cache_cloud_storage.go
pradeeppattamatta: @Jeffail added a new commit to update to consolidate the options logic into single function `getClientOptionWithCredential()` and call that from from all the files.

pradeeppattamatta:
Please Refer https://github.com/benthosdev/benthos/pull/2042 for new PR.

Created new PR #2042 since the past commits in this one were not signed off and I had trouble to go back and update those commits.

PR #1834, opened 2023-04-04
fix: issue 1823 improves tracer OTLP docs

cezar-tech:


internal/impl/otlp/tracer_otlp.go
```diff
@@ -19,27 +19,37 @@ import (
...
-		Field(service.NewStringMapField("tags").
-			Description("A map of tags to add to all tracing spans.").
-			Default(map[string]string{}).
-			Advanced())
+const (
+	svcNameKey = semconv.ServiceNameKey
+	svcVerKey  = semconv.ServiceVersionKey
+)
 
+var cliVersion = cli.Version
```
cezar-tech, line 27: Could this cause any bugs during the build/ testing ?
can I reference this variable this way ?

Jeffail:
One issue here, which I think is maybe why the behaviour is as it currently is right now, is that the field docs will now be bound to the version and date stamps, which are dynamically set at build time. We can't really get away with that as it'll break docs generation that precedes a release. It'd also be kind of annoying that the field spec will change for every new tag.

I think rather than making the defaults explicit via a default flag we should just explain the behaviour within the field description.

cezar-tech:
@Jeffail I didn't understand this phrase:
> is that the field docs will now be bound to the version and date stamps,

What line specifically do you refer to ?

I have changed more than the necessary to apply this fix actually, I could rollback some lines and change specifically the description like you said.

My issue was only this part that didn't talk about the defaults:
image

Jeffail:
> What line specifically do you refer to ?

What I mean is that the line `string(svcVerKey):  cliVersion,` will change constantly as it's set dynamically at build time, and when changes to docs occur it causes build errors as they're assumed stale. Rather than set an explicit (and dynamic) default value for the field docs I'd be happier to expand the description to state "A map of tags to add to all tracing spans. The tags `" + string(svcNameKey) + "` and `" + string(svcVerKey) + " are added automatically when omitted."

lucasoares:
> > What line specifically do you refer to ?
> 
> What I mean is that the line `string(svcVerKey): cliVersion,` will change constantly as it's set dynamically at build time, and when changes to docs occur it causes build errors as they're assumed stale. Rather than set an explicit (and dynamic) default value for the field docs I'd be happier to expand the description to state:
> 
> ```
> "A map of tags to add to all tracing spans. The tags `" + string(svcNameKey) + "` and `" + string(svcVerKey) + " are added automatically when omitted."
> ```

What if we set the version as something like `<benthos binary version>` just to be clear in the documentation? @cezar-tech

lucasoares (changes requested):
I have one problem with the current implementation. If we let things like this, we will not be able to change the service name and version by using OpenTelemetry's environment variables since the component configuration will always be set. We should only set the fixed name and version if the resource can't be built by the environment itself

`OTEL_RESOURCE_ATTRIBUTES` - https://opentelemetry.io/docs/specs/otel/resource/sdk/#specifying-resource-information-via-an-environment-variable

`OTEL_SERVICE_NAME` - https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration

Code example to check if the user has set these attributes:

```
// Check if the user has already set a service name and versrion tag by using environment variables
resourceData := resource.Environment().Iter()

for resourceData.Next() {
kv := resourceData.Attribute()

if kv.Key == semconv.ServiceNameKey {
	// do not set the service name
}

if kv.Key == semconv.ServiceVersionKey {
	// do not set the service version
}
}
```

cezar-tech:
Hi guys shall we get back into this ? I would like to help

PR #1835, opened 2023-04-06
feat: change azure_table_storage input to allow filters with interpolation

eduardodbr:
Simple PR to allow interpolation on azure_table_storage inputs filters. Good to use with env variables like:

```
input:
  azure_table_storage:
    storage_connection_string: ""
    table_name: "Table"
    filter: "PartitionKey eq '${PARTITION_KEY}'"
```


eduardodbr:
close this PR because this is already possible, my bad

PR #1836, opened 2023-04-06
rewrite gcp_pubsub output

disintegrator:
This change rewrites the gcp_pubsub output to use the new config API and to lean on the pubsub client's internal batching mechanism. This is referred to as the "internal buffer" in docs to distinguish it from Benthos' batching mechanism. The rewrite also benefits from added tests and a pubsub mock.

Stress test results on an M1 Pro (8-core, 16GB mem): 

![prometheus metric graph of `rate(output_sent[30s])` for today's gcp_pubsub output, peaking at around 1.7K messages per second, and the new one in this pull request, peaking at around 10K mesages per second](https://user-images.githubusercontent.com/678548/232035132-2202ad57-f798-403f-acfc-1f575fc7fdd5.png)

(`gc_gcp_pubsub` is the working name of the output in this PR that doesn't clash with the existig output)

### DIFF
--- a/internal/impl/gcp/output_pubsub.go
+++ b/internal/impl/gcp/output_pubsub.go
@@ -3,34 +3,22 @@ package gcp
 import (
 	"context"
 	"fmt"
-	"strings"
 	"sync"
-	"time"
-
-	"google.golang.org/api/option"
 
 	"cloud.google.com/go/pubsub"
-
-	"github.com/benthosdev/benthos/v4/internal/batch"
-	"github.com/benthosdev/benthos/v4/internal/bloblang/field"
-	"github.com/benthosdev/benthos/v4/internal/bundle"
-	"github.com/benthosdev/benthos/v4/internal/component"
-	"github.com/benthosdev/benthos/v4/internal/component/metrics"
-	"github.com/benthosdev/benthos/v4/internal/component/output"
-	"github.com/benthosdev/benthos/v4/internal/component/output/processors"
-	"github.com/benthosdev/benthos/v4/internal/docs"
-	"github.com/benthosdev/benthos/v4/internal/log"
-	"github.com/benthosdev/benthos/v4/internal/message"
-	"github.com/benthosdev/benthos/v4/internal/metadata"
+	"github.com/benthosdev/benthos/v4/public/service"
+	"github.com/sourcegraph/conc/pool"
+	"google.golang.org/api/option"
 )
 
-func init() {
-	err := bundle.AllOutputs.Add(processors.WrapConstructor(func(c output.Config, nm bundle.NewManagement) (output.Streamed, error) {
-		return newGCPPubSubOutput(c, nm, nm.Logger(), nm.Metrics())
-	}), docs.ComponentSpec{
-		Name:    "gcp_pubsub",
-		Summary: `Sends messages to a GCP Cloud Pub/Sub topic. [Metadata](/docs/configuration/metadata) from messages are sent as attributes.`,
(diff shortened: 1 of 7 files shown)


internal/impl/gcp/output_pubsub.go
```diff
@@ -3,34 +3,22 @@ package gcp
...
-	"github.com/benthosdev/benthos/v4/internal/component"
-	"github.com/benthosdev/benthos/v4/internal/component/metrics"
-	"github.com/benthosdev/benthos/v4/internal/component/output"
-	"github.com/benthosdev/benthos/v4/internal/component/output/processors"
-	"github.com/benthosdev/benthos/v4/internal/docs"
-	"github.com/benthosdev/benthos/v4/internal/log"
-	"github.com/benthosdev/benthos/v4/internal/message"
-	"github.com/benthosdev/benthos/v4/internal/metadata"
+	"github.com/benthosdev/benthos/v4/public/service"
+	"github.com/sourcegraph/conc/pool"
```
disintegrator, line 10: This library has a useful result pool, kind of like of [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled). If it's a no go then I'd be happy to put something together with errgroup or similar.

internal/impl/gcp/output_pubsub.go
```diff
@@ -52,221 +40,294 @@ Or delete all keys with:
...
+			service.NewDurationField("publish_timeout").
+				Default(defaults.Timeout.String()).
+				Example("10s").
+				Example("5m").
+				Example("60m").
+				Description("The maximum length of time to wait before abandoning a publish attempt for a message.").
+				Advanced(),
+			service.NewMetadataFilterField("metadata").
+				Optional().
+				Description("Specify criteria for which metadata values are sent as attributes."),
```
disintegrator, line 75: This is a breaking change because the old api used `metadata.exclude_prefixes` where is the new api uses `metadata.include_prefixes` and `metadata.include_patterns`. Should I manually implement the old config somehow?

internal/impl/gcp/output_pubsub_test.go
```diff
@@ -0,0 +1,244 @@
...
+
+	var batchErr *service.BatchError
+	require.ErrorAs(t, err, &batchErr, "error is not a batch error")
+	require.Equal(t, 2, batchErr.IndexedErrors(), "did not receive expected number of batch errors")
+
+	var errs []string
+	batchErr.WalkMessages(func(i int, m *service.Message, err error) bool {
+		errs = append(errs, err.Error())
+		return true
+	})
```
disintegrator, line 242: I can't quite figure out why WalkMessages isn't iterating over the two failed messages despite `batchErr.IndexedErrors()` returning 2...

disintegrator: Solved in #1837

disintegrator: Removed my cherry-picked commit since the fix is in main now. Thanks a lot @Jeffail!

internal/impl/gcp/output_pubsub.go
Jeffail: I might expand the public field definition so that you can toggle exclusive versus inclusive defaults. Ideally all components will eventually converge but we need to do it in stages (with backwards compatibility).

Jeffail (approved):
LGTM, thanks @disintegrator, going to switch to the new `NewMetadataExcludeFilter` to fill the gap in backwards compat.

PR #1837, opened 2023-04-07
fix service.BatchError implementation

disintegrator:
* Implementing Unwrap on *service.BatchError
* Fixing bug where (*service.BatchError).WalkMessages does not iterate any messages in a batch

### DIFF
--- a/internal/message/sort_group.go
+++ b/internal/message/sort_group.go
@@ -38,7 +38,8 @@ func NewSortGroupParts(parts []*Part) (*SortGroup, []*Part) {
 	return g, newParts
 }
 
-// NewSortGroup creates a new sort group to be associated with a.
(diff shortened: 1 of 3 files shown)


Jeffail:
Thanks @disintegrator, I think in retrospect adding `Unwrap` makes sense. We get away without it right now thanks to https://github.com/benthosdev/benthos/blob/main/public/service/errors.go#L79 but it's a fit flakey.

It's not valid, but my original reason for not including a public `Unwrap` (or similar stuff on any of the `public` types) was concern about people escaping into internals via something like `naughty := err.Unwrap().(interface{ IndexedErrors() int }).IndexedErrors()`. It's pretty dumb, I think I'm taking the concept of breaking changes too far especially here.

Jeffail (approved):

PR #1845, opened 2023-04-13
Use lib constants instead of having development credentials in the code

mfamador:
Using the constants provided by the Azure storage library Instead of hard-coding the azurite emulator credentials in the code.


Jeffail (approved):
LGTM, thanks @mfamador!

PR #1847, opened 2023-04-13
Prevent silent failures in the snowflake_put output

mihaitodor:
In the future, we might want to hook up and enable the internal logger from `github.com/snowflakedb/gosnowflake` and get the latest version which adds a bit of extra debug logging for certain failures as described [here](https://github.com/snowflakedb/gosnowflake/issues/701). I'll leave that for a follow-up PR, in case anyone asks about it.


Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #1850, opened 2023-04-14
Fix azure_table_storage hanging when there's no results

### LINKED ISSUE #1849: Input azure_table_storage hangs when the query result is empty
Given a simple pipeline with azure_table_storage as input:

```
input:
  azure_table_storage:
    storage_connection_string: "theConnString"
    table_name: ${TABLE_NAME}
    filter: "PartitionKey eq 'foo'"
    page_size: 1000

```

the pipeline hangs if the result from the query is an empty batch. This is probably the result of a regression introduced in version 4.12.0, the same pipeline works with 4.11.0 and older. 

I am trying to debug the issue but it does not seem to be an obvious one to me. I don't know if this affects other inputs, however, I tried sql_select with postgres and it works correctly.

I will keep trying to debug the issue but let me know if you can help.
Thanks

mfamador:
When there's an empty result it wasn't returning component.ErrTypeClosed and it wouldn't exit.
Don't know how it used to work on 4.11, though.
Fixes #1849 

cc: @eduardodbr


Jeffail (approved):
LGTM, thanks @mfamador!

PR #1856, opened 2023-04-18
add `sentry_capture` processor

disintegrator:
This change adds a `sentry_capture` processor that captures messages as events that are sent to Sentry (https://sentry.io/welcome/), a popular error and event reporting service.

Sample usage:

```yaml
input:
  generate:
    count: 10
    interval: 0
    mapping: |
      root = {
        "id": nanoid(),
        "type": "order_created",
        "customer_id": "CU1234",
        "amount": random_int(min:10, max:100),
        "currency": "USD",
        "created_at": now()
      }

pipeline:
  processors:
    - mapping: |
        root = if this.amount % 2 == 0 {
          throw("not a good outcome")
        }
    - catch:
      - label: capture_sentry_error
        sentry_capture:
          level: ERROR
          message: "${! error() }"
          tags:
            pipeline: order-processing-events
          context: |
            root = {
              "message": this.with("id", "type", "customer_id", "created_at")
            }

output:
  drop: {}
```

sentry-demo


internal/impl/sentry/processor_capture.go
```diff
@@ -0,0 +1,348 @@
...
+	"github.com/getsentry/sentry-go"
+)
+
+const (
+	transportAsync = "async"
+	transportSync  = "sync"
+)
+
+func getBenthosVersion() string {
+	bi, ok := debug.ReadBuildInfo()
```
Jeffail, line 21: I know it's gross but for now the best place to grab a version is probably from cli.Version directly: https://github.com/benthosdev/benthos/blob/main/internal/cli/run.go#L27, walking deps for module names unfortunately only seems to work for custom benthos builds where the module is imported separately, unless they fixed that recently.

disintegrator: make sense. updated!

Jeffail (approved):
Awesome, thanks @disintegrator!

PR #1865, opened 2023-04-24
add `changestream` support for mongodb input

cloorc:
An example for this feature is as following:

```yaml
# Common config fields, showing default values
input:
  label: ""
  mongodb:
    url: "mongodb://mongodb-0.mongodb-headless.develop.svc.cluster.local:27017,mongodb-1.mongodb-headless.develop.svc.cluster.local:27017/admin?readPreference=secondary&replicaSet=rs0&authSource=admin"
    database: "db"
    collection: "collection"
    username: "..."
    password: "..."
    operation: changestream
    resume_token_file: ~/.benthos/resume_token # make sure the directory exist
    query: ""
# Config fields, showing default values
output:
  label: ""
  file:
    path: "./output-file.txt"
    codec: lines
```

**NOTE**: Please make sure the `mongodb.*` parameters are correct.

Former configuration will follow mongodb change stream and write every changes into file `./output-file.txt` and a resume token bound to current changed document will be also written into `~/.benthos/resume_token` (**make sure the directory is already exist**) as soon as `benthos` has decoded this event.

### DIFF
--- a/internal/impl/mongodb/input.go
+++ b/internal/impl/mongodb/input.go
@@ -1,11 +1,16 @@
 package mongodb
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"io/fs"
+	"io/ioutil"
+	"os"
 
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
 
 	"github.com/benthosdev/benthos/v4/internal/impl/mongodb/client"
 	"github.com/benthosdev/benthos/v4/public/service"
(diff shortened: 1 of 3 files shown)


Jeffail:
Hey @cloorc, I think it might be worth separating this out into its own input since the logic forks quite a lot, maybe called `mongodb_changestream`. Also, rather than using a file for storing resume tokens it'd be better to use caching as that'd allow users to store the tokens anywhere (like redis, memcached, etc), there's an example of something similar here: https://github.com/benthosdev/benthos/blob/main/internal/impl/discord/input.go#L113

cloorc:
> Hey @cloorc, I think it might be worth separating this out into its own input since the logic forks quite a lot, maybe called `mongodb_changestream`. Also, rather than using a file for storing resume tokens it'd be better to use caching as that'd allow users to store the tokens anywhere (like redis, memcached, etc), there's an example of something similar here: https://github.com/benthosdev/benthos/blob/main/internal/impl/discord/input.go#L113

Yeah. Nice suggestion. I'm going to re-submit this PR in future.

PR #1868, opened 2023-04-24
fix: Resolve mongodb output batching issue

juniorrhis:
closes https://github.com/benthosdev/benthos/issues/1867


Jeffail (approved):
LGTM, thanks @juniorrhis! Ideally we need to add tests that cover this but happy with a fix for now.

PR #1872, opened 2023-04-27
re-export msgpack plugins via public interface

disintegrator:
The msgpack bloblang and processor plugins were not exposed through public/components package. This change re-exports these plugins so they can be cherry-picked in custom Benthos builds.


mnbbrown:
Thanks @disintegrator 🙌

Jeffail (approved):
Whoops! Thanks @disintegrator!

PR #1876, opened 2023-04-28
kafka: reword start_from_oldest sematics

filippog:
Clarify that Benthos will only re-consume old messages when a new consumer group is created.


Jeffail:
It also kicks in when the saved offset no longer exists

Jeffail:
also looks like you're committing from an ooooolllllld version, sorry didn't spot that on first glance

filippog:
> also looks like you're committing from an ooooolllllld version, sorry didn't spot that on first glance

doh! of course, sorry about the involuntary trip down memory lane

Jeffail (approved):

Jeffail:
LGTM, thanks @filippog!

Issue #882, opened 2021-09-23
Support Batching in the Azure Blob Storage output


leosunmo:
There's (probably) no real reason why it couldn't support batching just as well as the S3 output does.

Godocs for uploading a Block Blob: https://pkg.go.dev/github.com/Azure/azure-storage-blob-go/azblob#BlockBlobURL.Upload

It mentions that you can't update existing Block Blobs, and that you can stage Blobs before you commit them to the final Block Blob.
>To perform a partial update of a block blob, use StageBlock and CommitBlockList

Microsoft docs for `Put Blob`: https://docs.microsoft.com/en-gb/rest/api/storageservices/put-blob

mihaitodor:
@leosunmo Thank you for opening this issue! The initial implementation of this output made use of https://github.com/Azure/azure-sdk-for-go, which was labeled as deprecated, but due to a [lack of connection string support](https://github.com/Azure/azure-storage-blob-go/issues/244) in azure-storage-blob-go, we decided to still use the deprecated library. However, there was an update [here](https://github.com/Azure/azure-sdk-for-go/issues/14295#issuecomment-918511620) where, apparently https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob is what we should be using. I asked them for details, but it's been radio silence, so I guess I'll have to dig into it and see if it supports everything we need. I'll have a look at it soon and try to update Benthos to use the new lib.

Note to self: #615 should be addressed as part of this.

leosunmo:
Wow, didn't realise there were multiple versions of the Blob SDK. That looks like a bit of a mess :grimacing: 
I especially love how even the deprecated package is in preview still.

mihaitodor:
Small update: Looks like the azblob stuff [was migrated](https://github.com/Azure/azure-storage-blob-go#if-you-would-like-to-access-our-latest-go-sdk-please-refer-to-the-new-preview-azblob-package-here-if-you-would-like-more-information-on-azures-burgeoning-effort-to-coordinate-the-development-of-the-sdks-across-services-of-which-this-change-is-a-part-please-refer-to-this-article) over to the ~legacy~ current sdk and now resides here: https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/storage/azblob#readme We'll have to revisit the Azure Storage input and output and switch them to new API, which should allow us to leverage the batching functionality.

PR #1881, opened 2023-05-02
feat: migrate input azure_blob_storage to azure-sdk-for-go/storage/azblob

eduardodbr:
The input azure_blob_storage uses a deprecated SDK as you can see [here](https://github.com/Azure/azure-sdk-for-go#historical-releases). This PR migrates the input to the most recent [SDK](https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/storage/azblob). It was also added the option to authenticate via managed identity when no other auth method is provided (conn string, access key, or SAS token). No other code changes were made.

It would be nice to also have the review of the original contributor of this input @mihaitodor

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,9 @@ require (
 	cuelang.org/go v0.4.2
 	github.com/Azure/azure-sdk-for-go v68.0.0+incompatible
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.5.0
+	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.1.0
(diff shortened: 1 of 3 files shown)


mihaitodor:
Related: #882. Thanks @eduardodbr for looking into this! I was meaning to get to it, but, alas... There are some integration tests that are hardcoded to be skipped, but I think we can safely remove those hacks now and we should also be able to enable this test https://github.com/benthosdev/benthos/blob/463169c5cf6e511b494ac585722998a2a04d6c20/internal/impl/azure/integration_test.go#L145. If you wish to play with the tests, let me know. Otherwise, I might give it a shot later this week

Jeffail (approved):
LGTM, thanks @eduardodbr!

Issue #1155, opened 2022-03-04
Provide custom logger API for custom plugin builds


luxixing:
In stream mode, can I replace service.Logger with my own implementation? Because in the case of stream mode, I want the logs to be output to my own logging service, such as tcp logger, instead of the standard output

josephwoodward:
I've made a start on this here: https://github.com/benthosdev/benthos/pull/1883

mihaitodor:
Fixed in [v4.23.0](https://github.com/redpanda-data/connect/releases/tag/v4.23.0).

PR #1883, opened 2023-05-05
Provide support for using a custom logger via the `slog` logger interface

josephwoodward:
Closes: https://github.com/benthosdev/benthos/issues/1155

This change aims to address #1155 by giving users of Benthos more control over logging by adding an API that allows them to provide their own logging implementation, using the new `slog` API as a common interface.

### DIFF
--- a/.github/workflows/cross_build.yml
+++ b/.github/workflows/cross_build.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.19.x, 1.20.x, 1.21.x]
+        go-version: [1.20.x, 1.21.x]
         os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
     env:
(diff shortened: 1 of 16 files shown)


Jeffail:
Thanks @josephwoodward, looks great! Can we expect `golang.org/x/exp/slog` to be moved to `log/slog` in v1.21? If so maybe it'd be worth holding off so we're not stuck with the old package as merging now and then swapping them when `log/slog` is available would be a breaking API change.

Jeffail:
Arg I was hoping we could add a build constraint specifically to these files and continue supporting 1.20 but that hasn't been implemented yet: https://github.com/golang/go/issues/59033

There's a whole bunch of cool stuff in 1.21 that I want to mess around with so in the name of fun and coolness I think I'm going to do a release now with 1.20 supported and then jump to 1.21 and merge this PR.

Jeffail:
Merged via: https://github.com/benthosdev/benthos/commit/ee0000450413ad37685ecbfe0d1ef40d178d29fb

Thanks @josephwoodward!

PR #1889, opened 2023-05-08
Add pgzip support to compress, decompress, codec reader.

nibbleshift:
This adds pgzip support which enables the use of a parallel gzip library in codec reader, compress processor, and decompress processor.

In cases where you're dealing with gzip compressed data in 1MB or larger chunks this can lead to significant performance improvements.


Jeffail (approved):
LGTM, thanks @nibbleshift!

PR #1890, opened 2023-05-08
Add LRU Cache for DotPathToSlice

nibbleshift:
Cache DotPathToSlice results to avoid allocations and str repls.

Reduces memory allocations and reduces CPU utilization in that code path by about 30%.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -55,6 +55,7 @@ require (
 	github.com/gorilla/mux v1.8.0
 	github.com/gorilla/websocket v1.5.0
 	github.com/gosimple/slug v1.13.1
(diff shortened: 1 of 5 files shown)


Jeffail:
Can we get some benchmark tests that show the before/after? Ideally with a realistic config, maybe as an addition to https://github.com/benthosdev/benthos/blob/main/public/service/benchmark_test.go#L17

I'm little hesitant to add caches in deep areas of the codebase and a lot of these changes are in places I'd expect to be cold paths.

nibbleshift:
Perhaps we can discuss merging my PR in gabs instead https://github.com/Jeffail/gabs/pull/136

PR #1892, opened 2023-05-09
feat: Add SetPrintLoggerAtLevel method to StreamBuilder

togettoyou:
This commit adds a new method SetPrintLoggerAtLevel to StreamBuilder, which allows setting a custom logger supporting a simple Print-based interface for use by stream components, and specifying the logging level. This custom logger will override any logging fields set via configuration.

### DIFF
--- a/public/service/stream_builder.go
+++ b/public/service/stream_builder.go
@@ -125,6 +125,13 @@ func (s *StreamBuilder) SetPrintLogger(l PrintLogger) {
 	s.customLogger = log.Wrap(l)
 }
 
+// SetPrintLoggerAtLevel sets a custom logger supporting a simple Print based interface
+// to be used by stream components, and specifies the logging level. This custom logger
+// will override any logging fields set via config.
+func (s *StreamBuilder) SetPrintLoggerAtLevel(l PrintLogger, level int) {
+	s.customLogger = log.WrapAtLevel(l, level)
+}
+
 // HTTPMultiplexer is an interface supported by most HTTP multiplexers.
 type HTTPMultiplexer interface {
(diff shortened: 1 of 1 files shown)


Jeffail:
Hey @togettoyou, we have a slog based log option coming in https://github.com/benthosdev/benthos/pull/1883 that I think would maybe avoid the need to add this. My only gripe is that we'd end up with lots of `SetLoggerXXX` variants when almost everything can be achieved with the slog option, would that work for your use case?

togettoyou:
Thank you very much for your response, and I must say that the implementation of #1883 is indeed better.

I was actually hoping to create a function that encapsulates the execution of a Benthos task and allows for redirection of the execution logs. However, based on the API that is currently available, it seems that this is not supported. Here's an example:

```
func Run(ctx context.Context, conf string, w io.Writer) (chan error, error) {
	builder := service.NewStreamBuilder()

	err := builder.SetYAML(conf)
	if err != nil {
		return nil, err
	}

	// TODO Output Redirection
	builder.SetLogger(w)

	stream, err := builder.Build()
	if err != nil {
		return nil, err
	}

	e := make(chan error, 1)

	go func() {
		err = stream.Run(ctx)
		e <- err
	}()

	return e, nil
}
```

peczenyj:
Hey @togettoyou do you think you can sign your commits?

togettoyou:
Sure, I can sign my commits. Thank you for reminding me.

> Hey @togettoyou do you think you can sign your commits?

Jeffail:
I believe this is solved via https://github.com/benthosdev/benthos/pull/1883

PR #1894, opened 2023-05-10
fix definition for `label` and `processors` in CUE

disintegrator:
The logic used to determine if Benthos components can have an associated `label` or `processors` was broken and specifically for the #Processors produced an infinite recursion. The output was:

```cue
#Processor: or([ for name, config in #AllProcessors {
	(name): config
}])
#Processor: #Processor & {
	processors?: [...#Processor]
	label?: string
}
```

The code has been updated to correctly generate this augmentation and we now have the following:

```cue
#Processor: or([ for name, config in #AllProcessors {
	(name): config
}])
#Processor: {
	label?: string
}
```

### DIFF
--- a/internal/cuegen/component.go
+++ b/internal/cuegen/component.go
@@ -28,25 +28,6 @@ func doComponents(specs []docs.ComponentSpec, opts *componentOptions) ([]ast.Dec
 		fields = append(fields, field)
 	}
 
-	var addons []any
-	if opts.canLabel {
-		addons = append(
-			addons,
-			ast.NewIdent("label"),
-			token.OPTION,
-			ast.NewIdent("string"),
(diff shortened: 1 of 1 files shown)


internal/cuegen/component.go
```diff
@@ -28,25 +28,6 @@ func doComponents(specs []docs.ComponentSpec, opts *componentOptions) ([]ast.Dec
 		fields = append(fields, field)
 	}
 
-	var addons []any
```
disintegrator, line 31: notice that I never used this slice. I only checked its length in the original code. Sorry for the messyness.

Jeffail (approved):
LGTM, thanks @disintegrator!

PR #1897, opened 2023-05-12
document `unarchive` processor metadata handling

disintegrator:
This change updates the documentation for the `unarchive` processor to explicitly mention that metadata is preserved from the source message to resulting messages.


Jeffail (approved):
LGTM, thanks bae!





Uh, I mean @disintegrator

PR #1899, opened 2023-05-13
Add `redis_scan` input.

### LINKED ISSUE #1898: Redis input
I've created a custom redis input to walk through the keys matching a prefix and get their values.

If it's of interest I can add it to benthos, but before I'd like to discuss a few options:

1. since there's an additional call to get the values, should we have an option to get only the keys?
2. on which format should the messages be created:
2.1. when there's keys and values, return a json message? ex. {"my_key":"a_value"} ? 
2.2. when there's only keys, the message would be only its content? ex. "my_key"?
3. Would `redis` be a good name for the input?

mfamador:
Add the `redis_scan` input, to allow us to scan the redis keys and retrieve their values, optionally matching a blob-style pattern.

Ex.
```yaml
input:
  redis_scan:
    url: 'redis://localhost:6379'
    match: '*4*'
``` 

It generates a message for each key value pair in json format:

Ex.
```json
{"key":"100044","value":"1658881920000000000"}
``` 

Resolves #1898

### DIFF
--- a/internal/cli/studio/pull_runner_test.go
+++ b/internal/cli/studio/pull_runner_test.go
@@ -72,7 +72,7 @@ func testServerForPullRunner(
 
 		t.Logf("request: %v", expReq.path)
 
-		require.Equal(t, expReq.path, r.URL.Path)
+		require.Equal(t, expReq.path, r.URL.EscapedPath())
 
 		// Verify that our authorization tokens are punching through
 		assert.Equal(t, "aaa", r.Header.Get("X-Bstdio-Node-Id"))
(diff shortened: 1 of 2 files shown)


Jeffail (commented):
LGTM, any chance you can add an integration test to the mix?

mfamador:
Sure, I'll add them.

mfamador:
Actually, not having the corresponding output it's not as straight forward as I thought.
I probably need to write directly to redis instead of the output and not use the generic `integration` package?

Jeffail:
Yeah either something manual or it might be worth adding something similar to https://github.com/benthosdev/benthos/blob/main/internal/integration/stream_test_definitions.go#L755 where we can define input only tests for reuse, but manual is fine for now. Just anything that spins up the docker container and just makes sure data gets through is fine as a starting point.

PR #1901, opened 2023-05-13
feat: migrate output azure_blob_storage to azure-sdk-for-go/storage/azblob

eduardodbr:
The output azure_blob_storage uses a deprecated SDK as you can see [here](https://github.com/Azure/azure-sdk-for-go#historical-releases). This PR migrates the output to the most recent [SDK](https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/storage/azblob). It also added the option to authenticate via managed identity when no other auth method is provided (conn string, access key, or SAS token).


internal/impl/azure/output_blob_storage.go
```diff
@@ -198,22 +223,21 @@ func (a *azureBlobStorageWriter) WriteBatch(_ context.Context, msg message.Batch
 		if err != nil {
 			return fmt.Errorf("blob type interpolation error: %w", err)
 		}
-
-		c := a.client.GetContainerReference(containerStr)
-		b := c.GetBlobReference(pathStr)
-		if err = a.uploadBlob(b, blobTypeStr, p.AsBytes()); err != nil {
+		ctx := p.GetContext()
```
mihaitodor, line 226: I need to familiarise myself with how this context is used, but shouldn’t we use here the `ctx` parameter of the method?

eduardodbr: You are correct. I think I got used to retrieve the context from the message on custom processors and messed up here. Thanks!

mihaitodor:
Hey @eduardodbr, for some reason GitHub won't let me create PRs against your fork... Dunno why :( 

I took some time to port my changes on top of yours on [this branch](https://github.com/mihaitodor/benthos/tree/migrate_az_blob_storage_output_mihai) and I created [this PR](https://github.com/mihaitodor/benthos/pull/5) on my fork where you can leave comments if there's anything that doesn't look right or you'd rather I not change. Once you're happy with my changes, you should be able to just cherry-pick my commit(s). Note that I first rebased your branch on top of the latest `main` because we branched off from different points in history and I was getting confused with diffs...

Jeffail (approved):
LGTM, thanks @eduardodbr and @mihaitodor!

PR #1902, opened 2023-05-13
Expose WaitTimeSeconds for SQS on S3 input.

nibbleshift:
Expose SQS WaitTimeSeconds when using SQS with the S3 Input.


Jeffail (approved):
LGTM, thanks @nibbleshift!

Issue #1906, opened 2023-05-16
Add protobuf schema support to`schema_registry_encode` / `schema_registry_decode`


ye11ow:
Similar to #1150 , we can add protobuf schema support for these two functions as well.
More details: https://docs.confluent.io/platform/current/schema-registry/fundamentals/serdes-develop/index.html#wire-format

Sample code from kafka-go-client: https://github.com/confluentinc/confluent-kafka-go/blob/master/schemaregistry/serde/protobuf/protobuf.go#LL347C34-L347C34

ye11ow:
Hi @Jeffail , I already have some working code on our repo, should be able to create a PR for limited protobuf decode support

Jeffail:
Hey @ye11ow, this'd be awesome! I have limited availability for the next few weeks but I'll review the open PR ASAP.

ye11ow:
Thank you @Jeffail for the excellent work on this!

Jeffail:
Done via: https://github.com/benthosdev/benthos/commit/cb07600d5ccb0e1ac7399cdcb10e085d6c4aeab4
And: https://github.com/benthosdev/benthos/commit/a3d52e2428e6c33f26cf14435ec0a585ff2f8e7e

PR #1908, opened 2023-05-16
added basic protobuf support to `schema_registry_decode`

ye11ow:
Added basic protobuf support for `schema_registry_decode`. Tested with Confluent datagen connector.

Current limitations:
* It only supports protobuf definition with only ONE message type.
* [Schema reference](https://docs.confluent.io/platform/current/schema-registry/fundamentals/serdes-develop/serdes-protobuf.html#schema-references-in-protobuf) is not supported yet.

Related to #1906

### DIFF
--- a/internal/impl/confluent/processor_schema_registry_decode.go
+++ b/internal/impl/confluent/processor_schema_registry_decode.go
@@ -39,7 +39,7 @@ func schemaRegistryDecoderConfig() *service.ConfigSpec {
 		Description(`
 Decodes messages automatically from a schema stored within a [Confluent Schema Registry service](https://docs.confluent.io/platform/current/schema-registry/index.html) by extracting a schema ID from the message and obtaining the associated schema from the registry. If a message fails to match against the schema then it will remain unchanged and the error can be caught using error handling methods outlined [here](/docs/configuration/error_handling).
 
-Currently only Avro schemas are supported.
+Currently only Avro and Protobuf schemas are supported.
 
 ### Avro JSON Format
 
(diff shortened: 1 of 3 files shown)


Jeffail (commented):
Awesome @ye11ow! Any chance you'd be able to add the equivalent for `schema_registry_encode` as well?

ye11ow:
hey @Jeffail , I haven't started to work on the encode side yet. Our own use case only requires the decode part. I don't think it is difficult to implement it though. I create a separate PR later (maybe 1 or 2 weeks later).

Jeffail:
Hey @ye11ow, I decided to put a bit of work into adding references on top of this PR, I'm aiming to have encoding support as well. This PR will get merged at the same time which hopefully won't take long.

Jeffail:
Okay sorted, your bits are merged via: https://github.com/benthosdev/benthos/commit/cb07600d5ccb0e1ac7399cdcb10e085d6c4aeab4

And my bits on top: https://github.com/benthosdev/benthos/commit/a3d52e2428e6c33f26cf14435ec0a585ff2f8e7e

I've added references support (both protobuf and avro) and also removed the limitation of only one message per protobuf schema. The way that encoding works with multiple message definitions is a bit hacky so I've added caveats for it on the docs page.

Thanks for kicking this off @ye11ow! We should be covering a ton of new use cases now, just missing the JSONSchema functionality but I'm quite comfortable leaving it out for now.

ye11ow:
Hey @Jeffail, excellent work! Thank you so much for the extra work on the encoding and multiple message support! Really appreciate your time this part :-]

PR #1916, opened 2023-05-20
Resolves #1208 for kinesis streams

### LINKED ISSUE #1208: /ready behavior doesn't exactly match up with documentation
hey there! first, love the project, really appreciate all the hard work that's gone into it :)

edit: 

```shell
~ % benthos --version
Version: 3.65.0
Date: 2022-03-08T05:14:05Z
```

according to [the docs](https://www.benthos.dev/docs/components/http/about/#endpoints):

`/ready can be used as a readiness probe as it serves a 200 only when both the input and output are connected, otherwise a 503 is returned.`

what's interesting is that if one cannot connect to kinesis, the `/ready` endpoint returns `200 OK` after the error is printed (note the timestamps):

```shell
# benthos process
~ % benthos  --log.level trace -s http.debug_endpoints=true -s output.aws_kinesis.stream='ooooooooooooooooooooooooo' -s output.aws_kinesis.partition_key='${!json("id")}' -s output.aws_kinesis.region='us-east-1'
{"@timestamp":"2022-04-11T14:32:40-04:00","@service":"benthos","component":"benthos","level":"INFO","message":"Launching a benthos instance, use CTRL+C to close."}
{"@timestamp":"2022-04-11T14:32:40-04:00","@service":"benthos","component":"benthos","level":"INFO","message":"Listening for HTTP requests at: http://0.0.0.0:4195"}
{"@timestamp":"2022-04-11T14:35:59-04:00","@service":"benthos","component":"benthos.output","level":"ERROR","message":"Failed to connect to aws_kinesis: ResourceNotReady: exceeded wait attempts"}
```

```shell
# console
~ % http GET localhost:4195/ready
HTTP/1.1 503 Service Unavailable
Content-Length: 21
Content-Type: text/plain; charset=utf-8
Date: Mon, 11 Apr 2022 18:32:45 GMT

output not connected


~ % http GET localhost:4195/ready
HTTP/1.1 200 OK
Content-Length: 2
Content-Type: text/plain; charset=utf-8
Date: Mon, 11 Apr 2022 18:36:19 GMT

OK

```

seems like this behavior isn't quite what the docs describe - it's problematic as readiness probes switch from `unready` to `ready`, despite the (only) output being errored out.

leometzger:
Hi there,

Apparently, there is a bug in the validation of the '/ready' endpoint for kinesis streams when the stream does not exist. This problem may occur due to a misconfiguration.

The current behavior is that after the initial error is returned by the output connect method, benthos attempts to reconnect using exponential backoff. But, as the aws session has already been established, the kinesis output is considered ready even if the output is not connected.

This PR fixes this behavior for the output. It also applies a fix to input which had a similar issue.

Awesome project btw.

Bug report: #1208


Jeffail (approved):
Thanks @leometzger! Mostly LGTM, there's a bit of modification I want to make to `waitUntilStreamExists` but I'll do that on my side as I also need to rebase against changes made to the aws components recently.

PR #1918, opened 2023-05-21
Remove sync.Mutex/sync.WaitGroup from dispatchMessages

nibbleshift:
Remove WaitGroup and Mutex from dispatchMessages.
    
This PR removes the use of the Mutex and the WaitGroup in dispatchMessages.  Both the Mutex and WaitGroup appear to be unnecessary in this routine.  Additionally, the newPending slice is now allocated up-front to avoid multiple allocations as items are appended.


nibbleshift:
Gah, based the PR off the wrong branch....

PR #1928, opened 2023-05-27
Use the azblob package for the azure_blob_storage output

mihaitodor:
Also revamp the azure_blob_storage tests to remove some old hacks.

I think we might want to revamp the way the output writes batches to increase throughput as mentioned in #882, but I'll leave that for a future PR. Basically, we could try to stream all the messages in a batch in one go instead of writing each message individually.

@eduardodbr If you have some time, please have a look over this code and maybe try it out on your end. I want to make sure it still works for you.

### DIFF
--- a/internal/impl/azure/shared/service.go
+++ b/internal/impl/azure/shared/service.go
@@ -12,7 +12,9 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/storage"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/Azure/azure-sdk-for-go/sdk/data/aztables"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
 
 	"github.com/Azure/azure-storage-queue-go/azqueue"
 )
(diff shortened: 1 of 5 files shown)


eduardodbr:
Hi @mihaitodor I already opened [PR](#1901) to migrate the azure blob storage to the most recent azblob and also updated the integration tests to work. However, you do have some nice changes that can be added to that PR

internal/impl/azure/shared/service.go
```diff
@@ -121,3 +123,48 @@ func parseConnectionString(input string) (storageAccount, storageAccessKey strin
 	}
 	return accountName, accountKey, nil
 }
+
+// GetStorageClient creates the storage client for the provided credentials
+// TODO: Use a proper config struct for the parameters to avoid mixing them up
+func GetStorageClient(storageConnectionString, storageAccount, storageAccessKey, storageSASToken string) (*azblob.Client, error) {
```
eduardodbr, line 129: I would rename this to GetBlobStorageClient to be more explicit because Azure storage comprises blob, file and queue.

internal/impl/azure/shared/service.go
```diff
@@ -121,3 +123,48 @@ func parseConnectionString(input string) (storageAccount, storageAccessKey strin
...
+	var err error
+	if len(storageConnectionString) > 0 {
+		storageConnectionString := storageConnectionString
+		if strings.Contains(storageConnectionString, "UseDevelopmentStorage=true;") {
+			storageConnectionString = GetEmulatorConnectionString("10000", "10001", "10002")
+		}
+		// The UI doesn't add the AccountName parameter to the connection string for some reason...
+		if !strings.Contains(storageConnectionString, "AccountName=") {
+			storageConnectionString = storageConnectionString + ";" + "AccountName=" + storageAccount
+		}
```
eduardodbr, line 140: Not a big deal but are you sure about this? never seen it happening

mihaitodor:
@eduardodbr Oh, crap… I should’ve checked 🙈 Sorry about that! I saw the other one was merged and wanted to address the output for a while… I’ll have a look at your changes and send you a PR or smth.

internal/impl/azure/shared/service.go
mihaitodor: Dunno why it’s happening in my corporate account. Seems like a really weird omission on their end. I thought it’s quite harmless to add it automatically as a convenience, even though the error it gives should help the user figure out what’s missing.

mihaitodor:
Closing in favour of #1901 and https://github.com/mihaitodor/benthos/pull/5

PR #1930, opened 2023-05-27
Add ability to configure timeout when writing to Google Cloud Storage

josephwoodward:
Adds support for providing a timeout when uploading to Google Cloud Storage, following the same documentation as AWS S3.


Jeffail (approved):
LGTM, thanks @josephwoodward!

Jeffail:
Don't worry about CI failure, I'm about to rework the GCP components to the new plugin APIs so I'll fix it on my end.

PR #1935, opened 2023-06-01
shallow-copy in gcp_bigquery_select processor

disintegrator:
This change updates the `gcp_bigquery_select` processor to shallow copy incoming messages and mutate the copies rather than unintentionally mutate the original messages.

### DIFF
--- a/internal/impl/gcp/processor_bigquery_select.go
+++ b/internal/impl/gcp/processor_bigquery_select.go
@@ -169,6 +169,7 @@ func (proc *bigQuerySelectProcessor) ProcessBatch(ctx context.Context, batch ser
 	outBatch := make(service.MessageBatch, 0, len(batch))
 
 	for i, msg := range batch {
+		msg = msg.Copy()
 		outBatch = append(outBatch, msg)
 
 		var args []any


disintegrator:
I'm not sure if this is necessary. It was off the back of a conversation on Discord regarding how to preserve metadata and produce new messages in Benthos processors. It seems that the general guidance is that processors should shallow-copy messages and mutate the copies.

Jeffail:
Hey @disintegrator, there's no need for this as you're allowed to mutate messages within a processor directly. The need for copying is when you want to return more messages than you received (a sort of unarchive/expand processor) as in those cases you have the option of returning messages that originate from the provided messages (and adopt their tracing spans, metadata, etc) or brand new messages that are basically orphans. Brand new messages should be avoided when possible as they effectively terminate any potential for sending a `sync_response` derived from those new messages.

The API docs ought to make this stuff clearer, the challenge there is it's only really important in the context of batches and expansion which is usually a doozy to explain concisely.

disintegrator:
Thanks @Jeffail. That’s really useful to know!

PR #1945, opened 2023-06-05
Support connection names in NATS components

codegangsta:
This PR adds support for connection names in NATS components.

When running NATS in production, it's often nice to be able to assign names to connections, so you can effectively observe and monitor them. Instead of adding a `connection_name` field, I opted to use the `label` of the component to set the NATS connection name. If no label exists, the connection name will be an empty string, which is the default behavior for NATS anyway.


Jeffail (approved):
Nice, thanks @codegangsta. I haven't considered using labels for stuff like this in the past but it makes total sense.

PR #1947, opened 2023-06-06
fix message ordering support on gcp_pubsub output

disintegrator:
This change restores message ordering support on gcp_pubsub that broke as part of rewriting this output. The `*pubsub.Topic` type has a flag that needs to be set after acquiring the topic.


Jeffail (approved):
LGTM, thanks @disintegrator!

PR #1949, opened 2023-06-06
Update Oracle integration tests

### LINKED ISSUE #1942: Potential upgrade to Oracle Database 23c
Hi all,

First of all, I'd like to thank you for using my images!

I wanted to let you know that there is a new version of [Oracle Database (23c)](https://oracle.com/oracledbfree) out and that I've put new images together.
Alongside the new version comes a new name, so the new images can be found under [gvenzl/oracle-free](https://hub.docker.com/r/gvenzl/oracle-free).

The images are compatible with the one that is currently being used:
https://github.com/benthosdev/benthos/blob/67e996df9e3a1b2baeb8c37c8b245457163e9a01/internal/impl/sql/integration_test.go#L900-L901

Furthermore, we have put some optimizations in place that make the non-faststart `slim` image usually faster than the `slim-faststart` image (because the smaller image size downloads faster than the startup time optimizations in the `faststart` images).

mihaitodor:
Fixes #1942.

I tested it using the following commands on an M1 Macbook:

```shell
> brew install colima
> colima start --arch x86_64 --memory 4
> export DOCKER_HOST=unix:///Users/ntodor/.colima/docker.sock
> go test -run ^TestIntegrationOracle$ -count=1 ./internal/impl/sql
> colima stop
```

Details about Colima here: https://github.com/abiosoft/colima

### DIFF
--- a/internal/impl/sql/integration_test.go
+++ b/internal/impl/sql/integration_test.go
@@ -897,7 +897,7 @@ func TestIntegrationOracle(t *testing.T) {
 	pool.MaxWait = 3 * time.Minute
 
 	resource, err := pool.RunWithOptions(&dockertest.RunOptions{
-		Repository:   "gvenzl/oracle-xe",
+		Repository:   "gvenzl/oracle-free",
 		Tag:          "slim-faststart",
 		ExposedPorts: []string{"1521/tcp"},
 		Env: []string{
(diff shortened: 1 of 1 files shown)


Jeffail (approved):
LGTM, thanks @mihaitodor (and @gvenzl)

PR #1956, opened 2023-06-13
Azure storage account linting with development connection string

mfamador:
Don't fail on lint when there's a development connection string without an AccountName (UseDevelopmentStorage=true;)

### DIFF
--- a/internal/impl/azure/auth.go
+++ b/internal/impl/azure/auth.go
@@ -41,7 +41,7 @@ func azureComponentSpec(forBlobStorage bool) *service.ConfigSpec {
 		spec = spec.Field(service.NewStringField(bscFieldStorageSASToken).
 			Description("The storage account SAS token. This field is ignored if `" + bscFieldStorageConnectionString + "` or `" + bscFieldStorageAccessKey + "` are set.").
 			Default("")).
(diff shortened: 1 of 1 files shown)


Jeffail:
Do we need a patch release for this or can it wait a couple of weeks? It looks as though this only impacts test environments.

mfamador:
We're fine, we've created a custom tag, but anyone using azure storage on their integration tests will hit this.

mihaitodor:
As a workaround, they should be able to use `DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;` instead. Azurite shows `UseDevelopmentStorage=true;` as a potentially unsupported alternative [here](https://github.com/azure/azurite#connection-strings).

mfamador:
Sure, it will work as well.

Jeffail (approved):

PR #1960, opened 2023-06-14
Add `redis_scan` input.

### LINKED ISSUE #1898: Redis input

mfamador:
Add the `redis_scan` input, to allow us to scan the redis keys and retrieve their values, optionally matching a blob-style pattern.

Ex.
```yaml
input:
  redis_scan:
    url: 'redis://localhost:6379'
    match: '*4*'
``` 

It generates a message for each key value pair in json format:

Ex.
```json
{"key":"100044","value":"1658881920000000000"}
``` 

Resolves #1898

### DIFF
--- a/internal/impl/redis/integration_test.go
+++ b/internal/impl/redis/integration_test.go
@@ -189,6 +189,31 @@ input:
 		})
 	})
 
+	// SCAN
+	t.Run("scan", func(t *testing.T) {
+		t.Parallel()
+		template := `
+input:
+  redis_scan:
+    url: 'tcp://localhost:$PORT'
+    match: '*'
+`
+		client := redis.NewClient(&redis.Options{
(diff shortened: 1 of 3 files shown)


mfamador:
Added a small integration test.

Jeffail (approved):
LGTM, I might remove the max in flight parameter as I believe we should be fine with the standard back pressure of benthos.

PR #1969, opened 2023-06-20
Add MongoDB processor find-all operation

webfrank:
Hi, I've added find-all operation to MongoDB processor. Cannot use bson.MarshalExtJSON because for an array it returns the Error: WriteArray can only write a Array while positioned on a Element or Value but is positioned on a TopLevel


Jeffail:
Looks like we'll need to work around the array issue by iterating the returned BSON objects into a `[]json.RawMessage` and then marshal that into the final binary format. We also ideally need at least one unit test. I can do those two on my side when I merge but you need to add the sign off: https://github.com/benthosdev/benthos/pull/1969/checks?check_run_id=14410191847

webfrank:
Hi, don't know if I've done it correctly.

Jeffail:
Running `git commit --amend -s` would've done it, it just requires `Signed-off-by: foo <bar>`, looks like the PR got blown up from a merge (reporting 41 files changed).

PR #1971, opened 2023-06-20
Enable plugin authors to support connection backoffs

josephwoodward:
Closes: https://github.com/benthosdev/benthos/issues/885

This change adds the ability for plugin authors to return a `ErrBackOff` type that wraps the original error and allows authors to provide a `Wait` duration, causing the `Connect` function to pause for the duration in order slow down how quickly Benthos attempts to reconnect. It's based on your [your suggestion in this older PR](https://github.com/benthosdev/benthos/pull/886#issuecomment-933767400).

### DIFF
--- a/internal/component/errors.go
+++ b/internal/component/errors.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 )
 
 // ErrNotUnwrapped is returned in cases where a component was meant to be
@@ -47,6 +48,17 @@ var (
 
 	ErrFailedSend = errors.New("message failed to reach a target destination")
 )
+
+// ErrBackOff is an error returned that allows for a back off duration to be specified
+type ErrBackOff struct {
+	Err  error
+	Wait time.Duration
+}
+
+// Error returns the Error string.
+func (e ErrBackOff) Error() string {
+	return e.Err.Error()
+}
 
 //------------------------------------------------------------------------------
 
--- a/internal/component/input/async_reader.go
+++ b/internal/component/input/async_reader.go
@@ -124,7 +124,7 @@ func (r *AsyncReader) loop() {
 
 				var e component.ErrBackOff
 				if errors.As(err, &e) {
-					time.Sleep(e.Wait)
+					r.sleepWithCancellation(closeAtLeisureCtx, e.Wait)
 				}
 
 				nextBoff := r.connBackoff.NextBackOff()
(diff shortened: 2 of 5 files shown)


Jeffail:
Hey @josephwoodward, yeah this looks like the sort of thing I was suggesting, happy to have it added.

josephwoodward:
@Jeffail Great, it's a bit broken at the moment but I'll try and finish it off over the next few days.

internal/component/input/async_reader.go
```diff
@@ -122,6 +122,11 @@ func (r *AsyncReader) loop() {
 				r.mgr.Logger().Errorf("Failed to connect to %v: %v\n", r.typeStr, err)
 				mFailedConn.Incr(1)
 
+				var e component.ErrBackOff
+				if errors.As(err, &e) {
+					_ = r.sleepWithCancellation(closeAtLeisureCtx, e.Wait)
```
joseph-woodward-form3, line 127: @Jeffail Interested in your thoughts on this. Initially I was thinking a time.Sleep would be sufficient but it'd be good to be able to cancel it to which I found this solution which is a bit complicates things, perhaps unnecessarily.

internal/component/errors.go
```diff
@@ -48,6 +49,17 @@ var (
 	ErrFailedSend = errors.New("message failed to reach a target destination")
 )
 
+// ErrBackOff is an error returned that allows for a back off duration to be specified
+type ErrBackOff struct {
```
josephwoodward, line 53: Given this is internal I didn't feel a constructor function was necessary.

josephwoodward:
@Jeffail I think this is pretty much ready for review when you have the time. I was looking to add a test to [async_reader_test.go](https://github.com/benthosdev/benthos/blob/main/internal/component/input/async_reader_test.go) which will have to assert the test waited for the backoff duration, not sure if you had any better ideas on how to test this?

Jeffail (approved):
LGTM, there's a couple of things I'm going to move around in a follow up commit but it'll mostly stay the same. One significant change I'm thinking of making is to change the `ErrBackOff` to have a pointer receiver `Error` method. I don't think it's really necessary anymore with the new error handling stuff but it makes it consistent with our other error types.

Jeffail:
@josephwoodward here's the refactor: https://github.com/benthosdev/benthos/commit/00e641dd0efac8fdcb7e5a6dc458bd331a1239f7, if you feel strongly about any of the changes let me know as there's still time to make adjustments before the next release.

PR #1975, opened 2023-06-23
fix: azure table storage escape odata single quotes

davidandradeduarte:
With the current output we can't send single quote values to either PK/RK or it will fail.
We need to escape the single quote with another double quotes.
It's sad that we need to do this, one would think that the Azure SDK would do this for us. But right now it doesn't.

For example, something like this will fail:
```yaml
output:
  azure_table_storage:
    partition_key: "pk"
    row_key: "r'k"
```
```
HTTP/1.1 400 Bad Request
X-Content-Type-Options: nosniff
DataServiceVersion: 3.0;
Content-Type: application/json;odata=minimalmetadata;streaming=true;charset=utf-8

{"odata.error":{"code":"InvalidInput","message":{"lang":"en-US","value":"0:There is an unterminated string literal at position 30 in 'PartitionKey='pk',RowKey='r'k''.\nRequestId:01801433-1002-0089-55d3-a52db3000000\nTime:2023-06-23T13:06:34.0128916Z"}}}
```

And we need to do this instead:
```yaml
output:
  azure_table_storage:
    partition_key: "pk"
    row_key: "r''k"
```

This PR will only support the first approach, which is the one that I think makes more sense.

Let me know if this make sense. I will happily add a PR to the Azure SDK, but I'm not too confident that it will be merged soon.

### DIFF
--- a/internal/impl/azure/output_table_storage.go
+++ b/internal/impl/azure/output_table_storage.go
@@ -261,6 +261,8 @@ func (a *azureTableStorageWriter) execBatch(ctx context.Context, writeReqs map[s
 				var batch []aztables.TransactionAction
 				ne := len(entities)
 				for i, entity := range entities {
+					entity.PartitionKey = escapeODataString(entity.PartitionKey)
+					entity.RowKey = escapeODataString(entity.RowKey)
 					batch, err = a.addToBatch(batch, tt, entity)
 					if err != nil {
 						return err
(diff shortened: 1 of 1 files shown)


Jeffail:
Ooooof, yeah that's a doozy. I definitely don't like the prospect of double-escaping in cases where the user has already escaped their values. I think if there's little prospect of them implementing this within the lib itself then we should just add documentation for this particular snag and avoid the breaking change.

However, if they're definitely going to add this into the client library (and therefore we're on the hook for a breaking change either way) then it might be worth having a temporary implementation on our side so that we nip it in the bud, but then we also need to be hyper aware of upgrades because leaving that solution in place _after_ they fix it on their end we'd introduce a breaking change that ruins it for everyone.

PR #1976, opened 2023-06-23
feat: Interpolation support for exchange name

zaeemarshad:
## What

In our use case, benthos will be reading messages from file and the exchange name will be present as a field under the headers key. This change allows the exchange name to be interpolated and hence becomes dynamic.

Here's my sample benthos config for testing:

```yaml
logger:
  level: DEBUG
  format: logfmt
  add_timestamp: false
  static_fields:
    '@service': benthos

input:
  file:
    paths: [ /tmp/data/messages ]
    codec: lines

pipeline:
  processors:
    - mutation: |
        root = {}
        root.message = json("headers.message")
        meta exchange = json("headers.exchange")
    - log:
        message: 'Got exchange: ${!meta("exchange")}'

output:
  amqp_0_9:
    urls: [amqp://root:root@localhost:5672/]
    exchange: '${!meta("exchange")}'
    persistent: true
```

When I fed the following messages:

```json
{"headers":{"exchange":"test2", "message":"for test2", "routing_key":"test-rk"}}
{"headers":{"exchange":"test1", "message":"for test1", "routing_key":"test-rk"}}
```

the messages were published successfully and the exchanges were logged correctly:

```
INFO Running main config from specified file       @service=benthos path=/tmp/benthos.conf
INFO Consuming from file '/tmp/data/messages'      @service=benthos label="" path=root.input
ERRO HTTP Server error: listen tcp 0.0.0.0:4195: bind: address already in use  @service=benthos
INFO Got exchange: test2                           @service=benthos label="" path=root.pipeline.processors.1
INFO Got exchange: test1                           @service=benthos label="" path=root.pipeline.processors.1
INFO Sending AMQP messages to exchange: &{0xc0017b6d80}  @service=benthos label="" path=root.output
DEBU Waiting for pending acks to resolve before shutting down.  @service=benthos label="" path=root.input
DEBU Pending acks resolved.                        @service=benthos label="" path=root.input
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -86,6 +86,7 @@ require (
 	github.com/prometheus/client_golang v1.14.0
 	github.com/prometheus/common v0.39.0
 	github.com/pusher/pusher-http-go v4.0.1+incompatible
+	github.com/puzpuzpuz/xsync/v2 v2.4.1
 	github.com/quipo/dependencysolver v0.0.0-20170801134659-2b009cb4ddcc
 	github.com/rabbitmq/amqp091-go v1.4.0
 	github.com/rcrowley/go-metrics v0.0.0-20201227073835-cf1acfcdf475
--- a/go.sum
+++ b/go.sum
@@ -860,6 +860,8 @@ github.com/protocolbuffers/txtpbfmt v0.0.0-20201118171849-f6a6b3f636fc h1:gSVONB
 github.com/protocolbuffers/txtpbfmt v0.0.0-20201118171849-f6a6b3f636fc/go.mod h1:KbKfKPy2I6ecOIGA9apfheFv14+P3RSmmQvshofQyMY=
 github.com/pusher/pusher-http-go v4.0.1+incompatible h1:4u6tomPG1WhHaST7Wi9mw83Y+MS/j2EplR2YmDh8Xp4=
 github.com/pusher/pusher-http-go v4.0.1+incompatible/go.mod h1:XAv1fxRmVTI++2xsfofDhg7whapsLRG/gH/DXbF3a18=
+github.com/puzpuzpuz/xsync/v2 v2.4.1 h1:aGdE1C/HaR/QC6YAFdtZXi60Df8/qBIrs8PKrzkItcM=
+github.com/puzpuzpuz/xsync/v2 v2.4.1/go.mod h1:gD2H2krq/w52MfPLE+Uy64TzJDVY7lP2znR9qmR35kU=
 github.com/quipo/dependencysolver v0.0.0-20170801134659-2b009cb4ddcc h1:hK577yxEJ2f5s8w2iy2KimZmgrdAUZUNftE1ESmg2/Q=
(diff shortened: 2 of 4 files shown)


Jeffail (commented):
Hey @zaeemarshad, I'm in favour of adding interpolation here but calling `amqpChan.ExchangeDeclare` on every write is a big issue as that'll massively slow things down. Ideally we'd memoise the calls by the interpolated exchange name. Let me know if you're comfortable having a go at that otherwise I'll do it on my end when I have time.

zaeemarshad:
Hey @Jeffail, thanks for taking a look. The performance concern is fair but I don't know memoisation will fix the issue. Since, we are dealing with multiple go-routines, I think a shared cache structure probably at the package level will have to be created with the standard safety mechanisms against concurrent writes et all. What are your thoughts?

zaeemarshad:
Hi @Jeffail, apologies for going quite on this one. Please have a look at the updated PR. I have added a map and here's a log of messages showing the memoisation kicking in.

```
INFO Consuming from file '/Users/zaeem.arshad/benthos/messages'  @service=benthos label="" path=root.input
INFO Got exchange: test3                           @service=benthos label="" path=root.pipeline.processors.1
INFO Got exchange: test3                           @service=benthos label="" path=root.pipeline.processors.1
INFO Sending AMQP messages to exchange: &{0xc002378e40}  @service=benthos label="" path=root.output
INFO Got exchange: test3                           @service=benthos label="" path=root.pipeline.processors.1
DEBU Exchange test3 does not exist, declaring      @service=benthos label="" path=root.output
DEBU Exchange test3 exists in cache, not re-declaring  @service=benthos label="" path=root.output
INFO Got exchange: test3                           @service=benthos label="" path=root.pipeline.processors.1
DEBU Exchange test3 exists in cache, not re-declaring  @service=benthos label="" path=root.output
INFO Got exchange: test3                           @service=benthos label="" path=root.pipeline.processors.1
DEBU Exchange test3 exists in cache, not re-declaring  @service=benthos label="" path=root.output
INFO Got exchange: test3                           @service=benthos label="" path=root.pipeline.processors.1
DEBU Exchange test3 exists in cache, not re-declaring  @service=benthos label="" path=root.output
```

zaeemarshad:
@Jeffail Is there anything else that you think I can improve on?

Jeffail:
Hey @zaeemarshad, couple things came up but I'm fixing them after merge:

- Having a singleton to store declared exchange names is problematic as it prevents memory getting reclaimed in cases where the outputs are rotated dynamically, and it also has the unlikely problem of duplicate exchange names across different clusters would prevent one of them from being declared. I've fixed that by moving it into the component itself (I've also removed the package dependency as a lock should be fine).
- In cases where the exchange is static we need to declare it within the connect block. I realised when running the integration tests (which were broken with this PR) that they fail because the output needs to send data in order to declare the exchange, but the input cannot subscribe without it being declared, so a few items of data would almost always be lost. We can't solve that when the exchange is dynamic as it isn't known until that point, however I've added the ability to detect when the exchange is static and declare it in the connect block so that we remain backwards compatible with the old behaviour.

PR #1980, opened 2023-06-26
preserve *service.BatchError generated by plugins

disintegrator:
This change ensures that any `*service.BatchError`s that are returned from various plugins are preserved including their internal state. Previously, the air-gap layer of Benthos was intercepting errors generated by plugins and converting these between `*service.BatchError` and the internal `*batch.Error` types. These conversions were buggy because we would often lose the underlying `*message.SortGroup`. When plugin developers received `*service.BatchError`, in some cases, they were unable to iterate over the contained message errors with the `WalkMessages` method because the sort group was often `nil`.

### DIFF
--- a/public/service/errors.go
+++ b/public/service/errors.go
@@ -74,19 +74,16 @@ func (err *BatchError) Unwrap() error {
 // If the provided error is not nil and can be cast to an internal batch error
 // we return a public batch error.
 func toPublicBatchError(err error) error {
+	// Modern Benthos components that use the public service API will return a
+	// *service.BatchError type. We will preserve this if we encounter it.
+	var target *BatchError
+	if ok := errors.As(err, &target); ok {
+		return err
+	}
+
 	var bErr *batch.Error
 	if err != nil && errors.As(err, &bErr) {
 		err = &BatchError{wrapped: bErr}
-	}
-	return err
-}
-
-// If the provided error is not nil and can be cast to a public batch error we
-// return the internal batch error.
-func fromPublicBatchError(err error) error {
-	var bErr *BatchError
-	if err != nil && errors.As(err, &bErr) {
-		err = bErr.wrapped
 	}
 	return err
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 - HTTP components no longer ignore `proxy_url` settings when OAuth2 is set.
 - The `PATCH` verb for the streams mode REST API no longer fails to patch over newer components implemented with the latest plugin APIs.
 - The `nats_jetstream` input no longer fails for configs that set `bind` to `true` and do not specify both a `stream` and `durable` together.
+- Fixed a bug where `(*service.BatchError).WalkMessages` did not iterate over the underlying errors.
 
 ### Changed
 
--- a/public/service/errors_test.go
+++ b/public/service/errors_test.go
@@ -68,3 +68,10 @@ func TestMockWalkableError_OmitSuccessfulMessages(t *testing.T) {
 
 	require.Equal(t, err.IndexedErrors(), 2, "indexed errors did not match size of batch")
 }
+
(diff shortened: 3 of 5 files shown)


public/service/errors.go
```diff
@@ -74,19 +74,16 @@ func (err *BatchError) Unwrap() error {
...
+	var target *BatchError
+	if ok := errors.As(err, &target); ok {
+		return err
 	}
-	return err
-}
 
-// If the provided error is not nil and can be cast to a public batch error we
-// return the internal batch error.
-func fromPublicBatchError(err error) error {
```
disintegrator, line 86: `fromPublicBatchError` is no longer required because the public error type unwraps into the internal `*batch.Error` type.

public/service/errors.go
```diff
@@ -74,19 +74,16 @@ func (err *BatchError) Unwrap() error {
...
-	var bErr *BatchError
+	var bErr *batch.Error
 	if err != nil && errors.As(err, &bErr) {
-		err = bErr.wrapped
+		err = &BatchError{wrapped: bErr}
```
disintegrator, line 86: This existing bit of code is still problematic because we are not initializing the `BatchError` with a sort group. Any `BatchError`s returned from this code path are still not iterable. I'm not sure how we should get that into here and perhaps start using `NewBatchError` instead of this struct literal.

Jeffail:
This is a bit of doozy, I think we need to deprecate and warn against using the walk method as it currently exists. The problem here is that I implemented batch errors and realised afterwards (after a few releases) that in order to actually utilise them we need to establish the correct order of messages within the batch with respect to the original batch rather than the state that it was in during the error.

This is necessary for pipelines where you have, say, a broker followed by an output that has processors for mutating the batch. Imagine a `broker -> re-ordering processor -> output` chain where the output creates a batch error where only a given index failed and returns that upstream to the broker. The broker now has the ability to extract that batch error and use it as a way to determine that only a subset of the batch needs to be retried. The broker cannot naively use the indexes yielded by the processor because the re-ordering processor has changed them.

Therefore, what we need is for the broker to create a context-based sort group for the batch _before_ dispatching it to the output, then the batch error that is returned (which contains the batch at the time of the error being created) is able to reconcile the indexes it knows of with the sort group provided by the broker.

This doesnt work for the broker if the sort group was established by the child output, because that still yields indexes that do not align with the original batch it has knowledge of. This was the original reason why I didn't add the sort group to the public batch error during the last refactor because it only serves a value to plugin authors that interact directly with other public plugins (and therefore know the exact state of the batch it received).

In reality if you're executing a config that goes from `plugin output -> plugin output` and you're executing the child output as a resource, if you were to extract a `service.BatchError` you still cannot trust the sort group that it came with as the pipeline could be `plugin output -> re-ordering processor -> public output`, or even `plugin output -> native broker -> re-ordering processor -> public output` without your knowledge.

What you'd need to do instead is have a sort group handy extracted from the batch you're working with, and then if you receive a batch error you provide the walk method with your own sort group (essentially the same API as the internal implementation).

I should have made this clearer in https://github.com/benthosdev/benthos/pull/1837, I thought at the time this was for simpler use cases where you are testing components or maybe spawning child components directly rather than referencing them via config resources. I hadn't really considered the use case where a custom plugin is itself a high level broker.

I think in order to proceed we need to:
- Expose the ability to create a sort group to public plugin authors, documented as a sort of "advanced feature"
- Add a new walk method to the batch error that takes the sort group into account
- Deprecate the current walk method and mark it as problematic

This stuff is a nightmare to work on but let me know if you want to take a shot at it, otherwise I'm happy to take it.

Jeffail:
I went with this in the end instead: https://github.com/benthosdev/benthos/commit/5225e1a58a1bf102e05d6285e62df24591fb757e, added a whackload of comments that hopefully make things clearer, and the new APIs put an emphasis on the fact that you need to preserve some record of ordering relative to a particular batch if you want to re-associate the indexes from the error to it.

Issue #615, opened 2021-01-12
Remove ugly hack from Azure test


mihaitodor:
The time parsing bug has been [fixed](https://github.com/Azure/Azurite/issues/663) in Azurite (that was fast!), so the next release should allow us to drop [this](https://github.com/Jeffail/benthos/blob/4996fbffd0a43e25b24446c356db2b2314620ce9/lib/test/integration/azure_test.go#L42-L55) ugly hack. I'll make this change once they push a new release.

Unfortunately, we still need to [hack the host](https://github.com/Jeffail/benthos/blob/4996fbffd0a43e25b24446c356db2b2314620ce9/lib/test/integration/azure_test.go#L31) on each request if we want to use a random port for Azurite instead of letting it start on port 10000, which might clash with other things.

Jeffail:
Awesome, thanks @mihaitodor for raising that issue and for the follow up.

mihaitodor:
Before I can enable this test, there are a couple more bugs which needs to be fixed:
- [x] https://github.com/Azure/Azurite/issues/682 - Released in 3.11.0
- [ ] https://github.com/Azure/Azurite/issues/702 - I fear this might not get fixed anytime soon...

PR #1989, opened 2023-07-01
feat: migrate input/output azure_queue_storage to most recent SDK azqueue

### LINKED ISSUE #1084: Azure authentication via managed identity
Any interest in supporting [managed identity](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview) authentication for the Azure components? I'd be happy to attempt a PR to implement it.

eduardodbr:
This PR migrates the azure_queue_storage input and output to the most recent [SDK](https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/storage/azqueue). This means that benthos no longer needs the dependency for the old SDK.  It also ensures all azure inputs and outputs can be used with any of the available connection methods: connection string, accountName/accountKey, SASToken, or managed identity. 

The queue input/output was also allowing to set an env var `AZURITE_QUEUE_ENDPOINT_PORT` to access azurite when the connection string was set to `UseDevelopmentStorage=true;`. I maintained this behavior and now it is also possible to do the same with blob and table using `AZURITE_BLOB_ENDPOINT_PORT` and `AZURITE_TABLE_ENDPOINT_PORT`. However, this is not documented anywhere, neither the `UseDevelopmentStorage=true;` support nor the env var override. I don't know where is the best place to add those docs., so please let me know.

I believe you can close #1084 and #615


internal/impl/azure/auth.go
```diff
@@ -122,16 +108,48 @@ func getEmulatorConnectionString(blobServicePort, queueServicePort, tableService
 	)
 }
 
+const (
+	azuriteBlobPortEnv  = "AZURITE_BLOB_ENDPOINT_PORT"
+	azuriteQueuePortEnv = "AZURITE_QUEUE_ENDPOINT_PORT"
+	azuriteTablePortEnv = "AZURITE_TABLE_ENDPOINT_PORT"
+)
+
+func parseStorageConnectionString(storageConnectionString, storageAccount string) string {
```
mihaitodor, line 117: The name `parseStorageConnectionString` doesn't seem to fit here, since this function is used for all three services. I'd consider using the following signature instead: `func getConnectionString(connectionString, account string) string`. On the happy path, it returns the input `connectionString` as-is.

internal/impl/azure/input_queue_storage.go
```diff
@@ -119,60 +121,60 @@ func (a *azureQueueStorage) ReadBatch(ctx context.Context) (batch service.Messag
...
-				msgLag := 0
-				if approxMsgCount >= n {
-					msgLag = int(approxMsgCount - n)
-				}
-				part.MetaSetMut("queue_storage_message_lag", msgLag)
-			}
-			for k, v := range metadata {
-				part.MetaSetMut(k, v)
+	n := int32(len(dequeue.Messages))
+	props, _ := queueClient.GetProperties(ctx, nil)
```
mihaitodor, line 150: I wonder if it's safe to swallow this error (even though the original code did that as well)

mihaitodor (commented):
Nice job @eduardodbr! I left a few nitpicks

PR #1990, opened 2023-07-03
Add option to select pulsar topics with regex

treuherz:
Pulsar clients support consuming multiple topics dynamically by providing a regex (see [the TopicsPattern field](https://pkg.go.dev/github.com/apache/pulsar-client-go/pulsar#ConsumerOptions)). This change adds that ability as a config option for the Pulsar input.

Closes https://github.com/benthosdev/benthos/issues/1988

### DIFF
--- a/internal/impl/pulsar/input.go
+++ b/internal/impl/pulsar/input.go
@@ -21,16 +21,26 @@ const (
 func init() {
 	err := service.RegisterInput(
 		"pulsar",
-		service.NewConfigSpec().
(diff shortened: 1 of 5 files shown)


Jeffail (approved):
LGTM, thanks @treuherz! I'm going to remove the default values for `topics` and `topic_pattern` and instead mark them both as optional, just makes it a bit easier in future if we decide to deprecate the fields.

PR #2003, opened 2023-07-17
chore(deps): bump sarama to v1.40.0

dnwe:
Note: the module path has changed to github.com/IBM/sarama since ownership transitioned away from Shopify

This restores the original commit offsets AddBlock API as notified under IBM/sarama#2358


Jeffail (approved):
Fantastic, thanks for the fix and the follow up @dnwe!

PR #2004, opened 2023-07-17
fix incorrect CUE codegen for various components and fields

disintegrator:
This PR addresses a couple of bugs in the CUE schema generated by Benthos. Namely, the `switch` / `group_by` processors were not defined to take an array of clauses and the 2d-array of `workflow.order` (an any other similar fields) was defined as `[[...T]]` instead of `[...[...T]]`. The first problem was fixed by simplify the code generation logic for Benthos components. The latter required a specific fix.


Jeffail (approved):
LGTM, thanks @disintegrator!

PR #2006, opened 2023-07-21
add decimal support to gauge metrics

fouadkada:
# Context

The [metric](https://www.benthos.dev/docs/components/processors/metric/) processor accepts several types which one of them is `gauge`.

One expects that a decimal value could be set for a gauge.

In the current implementation, and because Benthos supports exporting metrics to multiple backends, and since not all of them support decimal gauge values, `int64` is the common denominator between all of them.

Of course, there are scenarios where one wants to set a gauge to a decimal value without resorting to workarounds, and some metrics platforms do support decimal gauges such as prometheus.

# Proposed solution

The current proposed solution breaks the existing `StatGauge` interface into two specialized interfaces, one for integers and another for floats.

All structs that implements this `StatGauge` now also need to implement the `StatGaugeFloater` interface. This way they also need to support decimal gauge metrics. In order to keep backward compatibility and since not all metrics platforms support decimal gauges, these new decimal functions forward there value to the existing integer functions, this way nothing breaks.

Prometheus is the only exception since it does support decimal gauges thus its implementation for decimal gauges uses actual decimal values.

An added value with this approach is that if another metric platform supports decimals(influxDB for example), the amount of changes and the implementation is straight forward.


internal/impl/pure/processor_metric.go
```diff
@@ -299,21 +299,42 @@ func (m *metricProcessor) handleCounterBy(val string, index int, msg message.Bat
...
+			return errors.New("value is negative")
+		}
+		parsedVal = i
+	} else {
+		f, err := strconv.ParseFloat(val, 64)
+		if err != nil {
+			return err
+		}
+		if f < 0 {
+			return errors.New("value is negative")
```
ggerritsen, line 315: Even though it doesn't happen yet elsewhere in this file, it might be nice to include the actual value in the error message, to make troubleshooting easier

internal/component/metrics/combine.go
```diff
@@ -66,19 +66,31 @@ type combinedGauge struct {
...
+func (c *combinedGauge) SetInt64(value int64) {
+	c.c1.SetInt64(value)
+	c.c2.SetInt64(value)
 }
 
-func (c *combinedGauge) Incr(count int64) {
-	c.c1.Incr(count)
-	c.c2.Incr(count)
+func (c *combinedGauge) SetFloat64(value float64) {
+	c.SetInt64(int64(value))
```
andrevmatos, line 75: This needs to passthrough value to both gauges:
```suggestion
	c.c1.SetFloat64(value)
	c.c2.SetFloat64(value)
```

internal/component/metrics/combine.go
```diff
@@ -66,19 +66,31 @@ type combinedGauge struct {
...
+}
+
+func (c *combinedGauge) IncrInt64(count int64) {
+	c.c1.IncrInt64(count)
+	c.c2.IncrInt64(count)
+}
+
+func (c *combinedGauge) IncrFloat64(count float64) {
+	c.IncrInt64(int64(count))
```
andrevmatos, line 84: Same:
```suggestion
	c.c1.IncrFloat64(count)
	c.c2.IncrFloat64(count)
```

internal/component/metrics/combine.go
```diff
@@ -66,19 +66,31 @@ type combinedGauge struct {
...
+func (c *combinedGauge) DecrInt64(count int64) {
+	c.c1.DecrInt64(count)
+	c.c2.DecrInt64(count)
 }
 
-func (c *combinedGauge) Decr(count int64) {
-	c.c1.Decr(count)
-	c.c2.Decr(count)
+func (c *combinedGauge) DecrFloat64(count float64) {
+	c.DecrInt64(int64(count))
```
andrevmatos, line 93:
```suggestion
	c.c1.DecrFloat64(count)
	c.c2.DecrFloat64(count)
```

internal/component/metrics/type.go
```diff
@@ -22,14 +22,34 @@ type StatTimer interface {
 // StatGauge is a representation of a single gauge metric stat. Interactions
 // with this stat are thread safe.
 type StatGauge interface {
-	// Set sets the value of a gauge metric.
-	Set(value int64)
+	StatGaugeInter
+	StatGaugeFloater
```
andrevmatos, line 26: The same treatment could be done for `StatCounter` (easier, passthrough to gauge) and maybe `StatTimer`

andrevmatos (commented):
Left some comments; the `combine.go` ones, specifically, is what I had to change for my Prometheus "real life custom metric" to work.

fouadkada:
closing this PR to make using this fork in a project easier since it seems the `replace` functionality of `go.mod` struggle with branches

internal/component/metrics/type.go
fouadkada: I will work on `StatCounter` next

andrevmatos (approved):
I've tested both `gauge` and `counter_by` with custom metrics, and this is working great, passing floats from messages down to Prometheus metrics exporter without casting to ints.
LGTM.

Jeffail:
Sorry, wrong button, didn't mean to close it 😅

andrevmatos:
@Jeffail `IncrFloat64` is specifically useful for `counter_by` where `value` is a float: we have the usecase where we need to accumulate some metric which is a fractional number, and hence `float64 counter_by` covers exactly this.

fouadkada:
> Hey @fouadkada, this change: https://github.com/benthosdev/benthos/pull/2006/files#diff-1b58fd65718887790715296f6406c57479dc74172143db9f63dd0aa3b280839bR62, is breaking and therefore we can't do it without a major version release. I think I'd prefer to minimize the impact of this PR by leaving all of the existing methods as they are and adding a new `SetFloat64` method specifically to the gauge metrics types.

Thanks a lot @Jeffail for you review, since the original interface that housed the `Incr()` method now has 2 child interfaces(1 for ints and the other one for floats) and since the `Incr() `has been replace with `IncrInt64` everywhere, I feel that this is not a breaking change. The implementors of this interface that do support ints and floats(prometheus for example) make use of both floats and ints implementations. The implementors that do not support floats, cast the float value into an int and act like `Incr()` used to act.

fouadkada:
> Hey @fouadkada, this change: https://github.com/benthosdev/benthos/pull/2006/files#diff-1b58fd65718887790715296f6406c57479dc74172143db9f63dd0aa3b280839bR62, is breaking and therefore we can't do it without a major version release. I think I'd prefer to minimize the impact of this PR by leaving all of the existing methods as they are and adding a new `SetFloat64` method specifically to the gauge metrics types.
> 
> I'm also not sure we need the `IncrFloat64` addition, is there a specific use case for that? If not then I think we ought to omit it as well for now.

@Jeffail the requested changes have been made, the breaking change of modifying `Incr()` to `IncrInt64()` has been reverted back, I would appreciate it if you can take another look at this please

public/service/metrics.go
```diff
@@ -125,12 +135,24 @@ type MetricsExporterGaugeCtor func(labelValues ...string) MetricsExporterGauge
 // MetricsExporterCounter represents a counter metric of a given name and
 // labels.
 type MetricsExporterCounter interface {
```
Jeffail, line 137: Modifying this interface is another breaking change, it means any existing implementations of `MetricsExporterCounter` will fail to compile if we merge this PR, which we can't have.

The backwards compatible way to do it would be to check for `interface{ IncrFloat64(count float64) }` and so on inside the airgap and fallback to the integer equivalents when it isn't implemented.

Jeffail (changes requested):
@fouadkada it's getting closer but there's still breaking changes here.

public/service/metrics.go
fouadkada: @Jeffail I'm sorry but I didn't fully get that, why would it fail to compile since current and previous versions of Benthos simply lack these interfaces and these newly added interfaces are implemented where needed(airGapGauge, airGapCounter...)?

I created a local sample application that makes use of these changes, that's a testing config file
```
http:
  enabled: true
  address: 0.0.0.0:4195
  root_path: /
  debug_endpoints: false

input:
  label: ""
  generate:
    mapping: |
      root.msg = "hello world"
      root.value = 10.2
    interval: 1s

pipeline:
  threads: -1
  processors:
    - metric:
        type: "counter_by"
        name: "hello_metric"
        value: ${! this.value }

output:
  drop: { }

metrics:
  prometheus:
    use_histogram_timing: true
    add_process_metrics: true
    add_go_metrics: true

shutdown_timeout: 20s
```
and the following is an extract of `go.mod` to make use of my changes:
```
module benthos-metrics

go 1.21

replace github.com/benthosdev/benthos/v4 v4.21.0 => /Users/fouad/workspace/personal/benthos

require github.com/benthosdev/benthos/v4 v4.21.0

require (
	...
)
```

even though Benthos version `4.21.0` lacks all of these changes, the application compiles and runs as expected.

@Jeffail I'd really appreciate it if you can elaborate a little bit more on this

Jeffail: @fouadkada if a plugin author has implemented the old `MetricsExporterCounter` and then you add more methods to it their implementation will no longer be valid. Anyone using that hypothetical plugin will have a failed build once they upgrade to the version of benthos containing your expanded version.

This stuff may seem minor, as it's only a quick fix for the plugin author to implement the new methods. However, the reality is these changes can be extremely problematic as it breaks semantic versioning and a module that imports both benthos among a few plugins doesn't have control over whether/when the plugin authors can update to your new interface.

There is no backwards compatible way to expand or change a public interface. You need to instead allow plugin authors to optionally implement the expanded set of methods and then internally make those checks yourself (or fall back in cases where it's not implemented).

If you're not sure how to do that then that's fine, I can do it myself when I next get a chance.

fouadkada: @Jeffail that makes a lot more sense now. The moment you mentioned custom plugin implementing the `MetricsExporterCounter` interface then I can now see why it's a breaking change.

I'd really appreciate it you can expand a little bit more on the following with an example or something then I can try to implement it myself:
> The backwards compatible way to do it would be to check for interface{ IncrFloat64(count float64) } and so on > inside the airgap and fallback to the integer equivalents when it isn't implemented.

Jeffail: Okay this is next on my priority list for PRs, I'll work through the breaking changes myself as I get it merged as it's hard to describe without just writing it.

PR #2014, opened 2023-07-27
Adding parse_jwt_rs methods

iquirino:
I am adding the jwt sign parse methods: 🔐

 - sign_jwt_es256  🥰
 - sign_jwt_es384  😍
 - sign_jwt_es512 🤩

 - parse_jwt_rs256  🥰
 - parse_jwt_rs384  😍
 - parse_jwt_rs512 🤩

 - parse_jwt_es256  🥰
 - parse_jwt_es384  😍
 - parse_jwt_es512 🤩

Thank you!


torfjor:
Hi @iquirino. Great work! I have a use case where I need to validate [Identity-Aware Proxy assertions](https://cloud.google.com/iap/docs/signed-headers-howto#iap_validate_jwt-go). They use ES256 as the signing method, which has become quite popular. Any chance you could take a stab at implementing that while you're "in the flow"?

Maybe we should cook up a complete example with a processor that fetches your issuer's JWKs, puts them in a cache and uses the cached JWK to validate a token?

iquirino:
> Hi @iquirino. Great work! I have a use case where I need to validate [Identity-Aware Proxy assertions](https://cloud.google.com/iap/docs/signed-headers-howto#iap_validate_jwt-go). They use ES256 as the signing method, which has become quite popular. Any chance you could take a stab at implementing that while you're "in the flow"? pray
> 
> Maybe we should cook up a complete example with a processor that fetches your issuer's JWKs, puts them in a cache and uses the cached JWK to validate a token?

Yes, I can do this, just wait a few minutes ;)

iquirino:
> Hi @iquirino. Great work! I have a use case where I need to validate [Identity-Aware Proxy assertions](https://cloud.google.com/iap/docs/signed-headers-howto#iap_validate_jwt-go). They use ES256 as the signing method, which has become quite popular. Any chance you could take a stab at implementing that while you're "in the flow"? pray
> 
> Maybe we should cook up a complete example with a processor that fetches your issuer's JWKs, puts them in a cache and uses the cached JWK to validate a token?

Hey @torfjor,

I've added ES/ECDSA sign and parse methods to my branch/PR.

;)

mihaitodor (changes requested):
Hey @iquirino, thank you for working on these extra methods! I finally got a chance to have a look over the code. Thanks for bearing with me! Awesome that you also added the ECDSA ones too!

Don't forget to `go test github.com/benthosdev/benthos/v4/cmd/tools/benthos_docs_gen/...` because it's set up to test some of the sample input + expected output code snippets in the docs. In this case, if you provide a sample input and the expected output, the sample input needs to be valid and the expected output needs to match. Unfortunately, that doesn't work well for cases where a long RSA key is needed as input, since that will make the docs look really ugly. That's why I omitted those in #1970. I'm not sure if @Jeffail has a good suggestion right now on how to add some usage examples that aren't picked up by the tests, but, if you really want to try adding some inline markdown, I guess you can do it in the `summary` field of `spec.Example()`, which is currently set to an empty string (see line 75 in `internal/impl/crypto/jwt_parse.go` in this PR) and then you can leave the `inputOutputs` field as not set when the `sampleSignature` isn't present.

peczenyj:
Hey @iquirino do you think you can sign your commits?

iquirino:
@peczenyj I'm sorry, I was too busy last weekend. 🤯
@mihaitodor I was unable to sign because I didn't added a GPG 🔐 Signature on my current computer. I've already fixed it.

On a talk with @Jeffail on Discord, I was able to manage a new type of example, a NotTestedExample. Which will be skipped by the bloblang documentation tests. 😶‍🌫️

The sign methods generates random signatures, which makes impossible to assert it.
https://stackoverflow.com/questions/16325057/why-does-rsa-encrypted-text-give-me-different-results-for-the-same-text 🥸

Thank you guys! 🤩

mihaitodor:
Thanks for the updates @iquirino. I think the tests look OK now, but I'll let @Jeffail have a closer look

Issue #1797, opened 2023-03-08
[BUG] Custom INPUT HTTP_SERVER doesn't support path variables.


silvioramalho:
## Context

Setting up a scenario with INPUT/HTTP_SERVER, with a variable in the first parameter of the path. 

* Example: "/{varName}"

For this, in the YAML file, it was necessary to assign an ADDRESS to the HTTP_SERVER, where Benthos creates a new HTTP_SERVER, different from the wide server.

```
http_server:
    address: 0.0.0.0:4197
    path: /{costomer_id}
    timeout: 5s
    allowed_verbs:
      - POST
    sync_response:
      status: "200"
      headers:
        Content-Type: application/json
```

Upon reviewing the code, I identified that this new HTTP_SERVER does not use the Gorilla Mux ROUTER to create routes. As a result, there is no extraction of variables included in the PATH. Therefore, the behavior of the wide server is different from that of an alternative HTTP_SERVER.

Jeffail:
Thanks for raising this @silvioramalho, should be fixed with https://github.com/benthosdev/benthos/commit/940af7b81ed3f561a70907e8d99e3c5987f8d67c

PR #2019, opened 2023-07-31
Fix HTTP server path prefix matching

### LINKED ISSUE #2018: http_server: Breaking change introduced with v4.14.0?
Hi,

I've just upgraded from v4.11.0 to v4.18.0.
This broke my streams that use the `http_server` input.

You can replicate it by starting Benthos in streams mode with these two files and the `-prefix-stream-endpoints=false` option:
[streams.zip](https://github.com/benthosdev/benthos/files/12215482/streams.zip)

This will make the HTTP route `/api/v2/entries` capture all requests made to `/api/v2/entries/{id}`. You can test that by making a GET request to "/api/v2/entries/foo" which will return "405 Method not allowed". This is because GET is not allowed by `/api/v2/entries`.

I think I found the change that introduces this behaviour: [#61a](https://github.com/benthosdev/benthos/commit/61aa1d860051ced3a6540a30cc3f08b88621fb99)
Specifically using `t.mux.PathPrefix(path)` creates this issue. That way, the first matching prefix captures all paths that have this prefix.
See also: https://stackoverflow.com/questions/28574875/go-gorilla-mux-not-routing-as-expected

So it doesn't seem to be possible to register routes that begin with a prefix that has already been registered before, such as "/foo" and "/foo/{name}". Is this correct?
This is a big limitation for us. Is this just an oversight or intented behaviour?

Thanks!

kabukky:
Fixes #2018 and should still support the use-case ( #1797 ) that introduced this behaviour in https://github.com/benthosdev/benthos/commit/61aa1d860051ced3a6540a30cc3f08b88621fb99

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -264,8 +264,8 @@ func (t *Type) RegisterEndpoint(path, desc string, handlerFunc http.HandlerFunc)
 			h(w, r)
 		})
 
-		t.mux.PathPrefix(path).Handler(wrapHandler)
-		t.mux.PathPrefix(t.conf.RootPath + path).Handler(wrapHandler)
+		t.mux.Path(path).Handler(wrapHandler)
+		t.mux.Path(t.conf.RootPath + path).Handler(wrapHandler)
 	}
 	t.handlers[path] = handlerFunc
 }


peczenyj:
Hey @kabukky do you think you can sign your commit?

PR #2022, opened 2023-08-02
Add v0_msg_exists_meta to Javascript

PKuebler:
Moin,

the function v0_msg_get_meta triggers a panic if the key is not found. Similar to Bloblang, I added a function v0_msg_exists_meta to allow meta keys to be optional.

v0_msg_get_meta:

Return error:
https://github.com/benthosdev/benthos/blob/main/internal/impl/javascript/functions.go#L215

Panic if error
https://github.com/benthosdev/benthos/blob/main/internal/impl/javascript/vm.go#L61

Greetings


Jeffail (approved):
LGTM, thanks @PKuebler!

PR #2024, opened 2023-08-03
chore(deps): bump franz-go to v1.14.3

mr-karan:
A critical consumer bug is fixed in newer franz-go update.

> This patch fixes an internal logic race that can be easily encountered when specifying exact offsets to consume from. If you encountered this bug, your consumer could just stop consuming for an indeterminite amount of time.

Ref: https://github.com/twmb/franz-go/blob/master/CHANGELOG.md#v1142


peczenyj:
Hey @mr-karan do you think you can sign your commits?

mr-karan:
@peczenyj Sure thing. Have pushed after signing.

PR #2028, opened 2023-08-04
DRAFT Add probabilistic in-memory caches bloom and cuckoo for dedup/cached with sharding

peczenyj:
I add support to two probabilistic filters (bloom and cuckoo) that can be used as cache on dedupe processor

optionally, such caches can dump and restore the state from filesystem. useful to avoid start the daemon without any data

Motivation: I need to perform a dedupe on a huge amount of data, and this pull request may save me memory / resources

this MR can be combined with https://github.com/benthosdev/benthos/pull/2123 in the future


internal/impl/pure/cache_cuckoo.go
```diff
@@ -0,0 +1,324 @@
...
+
+This cache can try to dump the content of the cache in disk during the benthos shutdown.
+
+Also, the cache can try to restore the state from an existing dump file. Errors will be ignored on this phase.
+
+This field accepts two kinds of value:
+
+If the path contains a single file with extension '.dat', it will be used for I/O operations.
+
+If the path constains a directory, we will try to use the most recent dump file (if any).
```
mihaitodor, line 80: Typo: `constains`

internal/impl/pure/cache_bloom.go
```diff
@@ -0,0 +1,498 @@
...
+
+This field accepts two kinds of value:
+
+If the path contains a single file with extension '.dat', it will be used for I/O operations.
+
+If the path constains a directory, we will try to use the most recent dump file (if any).
+
+If necessary, we will create a file with format 'benthos-bloom-dump.<timestamp>.dat'
+`).
+				Examples("/path/to/bloom-dumps-dir", "/path/to/bloom-dumps-dir/dump.dat").
```
mihaitodor, line 89: Is `/path/to/bloom-dumps-dir/dump.dat` a valid example given the description?
> If necessary, we will create a file with format 'benthos-bloom-dump.<timestamp>.dat'

internal/impl/pure/cache_cuckoo.go
```diff
@@ -0,0 +1,324 @@
...
+
+If necessary, we will create a file with format 'benthos-cuckoo-dump.<timestamp>.dat'
+`).
+				Examples("/path/to/cuckoo-dumps-dir", "/path/to/cuckoo-dumps-dir/dump.dat").
```
mihaitodor, line 84: Same question about the example

internal/impl/pure/cache_bloom.go
peczenyj: Hello, I will expand this section to be more clear.
In fact we can specify any file or directory but if we do not specify a file, we will generate one using the given convention + timestamp nanosecond to avoid conflicts.

my plan is to prepare flushes over time, to be possible, for instance, keep filters of 24 hours and use a sliding window of 7 days. It means if we can dedupe events in the last seven days.

also I’d like to add a periodic cleansing to wipe the cache should be nice

however… using files to manage state has some challenges. We may need one directory per benthos instance to avoid clashing, etc. I just want to add something really simple and if it doesn’t not fit, people can send a pull request to enhance it.

Using redis is “easier” because we can set a ttl on each bloom or cuckoo index.

I have a specific problem to solve that is dedup batches of entries without use a lot of memory.  However each time the benthos start it will find an empty filter…

we will see, besides dedupe I can’t see any other usage of bloom / cuckoo

internal/impl/pure/cache_cuckoo.go
peczenyj: Hello, I improve the examples for this section.

peczenyj:
@mihaitodor do you think it is ok now ?
@Jeffail what do you think?

Issue #1957, opened 2023-06-14
Support for interpolation on cached processor ttl field


margamraviteja:
`cache` processor support _interpolation_ on `ttl` field but the `cached` processor doesn’t support _interpolation_ on `ttl` field

Jeffail:
Done: https://github.com/benthosdev/benthos/pull/2031

PR #2031, opened 2023-08-08
add support to interpolation on field ttl on cached processor - fix issue #1957

peczenyj:
try to fix issue #1957


peczenyj:
Hey @Jeffail I note we are using the deprecated method String instead TryString in some packages. Do you want me change it here or perhaps in another PR?

Jeffail (approved):
LGTM, thanks @peczenyj!

PR #2032, opened 2023-08-09
Add secure option to open_telemetry_collector tracer

adamvr:
## Overview

Currently `benthos` always connects to open telemetry trace collectors insecurely (i.e. over http or insecure gRPC), which causes tracing to fail if the trace collector is expecting a secure connection.

## Description

This PR adds the `secure` option to the `open_telemetry_collector` tracer, which configures the otel trace collector to use HTTPS or gRPC over TLS if set to true.

## Other thoughts

- An alternative to a new option would be to specify secure / insecure as part of the tracer URL (e.g. `url: https://otlp.collector.com`). I decided against this since specifying them separately better matches `otlptrace.New`'s API and avoids having to do a bunch of dumb URL parsing, string matching and invalid schema handling.
- Thanks for maintaining such a great library @Jeffail, @mihaitodor and all the others! :heart:

### DIFF
--- a/internal/impl/otlp/tracer_otlp.go
+++ b/internal/impl/otlp/tracer_otlp.go
@@ -25,11 +25,17 @@ func init() {
 			service.NewURLField("url").
 				Description("The URL of a collector to send tracing events to.").
 				Default("localhost:4318"),
+			service.NewBoolField("secure").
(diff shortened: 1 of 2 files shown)


Jeffail (approved):
LGTM, thanks @adamvr!

Issue #2033, opened 2023-08-09
Panic under go 1.21


matthewhughes-uw:
Hello, I noticed a panic when trying to run some code that used this library under Go 1.21. This can be reproduced with the test suite in this repository, the tests run fine on 1.20:

```console
$ go version
go version go1.20.2 linux/amd64
ok  	github.com/benthosdev/benthos/v4/cmd/tools/benthos_docs_gen	1.077s
```

But there's a panic on 1.21:

```console
$ go install golang.org/dl/go1.21.0@latest
$ go1.21.0 download
$ go1.21.0 test github.com/benthosdev/benthos/v4/cmd/tools/benthos_docs_gen 
panic: parse mapping: line 6 char 45: unrecognised method 'parse_duration'
  |
6 | let backfill_seconds = this.backfill_period.parse_duration() / 1000000000
  |                                             ^---

goroutine 1 [running]:
github.com/benthosdev/benthos/v4/internal/impl/twitter.init.0()
	/home/mjh/src/benthos/internal/impl/twitter/init.go:15 +0x45
FAIL	github.com/benthosdev/benthos/v4/cmd/tools/benthos_docs_gen	0.229s
FAIL
```

I'm reporting mostly for visibility, the simple workaround is of course to just use Go 1.20 until this library explicitly supports 1.21

matthewhughes934:
I think the culprit is this changed in Go 1.21(https://tip.golang.org/doc/go1.21)

> Package initialization order is now specified more precisely. The new algorithm
> is:
> 
>   - Sort all packages by import path.
>   - Repeat until the list of packages is empty:
>       - Find the first package in the list for which all imports are already
>         initialized.
>       - Initialize that package and remove it from the list.

`parse_duration` is loaded in an `init` function https://github.com/benthosdev/benthos/blob/7c7fe5fe23283b70bc1bca670bfc2d743c926f02/internal/impl/pure/bloblang_time.go#L163 so before I guess it worked based on the not very well defined ordering of these nationalizations.

I can get those tests to run by hacking things to force the init that registers that function to run first:

```patch
diff --git a/internal/impl/twitter/init.go b/internal/impl/twitter/init.go
index bdf09850e..b79b3653c 100644
--- a/internal/impl/twitter/init.go
+++ b/internal/impl/twitter/init.go
@@ -3,6 +3,7 @@
 import (
        _ "embed"
 
+       _ "github.com/benthosdev/benthos/v4/internal/impl/pure"
[...]

PR #2034, opened 2023-08-10
Fix init-time panic in Go 1.21

matthewhughes934:
See issue #2033 for details, the quick version is there was an implicit reliance on package initialization ordering, but this ordering was changed in Go 1.21. The fix is to add an explicit import to make this force the desired order.

### DIFF
--- a/internal/impl/twitter/init.go
+++ b/internal/impl/twitter/init.go
@@ -2,6 +2,9 @@ package twitter
 
 import (
 	_ "embed"
+	// bloblang functions are registered in init functions under this package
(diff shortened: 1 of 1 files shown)


matthewhughes934:
It might be useful to have tests running under 1.21 to validate this change/prevent regressions. Maybe the test job could use a version matrix?

Jeffail (approved):
LGTM, thanks @matthewhughes934! We used to have a version matrix for running the tests against but we ran out of space for the dependency caches. At some point for this next release I'll update all the tests and scripts to run against 1.21.

PR #2035, opened 2023-08-11
Add s3_etag to AWS S3 Input meta

seeya:
S3 Object contains an ETag. The value is currently not available using `root_meta("s3_etag")`.


internal/impl/aws/input_s3.go
```diff
@@ -773,6 +773,9 @@ func s3MsgFromParts(p *s3PendingObject, parts []*message.Part) message.Batch {
 		if p.obj.VersionId != nil && *p.obj.VersionId != "null" {
 			part.MetaSetMut("s3_version_id", *p.obj.VersionId)
 		}
+		if p.obj.ETag != nil {
+			part.MetaSetMut("s3_etag", strings.ReplaceAll(*p.obj.ETag, `"`, ""))
```
Jeffail, line 777: What's the goal here? If you want to unquote the contents then`strings.ReplaceAll` isn't enough as it wouldn't unescape the inner contents properly, we could potentially use https://pkg.go.dev/strconv#Unquote but personally I'd rather just pass the string through unmodified and let users decide if they want to map it into a different form.

Jeffail (commented):
Hey @seeya, just one minor comment but happy to merge. However, it looks as though this PR has been mixed in with some other changes that aren't relevant.

PR #2036, opened 2023-08-12
MongoDB input - add batching

boorad:
This PR converts the MongoDB Input component into a `BatchInput`.  Its performance is greatly improved on collections with large document counts.

This implementation does not use the usual BatchPolicy mechanism, but rather conforms the configuration closer to `find` and `aggregate` options in the MongoDB driver.  For the `find` operation, `limit` and `sort` options are also implemented.

### DIFF
--- a/internal/impl/mongodb/input.go
+++ b/internal/impl/mongodb/input.go
@@ -6,6 +6,7 @@ import (
 
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
 
 	"github.com/benthosdev/benthos/v4/public/service"
 )
@@ -14,6 +15,7 @@ import (
 const (
 	FindInputOperation      = "find"
 	AggregateInputOperation = "aggregate"
+	DefaultBatchSize        = 1000
 )
 
 func mongoConfigSpec() *service.ConfigSpec {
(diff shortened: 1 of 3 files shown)


internal/impl/mongodb/input.go
```diff
@@ -42,23 +44,39 @@ func mongoConfigSpec() *service.ConfigSpec {
 		Field(service.NewBloblangField("query").
 			Description("Bloblang expression describing MongoDB query.").
 			Example(`
-  root.from = {"$lte": timestamp_unix()}
-  root.to = {"$gte": timestamp_unix()}
-`))
+root.from = {"$lte": timestamp_unix()}
+root.to = {"$gte": timestamp_unix()}
+`)).
+		Field(service.NewAnyMapField("options").
```
Jeffail, line 50: I'd prefer to have explicit fields rather than a generic option as it makes documentation and linting more explicit, is there a specific reason for using an AnyMap here?

internal/impl/mongodb/input.go
```diff
@@ -135,33 +176,111 @@ func (m *mongoInput) Connect(ctx context.Context) error {
...
+				return nil
+			}, nil
+		}
+	}
+	return nil, nil, service.ErrEndOfInput
 }
 
 func (m *mongoInput) Close(ctx context.Context) error {
 	if m.cursor != nil && m.client != nil {
+		m.logger.Infof("Got %d documents from '%s' collection", m.count, m.collection)
```
Jeffail, line 217: This is nice to have but I think it ought to be a `Debugf` level log.

Jeffail (commented):
Hey @boorad, looks good overal and extra thanks for the documentation fixes. Left two comments. I suspect the reasoning for the AnyMap is because certain fields are only relevant for certain operations, but rather than make the config structure generic and vague I prefer to add all the available fields explicitly with a caveat "Only relevant when the operation is..." at the end of the description.

internal/impl/mongodb/input.go
boorad: Fine by me.  Changed.

internal/impl/mongodb/input.go
boorad: 👍

Jeffail:
Thanks @boorad, merged via: https://github.com/benthosdev/benthos/commit/3cb87614be282d6cd0a92cb2353420d11f87010c, I'm not sure why this PR didn't update with the merge status.

PR #2038, opened 2023-08-12
Added dependabot

Neo2308:
* Added dependabot to keep go dependencies & actions updated.


Jeffail (approved):
LGTM, thanks @Neo2308, I tweaked some of the settings as I'm not sure how intense it's going to be so leaning on lower frequency PRs for now and then we can tune it later.

PR #2047, opened 2023-08-17
Support extra endpoint parameters in oauth

rvangsgaard:


Jeffail:
Hey @rvangsgaard, looks good overall but is there a reason to allow `any` values rather than `string`? It looks as though everything gets stringified regardless.

rvangsgaard:
> Hey @rvangsgaard, looks good overall but is there a reason to allow `any` values rather than `string`? It looks as though everything gets stringified regardless.

There is no reason to allow any values. This is now fixed.

Jeffail (approved):
LGTM, thanks @rvangsgaard!

PR #2063, opened 2023-08-17
feat: add custom delimiter option to csv-safe codec

ryanhallcs:
Adds a custom delimiter options `csv-safe:x` to the csv-safe codec. Uses the logic of `csv:x` as a baseline


Jeffail (approved):
LGTM, thanks @ryanhallcs!

PR #2084, opened 2023-08-28
Cleanup bloblang linting

mihaitodor:
- Allow comments in single root and directly imported bloblang mappings

```yaml
input:
  generate:
    count: 1
    mapping: root.foo = "bar"
  processors:
    - mapping: |
        #!blobl

        from "test.blobl"
    - switch:
      - check: |
          #!blobl

          this.foo == "bar"
        processors:
          - noop: {}
```

- Explicitly reject content after a directly imported mapping (`from` statement)
- Add tests for directly imported mappings (`from` statement)
- Add bloblang mapping and interpolated field linter tests
- Deflake `watcher` test

### DIFF
--- a/internal/bloblang/parser/combinators.go
+++ b/internal/bloblang/parser/combinators.go
@@ -653,14 +653,6 @@ func DiscardAll(parser Func) Func {
 	}
 }
 
-// Nothing is a parser that always returns an error. This can be useful for
-// disabling parsers.
-func Nothing() Func {
-	return func(input []rune) Result {
-		return Fail(NewError(input, "nothing"), input)
-	}
-}
-
 // MustBe applies a parser and if the result is a non-fatal error then it is
 // upgraded to a fatal one.
 func MustBe(parser Func) Func {
--- a/internal/bloblang/parser/mapping_parser_test.go
+++ b/internal/bloblang/parser/mapping_parser_test.go
@@ -44,6 +44,10 @@ func TestMappingErrors(t *testing.T) {
    `,
 			errContains: `line 2 char 4: expected import, map, or assignment`,
 		},
+		"comment with no mapping": {
+			mapping:     `# foobar`,
+			errContains: `line 1 char 1: expected import, map, or assignment`,
+		},
 		"double mapping": {
 			mapping:     `foo = bar bar = baz`,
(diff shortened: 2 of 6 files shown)


internal/bloblang/parser/combinators.go
```diff
@@ -653,14 +653,6 @@ func DiscardAll(parser Func) Func {
 	}
 }
 
-// Nothing is a parser that always returns an error. This can be useful for
-// disabling parsers.
-func Nothing() Func {
-	return func(input []rune) Result {
-		return Fail(NewError(input, "nothing"), input)
-	}
-}
```
mihaitodor, line 662: Doesn't seem to be used anywhere ¯\_(ツ)_/¯

internal/config/watcher_test.go
```diff
@@ -36,10 +36,11 @@ output:
 	rdr := newDummyReader(confFilePath, nil)
 
 	changeChan := make(chan struct{})
+	once := sync.Once{}
 	var updatedConf stream.Config
 	require.NoError(t, rdr.SubscribeConfigChanges(func(conf *Type) error {
 		updatedConf = conf.Config
-		close(changeChan)
+		once.Do(func() { close(changeChan) })
```
mihaitodor, line 43: Not sure why this handler got executed more than once on my M1... I can't reproduce it on demand

internal/bloblang/parser/mapping_parser_test.go
```diff
@@ -446,6 +474,25 @@ root = this.apply("foo")`,
...
+"foo" == "bar"`,
+			input: []part{
+				{Content: ``},
+			},
+			output: part{
+				Content: `false`,
+			},
+		},
 		"test imported map": {
 			mapping: fmt.Sprintf(`import "%v"
```
mihaitodor, line 497: Should Benthos reject `mapping: import "foobar.blobl"`? Currently, it's equivalent to `noop`.

Jeffail: Good question, probably as it's likely unintentional that someone would do this. I'd probably expose this as a lint warning rather than error which we dont have for bloblang right now. Maybe lets leave this for now.

Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #2092, opened 2023-09-01
Allow to manually specify the number of partitions for new kafka topics

### LINKED ISSUE #2088: Possibility to set the number of partitions for new topics created by kafka output
I would like to add the possibility to set the number of partitions for topics created by kafka output.

Let me explain my use case:
Our application runs on Kubernetes, where we have a Redpanda instance alongside multiple Benthos instances and other stuff. The problem with running Redpanda in k8s with Helm, is that one cannot set a default number of partitions for new topics (as seen in [this](https://github.com/redpanda-data/helm-charts/blob/af6c466e3d5e1a4d1fc366430bc97b0d445ae4c7/charts/redpanda/values.yaml#L940C20-L940C20) line).
In other components of our application we use the Sarama library to generate and manage consumers and producers, and it's possible to configure the number of partitions when creating a new kafka topic. Unfortunately, the data processed through Benthos does not come from or goes to any of those other components, therefore when producing new kafka topics from Benthos, Redpanda uses it's default value of 1 partition per topic.

It would be great to implement a `partitions` parameter that, if set, creates the new topic with the set number of partition, using the [`ClusterAdmin`](https://pkg.go.dev/github.com/Shopify/sarama@v1.38.1#ClusterAdmin) interface of sarama. It would also be possible to cache the topic names for increased speed. An example implementation can be seen in our sarama-wrapper, where in [this function](https://github.com/united-manufacturing-hub/Sarama-Kafka-Wrapper/blob/8f7f860c384b1afac4434d394b98f10af9a108f3/pkg/kafka/client.go#L254) we check if a topic exists and eventually create it before enqueueing the message (we use an async producer).
Of course, if the parameter is not set (or set to 0 maybe), then it would just not set the `NumPartitions` filed in the `sarama.TopicDetail` struct before creating the topic, thus using whatever default value is set in the kafka broker.

It would be nice to have some feedback about this functionality, and if it would be well received by the users, so that I could start working on the implementation.

DaruZero:
This PR adds two new fields for the kafka output.
- `partitions_per_new_topic`: the number of partition to generate if the topic is not existing. This value overrides the default configuration of the kafka broker. If the value is `0` (set by default), it is ignored, and the topic will be created normally.
- `topic_replication_factor`: the replication factor to assign to the topic. This value overrides the default configuration of the kafka broker. If `partitions_per_new_topic` is set to `0`, it is ignored. Must be an odd number (`1` by default)

I am not sure if and how I should update the integration tests. In case, let me know and I'll update the PR.

Closes #2088

### DIFF
--- a/internal/impl/kafka/output_sarama_kafka.go
+++ b/internal/impl/kafka/output_sarama_kafka.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/IBM/sarama"
 	"github.com/cenkalti/backoff/v4"
+	"golang.org/x/sync/syncmap"
 
 	batchInternal "github.com/benthosdev/benthos/v4/internal/batch"
 	"github.com/benthosdev/benthos/v4/internal/batch/policy"
@@ -68,6 +69,8 @@ Unfortunately this error message will appear for a wide range of connection prob
 			docs.FieldString("key", "The key to publish messages with.").IsInterpolated(),
 			docs.FieldString("partitioner", "The partitioning algorithm to use.").HasOptions("fnv1a_hash", "murmur2_hash", "random", "round_robin", "manual"),
 			docs.FieldString("partition", "The manually-specified partition to publish messages to, relevant only when the field `partitioner` is set to `manual`. Must be able to parse as a 32-bit integer.").IsInterpolated().Advanced(),
+			docs.FieldInt("partitions_per_new_topic", "If a topic is created and this value is greater than zero then this number of partitions will be used for the topic.").Optional().Advanced(),
+			docs.FieldInt("topic_replication_factor", "The replication factor for the newly created topics. If `partitions_per_new_topic` is greater than zero this field is required, otherwise is ignored").Advanced(),
 			docs.FieldString("compression", "The compression algorithm to use.").HasOptions("none", "snappy", "lz4", "gzip", "zstd"),
 			docs.FieldString("static_headers", "An optional map of static headers that should be added to messages in addition to metadata.", map[string]string{"first-static-header": "value-1", "second-static-header": "value-2"}).Map(),
 			docs.FieldObject("metadata", "Specify criteria for which metadata values are sent with messages as headers.").WithChildren(metadata.ExcludeFilterFields()...),
@@ -131,6 +134,7 @@ type kafkaWriter struct {
 	topic     *field.Expression
 	partition *field.Expression
 
+	admin       sarama.ClusterAdmin
 	producer    sarama.SyncProducer
 	compression sarama.CompressionCodec
 	partitioner sarama.PartitionerConstructor
@@ -183,6 +187,10 @@ func NewKafkaWriter(conf output.KafkaConfig, mgr bundle.NewManagement) (output.A
 		return nil, fmt.Errorf("failed to parse parition expression: %v", err)
 	}
 	if k.backoffCtor, err = conf.Config.GetCtor(); err != nil {
+		return nil, err
+	}
+
+	if k.admin, err = sarama.NewClusterAdmin(conf.Addresses, sarama.NewConfig()); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


internal/impl/kafka/output_sarama_kafka.go
```diff
@@ -68,6 +69,8 @@ Unfortunately this error message will appear for a wide range of connection prob
 			docs.FieldString("key", "The key to publish messages with.").IsInterpolated(),
 			docs.FieldString("partitioner", "The partitioning algorithm to use.").HasOptions("fnv1a_hash", "murmur2_hash", "random", "round_robin", "manual"),
 			docs.FieldString("partition", "The manually-specified partition to publish messages to, relevant only when the field `partitioner` is set to `manual`. Must be able to parse as a 32-bit integer.").IsInterpolated().Advanced(),
+			docs.FieldInt("partitions_per_new_topic", "If a topic is created and this value is greater than zero then this number of partitions will be used for the topic.").Optional().Advanced(),
+			docs.FieldInt("topic_replication_factor", "The replication factor for the newly created topics. If `partitions_per_new_topic` is greater than zero this field is required, otherwise is ignored").Advanced(),
```
Jeffail, line 73: Been thinking over the config both in terms of intuition when opting into manual topic creation and also the assumed behaviour when these fields aren't set. The current behaviour is that topic creation is left up to your broker configuration, where it may be enabled or not.

If a user wishes to opt-into this new functionality it's not entirely necessary for them to set both the partitions and the number of replicas as they may want to stick with the broker defaults, and this applies to either field.

With that in mind I think there are two issues:
1. The presence of these fields could be confusing as users may no longer assume that the broker defaults would be respected if they omit these fields
2. Using partitions_per_new_topic > 0 as the determining factor as to whether manual creation is unintuitive for users that want to use the broker default for partitions but a manual setting of replicas

I think a more intuitive approach would be to add a config block that is more explicit:

```
    custom_topic_creation:
      enabled: false
      partitions: -1
      replication_factor: -1
```

The name `custom_topic_creation` makes it clearer that the default is to defer to whatever the broker sets. Having an explicit `enabled` field removes any ambiguity around which fields are necessary for enabling this functionality, and setting defaults for both fields at -1 aligns it with other APIs in the kafka ecosystem and is a more explicit signal of "use the broker settings".

internal/impl/kafka/output_sarama_kafka.go
```diff
@@ -296,6 +304,14 @@ func (k *kafkaWriter) buildUserDefinedHeaders(staticHeaders map[string]string) [
 
 //------------------------------------------------------------------------------
 
+// topicCache serves as a cache for topics that have been created during the
+// lifetime of this writer. This is to avoid unnecessary calls to the Kafka
+// cluster to create topics that already exist.
+//
+// The map is keyed by topic name and the value is a boolean indicating whether
+// the topic exists or not.
+var topicCache = syncmap.Map{}
```
Jeffail, line 313: Two issues with this:

1. The memory is never reclaimed. In a long running service where inputs and outputs are spawned and torn down dynamically we would slowly grow this cache without any mechanism for clearing memory.
2. The cache being global means that multiple kafka outputs connecting to different clusters would conflict

Both issues can be solved by moving the cache into the `*kafkaWriter`. A config containing multiple kafka outputs would only double the number of look ups per topic which isn't a big deal.

Jeffail (changes requested):
Hey @DaruZero, looks good overall but there's two changes I'd want to make. I can potentially do it on my end after merging but I can't say when I'll get around to it.

internal/impl/kafka/output_sarama_kafka.go
DaruZero: I agree that a whole config object for this functionality, like the one you proposed, is definitely more intuitive for the user. The only thing is that unfortunately the function `sarama.ClusterAdmin.CreateTopic()` requires a `sarama.TopicDetail` object that has both `NumPartitions` and `ReplicationFactor` set. That is understandable, because the default value of one could cause conflicts with the new value of the other, for example setting the number of partitions to 2 for a new topic but having the replication factor set to 3 by default.

So I would implement the changes that you proposed, but still require both `partitions` and `replication_factor` to be set

internal/impl/kafka/output_sarama_kafka.go
DaruZero: Yep, that's right, I overlooked this. I'll fix it right away

DaruZero:
@Jeffail any thoughts on the latest changes?

Jeffail (approved):
LGTM, thanks @DaruZero!

PR #2093, opened 2023-09-03
fix version attribute on default_ttl field at cache ttlru

peczenyj:
hello @Jeffail 
This change is just documentation


Jeffail (approved):
LGTM, thanks @peczenyj!

PR #2094, opened 2023-09-04
Added support for JSON schemas within the schema registry

calmera:
No processing is done against the actual message content except for popping/prepending the magic bytes for the registry and validating the message contents against the schema obviously.


Jeffail (approved):
LGTM, thanks @calmera! Can you add the DCO sign off to your commits? Running `git commit --amend -s` and then force-pushing should be enough.

PR #2111, opened 2023-09-12
Fix deprecated interpolated string methods

peczenyj:
since the methods Bytes and String are considered deprecated in favor of TryBytes and TryString, I search and update some modules and tests 

enjoy


Jeffail (approved):
LGTM, thanks @peczenyj!

PR #2118, opened 2023-09-16
Add skip BOM for CSV Reader

webfrank:
Files exported from Excel, and probably other tools, add Byte Order Mark (BOM) at the beginning of the CSV file.

I've used library https://github.com/dimchansky/utfbom to detect such mark and remove, otherwise first header will include the mark, which is not printable, and will not be possible to access the value.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -36,6 +36,7 @@ require (
 	github.com/couchbase/gocb/v2 v2.6.3
 	github.com/denisenkom/go-mssqldb v0.12.3
 	github.com/dgraph-io/ristretto v0.1.1
+	github.com/dimchansky/utfbom v1.1.1
 	github.com/dop251/goja v0.0.0-20220815083517-0c74f9139fd6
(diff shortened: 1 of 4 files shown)


Jeffail:
@webfrank we can probably avoid the dependency as it looks as though we've already implemented a BOM skipper for the input codecs: https://github.com/benthosdev/benthos/blob/31049847779559e08b5bded6a77145bf64c9c2ad/internal/codec/reader.go#L201

peczenyj:
I think we may find such issue in many other components, not only csv.
Perhaps this can be a global option such as `—skip-bom=[true|false]` and it can be true by default 

then all readers may skip BOM

Also, this can be a per-component configuration where we can enable or disable using the default value the option defined above.

that’s my two cents

peczenyj:
Ops I did no see the last reply, sorry

webfrank:
Having skipbom in the code this PR is no more required.

See my comment here https://github.com/benthosdev/benthos/issues/2117#issuecomment-1722413626

PR #2128, opened 2023-09-19
Add MetadataFilter field for NATS outputs

alexthecarp:
## Motivation
NATS supports metadata in the form of Headers (in some configurations, [JetStream](https://docs.nats.io/nats-concepts/jetstream/headers) and [classic](https://docs.nats.io/release-notes/whats_new/whats_new_22#message-headers) since 2.2.0 ).
Benthos currently supports a `headers` object that can add explicit, interpolated headers. 

In this PR I'm adding functionality that allows passing through any `meta()` headers, using the MetadataFilter functionality that's built-in to Benthos (similar to Kafka outputs). 

Most of the heavylifting is already done by the platform so the changes are minimal. 

NOTE: This doesn't change the semantics of the existing configurations, since by default we still don't pass anything (thus it is backwards compatible). A user is required to explicitly add any headers they want, or use a catch-all regular expression:
```    
metadata:
  include_patterns: 
    - ".*"
```

### DIFF
--- a/internal/impl/nats/output.go
+++ b/internal/impl/nats/output.go
@@ -36,6 +36,9 @@ func natsOutputConfig() *service.ConfigSpec {
 				"Content-Type": "application/json",
 				"Timestamp":    `${!meta("Timestamp")}`,
 			})).
+		Field(service.NewMetadataFilterField("metadata").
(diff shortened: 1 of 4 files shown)


Jeffail (approved):
LGTM, thanks @alexthecarp!

PR #2129, opened 2023-09-20
Add log line with Benthos version

difabion:
Adds a standard log output on startup identifying Benthos version.

On untagged builds, this displays the commit hash; otherwise, the tag is shown.


Jeffail (approved):
LGTM, thanks bae, I've slightly modded it so that we add version as a field to the config logs to save us a line at init time.

PR #2130, opened 2023-09-21
Allow redundant urls for amqp 1

### LINKED ISSUE #2106: Allow redundant URLs for AMQP_1
We'd like the same option as amqp_0_9 (see #912 ) for amqp_1 regarding specifying multiple urls for our amqp-clusters
(our prio is the input)

peczenyj:
This should fix #2106 

to be backward compatible, both fields `url` and `urls` will co-exist for some time,

by default we will parse `urls` and if it is empty we will try to use field `url` as fallback.

IMPORTANT: we need to fill the field Version with the right value instead the placeholder T.B.D (to be defined) and regenerate docs before release a new version.

enjoy

### DIFF
--- a/internal/impl/amqp1/config.go
+++ b/internal/impl/amqp1/config.go
@@ -11,6 +11,7 @@ import (
 const (
 	// Shared
 	urlField      = "url"
+	urlsField     = "urls"
 	tlsField      = "tls"
 	saslField     = "sasl"
 	saslMechField = "mechanism"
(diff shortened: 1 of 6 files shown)


Jeffail (approved):
LGTM, thanks @peczenyj! I'm going to mark the `url` field as deprecated to avoid some confusion. Also going to add a linting rule to ensure at least one is set.

PR #2160, opened 2023-10-11
fix(cue): allow labels on caches and rate limits

disintegrator:
This change fixes the CUE schema emitted by `benthos list --format cue` such that cache and rate-limit resources can take a `label` config. This is necessary so that these resources are addressable in configs.


tgmendes (approved):
Thanks for doing this mate 🙌

Jeffail (approved):
LGTM, thanks @disintegrator!

PR #2161, opened 2023-10-12
Fix string formatting when setting custom slog logger

josephwoodward:
Currently when using the `Modular` logging interface to format strings you get the following output:

```golang
func TestSlogMessageFormattingTest(t *testing.T) {
	var buf bytes.Buffer
	h := slog.NewTextHandler(&buf, &slog.HandlerOptions{ReplaceAttr: clearTimeAttr, Level: slog.LevelDebug})
	s := slog.New(h)

	var logger Modular = NewBenthosLogAdapter(s)
	logger.Warnf("Hello %s", "World")

	expected := "time=\"\" level=WARN msg=\"Hello %s\" !BADKEY=World\n"
	assert.Equal(t, expected, buf.String())
}
```

This change fixes that and instead returns the expected log line:

```
"time=\"\" level=WARN msg=\"Hello World\"\n"
```

### DIFF
--- a/internal/log/slog.go
+++ b/internal/log/slog.go
@@ -1,6 +1,7 @@
 package log
 
 import (
+	"fmt"
 	"os"
 
 	"log/slog"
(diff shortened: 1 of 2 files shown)


Jeffail (approved):
LGTM, thanks @josephwoodward!

PR #2162, opened 2023-10-12
slog: only include for golang 1.21+

srenatus:
This puts the slog-related bits into files that carry a build constraint of go1.21, i.e. are built only with golang 1.21 _and higher_.

Effectively, it allows us to keep `go 1.20` in go.mod, allowing benthos to be used in projects that still support golang 1.20.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -313,4 +313,4 @@ require (
 	modernc.org/token v1.1.0 // indirect
 )
 
-go 1.21
+go 1.20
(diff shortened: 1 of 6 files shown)


Jeffail (approved):
Oh nice, thanks for setting this up @srenatus, I think I mistakenly thought this wasn't available yet due to https://github.com/golang/go/issues/59033 remaining open but I guess it works.

PR #2163, opened 2023-10-12
Add ts_between method

madskrogh:
This pr adds the **ts_between** method to bloblang, requested by benthosdev#1982.

This is my first contribution to benthos so please feel free to comment on anything that isn't in line with the style and conventions of the project. 🙏

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 
 - The `amqp_0_9` output now supports dynamic interpolation functions within the `exchange` field.
 - Field `custom_topic_creation` added to the `kafka` output.
+- New bloblang timestamp method `ts_between`.
 
 ## 4.22.0 - 2023-10-03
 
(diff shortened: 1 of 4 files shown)


Jeffail (approved):
LGTM, thanks @madskrogh! I've just added some APIs for defining timestamp parameters: https://github.com/benthosdev/benthos/commit/17fbe827237883571dab1c2e4c406bdd716dca0c and I'll adapt this new plugin to use them as it'll make it a little bit more flexible.

I might also rename this to `ts_sub`, as it'd then be similar to the `time` package method. If in future we plan to add the ability to substract a duration string then it'll be called `ts_sub_duration` so we won't be at risk of collisions there.

Jeffail:
I also realised that if we're renaming to `ts_sub` then we need to remove the abs behaviour. To help with that I've fixed the .abs method so that integers input will yield integers out: https://github.com/benthosdev/benthos/commit/c70122476065723da728522329fa784280456abb

madskrogh:
Great!

PR #2166, opened 2023-10-13
Add tracing support to nats components

### LINKED ISSUE #2153: Extract tracing information for NATS input
I have a simple benthos configuration that is using nats as an input, and I would like to be able to extract the tracing information from the nats request. The incoming nats message has the opentelemetry `traceparent` and `tracestate` headers for propagating the tracing information.

Would it be possible to add support for extracting the tracing information from the nats headers?

I would be happy to submit a PR if this seems reasonable.

mattpgray:
This pull request adds `extract_tracing_map` and `inject_tracing_map` parameters to the nats inputs and output: `nats`, `nats_jetstream` and `nats_stream`.

I could not run the full set of integration tests locally, but the nats integration tests all passed for me.

There were a couple of things that were not completely clear to me:

- The version of the fields. I set this to 4.23.0 which I hope is correct
- The level of testing for something like this. I can see that the tracing inputs and output wrappers are tested, but I could not see a test for kafka tracing and I was basing the nats version on that. I performed a manual test on the nats input and output and it worked well.

resolves #2153


Jeffail (approved):
LGTM, thanks @mattpgray! Yeah I think that's the tidiest way you could set the versions, it's always speculative as you have no idea when authoring the docs how soon the PR will get merged. I generally update the version stamps as I merge them.

For testing purposes I've got some manual integration tests for the kafka bits which I think is fine for now. Before marking these fields as stable we'll need to add an integration test that can be shared by all components using them, but I was waiting for more components to utilitise them first, which itself is waiting for more feedback.

PR #2168, opened 2023-10-16
Update service.go

gramian:
Change `version` key in initial log entry to `benthos_version`.


Jeffail (approved):
LGTM, thanks @gramian, the DCO failure is because you need to sign commits with the `-s` flag for the sign off blurb to be in the commit messages. Doesn't really matter for this one though.

gramian:
Noted. Thanks.

PR #2169, opened 2023-10-16
Support container SAS tokens

mfamador:
Currently only storage account SAS token are supported, this PR allow us to use container SAS token as well.


Jeffail (approved):
Cool, LGTM, are there any docs we ought to update on this or is it something a typical azure user would just know to use?

mfamador:
No, it's just a change under the hoods, no docs updates.

mfamador:
It would fail before this PR is anyone tried to use a container SAS

PR #2174, opened 2023-10-18
Fix the kafka_franz input metadata types

mihaitodor:
These changes should get the data types of the `kafka_franz` input metadata to match the `kafka` input ones.

NB: This is technically a breaking change, but I don't think it's going to annoy a lot of people...


Jeffail (approved):
LGTM, whoops! Thanks @mihaitodor!

PR #2179, opened 2023-10-20
feat: add message delay seconds to aws_sqs output

bira37:
Add `delay_seconds` field to aws_sqs output to make possible to configure delay per individual message, as described in  [AWS SQS Docs](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-delay-queues.html)


Jeffail (approved):
LGTM, thanks @bira37!

PR #2196, opened 2023-10-30
Add `azure_file_share` output

mfamador:
Add the `azure_file_share` output.

```yml
output:
  azure_file_share:
    storage_connection_string: '<REDACTED>'
    share_name: 'foo'
    path: 'bar/baz.csv'
```
or
```yml
output:
  azure_file_share:
    storage_account: 'storageaccountfoo'
    share_name: 'foo'
    storage_sas_token: '<REDACTED'
    path: 'bar/baz.csv'
```

Azurite doesn't support azure file share yet, so there's no integration tests for now. I've tested it manually using storage connection string, storage SAS token and share SAS token as well.
https://github.com/Azure/Azurite/issues/113


internal/impl/azure/output_file_share.go
```diff
@@ -0,0 +1,209 @@
...
+	if conf.Path, err = pConf.FieldInterpolatedString(fsFieldPath); err != nil {
+		return
+	}
+	return
+}
+
+func fsSpec() *service.ConfigSpec {
+	return azureComponentSpec().
+		Beta().
+		Version("3.24.0").
```
mihaitodor, line 59: This needs to be updated

mihaitodor (approved):
Thanks @mfamador! I think that looks good, just left one small nitpick. I might take a stab at doing a bit of refactoring around the auth stuff once I get back to the CosmosDB work, hopefully before the end of the year.

internal/impl/azure/output_file_share.go
mfamador: Thanks! Fixed.

PR #2198, opened 2023-11-01
[kafka input] Allow to configure the offset retention duration

mfamador:
Allow to configure the offset retention duration, If not set, disabled (in which case the `offsets.retention.minutes` option on the broker will be used).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+## 4.24.0 - 2023-11-24
+
 ### Added
 
 - Field `idempotent_write` added to the `kafka_franz` output.


internal/impl/kafka/input_sarama_kafka_parts.go
```diff
@@ -117,7 +117,7 @@ func (k *kafkaReader) offsetVersion() int16 {
 	// - 4 (kafka 2.0.0 and later)
 	var v int16 = 1
 	// TODO: Increase this if we drop support for v0.8.2, or if we allow a
-	// custom retention period.
+	// custom offsetRetention period.
```
Jeffail, line 120: @mfamador have you tested custom retention periods with explicit partition consumers? pretty sure we'd need to increase the offset version here.

mfamador: Not with explicit partition consumers, no. Only with balanced consumer groups, it solved me the issue when the retention period was wrongly set to almost 0(?) on older kafka versions.

I didn't remember we could even have explicit partition consumers with this input, but I see we can specify it within the topic name (foo:0-10).

I'll test it an increase the version, sure.

PR #2204, opened 2023-11-03
Add more gocql cluster options

peczenyj:
Hello

Here I am adding new options to cassandra input and output

first, we differentiate query **timeout** from **connect timeout**

then I add the option to use token aware host policy (default false)

and, optional, ask to shuffle replicas or not (default false)

also, we can now enable snap compression and set the default idempotence (useful for input)

in the end, I am wrapping the manager logger with a field "cassandra_input" or "casssandra_output" followed by the label to substitute the gocql standard logger by one that send everything to debug

enjoy


peczenyj:
Hey @Jeffail why the input and output codes and some options are So different for cassandra?

PR #2218, opened 2023-11-07
Dependabot ignore bluemonday

mihaitodor:
This package is noisy and it also requires bumping up the minimum Go version which will have to wait a bit longer.

Also reflow `dependabot.yaml`.


Jeffail (approved):
LGTM, hopefully we won't need this for long but it'd be good to keep the ignore section there so we can quickly add/remove packages in future.

PR #2230, opened 2023-11-10
Create the sarama ClusterAdmin with the actual sarama configuration.

mfamador:
Using sarama 1.42.0, we're not able to connect a sarama output to a kafka 1.0.0/ Event hubs. 
Probably it happens with all but the default version.
Using the actual sarama config to create the ClusterAdmin we can now connect to any kafka successfuly.

Fixes https://github.com/benthosdev/benthos/issues/2199


Jeffail (approved):
LGTM, thanks @mfamador!

PR #2245, opened 2023-11-15
Enhance protobuf resolution to register first level of nested messages

mxcoder:
Hello, in this PR I'd like to propose a change that would allow users to utilize nested messages from `.proto` files.
This case is useful whenever you need to create partial serializations from legacy protobuf schemas.

With a protobuf like:
```protobuf
syntax = "proto3";
package testing;

import "person.proto";

message House {
  message Mailbox {
    string color = 1;
    string identifier = 2;
  }
  repeated testing.Person people = 1;
  string address = 2;
  Mailbox mailbox = 3;
}
```
This change would allow to use the protobuf processor with a `message` setting like:
```yaml
protobuf:
  operator: from_json
  import_paths: ["./']
  message: testing.House.Mailbox
```

### DIFF
--- a/config/test/protobuf/schema/house.proto
+++ b/config/test/protobuf/schema/house.proto
@@ -4,6 +4,11 @@ package testing;
 import "person.proto";
 
 message House {
+  message Mailbox {
+    string color = 1;
+    string identifier = 2;
+  }
   repeated testing.Person people = 1;
   string address = 2;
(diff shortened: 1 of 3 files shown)


Jeffail (approved):
LGTM, thanks @mxcoder!

mxcoder:
Thanks for accepting :)

PR #2263, opened 2023-11-20
Read until idle timeout

### LINKED ISSUE #2249: Update read_until to close the input after an idle timeout
`read_until` closes the input once a consumed message passes a specific condition based on a query. Here I propose a new field `idle_timeout` that specifies the maximum amount of time without receiving new messages after which the input is closed.

```
input:
  label: ""
  read_until:
    input: null
    check: ""
    restart_input: false
    idle_timeout: "5s"
```

The configuration of 'check' and 'idle_timeout' could be used either independently or collectively, resulting in the termination of the input upon the first occurrence.

eduardodbr:
fixed #2249

### DIFF
--- a/internal/impl/pure/input_read_until.go
+++ b/internal/impl/pure/input_read_until.go
@@ -26,9 +26,11 @@ func init() {
 	}), docs.ComponentSpec{
 		Name: "read_until",
 		Summary: `
-Reads messages from a child input until a consumed message passes a [Bloblang query](/docs/guides/bloblang/about/), at which point the input closes.`,
+Reads messages from a child input until a consumed message passes a [Bloblang query](/docs/guides/bloblang/about/), at which point the input closes. It is also possible to configure a timeout after which the input is closed if no new messages arrive in that period.`,
 		Description: `
 Messages are read continuously while the query check returns false, when the query returns true the message that triggered the check is sent out and the input is closed. Use this to define inputs where the stream should end once a certain message appears.
+
+If the idle timeout is configured, the input will be closed after that period of time if no new messages arrive. Use this field if you want to empty out and close an input that doesn't have a logical end.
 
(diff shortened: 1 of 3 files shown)


internal/impl/pure/input_read_until.go
```diff
@@ -104,9 +123,15 @@ func newReadUntilInput(conf input.Config, mgr bundle.NewManagement, log log.Modu
...
+	var idleTimeout time.Duration = -1
+	if len(conf.ReadUntil.IdleTimeout) > 0 {
+		if idleTimeout, err = time.ParseDuration(conf.ReadUntil.IdleTimeout); err != nil {
+			return nil, fmt.Errorf("failed to parse idle_timeout string: %v", err)
+		}
+	}
 
-	if check == nil {
-		return nil, errors.New("a check query is required")
+	if check == nil && idleTimeout < 0 {
```
mihaitodor, line 133: Wouldn't it be better to have `<= 0` and initialise `idleTimeout` to `0` instead of `-1`? Not sure if `0` duration makes sense here.

eduardodbr: That also works. At first, my goal was to set it as negative to disable it, but I later realized that I didn't need to do that because we can use the empty string as a default value, so I guess I just forgot to change it to 0 since it still works like this. I'll update it, thanks

Jeffail (approved):
LGTM, thanks @eduardodbr. One minor concern is whether there's a significant hit to having unbounded timers allocated from a busy input. Normally I'd rework this to use `time.NewTimer` which we can stop but I also remember reading about an optimisation in Go where that's no longer necessary, I can't find it right now though so maybe I dreamt it. I might rework it just incase.

mihaitodor:
Nice catch @Jeffail! I think that timer enhancement is still WIP: https://github.com/golang/go/issues/61542

PR #2264, opened 2023-11-21
Config Passthrough for disable of idempotent write with Franz Kafka producer

youngjosh:
In certain cases you may wish to disable idempotent write for the Franz-Go Kafka client, for example it requires additional cluster privileges (`IDEMPOTENT_WRITE` on `CLUSTER`), and can have repercussions with regards to retrying failed messages. This currently defaults to enabled on the Franz-Go client so I have created a config passthrough as below.

The default behaviour of the other bundled Sarama Kafka client does not enable Idempotent write by default, so it requires less permissions than the Franz-Go client to produce to a cluster. Allowing this config passthrough allows for easier migration from one Kafka client to the other.

You can read more about the `DisableIdempotentWrite` option [here](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#DisableIdempotentWrite).

### DIFF
--- a/internal/impl/kafka/output_kafka_franz.go
+++ b/internal/impl/kafka/output_kafka_franz.go
@@ -56,6 +56,10 @@ This output often out-performs the traditional ` + "`kafka`" + ` output as well 
 			Description("A rack identifier for this client.").
 			Default("").
 			Advanced()).
+		Field(service.NewBoolField("disable_idempotent_write").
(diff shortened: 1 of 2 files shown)


Jeffail (approved):
LGTM, thanks @youngjosh. I think I might flip the wording so that the field is `idempotent_write` with a default of `true`.

PR #2270, opened 2023-11-26
Prevent redundant sudo calls in the install script

### LINKED ISSUE #2269: why benthos install script needs sudo ?
https://github.com/benthosdev/benthos/blob/4da46671f378688023d24d28598b356ed395d41f/website/static/sh/install#L148

I want to install benthos in my user profile but I don't want the install script to require admin privileges.

I used asdf plugins, but in the end it invokes this script. I pressed CTRL+C.

Suggestion : check if install path is user owned, then ask use sudo if not.

mihaitodor:
Fixes #2269.

While it's a bit cryptic, I think this approach should be portable across the OSes we care about. Details [here](https://unix.stackexchange.com/questions/269141/getting-the-owner-of-the-folder-with-the-if-command).

### DIFF
--- a/website/static/sh/install
+++ b/website/static/sh/install
@@ -145,6 +145,10 @@ install_benthos()
 	chmod +x "$PREFIX/tmp/$benthos_bin"
 
 	echo "Putting benthos in $benthos_install_path (may require password)"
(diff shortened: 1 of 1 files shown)


Jeffail (approved):
Wow that's yummy, thanks @mihaitodor!

PR #2285, opened 2023-12-09
Fix ineffective break statement in dynamic output

mihaitodor:
I noticed this by running [staticcheck](https://staticcheck.dev). Unfortunately, golangci-lint is still using a deprecated version of this linter. See [here](https://github.com/golangci/golangci-lint/issues/853) and [here](https://github.com/golangci/golangci-lint/issues/357).


Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #2286, opened 2023-12-10
Add `proxy_url` field to the websocket components

### LINKED ISSUE #2278: websocket input does not support proxy
I'm trying to connect to a websocket server via a proxy but the input does not seem to support the `proxy_url` config option and does not pick the `http_proxy` environment variable either.

I expected to be able to set `proxy_url` same way I do for the `http_client` input (https://www.benthos.dev/docs/components/inputs/http_client/#proxy_url).

mihaitodor:
Fixes #2278.

Note: The `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` environment variables can be used as fallback when this field isn't set. Details here: <https://pkg.go.dev/golang.org/x/net/http/httpproxy#FromEnvironment>.

I tested this using an echo NodeJS server:
```node
const WebSocket = require('ws');

const wss = new WebSocket.Server({ port: 8080 });

wss.on('connection', function connection(ws) {
  ws.on('message', function incoming(message) {
    console.log('received: %s', message);
  });

  ws.send('hello from the server!');
});
```

that I ran on a remote GitPod box like so:

```shell
> pip install ws
> node test.js
```

It exposes the server via a https endpoint, so then I am able to interact with it using this config:

```yaml
input:
  websocket:
    url: wss://8080-mihaitodor-benthos-4exudlgyttk.ws-eu106.gitpod.io
  processors:
    - log:
        message: ${! content() }

output:
  websocket:
    url: wss://8080-mihaitodor-benthos-4exudlgyttk.ws-eu106.gitpod.io
```

Then I ran [`mitmweb`](https://mitmproxy.org/) locally and I configured the new `proxy_url` field like so:

```yaml
input:
  websocket:
    url: wss://8080-mihaitodor-benthos-4exudlgyttk.ws-eu106.gitpod.io
    proxy_url: http://localhost:8080
    tls:
      enabled: true
      skip_cert_verify: true
  processors:
    - log:
        message: ${! content() }

output:
  websocket:
    url: wss://8080-mihaitodor-benthos-4exudlgyttk.ws-eu106.gitpod.io
    proxy_url: http://localhost:8080
    tls:
      enabled: true
      skip_cert_verify: true
```

I also had to set `skip_cert_verify: true` because Benthos can't validate the cert from `8080-mihaitodor-benthos-4exudlgyttk.ws-eu106.gitpod.io` when mitmproxy acts as a man-in-the-middle proxy.

### DIFF
--- a/internal/impl/io/input_websocket.go
+++ b/internal/impl/io/input_websocket.go
@@ -40,6 +40,9 @@ func websocketInputSpec() *service.ConfigSpec {
 			service.NewURLField("url").
 				Description("The URL to connect to.").
 				Example("ws://localhost:4195/get/ws"),
+			service.NewURLField("proxy_url").
+				Description("An optional HTTP proxy URL.").
+				Advanced().Optional(),
 			service.NewStringField("open_message").
 				Description("An optional message to send to the server upon connection.").
 				Advanced().Optional(),
@@ -92,12 +95,13 @@ type websocketReader struct {
 
 	lock *sync.Mutex
 
-	client     *websocket.Conn
-	urlParsed  *url.URL
-	urlStr     string
-	tlsEnabled bool
-	tlsConf    *tls.Config
-	reqSigner  httpclient.RequestSigner
+	client         *websocket.Conn
+	urlParsed      *url.URL
+	urlStr         string
+	proxyURLParsed *url.URL
+	tlsEnabled     bool
+	tlsConf        *tls.Config
+	reqSigner      httpclient.RequestSigner
 
 	openMsgType wsOpenMsgType
 	openMsg     []byte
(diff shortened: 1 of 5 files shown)


internal/impl/io/input_websocket.go
```diff
@@ -154,14 +163,17 @@ func (w *websocketReader) Connect(ctx context.Context) error {
 	}
 
 	var client *websocket.Conn
+
+	dialer := *websocket.DefaultDialer
```
mihaitodor, line 167: This is an interesting gotcha. When `w.tlsEnabled == true`, the existing implementation created a new `websocket.Dialer` instance instead of copying `websocket.DefaultDialer`, which [sets](https://github.com/gorilla/websocket/blob/01b0aaed92c975e5e740c6202e7ff8e5a85085ce/client.go#L144-L147) `Proxy: http.ProxyFromEnvironment` and `HandshakeTimeout: 45 * time.Second`. I think a shallow copy should suffice in this case.

mterron:
Can we merge this one @Jeffail ?

Jeffail:
Moved over to https://github.com/redpanda-data/connect/pull/2286

mihaitodor:
Merged in https://github.com/redpanda-data/benthos/pull/29 and will be available in v4.29.1.

PR #2290, opened 2023-12-11
add `kv` processor

harshit-sp:
This pull request introduces a `Key-Value (KV)` processor designed to transform messages formatted as 'foo=bar john=doe' into a well-structured JSON representation.

**Sample usage:**
```yaml
pipeline:
  processors:
  - kv:
      field_split: " "
      value_split: "="
```
**Related issue:** [2050](https://github.com/benthosdev/benthos/issues/2050)


eduardodbr:
Hey @harshit-sp

New processors should use the `public/service` APIs, the way you implemented it is being deprecated. 
```
// Config is the all encompassing configuration struct for all processor types.
// Deprecated: Do not add new components here. Instead, use the public plugin
// APIs. Examples can be found in: ./internal/impl.
```

Check PR #1856  on how to properly do it.

If we want to extend this to something more complex like the [elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/kv-processor.html) implementation, I'm ok with having it as a processor. However, if we only want to have it as a simple funcion like this PR implements, we should consider adding this as a bloblang string method instead of a processor.

```
root.line = this.line.kv(" ","=")

# In:  {"line":"level=info msg=test-message"}
# Out: {"line":{"level":"info","msg":"test-message"}}
```

harshit-sp:
Okay, I will update accordingly.

internal/impl/kv/processor_kv.go
```diff
@@ -0,0 +1,103 @@
...
+		logger:            logger,
+		pairDelimiter:     pairDelimiter,
+		keyValueSeparator: keyValueSeparator,
+		targetField:       targetField,
+	}, nil
+}
+
+func newKVProcessorConfig() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Version("4.16.0").
```
eduardodbr, line 62: This should be changed to 4.25 which is the next release

internal/impl/kv/processor_kv.go
```diff
@@ -0,0 +1,103 @@
...
+			key := parts[0]
+			value := parts[1]
+			obj[key] = value
+		}
+	}
+
+	if k.targetField != "" {
+		msg.SetStructured(map[string]any{k.targetField: obj})
+	} else {
+		msg.SetStructured(obj)
```
eduardodbr, line 95: If the input message has neither pairDelimiter nor keyValueSeparator the output will be empty. Is this the expected behavior? 
I think it would be better if it just maintains the original input instead of an empty value.

internal/impl/kv/processor_kv.go
```diff
@@ -0,0 +1,103 @@
...
+		pairDelimiter:     pairDelimiter,
+		keyValueSeparator: keyValueSeparator,
+		targetField:       targetField,
+	}, nil
+}
+
+func newKVProcessorConfig() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Version("4.16.0").
+		Summary("This processor is designed to convert messages formatted as 'foo=bar bar=baz' into a structured JSON representation.").
```
eduardodbr, line 63: This summary is specific for a single use case and this processor is generic. The documentation should be generic as well. 

The output of this processor can be unexpected depending on the input message. The documentation should also address that

internal/impl/kv/processor_kv.go
harshit-sp: updated

internal/impl/kv/processor_kv.go
harshit-sp: Maintained the original log.

internal/impl/kv/processor_kv.go
harshit-sp: Made the summary generic.

PR #2292, opened 2023-12-11
Fix logging in amqp_0_9 components

mihaitodor:
Currently, users will see a log message similar to this when the `amqp_0_9` output attempts to connect to a server: "Sending AMQP messages to exchange: \u0026{0xc0027123c0}".

This regression was introduced in #1976.

Note: I took the liberty to standardise some of the other log messages in the `amqp_0_9` components.


Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #2304, opened 2023-12-24
NATS: Refactor errors so that defer can cleanup

alexthecarp:
This fixes a connection leak on the NATS Jestream input when there is an error with the consumer. 
See https://github.com/benthosdev/benthos/issues/2303 for A LOT more details. 

TL;DR: There is an `err` variable that shadows the top level one and the deferred cleanup function misses the error. With this fix we don't rely on closures to check the error. 

There are 4 other sites in the code where I'm doing the same refactor. There don't seem to be any bugs there but it's nice to be consistent and also to help future PRs not introduce the same error.

### DIFF
--- a/internal/impl/nats/input_jetstream.go
+++ b/internal/impl/nats/input_jetstream.go
@@ -221,7 +221,7 @@ func newJetStreamReaderFromConfig(conf *service.ParsedConfig, mgr *service.Resou
 
 //------------------------------------------------------------------------------
 
-func (j *jetStreamReader) Connect(ctx context.Context) error {
+func (j *jetStreamReader) Connect(ctx context.Context) (err error) {
 	j.connMut.Lock()
 	defer j.connMut.Unlock()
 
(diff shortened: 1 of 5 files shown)


alexthecarp:
gentle ping @Jeffail , we're running into this quite often in our environment and it's making NATS very unhappy

Jeffail:
Hey @alexthecarp, going to merge these pretty soon but it's going to be blocked on release candidates for the current refactoring work that's going on as it's a large body of work. If the fix is urgent enough we could make a patch release and get it out in the next week or so potentially, does that sound like something you need?

alexthecarp:
hi @Jeffail thanks for the quick reply! I think maybe a quick patch release might make sense if it's not too much trouble (the changes themselves seem benign enough to not worry about new bugs ? but that's of course your call). 

Either way, our setup can handle building Benthos from source but we try to not keep private patches, so even just merging them would help (this way I can build at the merge commit until the next release).

Jeffail (approved):
LGTM, thanks @alexthecarp!

PR #2306, opened 2023-12-24
Nanomsg input: Refactor to not leak socket on error

alexthecarp:
This is another instance of https://github.com/benthosdev/benthos/issues/2303. 

In this case, we leak the socket if there are any errors when setting options on subfilters.


Jeffail (approved):
LGTM, thanks @alexthecarp!

PR #2308, opened 2023-12-24
Zeromq: Refactor error for better cleanup

alexthecarp:
This is an instance of https://github.com/benthosdev/benthos/issues/2303. 

For the `input`, this should fix a potential issue where we don't close the socket if we fail any of the `SetSubscribe` calls for any subfilter. 

The `output` seems fine, I'm just changing it for completeness.


Jeffail (approved):
LGTM, thanks @alexthecarp!

PR #2326, opened 2024-01-11
Changelog Support

calmera:
This PR introduces two different methods to bloblang:

- `diff` allows the current value to be compared to the one passed as an argument, returning a `changelog` containing the operations needed for converting the current value to the one passed.
- `patch` allows for the passed `changelog` to be applied to the current value

This functionality is particularly interesting when converting from full state messages to delta messages


Jeffail (approved):
LGTM, very cool feature, thanks @calmera!

PR #2330, opened 2024-01-13
Fix panic when calling `zip()` in Bloblang

mihaitodor:
This change prevents the Bloblang `zip()` method from panicking when invoked without arguments.


internal/impl/pure/bloblang_objects.go
```diff
@@ -125,6 +125,9 @@ If a key within a nested path does not exist then it is ignored.`).
 			}
```
Jeffail, line 125: I'd say it's probably more intuitive to return an error here when no arguments are present as this'd indicate misuse of the method, maybe someone assuming this method creates a zip archive.

mihaitodor: Good call! I was on the fence about it, but I agree that returning an error is a better idea. Fixed 🎉

Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #2332, opened 2024-01-15
Enable pprof /allocs endpoint

filippog:
This profile type was missing, from https://pkg.go.dev/runtime/pprof

```
The allocs profile is the same as the heap profile but changes the
default pprof display to -alloc_space, the total number of bytes
allocated since the program began (including garbage-collected bytes).
```


Jeffail (approved):
LGTM, thanks @filippog!

PR #2334, opened 2024-01-17
Add new processors  processor and its test in Benthos

calmera:
This commit introduces a new processors processor in the Benthos package for batching multiple sub-processors together under a single processor. This new processor will allow to retain message batches and improve the readability and testability of configurations. Additionally, a new unit test has been created to ensure the processor's functionality.

### DIFF
--- a/internal/impl/pure/processor_processors.go
+++ b/internal/impl/pure/processor_processors.go
@@ -1,0 +1,118 @@
+package pure
+
+import (
+	"context"
+
+	"github.com/benthosdev/benthos/v4/internal/bundle"
+	"github.com/benthosdev/benthos/v4/internal/component/interop"
+	"github.com/benthosdev/benthos/v4/internal/component/processor"
+	"github.com/benthosdev/benthos/v4/internal/log"
+	"github.com/benthosdev/benthos/v4/internal/message"
+	"github.com/benthosdev/benthos/v4/public/service"
+)
+
+func processorsProcSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Categories("Composition").
+		Stable().
+		Summary(`A processor grouping several sub-processors while retaining batches`).
+		Description(`
+Quite often you will come across a case where you want to collect several processors under a single processor, whether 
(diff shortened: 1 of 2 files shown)


internal/impl/pure/processor_processors.go
```diff
@@ -0,0 +1,118 @@
+package pure
+
+import (
+	"context"
+
+	"github.com/benthosdev/benthos/v4/internal/bundle"
+	"github.com/benthosdev/benthos/v4/internal/component/interop"
+	"github.com/benthosdev/benthos/v4/internal/component/processor"
+	"github.com/benthosdev/benthos/v4/internal/log"
+	"github.com/benthosdev/benthos/v4/internal/message"
```
mihaitodor, line 10: Thanks for working on this! I wonder if it would be possible to avoid using the internal APIs and implement it using `service.ExecuteProcessors` just like the `cached` processor: https://github.com/benthosdev/benthos/blob/74a82c0e425e154fd88eeabcc4723ea1a77ce205/internal/impl/pure/processor_cached.go

Jeffail: @mihaitodor probably best not to, in cases like these where you're wrapping other processors I think core stuff like `try` and `catch` should be internal, it'll reduce allocs and have a small performance improvement over the public APIs. It won't be perceptible in most processors but it'll make a difference when the inner processors are mappings.

mihaitodor: Oh, good to know 👍

calmera:
I keep on getting issues on the documentation generation. Any ideas?
```Documentation for 'processors' has changed, updating: website/docs/components/processors/processors.md
website/docs/components/processors/processors.md(43,1) yaml: line 8: found character that cannot start any token
make: *** [Makefile:114: docs] Error 1
```

```yaml
pipeline:
  processors:
    - label: my_super_feature
      processors:
        - log:
            message: "Let's do something cool"
		- archive:
			format: json_array
		- mapping: root.items = this
		- http:
			url: "https://example.com/endpoint"
	        verb: POST
```

Jeffail (approved):
LGTM, thanks @calmera!

PR #2335, opened 2024-01-18
Unify jwt package

tprestegard:
Replace all uses of `golang-jwt/jwt` (v3.2.2+incompatible) with `golang-jwt/jwt/v4` in order to unify the jwt package used throughout benthos.

v3.2.2+incompatible still appears as an indirect dependency.


tprestegard:
I ran the unit tests, linting, and format checks successfully. I tried running the integration tests, but many of them failed. There didn't appear to be any integration tests in the packages that were affected by this change, however.

Jeffail (approved):
LGTM, thanks @tprestegard, don't worry about the integration tests, they don't block PRs as they're extremely heavy, and running all those services is quite error prone.

PR #2345, opened 2024-01-24
More advanced testing output

calmera:
These changes add more advanced testing output capabilities beyond the one that exists today. It sets the scene for having multiple test output formatters and giving users the choice between something human or machine readable.

A `--legacy` flag (currenly set to true by default) allows users to indicate which runner they want to use; the legacy one or the neo one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,12 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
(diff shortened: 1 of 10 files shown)


mihaitodor:
Would you mind rebasing this and force pushing to the same branch so the diff doesn't include recent commits from main? It's a bit hard to read the diff right now.

calmera:
Yeah it is a bit of a mess. I might drop this PR, take a diff of the changes I made and re-apply a different PR just to make things clear.

PR #2347, opened 2024-01-27
fix connection leakage

gangtao:
in my test, the nats connection was not released even when the pipeline is deleted.

this fix will release the connection.


Jeffail (approved):
LGTM, thanks @gangtao!

PR #2352, opened 2024-01-30
Add NATS Service Input

marshauf:
These changes add a Benthos input component, which allows publishing Service Endpoints inside NATS.
The input component receives request from NATS in form of messages and allows replying with a sync response component or returning an error with reject output.

The documentation refers to NATS Services as Microservices. I heard that NATS/Synadia plans to drop the "Micro" and just call them Services. I am happy to change the description.

Based on capabilities of https://pkg.go.dev/github.com/nats-io/nats.go/micro.

The service statistic endpoint does not report the correct execution time, because the request gets put into a channel and the handler function returns directly. I had to turn the push based nature of the micro.Service into a pull/read based approach, since the Input interface is pull based.

An alternative approach would be to implement most of the capabilities of the micro package. But that seemed to much work, just to get correct service statistics. Or make changes inside the micro package itself.

I tried running integration tests, but they just didn't do anything. That is probably because I do not have docker, but Podman. Even setting an alias didn't help. I tested the component by hand and it is currently in use in a development environment with several endpoints being served by one Benthos stream.

I am happy to hear any feedback and explain more.


codegangsta:
👋 Hey! I know this is a bit late, but I'll be taking a look at this as well as the req/reply processor this week

marshauf:
Hi, thank you for taking a look at it.

It maybe that this processor has a memory leak. At least I noticed the memory consumption grew slowly over days.
I am also not sure if the connection handling is correct. Cases like graceful shutdown and reconnecting after a disconnection from s NATS Server.

PR #2353, opened 2024-01-30
Add NATS KV Cache Component

marshauf:
Adds a Benthos Cache component, which allows caching values inside a NATS Jetstream Key Value Bucket, referenced by a key.
TTL settings from Benthos are ignored, because NATS KV does not support TTL on the key value level, currently.

Developed alongside  #2352 and I am also happy to hear any feedback.


Jeffail (approved):
LGTM, thanks @marshauf!

PR #2357, opened 2024-01-31
Add credit property to amqp1 input

andreas-habel:
Exposing the `credit` property of the amqp ReceiverOptions allows for a performance increase when receiving messages.
My tests with a local broker went up from ~50 msg/sec with `credit = 1` to 1k or more msg/sec with `credit = 64`
It's similar to existing `max_in_flight` properties, but to keep the parameter recognizable for amqp1 users, we will stick to `credit`.

Since the default of the property has always been `1`, this should not introduce any issues to existing flows.

The description of the `amqp1` component has been externalized into a separate markdown file and is reinjected with `go:embed` for much better readability (IMHO).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 - Field `address_cache` added to the `socket_server` input.
-- Field `read_header` added to the `amqp_1` input.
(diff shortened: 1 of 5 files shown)


Jeffail (approved):
LGTM, thanks @andreas-habel! And thanks for making those tweaks! 😃

Going to merge now! Look out!

PR #1583, opened 2022-11-22
Add CosmosDB components

mihaitodor:
In the future, it might be worth injecting https://github.com/btnguyen2k/gocosmos as a driver for the `sql_*` components.

TODO:
- [ ] Add support for pagination and querying across partition keys to the input
- [ ] Make the partition key optional for the output and query the container to read it
- [ ] Add a processor too...

PS: This needs more work before merging... I might get to finish it eventually.

PR #2358, opened 2024-02-01
Add CosmosDB components and gocosmos SQL driver

mihaitodor:
This is a revamp of #1583, where I did the initial exploration to add these components. Since then, [gocosmos](https://pkg.go.dev/github.com/microsoft/gocosmos) became an official Microsoft SQL driver for CosmosDB and it is currently marked as experimental, so I decided to include it as well because it comes with some extra goodies, such as support for [hierarchical partition keys](https://learn.microsoft.com/en-us/azure/cosmos-db/hierarchical-partition-keys) and [cross-partition queries](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/how-to-query-container#cross-partition-query), which are not currently supported by the [`azcosmos`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos) driver.


public/service/message.go
```diff
@@ -501,6 +501,20 @@ func (b MessageBatch) TryInterpolatedString(index int, i *InterpolatedString) (s
 	return i.expr.String(index, msg)
 }
 
+// TryInterpolatedAny resolves an interpolated string expression on a message
+// batch, from the perspective of a particular message index.
+//
+// This method allows interpolation functions to perform windowed aggregations
+// across message batches, and is a more powerful way to interpolate strings
+// than the standard .String method.
+func (b MessageBatch) TryInterpolatedAny(index int, i *InterpolatedString) (any, error) {
```
mihaitodor, line 510: I hope this is a reasonable addition. While it might be a bit tricky to troubleshoot if the expression returns an unexpected type, I think it's too cumbersome to have a separate field for the type and then parse the string returned by interpolation to match that type. For Patch operations, users might want to add an object or array field and the API doesn't allow me to pass those as strings.

internal/impl/azure/output_cosmosdb.go
```diff
@@ -0,0 +1,151 @@
...
+
+The `+"`partition_keys`"+` field must resolve to the same value(s) across the entire message batch.
+`+cosmosdb.CredentialsDocs+cosmosdb.BatchingDocs)).
+		Footnotes(cosmosdb.EmulatorDocs).
+		Fields(cosmosdb.ContainerClientConfigFields()...).
+		Field(cosmosdb.PartitionKeysField(true)).
+		Fields(cosmosdb.CRUDFields(false)...).
+		Field(service.NewBatchPolicyField(cdboFieldBatching)).
+		Field(service.NewOutputMaxInFlightField()).
+		LintRule("root = []"+cosmosdb.CommonLintRules+cosmosdb.CRUDLintRules).
```
mihaitodor, line 36: This is a tiny bit awkward, so I'm kinda' wondering if maybe it's worth having a `LintRules()` API which does `root = []` implicitly and takes as input a variadic list of strings, each containing some bloblang snippet with one or more `root."-" = ...` and maybe extra variables. I don't really like it, though, since it would be hard to spot bugs in the composed mapping.

public/service/message.go
mihaitodor: Removed this and switched to full bloblang mapping fields as discussed during the Blobtalk.

public/service/message.go
```diff
@@ -366,6 +367,34 @@ func (m *Message) BloblangQuery(blobl *bloblang.Executor) (*Message, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	switch res.(type) {
+	case value.Delete:
+		return nil, bloblang.ErrRootDeleted
+	case value.Nothing:
+		return nil, nil
+	}
```
mihaitodor, line 394: I'm not sure what's the most user-friendly behaviour here, but I imagine in some cases they might want to be abel to detect that someone added `root = deleted()` in the config. I did the same for `MessageBatch. BloblangQueryValue()`.

public/service/message.go
```diff
@@ -453,6 +482,41 @@ func (b MessageBatch) BloblangQuery(index int, blobl *bloblang.Executor) (*Messa
...
+// batches.
+func (b MessageBatch) BloblangQueryValue(index int, blobl *bloblang.Executor) (any, error) {
+	uw := blobl.XUnwrapper().(interface {
+		Unwrap() *mapping.Executor
+	}).Unwrap()
+
+	msg := make(message.Batch, len(b))
+	for i, m := range b {
+		msg[i] = m.part
+	}
```
mihaitodor, line 499: Is this necessary here? I left it in since I don't know exactly why `BloblangQuery()` does it, but I'm guessing in that case it's relevant because it returns messages.

Jeffail: Yeah we need to provide the batch as execution context, it's how the `.from` methods work.

Jeffail (approved):
Nice, LGTM, thanks @mihaitodor. I've added a commit on top that does some minor adjustments, one of them is renaming the map based fields so that they have a `_map` suffix, just a heads up in case you have demos with the old field names.

PR #2363, opened 2024-02-02
fix: always bind a given stream when SubscribeSync()

Anton-Frank:
When trying to use nats_jetstream input plugin, this was not possible for streams that were mirrored from a stream in a different jetstream domain. 

This is fixed by adding the given stream name always to the options when calling SubscribeSync().

What is not solved is the issue, that the stream cannot be resolved from the subject name, if the stream name is not explicitly given and the stream is a mirror of another stream from a different jetstream domain. 

In this case, the documentation "Either a subject or stream must be specified" is not valid, because both settings need to be set.


Jeffail:
Thanks @Anton-Frank, that's a doozy to diagnose so I really appreciate your investigation. I'll probably add a paragraph to the general docs rather than the specific field docs to try and explain the requirements for mirrored streams.

Jeffail (approved):

PR #2370, opened 2024-02-09
Punt unused field

mihaitodor:
Also clean up .golangci.yml a bit.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -27,10 +27,11 @@ linters-settings:
       - ifElseChain # Mostly false positives
       - ptrToRefParam # False positives?
       - importShadow # Probably not worth the hassle...
-  depguard:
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -27,10 +27,11 @@ linters-settings:
       - ifElseChain # Mostly false positives
       - ptrToRefParam # False positives?
       - importShadow # Probably not worth the hassle...
-  depguard:
-    include-go-root: true
-    packages:
-      - "io/ioutil"
+  # depguard:
```
mihaitodor, line 30: Leaving this here as an example. We can re-enable this linter (it's commented out below) when we'll need it.

Jeffail (approved):
LGTM

Issue #1167, opened 2022-03-10
New output component for BigQuery for the new Write API


Locustv2:
As the current (BigQuery component)[https://www.benthos.dev/docs/components/outputs/gcp_bigquery] is still experimental, it would be good to replace it with a new component using the Write API from Big Query: (BigQuery Write API explained: An overview of the Write API)[https://cloud.google.com/blog/topics/developers-practitioners/bigquery-write-api-explained-overview-write-api]

The advantage of using this api is that it uses Protobuf data. Which means it's faster as it will be moving binary data around.

Here are some additional material: (Learn how to stream JSON data into BigQuery using the new BigQuery Storage Write API)[https://cloud.google.com/blog/topics/developers-practitioners/learn-how-stream-json-data-bigquery-using-new-bigquery-storage-write-api]

irfanhabib:
The link appears to be dead! But I would be interested in this too.  New link: 
https://cloud.google.com/bigquery/docs/write-api

PR #2375, opened 2024-02-11
gcp: Switch BigQuery output to Storage Write API.

grosskur:
Currently, the BigQuery output uses the "load jobs" API to batch load data, which works but has limitations when used to make frequent updates as is typically done in a streaming system. For example, the default quota is 1,500 load jobs per table per day, which means you can only create ~1 load job per minute. See:

https://cloud.google.com/bigquery/quotas#load_jobs

The Storage Write API is a newer API that works for both batch and streaming and has some advantages. See:

https://cloud.google.com/bigquery/docs/write-api#advantages

This change switches the BQ output to use the Storage Write API. At a high level, it now creates a "managed stream" and then repeatedly calls `AppendRows` on it to send batches of data. Each row is a seralized protobuf message, and the API uses gRPC bi-directional streaming:

https://github.com/googleapis/googleapis/tree/master/google/cloud/bigquery/storage/v1

Notes:

- Only simple JSON messages have been tested with `STRING`, `INTEGER`, and `DATETIME` fields. CSV messages have not been tested.

- The type of each field in a message needs to be mapped to a protobuf field type for transport, which is handled using reflection and heuristics. To make this work reliably, the BQ table must exist in order to infer the target schema.

- The "default" stream (at-least-once delivery) is always used. Using an application-created stream (exactly-once delivery, ACID features) is not supported.

- The options `write_disposition`, `create_disposition`, `ignore_unknown_values`, `max_bad_records`, and `auto_detect` are no longer supported because they were features of the load jobs API. Some of them could potentially be reimplemented.

- The option `job_labels` is no longer supported because it is specific to the load jobs API.

Fixes: #1167

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -129,6 +129,7 @@ require (
 	golang.org/x/sync v0.6.0
 	golang.org/x/text v0.14.0
 	google.golang.org/api v0.157.0
+	google.golang.org/grpc v1.60.1
 	google.golang.org/protobuf v1.32.0
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1
 	gopkg.in/yaml.v3 v3.0.1
@@ -312,7 +313,6 @@ require (
 	google.golang.org/genproto v0.0.0-20240102182953-50ed04b92917 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20240102182953-50ed04b92917 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20240116215550-a9fa1716bcac // indirect
-	google.golang.org/grpc v1.60.1 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/jcmturner/aescts.v1 v1.0.1 // indirect
 	gopkg.in/jcmturner/dnsutils.v1 v1.0.1 // indirect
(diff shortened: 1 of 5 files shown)


grosskur:
This is PR is still pretty preliminary, but I wanted to send it in case it's useful. It doesn't have as much functionality as the current `gcp_bigquery` output that uses the load jobs API, so maybe it could be added under a different name, e.g., `gcp_bigquery_write` so as not to disrupt existing users. Open to any/all thoughts and feedback.

Jeffail:
Hey @grosskur, yeah I think a separate output type specifically for this api would be easier for us than trying to reach parity with the current behaviour, we could document the advantages/disadvantages and leave it to users to decide. I think `gcp_bigquery_write` works as a name.

grosskur:
Cool---thanks for the feedback. I'll spend some time and get it cleaned up as a separate output called `gcp_bigquery_write`.

bir:
@grosskur Any progress?  We are also interested in leveraging BQ

firegrass:
Hey @grosskur are you still using this streaming api? Does this code work? Thanks in advance.

firegrass:
Just a quick update. I have this working here if anyone wants to copy it.

https://github.com/mooselabs/redpanda-connect-plugin-gcp-bigquery-stream

PR #2386, opened 2024-02-17
fix(tracing): usage of LogKV and record error

### LINKED ISSUE #2384: OpenTelemetry tracing errors should use RecordError function
OpenTelemetry contains a function called `RecordError` on the span struct to record errors on the spans.

Currently, this is how we are storing errors on spans:
```
		span.SetTag("error", "true")
		span.LogKV(
			"event", "error",
			"type", err.Error(),
		)
```

This doesn't follow OpenTelemetry's specification on error handling, and then spans will not be tagged as an "error" on tracing visualization platforms.

Since we have this on the docs:
```
WARNING: Although the configuration spec of this component is stable the format of spans, tags and logs created by Benthos is subject to change as it is tuned for improvement.
```

We can change the behavior, right?

### LINKED ISSUE #2385: Tracing implementation LogKV usage is incorrect throughout the codebase
The current implementation for the `LogKV` function is the following:

```
// LogKV adds log key/value pairs to the span.
func (s *Span) LogKV(name string, kv ...string) {
	if s == nil {
		return
	}
	var attrs []attribute.KeyValue
	for i := 0; i < len(kv)-1; i += 2 {
		attrs = append(attrs, attribute.String(kv[i], kv[i+1]))
	}
	s.w.AddEvent(name, trace.WithAttributes(attrs...))
}
```

This means it will create an event `name` (the first function parameter) and it will also create `len(parameters - 1)/2` attributes.

This means we should see in our codebase only usages with an odd number of parameters, which is not currently true (actually every call is invalid):

```go
// processor_dedupe.go:151
ctx.Span(i).LogKV("event", "dropped", "type", "deduplicated")

// processor_dedupe.go:157
ctx.Span(i).LogKV("event", "dropped", "type", "deduplicated")

// processor_group_by_value:102
ctx.Span(i).LogKV("event", "grouped", "type", v)

// processor_group_by:150
ctx.Span(i).LogKV("event", "grouped", "type", groupStr)

// processor_group_by:157
ctx.Span(i).LogKV("event", "grouped", "type", "default")

// processor_while:156
ctx.Span(i).LogKV("event", "loop")
```

This means we are discarding the last parameter of these calls.

Go Playground showing the code execution:
https://go.dev/play/p/yzi491xq1Ob

lucasoares:
Fixes #2385
Fixes #2384


internal/component/processor/error.go
```diff
@@ -19,10 +19,6 @@ func MarkErr(part *message.Part, span *tracing.Span, err error) {
 		span = tracing.GetActiveSpan(part)
 	}
 	if span != nil {
-		span.SetTag("error", "true")
-		span.LogKV(
-			"event", "error",
-			"type", err.Error(),
-		)
+		span.RecordError(err, true)
```
lucasoares, line 22: I'm recording the stacktrace by default but this should be configurable.

What is the best approach to let users configure if want to record stacktrace or not in this case, since this the places where `RecordError` function is used can't reach the `traces` configuration?

internal/tracing/v2/span.go
```diff
@@ -43,6 +44,16 @@ func (s *Span) LogKV(name string, kv ...string) {
 	s.w.AddEvent(name, trace.WithAttributes(attrs...))
 }
 
+// RecordError adds an error to the span.
+func (s *Span) RecordError(err error, recordStackTrace bool) {
+	if s == nil || err == nil {
+		return
+	}
+
+	s.w.RecordError(err, trace.WithStackTrace(recordStackTrace))
```
lucasoares, line 53: For documentation: https://opentelemetry.io/docs/languages/go/instrumentation/#record-errors

PR #2403, opened 2024-02-29
Added direct support for oauth2 tokens in kafka using franz-lib

cthtrifork:
Initially I thought it was not supported, but I talked to the author of the library a while ago and got it confirmed:
https://github.com/twmb/franz-go/issues/482

I then implemented the setup in this project:
https://github.com/redpanda-data/console/issues/699

And would like to do the same here.

I prefer not to use the external cache mechanism for oauth2 tokens, also from a security standpoint of having the token stored several places.

### DIFF
--- a/internal/impl/kafka/sasl.go
+++ b/internal/impl/kafka/sasl.go
@@ -320,6 +320,7 @@ func SaramaSASLField() *service.ConfigField {
 			Secret(),
 		service.NewStringField(saramaFieldSASLAccessToken).
 			Description("A static OAUTHBEARER access token").
+			Secret().
 			Default(""),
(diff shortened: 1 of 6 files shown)


cthtrifork:
I am not sure I understand the framework completely yet. Can we store the token between requests without get a new token on each message-in-flight?

cthtrifork:
I have now tested the implementation and I can see it requests a new token on each message, so I guess we need to rely on a cache and hope there is no security issue.

Closing this.

cthtrifork:
> I am not sure I understand the framework completely yet. Can we store the token between requests without getting a new token on each message-in-flight?

I have tested some more and it seems to work. When running parallism, it will naturally get more tokens

PR #2417, opened 2024-03-06
added support for `new` and `last_per_subject` delivery strategies

calmera:
NATS supports the retrieval of last messages per subject and only new messages. Now benthos does as well


Jeffail (approved):
LGTM, thanks @calmera! You are the light that draws me ever closer towards the end of this tunnel of despair.

PR #2433, opened 2024-03-22
Skip processing SFTP file if the cache.Get returns an unknown error

dmateusp:


Jeffail (approved):
LGTM, thanks @dmateusp!

PR #2436, opened 2024-03-24
SFTP: remove mutex lock/unlock in ack

joffreychambrin:


mihaitodor:
Thank you for your contribution! This should be addressed via #3498 which should get merged soon.

PR #2459, opened 2024-03-30
Fix DockerHub REST API examples

mihaitodor:
Fixes #2422.

Looks like DockerHub now rejects the default `Content-Type:application/octet-stream` header which Benthos sends by default with HTTP error 406.


Jeffail (approved):
LGTM, thanks @mihaitodor

PR #2464, opened 2024-03-31
Prepare for pr 2458: add linter tagalign and fix source code

peczenyj:
this is a preparation for pull request https://github.com/benthosdev/benthos/pull/2458

just run linter tagalign and run autofix

enjoy


Jeffail:
I think I'm going to pass on this one, mostly because these struct based config blocks are on the way out anyway.

PR #2488, opened 2024-04-03
Fix deadlock in nats_kv cache

mihaitodor:
The process will exit if connect() fails, but it's probably best to up resources anyway.

### DIFF
--- a/internal/impl/nats/cache_kv.go
+++ b/internal/impl/nats/cache_kv.go
@@ -89,16 +89,16 @@ func (p *kvCache) connect(ctx context.Context) error {
 	}
 
 	var err error
-
-	defer func() {
-		if err != nil {
-			p.disconnect()
-		}
-	}()
-
 	if p.natsConn, err = p.connDetails.get(ctx); err != nil {
 		return err
 	}
+
+	defer func() {
+		if err != nil {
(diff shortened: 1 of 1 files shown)


internal/impl/nats/cache_kv.go
```diff
@@ -89,23 +89,18 @@ func (p *kvCache) connect(ctx context.Context) error {
...
-		}
-	}()
-
 	if p.natsConn, err = p.connDetails.get(ctx); err != nil {
 		return err
 	}
 
 	var js nats.JetStreamContext
 	if js, err = p.natsConn.JetStream(); err != nil {
+		p.natsConn.Close()
```
Jeffail, line 98: We need `p.natsConn = nil` as well here

internal/impl/nats/cache_kv.go
```diff
@@ -89,23 +89,18 @@ func (p *kvCache) connect(ctx context.Context) error {
...
 		return err
 	}
 
 	if p.kv, err = js.KeyValue(p.bucket); err != nil {
+		p.natsConn.Close()
```
Jeffail, line 103: Same here

Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #2491, opened 2024-04-05
Fix `azure_blob_storage` input pagination

mihaitodor:
I wonder if this is the right way to implement it... I noticed that we can cache the pager that's returned by `NewListBlobsFlatPager()` which I think avoids some extra calls, but it needs testing with a large bucket that has over 5k items due to the way `MaxResults` is implemented (details [here](https://learn.microsoft.com/en-us/rest/api/storageservices/list-blobs?tabs=microsoft-entra-id#uri-parameters)).

### DIFF
--- a/internal/impl/azure/input_blob_storage.go
+++ b/internal/impl/azure/input_blob_storage.go
@@ -150,7 +150,6 @@ func newAzureObjectTarget(key string, ackFn codec.ReaderAckFn) *azureObjectTarge
 //------------------------------------------------------------------------------
 
 func deleteAzureObjectAckFn(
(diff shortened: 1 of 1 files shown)


mihaitodor:
@Jeffail @mfamador [said](https://discord.com/channels/746368194196799589/1225448248316137604/1225732683469164565) on Discord that it did the trick.

mfamador:
Yes, this fixes the issue. Thanks, @mihaitodor !

Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #2494, opened 2024-04-06
Silence some output logs

### LINKED ISSUE #2489: [elasticsearch] Suppress credentials in logs
**Version**

v4.26.0

**Problem**

Currently username and password are printed in the logs for elasticsearch.

```
level=info msg="Sending messages to Elasticsearch index at urls: [https://elastic:espass@elasticsearch-master.ns-ext:9200]" @service=benthos label=o_kafka_es_c path=root.output_resources
```

**Feature request**

Credentials should be masked in the logs.

Thank you.

mihaitodor:
Fixes #2489.

Not sure if logging this entry is really all that useful TBH. Maybe we should remove it entirely? It doesn't seem to be consistent across all outputs anyway.

### DIFF
--- a/internal/impl/aws/output_dynamodb.go
+++ b/internal/impl/aws/output_dynamodb.go
@@ -234,7 +234,7 @@ func (d *dynamoDBWriter) Connect(ctx context.Context) error {
 	}
 
 	d.client = client
-	d.log.Infof("Sending messages to DynamoDB table: %v\n", d.conf.Table)
+	d.log.Debugf("Sending messages to DynamoDB table: %v\n", d.conf.Table)
 	return nil
 }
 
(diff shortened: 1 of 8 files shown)


Jeffail:
Personally I think it's nice to have info logs (just a single one) for each active input and output just so that you know at a glance:

A: The config you have applied is working correctly
B: It's actively connecting to external services

We don't necessarily need the log to contain actual details about the connection though, just something vague like "NATS output connected" would be enough. We could potentially do this at the async output level so it doesn't need to be implemented for every single output though.

If you leave this with me I might have a look at doing it today actually.

Jeffail:
Got it here: https://github.com/benthosdev/benthos/commit/d5613d4ea772788472fb5930a73092656b267ffc

PR #2495, opened 2024-04-08
Enhance `nats_kv` processor

mihaitodor:
This is a slightly-breaking change to the nats_kv processor which is still beta. It alters the behaviour of the `keys` operation which now returns an array of keys instead of a batch of messages, as discussed with @codegangsta.

I'm a bit concerned that the history operation still returns a batch of messages, which might confuse people. Happy to modify that one as well to return an array of objects if you'd like. In the current implementation, some of the fields of each history item are emitted as message metadata, which is a bit odd anyway, since they're not identical across the batch...

PS: I also took the liberty to clean up the code a tiny bit.


internal/impl/nats/processor_kv.go
```diff
@@ -286,18 +311,35 @@ func (p *kvProcessor) Process(c context.Context, msg *service.Message) (service.
...
+			select {
+			case entry := <-watcher.Updates():
+				if entry == nil {
+					break loop
+				}
+				keys = append(keys, entry.Key())
+			case <-ctx.Done():
+				return nil, fmt.Errorf("watcher update loop exited prematurely: %s", ctx.Err())
+			}
 		}
```
mihaitodor, line 336: I'm a bit concerned about this approach, because it seems to block indefinitely in some cases. For example, try setting `keys_filter: '"001.>"'`. Not sure if this is by design, or a bug, but I think we might need to consider adding a `timeout` field that lets us error out if a timer expires while we wait for updates. WDYT?

internal/impl/nats/processor_kv.go
```diff
@@ -275,7 +300,7 @@ func (p *kvProcessor) Process(c context.Context, msg *service.Message) (service.
 		p.addMetadata(m, key, 0, nats.KeyValuePurge)
 		return service.MessageBatch{m}, nil
 
-	case kvHistory:
+	case kvpOperationHistory:
 		entries, err := kv.History(key)
```
mihaitodor, line 304: I just noticed that `kv.History()` supports wildcards in the `key`. Not sure if this is by design @codegangsta, but I think it would be better to adjust the docs and pass in `keys_filter` instead of `key` for this operation as well. Also, see the comment about changing this one to return an array of objects instead of a batch of messages.

internal/impl/nats/processor_kv.go
```diff
@@ -286,18 +311,35 @@ func (p *kvProcessor) Process(c context.Context, msg *service.Message) (service.
...
 		if err != nil {
 			return nil, err
 		}
-		batch := service.MessageBatch{}
-		for _, key := range keys {
-			m := service.NewMessage([]byte(key))
-			m.MetaSetMut(metaKVBucket, p.bucket)
-			batch = append(batch, m)
+		defer func() {
+			_ = watcher.Stop()
```
mihaitodor, line 321: Does it make sense to return (or log) an error here? The Nats client library ignores it completely: https://github.com/nats-io/nats.go/blob/46be7408f6630569929044fcad459ba0844a296d/kv.go#L870

codegangsta:
Taking a peek at this right now

internal/impl/nats/processor_kv.go
codegangsta: Interesting you are getting blocking on that value, I wonder if something is breaking when you send NATS a subject it doesn't like.

All that said, it's entirely possible for the watcher to never end, depending on how fast keys are being inserted. More of an edge case, but one concession we could make is to either have a configurable timeout, or stop the watcher when a specific time is reached, or a combination of the two.

I don't particularly like the idea that a simple timeout would result in receiving a partial set of keys. so I agree an error would be best in the simple timeout case.

That said, I kinda do like the idea of just returning the keys up to time that we requested it, would result in potentially more stale data for more timeliness on the retrieval side.

This is definitely more of an edge case where we have a really hot kv at the moment the watcher is created. I'm cool with any of the solutions.

internal/impl/nats/processor_kv.go
codegangsta: Most of the errors have to do with race conditions, having to do with:

1. The connection being closed or draining
2. The underlying subscription already being unsubscribed or
3. The underlying consumer already being deleted

All possible but not particularly actionable for this use case, and generally exceptional. **Maybe** a DEBUG log could be useful here someday but I can't think of a reason why it would be at the moment.

website/docs/components/processors/nats_kv.md
```diff
@@ -225,6 +226,26 @@ revision: "42"
 revision: ${! @nats_kv_revision }
 ```
 
+### `keys_filter`
```
codegangsta, line 229: I'm going back and forth with the idea of introducing this vs reusing the `key` field. My hangup is that at the end of the day they are serving the same purpose for any `watch` related options, so `History`, `Keys` and `Watch` in the input. I wonder if the `key` field should be used for all of it. WDYT

internal/impl/nats/processor_kv.go
codegangsta: We should absolutely make it consistent, I added some questions about the use of `keys_filter` in another comment to discuss.

internal/impl/nats/processor_kv.go
mihaitodor: I think I'll go with the more conservative route and just return an error if the timeout lapses. People should take that as a sign that they either need to bump up the timeout or use a wildcard which selects fewer keys. We could enable partial results later on with a separate flag if anyone needs that functionality.

I'll slap together some quick code which shows how `'"001.>"'` blocks and I'll open an issue here https://github.com/nats-io/nats.go.

codegangsta (changes requested):
Added some questions and thoughts.

One overall thought is that while we are here, we should at least upgrade to the new version of the jetstream client, which is now in it's own package: `github.com/nats-io/nats.go/jetstream`.

For KV there shouldn't be much in the way of changes, I think we can circle back on the other jetstream components in another PR, as the more significant changes are in how consumers work.

I can ofc help with the implementation, just lmk

internal/impl/nats/processor_kv.go
mihaitodor: Sounds good, I added a debug log just in case someone needs to troubleshoot it.

internal/impl/nats/processor_kv.go
mihaitodor: Issue opened here: https://github.com/nats-io/nats.go/issues/1608

mihaitodor:
@codegangsta I updated the `nats_kv` components code to use the new `github.com/nats-io/nats.go/jetstream` in 5c4f02f68. It's nice to see that the API calls are context-aware now. Wasn't sure if it's ready for prime-time, but it seems to work well. I'll leave the upgrade of the `nats_jetstream` components for a separate PR.

I have a few more comments to address. Stay tuned :)

website/docs/components/processors/nats_kv.md
mihaitodor: Yeah, I'm leaning towards that as well. My initial thought was that it might be confusing for users if they set a wildcard for operations which don't support them, but they should get an error in that case anyway. I'll just document that the `history` and `keys` operations support wildcards. Should've done this from the start.

mihaitodor:
OK, I went ahead and also adjusted the `history` operation to return a list of objects instead of a batch of messages and documented the fields that are returned. See 1a8d57b. Guess I can DRY up the code a bit more if need be, but I'm curious if you think this breaking change makes sense.

internal/impl/nats/processor_kv.go
mihaitodor: And I also added a `timeout` field in 83c662e.

codegangsta: I think that's a good solution for most cases!

website/docs/components/processors/nats_kv.md
codegangsta: Yeah. Looking at the code I'm leaning more toward that! Nice call

codegangsta (approved):
Looking good!

Wish there was a release when this is merged since I had to do some ungodly acts in my next video to get keys. Thankfully I won't be exposing any breaking changes there. Thanks so much for doing this. Hit me up on discord if you are interested in doing the `nats_jetstream` component updates since I do infact have a lot of opinions there, and may be working on some this week if you want to move on to other things.

Thanks again for updating to the newer jetstream api

mihaitodor:
Awesome, thanks @codegangsta! I'll defer to @Jeffail for the rest of the review. Hope he's OK with the small breaking changes :)

> Hit me up on discord if you are interested in doing the `nats_jetstream` component updates since I do infact have a lot of opinions there, and may be working on some this week if you want to move on to other things.

Sounds good! I might take a stab at it this week.

Jeffail (approved):
LGTM, thanks @mihaitodor!

Issue #2190, opened 2023-10-26
Schema_registry_encode subject not found when contains "/"


nrabe1010:
The `schema_registry_encode` processor with protobuf data cannot find schemas that include a `/` in the subject name. This affects all versions as far as I can tell. My guess is this is due to a bug with the subject name in the registry client request not being escaped properly.

```
level=error msg="schema subject 'proto/common/envelope.proto' not found by registry" @service=benthos label="" path=root.pipeline.processors.0
```
<img width="1487" alt="registrySubject" src="https://github.com/benthosdev/benthos/assets/22421773/19071fbc-6ea7-42b5-a53b-810f64c82ea2">

Here is a sample proto definitions I encountered this with. I can put together a gist of an example of this if requested.

(registered as `/proto/common/envelope.proto`)
```
syntax = "proto3";

package common;

message Envelope {
  string collectorId = 6;
}

```

(registered as product-value)
```
syntax = "proto3";

package raw;

import "proto/common/envelope.proto";

message Product {
  common.Envelope envelope = 1;
  optional uint64 aNumber = 2;
}
```

Jeffail:
Hey @nrabe1010, thanks for the info, you're correct in that the subject URL paths were not being escaped. I've got a fix here: https://github.com/benthosdev/benthos/commit/7ae5fe618d1e0f501c3d61dfac63c67938d24667 which will be in the next release. If you get a chance to try it out with your use case please let me know how it goes.

nrabe1010:
Thanks @Jeffail for the quick turnaround.

I was able to test this today and noticed the fix partially works. The referenced schema is not found, however, if I register two versions of the referenced `proto/common/envelope.proto` schema, one as is, and the other with a name as `proto%2Fcommon%2Fenvelope.proto`, Benthos is able to ser/de without issue.

Jeffail:
Argh, thanks for testing @nrabe1010, I don't suppose there's a schema registry API spec that I can reference that you know of? If we're not meant to URL escape the subject then I'm not sure what it would expect us to do.

nrabe1010:
I am not aware of one. I haven't been able to do too much digging into this, but did notice the following behavior:
```
curl -u admin:local-dev "http://127.0.0.1:30081/subjects/proto/common/envelope.proto/versions/latest"
# 404, definitely not an endpoint so this is expected

[...]

PR #2498, opened 2024-04-08
don't escape paths twice when using a schema registry

huttotw:
In the case where we had a subject like `my/deeply/nested/message.proto` it was being doubly escaped like:
```
/subjects/my%252Fdeeply%252Fnested%252Fmessage.proto/versions/1
```
when it should be:
```
/subjects/my%2Fdeeply%2Fnested%2Fmessage.proto/versions/1
```

Note that by default Go will unescape in the `Path` field of `url.URL`, to access the raw path you can use `RawPath`. Not sure if we should change [this test func ](https://github.com/huttotw/benthos/blob/7b9988a291b3a27bf6e6ddbd6aa0bd3b60bb9125/internal/impl/confluent/processor_schema_registry_decode_test.go#L87)to use the `RawPath` or not?


Jeffail:
Hey @huttotw, is there a formal spec for how a schema registry server expects these paths to be formed? We've had issues with this before: https://github.com/benthosdev/benthos/issues/2190 but I find it very difficult to get this "right" when we're having to guess based on different implementations of the server.

huttotw:
@Jeffail thanks for the quick response!

I'm not certain that the bug was fixed in #2190, I tried to reproduce the scenario [here](https://go.dev/play/p/AHZj-sve905). Let me know what you think!

I'm referencing Confluent's docs here, but the only conclusion I can draw is that subject should be treated as an opaque identifier:

https://docs.confluent.io/platform/current/schema-registry/develop/api.html#get--subjects-(string-%20subject)-versions-(versionId-%20version)

Nevertheless, I don't think we want to escape twice; that seems a bit unexpected from all sides. It looks like we should be using the `JoinPath` method instead of manually setting the `Path` based on a `JoinPath` result, there are some other things happening under the hood it looks like when Path gets set via `JoinPath`.

Issue #2232, opened 2023-11-11
NATS Request/Reply support


calmera:
Nats supports the Request/Reply paradigm which offers semi-synchronous communication. We currently already have sync_response working for http_server, but it would be great to have the same working for NATS Request/Reply.

twmb:
Closing as a niche feature request. The NATS input/output exist but request/reply is a specialized pattern. If there is renewed interest, feel free to open a new issue or PR.

PR #2512, opened 2024-04-14
Add nats_reply_subject to NATS message metadata

Zambito1:
This is not an ideal solution to #2232, but this change does make it possible to reply to NATS requests. Just add a NATS output that publishes on the subject `${! metadata("nats_reply_subject") }`. Even with the more ideal `sync_response` implementation, I think this would still be good to have.

### DIFF
--- a/internal/impl/nats/input.go
+++ b/internal/impl/nats/input.go
@@ -24,6 +24,7 @@ This input adds the following metadata fields to each message:
 
 ` + "``` text" + `
 - nats_subject
+- nats_reply_subject
 - All message headers (when supported by the connection)
 ` + "```" + `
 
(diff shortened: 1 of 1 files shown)


Jeffail:
Hey @Zambito1, LGTM but can you please appease the benevolent DCO bot with `git commit --amend -s` and then force push?

Zambito1:
the DCO bot seems to like me now

Jeffail (approved):
Thanks @Zambito1, the bot may indeed like you, but I love you (like a distant relative) and so if I were ever to come at odds with the bot I would expect you to swear loyalty to myself alone.

ianschenck:
Simply adding a NATS output that ships messages to that subject results in a race between ACKing the message and sending the reply message. NATS overloads the Reply field on msg, and uses it for ACKing/NACKing as well as request-reply, so you randomly will receive either `+ACK` or the message from the output.

Zambito1:
Yeah, I ran into this also @ianschenck. This definitely shouldn't be used for REQ/REP in production.

PR #2515, opened 2024-04-14
Fix parse_form_url_encoded bloblang method

mihaitodor:
Use correct data type when a query parameter is repeated multiple times.


Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #2520, opened 2024-04-15
Add `persistent` field to the `amqp_1` output

mihaitodor:
The `amqp_0_9` has this field already: https://www.benthos.dev/docs/components/outputs/amqp_0_9/#persistent

Implementation inspired from here: https://github.com/Azure/go-amqp/issues/11

This omission was reported on Slack: https://gophers.slack.com/archives/CLWCBK7FY/p1713189832118529

TODO:
- [ ] Wait for confirmation that it works as expected

### DIFF
--- a/internal/impl/amqp1/config.go
+++ b/internal/impl/amqp1/config.go
@@ -28,6 +28,7 @@ const (
 	targetAddrField  = "target_address"
 	appPropsMapField = "application_properties_map"
 	metaFilterField  = "metadata"
+	persistentField  = "persistent"
 )
 
 // ErrSASLMechanismNotSupported is returned if a SASL mechanism was not recognized.
--- a/internal/impl/amqp1/output.go
+++ b/internal/impl/amqp1/output.go
@@ -55,6 +55,10 @@ This output benefits from sending multiple messages in flight in parallel for im
 			saslFieldSpec(),
 			service.NewMetadataExcludeFilterField(metaFilterField).
 				Description("Specify criteria for which metadata values are attached to messages as headers."),
+			service.NewBoolField(persistentField).
+				Description("Whether message delivery should be persistent (transient by default).").
+				Advanced().
+				Default(false),
 		).LintRule(`
 root = if this.url.or("") == "" && this.urls.or([]).length() == 0 {
   "field 'urls' must be set"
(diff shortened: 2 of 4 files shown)


internal/impl/amqp1/config.go
```diff
@@ -28,6 +28,7 @@ const (
 	targetAddrField  = "target_address"
 	appPropsMapField = "application_properties_map"
 	metaFilterField  = "metadata"
+	persistentField  = "persistent"
```
timo102, line 31: The naming of the field could be confusing in the context of the AMQP 1.0 standard. The AMQP 0.9 standard distinguishes between "durable" and "persistent."  These are different terms with slightly different meaning. 

- Durable: A server resource that survives a server restart.
- Persistent: A message that the server holds on reliable disk storage and MUST NOT lose after a server restart.

The AMQP 1.0 standard only "knows" durability (durable). My conclusion would be: it should be rather named "durable"

https://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-header

https://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-header

internal/impl/amqp1/output.go
```diff
@@ -180,7 +190,11 @@ func (a *amqp1Writer) Connect(ctx context.Context) (err error) {
 	}
 
 	// Create a sender
-	if sender, err = session.NewSender(ctx, a.targetAddr, nil); err != nil {
+	var opts *amqp.SenderOptions
+	if a.persistent {
+		opts = &amqp.SenderOptions{SettlementMode: amqp.SenderSettleModeUnsettled.Ptr()}
```
timo102, line 195: The default value for the SettlementMode is "MixedMode":

```
	// Sender MAY send a mixture of settled and unsettled deliveries to the receiver.
	SenderSettleModeMixed SenderSettleMode = encoding.SenderSettleModeMixed
```
https://github.com/Azure/go-amqp/blob/1003610b1e679a884a60f9011c37e55c6eb07338/const.go#L20

Changing the mode to "Unsettled" could be unnecessary.

mihaitodor:
Thanks for the review @timo102! These changes were meant more as a test based on a [support conversation](https://gophers.slack.com/archives/CLWCBK7FY/p1713189832118529) on the Golang Gophers Slack. I also saw that someone else suggested something similar in #2530 and I'm happy to close this one and move the discussion there.

PR #2526, opened 2024-04-16
Fix sensitive field scrubber

mihaitodor:
Don't throw an error when scrubbing empty password fields.


Jeffail (approved):
LGTM, thanks @mihaitodor! Also thanks for cleaning up the regex :heart:

PR #2556, opened 2024-04-30
Support NAK delay on `nats_jetstream` input

mfamador:
Add an optional parameter to make NATS Jetstream to delay the redelivery of all msgs when negatively acknowledged. It can be used as a retry mechanism when a transient error occurs.

```yml
input:
  nats_jetstream:
    urls:
      - localhost:4222
    subject: foo
    durable: bar
    nak_delay: 5m

pipeline:
  processors:
  -  . . . 

output:
  switch:
    cases:
    - check: 'errored()'
      output:
        reject: "rejecting due to processing error: ${! error() }"
    - output:
        resource: output_ok
``` 

Additionally, we can use a header on each msg to define the unix epoch timestamp until when we want to delay the msg processing. By default is `nak_delay_until` but we can set it with another name using the parameter `nak_delay_until_header`

```yaml
input:
  nats_jetstream:
    subject: foo
    durable: bar

pipeline:
  processors:
  - mapping: throw("transient error")
    
output:
  switch:
    cases:
    - check: '@nak_delay_until.number() - timestamp_unix() > 0'
      output:
        reject: "not time to process it yet ${! @time_to_process }"
              
    - check: 'errored()'
      output:
        nats_jetstream:
          subject: foo
          headers:
            nak_delay_until: ${! (timestamp_unix() + @num_retries * 10).int64() }
            num_retries: ${! @num_retries }
        processors:
        - mapping: meta num_retries = @num_retries.number().or(0) + 1
            
    - output:
        drop: { } # our regular output
        processors:
```

It was also added an opt-in parameter to create the stream and subject if they not exist and another to specify the number of stream replicas when in clustered mode.

### DIFF
--- a/internal/impl/nats/input_jetstream.go
+++ b/internal/impl/nats/input_jetstream.go
@@ -19,7 +19,6 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


mfamador:
Removed the part where we try to update the stream if we change the subject. This way it has the same behaviour as without the create_if_not_exists flag on which it fails if we don't change the durable/consumer as well.

mfamador:
Well, I think it's time to close this one.

PR #2566, opened 2024-05-07
Dynamic credentials support for SQL components

jfrconley:
Currently, there is no way to change the credentials used by any of the SQL components without updating the configuration and restarting the stream. This is a problem for use cases where credentials must be rotated on a regular basis (AWS RDS IAM, CockroachDB JWT,  Hashicorp Vault temporary credentials).

To resolve this, I have implemented an additional `dynamic_credentials` configuration block supported by all SQL components (other than deprecated ones). This property allows specifying a cache resource and a key with which to retrieve credentialing information. The value stored in this key is used to supply interpolation parameters to the dsn field, which has now been made dynamic. The cache is queried and the dsn is regenerated when a new connection is created. Init statements and files are only run once, and are not reapplied when creating a new connection. This has the advantage of working with all drivers, regardless of underlying implementation. A side effect of this approach is that idle connections with stale credentials are not proactively removed. The settings `conn_max_idle` and `conn_max_idle_time` can be used counteract this based on use case requirements.

Example:
```yaml
input:
  sql_select:
    driver: postgres
    // You can use interpolation to insert properties from the retrieve cache item
    dsn: "postgres://${! username }:${! password }@localhost/testdb"
    table: test
    columns: [ "*" ]
    dynamic_credentials:
      cache: cred_cache
      cache_key: input_db_creds

cache_resources:
  - label: cred_cache
    memory:
      init_values:
        input_db_creds: '{"username": "testuser", "password": "testpass" }'  
```

This setup can pair well with the `cached` processor to refresh credentials on a regular interval.


internal/impl/sql/dynamic_credential_wrapper.go
```diff
@@ -0,0 +1,106 @@
...
+	manager            *service.Resources
+}
+
+func NewDynamicCredentialConnector(manager *service.Resources, driverName string, dsn *service.InterpolatedString, dynamicCredentials *DynamicCredentials) (*DynamicCredentialConnector, error) {
+	driveri, err := sql.Open(driverName, "")
+	if err != nil {
+		return nil, err
+	}
+	resolvedDriver := driveri.Driver()
+	_ = driveri.Close()
```
jfrconley, line 29: This is a bit of a hack to get the underlying driver using our driver string. The stdlib sql package specified that no connections should be created when calling open, so this should be generally safe. Tests for all current drivers are written and pass.

jfrconley:
@Jeffail Would it be possible to get a review on this and #2578? Both of these together enable our dynamic token use case

PR #2588, opened 2024-05-14
Update docs to use the new `metadata()` bloblang function

mihaitodor:
The `meta()` function is deprecated since v4.15.0, but it can still come in handy as a shortcut for `metadata("foo").string()` and `metadata().map_each(m -> m.string())` can be replaced with a call to `meta()`.


twmb:
I'm assuming this is a bit out of date... closing, also Ash is working on some v5 stuff that will likely redo some of this?

PR #2590, opened 2024-05-15
Add `max_retries` added to the `retry` processor

### LINKED ISSUE #2544: Add additional field "max_retries" for retry processor
In benthos version 4.27, a new "retry" processor was introduced. But, unfortunately, there is no opportunity in it (or did I miss it?) make a fixed number of retries, as is done in a similar output retry. Please add an additional field "max_retries": the maximum number of retries before giving up on the request (if set to zero there is no discrete limit)

mihaitodor:
Also add `retry_count` and `backoff_duration` metadata fields.

Fixes #2544.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@ Changelog
 
 All notable changes to this project will be documented in this file.
 
+## 4.28.0 - TBD
+
+### Added
+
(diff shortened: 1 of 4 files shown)


mihaitodor:
Rebased in https://github.com/redpanda-data/benthos/pull/40.

Issue #917, opened 2021-10-25
Pulsar Component - missing features


crtomirmajer:
Pulsar component is missing some important features:
- support for Token/OAuth2 Authentication, currently [none](https://github.com/Jeffail/benthos/blob/master/internal/impl/pulsar/input.go#L114-L118). I suggest adding the following:

```
pulsar:
    auth:
        oauth2:
          audience: "urn:sn:pulsar:x:y"
          issuer_url: "https://auth.xyz"
          private_key_file: /this/folder/key.json
        token:
          token: <base64-token-here> 
```

- configurable subscription type, currently hardcoded to [Shared](https://github.com/Jeffail/benthos/blob/master/internal/impl/pulsar/input.go#L125). My proposal is to add `subscription_type` option:
```
pulsar:
   subscription_type: <type-here>  # one of: shared, key_shared, exclusive, failover
```
- message attributes, such as `id`, `publishTime`, `redeliveryCount`, etc. (see [Message interface](https://github.com/apache/pulsar-client-go/blob/master/pulsar/message.go#L71-L123))

srenatus:
Another thing that's currently missing is exposing SubscriptionInitialPosition in [ConsumerOptions](https://pkg.go.dev/github.com/apache/pulsar-client-go@v0.12.1/pulsar#ConsumerOptions)
```
// SubscriptionInitialPosition is the initial position at which the cursor will be set when subscribe
// Default is `Latest`
SubscriptionInitialPosition
```

As far as I understand, it's the pulsar equivalent of setting the offset to start consuming from in Kafka.

Unless someone objects, I'll try to put a PR up this week 🤞

PR #2592, opened 2024-05-15
pulsar/input: add subscription_initial_position

srenatus:
For #917.

I wasn't sure how the integration tests worked... so I've not touched those. Tested with my local app, FWIW. 🤞


Jeffail (approved):
LGTM, thanks @srenatus!

PR #2593, opened 2024-05-16
Supress make warning present on Benthos forks

mkysel:
If you develop Benthos on a fork, you won't have any _git tags_ present by default.

The git describe throws an annoying error which shows for any `make` command
```
$ git describe --tags
fatal: No names found, cannot describe anything.
```

such as:
```
$ make
fatal: No names found, cannot describe anything.
make: Nothing to be done for `all'.
```

Supressing the error message makes the make output way more pleasant


Jeffail (approved):
Nice, thanks @mkysel!

PR #2597, opened 2024-05-17
Handle hash commands such as HGETALL in redis processor

### LINKED ISSUE #2282: Redis processor fails with HGETALL
Hi,
  I was trying to use redid processor with command HGETALL to retrieve all the fields and values inside an hashset but the processor return always nothing (it prints a blank line in the log)

This is a sample test:

```
input:
  generate:
    count: 1
    mapping: root = ""

pipeline:
  processors:
  - redis:
      url: redis://localhost:6379
      retries: 0
      command: hgetall
      args_mapping: root = ["demo:list:whitelist"]
  - log:
      message: ${! meta() }
  - catch:
    - log:
        message: Not Found
```

and output is:

```
INFO Running main config from specified file       @service=benthos benthos_version=4.24.0 path=whitelist.yaml
INFO {}                                            @service=benthos label="" path=root.pipeline.processors.1

```

From reds cli:

```
HGETALL demo:list:whitelist

1) "AA123BB"
2) "Car 1"
3) "CC456DD"
4) "Car 2"
5) "MPC7772"
6) "Volkswagen"
```

Using HGET command with key and field works as expected.

mkysel:
The crux of the reported issue was that https://github.com/benthosdev/benthos/blob/main/public/service/message.go#L262 requires an array of `[string]any` but the redis library returns `[any]any` which does not get processed right.

Once all the types are cast to string, it behaves as expected.

Added unit tests for HSet HGet and HGetAll.

All existing Redis unit tests pass.

Fixes #2282

### DIFF
--- a/internal/impl/redis/processor.go
+++ b/internal/impl/redis/processor.go
@@ -351,6 +351,19 @@ func (r *redisProc) execRaw(ctx context.Context, index int, inBatch service.Mess
 		return err
 	}
 
+	if structured, ok := res.(map[any]any); ok {
+		m2 := make(map[string]any, len(structured))
+
+		for key, value := range structured {
+			typeCast, ok := key.(string)
+			if !ok {
+				return fmt.Errorf("expected a string, got: %T", key)
+			}
+			m2[typeCast] = value
+		}
+		res = m2
+	}
+
 	msg.SetStructuredMut(res)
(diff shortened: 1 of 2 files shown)


Jeffail (approved):
Brilliant, thanks @mkysel!

PR #2602, opened 2024-05-22
pulsar/oauth2: fix config lookup, add "scope"

srenatus:
These are two things I had to adjust to be able to use Benthos with Pulsar and Keycloak.

I think the config lookup is a straight bug -- it can't work if you enable linting the config, can it? The linter should bail if you set "oauth" instead of "oauth2" and if you set "oauth2", the auth config builder will fail to pick up the values.


Jeffail (approved):
Looks great, thanks for fixing this @srenatus, and also thanks for updating the docs as well. We're hopefully adding the ability to auto generate the docs back in soon.

PR #2603, opened 2024-05-23
Point S3 input documentation to scanner instead of codec

KasonBraley:
The existing `codec` link pointed to something non-existent. It was replaced with the `scanner` component in 95a5055e25b6c4af187a9c86e4e836e0de676a3e.


JakeSCahill:
@KasonBraley I've carried your PR over to https://github.com/redpanda-data/rp-connect-docs/pull/9

We will be adding contribution docs here soon

PR #2619, opened 2024-05-30
Add flightsql SQL driver

mkysel:
This is a small step in the Apache Arrow / FlightSQL direction. It is the minimal working improvement that is both easy and useful.

FlightSQL Driver: https://pkg.go.dev/github.com/apache/arrow/go/v14/arrow/flight/flightsql/driver#section-readme

I tested it against the new InfluxDBv3 which is flightsql/arrow compatible.

Docs for Influx here: https://www.influxdata.com/blog/two-methods-connecting-influxdb-3-0/

Example config:
```
input:
  sql_select:
    driver: flightsql
    dsn: flightsql://us-east-1-1.aws.cloud2.influxdata.com:443?bucket-name=get-started&tls=enabled&token=[token]
    table: census
    columns: [ '*' ]
```

I haven't yet figured out variable binding, but here are some breadcrumbs https://github.com/apache/arrow/pull/40311


public/components/sql/package.go
```diff
@@ -10,6 +10,7 @@ import (
 
 	// Import all (supported) sql drivers.
 	_ "github.com/ClickHouse/clickhouse-go/v2"
+	_ "github.com/apache/arrow/go/v14/arrow/flight/flightsql/driver"
```
zeroshade, line 13: Why v14 instead of v16?

zeroshade:
There might be benefit for a wider scope and potential if you target [ADBC](https://pkg.go.dev/github.com/apache/arrow-adbc/go/adbc) and then just use the FlightSQL driver for it. 

That way it can be easily adapted to also allow using any of the ADBC drivers, such as snowflake, with minimal future changes.

Just an idea

public/components/sql/package.go
mkysel: bumped

mkysel:
One of the hiccups with this implementation is that not all FlightSQL endpoints support `variable/parameter binding`s. Those that do, use different dialects that either use `$` or `?`.

There are no integration tests since I tested this against `influxv3` which does not have a docker image yet. I suspect @zeroshade might have ideas on what the easiest integration test might be against.

zeroshade:
The open source Dremio docker image is likely the easiest integration test I can think of. Have a look at https://github.com/apache/arrow-adbc/blob/715fe2459a836527602383c50ecd40b8d520bb41/.github/workflows/integration.yml#L106 which is the CI workflow used for the flightsql integration testing for ADBC for examples of how to set it up.

PR #2628, opened 2024-06-02
Fix all occurrences of Redis port number, using 6379.

olgeni:
Remade it, because the first pull request got confused about life...


Jeffail (approved):
LGTM, thanks @olgeni!

PR #2630, opened 2024-06-03
Add free and enterprise import bundles

Jeffail:
This PR adds two new packages to `public/bundle`:

- `free`, which imports all component plugins that are licensed as Apache v2
- `enterprise`, which imports everything that is in `free` as well as all enterprise licensed component plugins

The intention here is to give users that are creating custom builds a clearer entrypoint with explicit licensing. There's more work to be done regarding licensing, where ideally we need clarity for every single public package, but this at least unblocks the largest use case which is people generally wanting everything in their build.

I believe we'll need to perform a follow up task once this is merged to tag and release and make both of these packages into explicit modules, at which point they should appear with clear licensing within gopkg.

### DIFF
--- a/public/bundle/enterprise/LICENSE
+++ b/public/bundle/enterprise/LICENSE
@@ -1,0 +1,408 @@
+**Redpanda Community License Agreement**
+
+Please read this Redpanda Community License Agreement (the “Agreement”)
+carefully before using Redpanda (as defined below), which is offered by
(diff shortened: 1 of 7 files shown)


twmb (approved):

srenatus:
I'm thankful for this! 👏 One small twist we could consider, too: have two packages, `enterprise` and `free`, and have the `enterprise` package import `free` + `X`, just so that `X` is on-the-nose obvious?

PR #2637, opened 2024-06-06
Enable exported rule for the revive linter

mihaitodor:
Would be nice to also enable [`unused-parameter`](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#unused-parameter) and probably some others as well. It will take a bit of work though...


Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #2638, opened 2024-06-06
Improve error message for invalid PubSub Attributes

AndreasBergmeier6176:
PubSub only accepts UTF-8 as attribute values.
Thus check validity instead of potentially provoking internal error in PubSub client.
Also use that opportunity to explicitly state, which attribute had the problematic data.


Jeffail (approved):
LGTM, thanks @AndreasBergmeier6176 and @mihaitodor

PR #2652, opened 2024-06-12
Add Splunk components

### LINKED ISSUE #1133: [Feature] Splunk Output Component
Hey 👋 ,

I really like your project. Do you think we could have the change to add an output component to send things inside Spunk?

I was looking at other project like `fluent-bit`.

https://docs.fluentbit.io/manual/pipeline/outputs/splunk

This could be a great addition imo 😄

mihaitodor:
This replaces the existing `splunk_hec` output with a Go implementation and it introduces a `splunk` input as well as integration tests.

Fixes #1133.

TODO:
- [x] Add some extra error handling.


internal/impl/splunk/input.go
```diff
@@ -0,0 +1,181 @@
...
+)
+
+//------------------------------------------------------------------------------
+
+func inputSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		// Stable().
+		Version("4.29.1").
+		Categories("Services").
+		Summary(`Publishes messages to Splunk.`).
```
Jeffail, line 33:
```suggestion
		Summary(`Consumes messages from Splunk.`).
```

internal/impl/splunk/input.go
```diff
@@ -0,0 +1,181 @@
...
+		Description(service.OutputPerformanceDocs(true, true)).
```
Jeffail, line 34:
```suggestion
```

internal/impl/splunk/input.go
```diff
@@ -0,0 +1,181 @@
...
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+
+	time.Sleep(5 * time.Second)
+
+	resp, err := i.client.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to execute http request: %s", err)
+	}
+
+	// TODO: HTTP ERROR HANDLING
```
Jeffail, line 128: More granular error handling is a nice-to-have but what we have here seems reasonable to me as a starting point.

internal/impl/splunk/input.go
```diff
@@ -0,0 +1,181 @@
...
+	payload.Set("output_mode", "json")
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, i.url, strings.NewReader(payload.Encode()))
+	if err != nil {
+		return fmt.Errorf("failed to construct http request: %s", err)
+	}
+	req.SetBasicAuth(i.user, i.password)
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+
+	time.Sleep(5 * time.Second)
```
Jeffail, line 121:
```suggestion
```

internal/impl/splunk/output.go
```diff
@@ -0,0 +1,203 @@
...
+	req.Header = header
+	req.ContentLength = int64(payload.Len())
+
+	resp, err := o.client.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to execute http request: %s", err)
+	}
+	defer resp.Body.Close()
+
+	// TODOOOOOOOOOOOOOOOOOOOOOOOOO ERROR HANDLING
```
Jeffail, line 198: As above, I think more granular error handling is great but this is fine for now.

Jeffail (commented):
Looking good, left a couple of comments and we also need to add the enterprise license headers to each of the source files.

internal/impl/splunk/input.go
mihaitodor: I'll run a few more tests and see if it's worth adding some checks to aid users.

mihaitodor: I ended up checking the status as well. OTEL is doing some extra handling [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/2a3c4933fb656aa38618935f62d197091f998107/exporter/splunkhecexporter/hec_worker.go#L65-L72) for status 429 or 502, but I'm not sure if that's relevant for Connect so I didn't replicate it for now.

internal/impl/splunk/input.go
```diff
@@ -0,0 +1,181 @@
...
+
+	if i.reader == nil {
+		return nil, nil, service.ErrNotConnected
+	}
+
+	line, err := i.reader.ReadBytes('\n')
+	if err != nil {
+		if err == io.EOF {
+			return nil, nil, service.ErrEndOfInput
+		}
```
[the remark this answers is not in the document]
mihaitodor: Good call! I think I only need to set the `reader` (`bufio.Reader`) to `nil` because only the underlying http `response.Body` can be closed.

Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #2653, opened 2024-06-12
Add Processor for Cloud Error Reporting

abergmeier:
Expose this as gcp_errorreporting.


rockwotj:
Should this be an output?

PR #2655, opened 2024-06-13
Fix CI docs check and do some cleanup

mihaitodor:
Currently, we don't detect if a new component got added without having its docs committed to the repo. With `git diff-index` we can only detect files which git is aware of. Some details here: https://stackoverflow.com/questions/2657935/checking-for-a-dirty-index-or-untracked-files-with-git

Also clean up govulncheck, since Dependabot seems to be good enough for this and golangci-lint doesn't want to add this linter to their suite.

TODO: In the future, detect if any docs need to be removed when / if components are deleted.


Jeffail (approved):
Nice, thanks @mihaitodor

mihaitodor:
OK, build passes now. Had to enable CGo for `make docs` 😁

PR #2656, opened 2024-06-13
Add deprecated fields to splunk output

Jeffail:
Adds the old deprecated splunk fields as dormant so that linters don't complain about old configs.

I've also updated the changelog, upgraded benthos and generated the latest docs ready for v4.30.0.


mihaitodor (approved):
:shipit:

PR #2665, opened 2024-06-18
Add `timestamp` field to the `kafka` and `kafka_franz` outputs.

### LINKED ISSUE #398: Kafka Output Timestamp
Currently using Kafka output to send messages for later processing with Kafka topic configuration set to CreateTime. Is it possible to extend the Kafka output to use an interpolation function when setting Kafka message time? 

Ideally, this would be a metadata key to avoid looking into payload again.

mihaitodor:
Fixes #398.

Note: In #398, the broker `CreateTime` mode is mentioned and I believe this is the mode that's being used by both the `kafka` and `kafka_franz` output. See the comment from Sarama here: https://github.com/IBM/sarama/blob/d2246cc9b9df113f3ffed7c442fd81e8b9cf736e/async_producer.go#L181-L192. Also, see this issue in franz-go: https://github.com/twmb/franz-go/issues/166.

### DIFF
--- a/internal/impl/kafka/integration_sarama_test.go
+++ b/internal/impl/kafka/integration_sarama_test.go
@@ -20,6 +20,8 @@ import (
 	"github.com/redpanda-data/connect/v4/internal/impl/kafka"
 )
 
+// TestIntegrationSaramaCheckpointOneLockUp checks that setting `checkpoint_limit: 1` on the `kafka` input doesn't lead to lockups.
+// Note: This test will take 10 minutes to complete unless you specify the `-timeout` flag explicitly. If you set `-timeout 0`, it will complete in a minute.
 func TestIntegrationSaramaCheckpointOneLockUp(t *testing.T) {
 	integration.CheckSkipExact(t)
 	t.Parallel()
@@ -35,7 +37,7 @@ func TestIntegrationSaramaCheckpointOneLockUp(t *testing.T) {
 	kafkaPortStr := strconv.Itoa(kafkaPort)
 
 	options := &dockertest.RunOptions{
-		Repository:   "docker.vectorized.io/vectorized/redpanda",
+		Repository:   "redpandadata/redpanda",
 		Tag:          "latest",
 		Hostname:     "redpanda",
 		ExposedPorts: []string{"9092"},
@@ -43,7 +45,11 @@ func TestIntegrationSaramaCheckpointOneLockUp(t *testing.T) {
 			"9092/tcp": {{HostIP: "", HostPort: kafkaPortStr}},
 		},
 		Cmd: []string{
-			"redpanda", "start", "--smp 1", "--overprovisioned", "",
+			"redpanda",
+			"start",
+			"--node-id 0",
+			"--mode dev-container",
+			"--set rpk.additional_start_flags=[--reactor-backend=epoll]",
 			"--kafka-addr 0.0.0.0:9092",
 			fmt.Sprintf("--advertise-kafka-addr localhost:%v", kafkaPort),
 		},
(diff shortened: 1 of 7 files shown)


internal/impl/kafka/integration_sarama_test.go
```diff
@@ -35,15 +37,19 @@ func TestIntegrationSaramaCheckpointOneLockUp(t *testing.T) {
...
-		Repository:   "docker.vectorized.io/vectorized/redpanda",
+		Repository:   "redpandadata/redpanda",
 		Tag:          "latest",
 		Hostname:     "redpanda",
 		ExposedPorts: []string{"9092"},
 		PortBindings: map[docker.Port][]docker.PortBinding{
 			"9092/tcp": {{HostIP: "", HostPort: kafkaPortStr}},
 		},
 		Cmd: []string{
-			"redpanda", "start", "--smp 1", "--overprovisioned", "",
```
mihaitodor, line 46: This doesn't seem to work any more, at least not via [Colima](https://github.com/abiosoft/colima) on an M3.

internal/impl/kafka/integration_sarama_test.go
```diff
@@ -35,15 +37,19 @@ func TestIntegrationSaramaCheckpointOneLockUp(t *testing.T) {
...
+			"redpanda",
+			"start",
+			"--node-id 0",
+			"--mode dev-container",
```
mihaitodor, line 51: Docs here: https://docs.redpanda.com/current/reference/rpk/rpk-redpanda/rpk-redpanda-mode/#development-mode Among other things, it sets `--overprovisioned` to `true`.

```
> redpanda start --mode help
...
Mode uses well-known configuration properties for development or tests
environments:
--mode dev-container
    Bundled flags:
        * --overprovisioned
        * --reserve-memory 0M
        * --check=false
        * --unsafe-bypass-fsync
    Bundled cluster properties:
        * auto_create_topics_enabled: true
        * group_topic_partitions: 3
        * storage_min_free_bytes: 10485760 (10MiB)
        * topic_partitions_per_shard: 1000
        * fetch_reads_debounce_timeout: 10
        * group_initial_rebalance_delay: 0
        * log_segment_size_min: 1
        * write_caching_default: true
```

internal/impl/kafka/integration_test.go
```diff
@@ -356,3 +364,76 @@ input:
 		integration.StreamTestOptVarSet("VAR1", ""),
 	)
 }
+
+func TestIntegrationKafkaOutputFixedTimestamp(t *testing.T) {
```
mihaitodor, line 368: It would be nice to maybe avoid adding an extra test. I was thinking to maybe introduce a new `StreamTestReadsMetadata()` which writes something to the output and then launches the input similar to `StreamTestOpenCloseIsolated()` and then checks if the metadata of the message read from the input contains the metadata keys and values which were passed via a (variadic) parameter.

internal/impl/kafka/integration_test.go
```diff
@@ -356,3 +364,76 @@ input:
...
+    timestamp: 666
+
+input:
+  kafka_franz:
+    seed_brokers: [ localhost:$PORT ]
+    topics: [ topic-$ID ]
+    consumer_group: "blobfish"
+  processors:
+    - mapping: |
+        root = if metadata("kafka_timestamp_unix") != 666 { "error: invalid timestamp" }
```
mihaitodor, line 425: This is a bit messy... Would you be open to updating `messageMatch()` [here](https://github.com/redpanda-data/benthos/blob/4edb4b29808fae57d97f166670e65248d02c1173/public/service/integration/stream_test_helpers.go#L554) to also check that for the received `message.Part` `ErrorGet()` returns a nil error?

Jeffail: Yeah I think that'd be fine, it's a niche that any existing tests would break as a result and it's not a big deal to have minor behavioural changes in the testing suite.

Jeffail (approved):
Looks great, thanks @mihaitodor, and extra ❤️ for fixing up the integration tests.

Play-AV:
Thanks for this <3

For redis streams > kafka
`meta timestamp = (this.redis_stream.split("-").index(0).int64()/1000).ts_unix()`
seemed to work

PR #2671, opened 2024-06-20
Add explicit license headers to each source file

Jeffail:
Since we're dual license now it makes sense to add explicit headers for each source file. I've tried to ease us into this by adding a script (first commit) that will automatically detect files missing a copyright header and adds it (all files Apache 2.0 by default).

We can eventually add a CI check similar to the stale docs detector where missing headers are flagged, but for now I don't want to block PRs on it as it's easy enough to do when I update the CHANGELOG for new releases.


mihaitodor (approved):
:shipit:

PR #2696, opened 2024-07-12
Add status events to topic writer

Jeffail:
This PR adds a new status event emitter to connect, which periodically sends protobuf messages to a configured topic that describes the health of the running instance. This can be used by external processes that consume the topic in order to build a model of how many instances are running a given pipeline identifier and whether they are managing it successfully.


proto/redpanda/api/connect/v1alpha1/status.proto
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// StatusEvent describes the current state of an individual connect instance,
+// which is self-reported periodically.
+message StatusEvent {
+  enum Type {
+    // The status has not been specified.
+    TYPE_UNSPECIFIED = 0;
+    // An instance has parsed a config and is now attempting to run a pipeline.
+    TYPE_INITIALISING = 1;
```
simon0191, line 26: Let's use american english spelling. It's the standard for our docs and the rest of our code.

https://github.com/redpanda-data/docs-site/blob/main/meta-docs/STYLE-GUIDE.adoc

proto/redpanda/api/connect/v1alpha1/status.proto
```diff
@@ -0,0 +1,42 @@
+syntax = "proto3";
+
+package redpanda.api.connect.v1alpha1;
+
+option go_package = "internal/protoconnect";
+
+// ConnectionError describes a specific connection failure.
+message ConnectionError {
+  string message = 1; // The error message.
+  repeated string path = 2; // The path of the connector in the config.
```
simon0191, line 10: why repeated?
if repeated, shouldn't it be `paths`?

cmd/redpanda-connect/main.go
```diff
@@ -39,9 +42,14 @@ func redpandaTopLevelConfigField() *service.ConfigField {
 }
 
 func main() {
-	rpLogger := enterprise.NewTopicLogger()
+	instanceID, err := uuid.NewV4()
```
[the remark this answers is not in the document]
Jeffail: Sounds good, done.

proto/redpanda/api/connect/v1alpha1/status.proto
Jeffail: Fixed, my bad

proto/redpanda/api/connect/v1alpha1/status.proto
Jeffail: Yeah good point, there's two ways of expressing a path, one would be a list of nested fields and the other is a string in some encoded path notation. So for example, the config:

```yaml
input:
  generate:
    count: 10
    interval: 1s
    mapping: 'root.name = fake("name")'
```

The path to the mapping could either be a list of field names `[]{"input", "generate", "mapping"}`, or it could be an encoded path, which in connect is normally a slight deviation from a JSON pointer (https://docs.redpanda.com/redpanda-connect/configuration/field_paths/) and would be `input.generate.mapping` in this case.

I chose to expose the path as a list because it means consumers do not need to understand the notation we're using (or how we do escapes), but what the list represents is a singular path made up of multiple elements.

So yes it's a repeated field and therefore describing it as a plurality makes sense, but it also isn't multiple paths so we'd need to call it something else.

Two options:
- A non-repeated field called `path` which is in our own path notation
- A repeated field called something else like `path_segments`

simon0191: I'd vote for non-repeated field called `path` which is in our own path notation.
It's simple and can be easily documented pointing to https://docs.redpanda.com/redpanda-connect/configuration/field_paths

internal/impl/kafka/enterprise/integration_test.go
```diff
@@ -0,0 +1,226 @@
...
+	t.Parallel()
+
+	pool, err := dockertest.NewPool("")
+	require.NoError(t, err)
+
+	kafkaPort, err := integration.GetFreePort()
+	require.NoError(t, err)
+
+	kafkaPortStr := strconv.Itoa(kafkaPort)
+	options := &dockertest.RunOptions{
```
simon0191, line 99: FYI, there's a redpanda testcontainer.
The interface is very simple.

```go
ctx := context.Background()

redpandaContainer, err := redpanda.Run(ctx,
    "docker.redpanda.com/redpandadata/redpanda:v23.3.3",
    redpanda.WithEnableSASL(),
    redpanda.WithEnableKafkaAuthorization(),
    redpanda.WithEnableWasmTransform(),
    redpanda.WithNewServiceAccount("superuser-1", "test"),
    redpanda.WithNewServiceAccount("superuser-2", "test"),
    redpanda.WithNewServiceAccount("no-superuser", "test"),
    redpanda.WithSuperusers("superuser-1", "superuser-2"),
    redpanda.WithEnableSchemaRegistryHTTPBasicAuth(),
)
if err != nil {
    log.Fatalf("failed to start container: %s", err)
}

// Clean up the container
defer func() {
    if err := redpandaContainer.Terminate(ctx); err != nil {
        log.Fatalf("failed to terminate container: %s", err)
    }
}()
```

https://golang.testcontainers.org/modules/redpanda/

simon0191 (approved):
LGTM

mihaitodor (approved):
LGTM 🐑 🚀

PR #2698, opened 2024-07-12
Implement Redpanda Data Transform Processor

rockwotj:
This adds a processor for Redpanda Data Transforms as a Redpanda connect
processor. The advantage of this processor over the existing wasm
processor is that Data Transforms has an existing SDK and developer
experience through `rpk transform init && rpk transform build`.

We implement the ABI contract for data transforms using Wazero, which
fairly closely follows the implementation within the broker. From an
implementation prespective we also have a single ABI contract no matter
which guest language is being used (Go/Rust/JavaScript/etc).

We support users specifying how to map kafka terms from their existing
messages, using similar methods to the kafka input/output plugins that
already exist within Redpanda.

For testing we check in an uppercase transform to generate a binary
using `rpk`.

Note that this PR does not implement the schema registry support within
data transforms. That will come in another PR.


internal/impl/redpanda/processor_data_transforms.go
```diff
@@ -0,0 +1,414 @@
...
+	if _, err = wasi_snapshot_preview1.Instantiate(ctx, r); err != nil {
+		return
+	}
+	cfg := wazero.NewModuleConfig().
+		WithStartFunctions().
+		WithArgs("transform").
+		WithName("transform").
+		WithEnv("REDPANDA_INPUT_TOPIC", "benthos")
+	for i := 0; i < 8; i += 1 {
+		cfg = cfg.WithEnv(fmt.Sprintf("REDPANDA_OUTPUT_TOPIC_%d", i), fmt.Sprintf("output_%d", i))
```
rockwotj, line 236: I just created dummy values here - I'm not sure if there should be anything special

internal/impl/redpanda/processor_data_transforms.go
```diff
@@ -0,0 +1,414 @@
...
+	// Notify the guest that it has data to process
+	r.guestChan <- nil
+	// Wait for the guest to process everything
+	select {
+	case <-r.hostChan:
+	case <-time.After(r.cfg.timeout):
+		_ = r.mod.Close(ctx)
+		drainChannel(r.hostChan)
+	}
+	return r.outputBatch, r.procErr
```
rockwotj, line 396: I noticed the other module called SetError on each individual message and never returned an error here - is that a better practice? I'm assuming this will stop the pipeline while the service.Message.SetError way does not?

internal/impl/redpanda/processor_data_transforms.go
```diff
@@ -0,0 +1,414 @@
...
+		_, err := start.Call(context.Background())
+		if !engine.mod.IsClosed() {
+			_ = engine.mod.Close(context.Background())
+		}
+		if err == nil {
+			err = sys.NewExitError(0)
+		}
+		engine.procErr = err
+		close(engine.hostChan)
+	}()
```
rockwotj, line 257: The engine running in the background is a necessary step with how our ABI is structured, but does make things a little more complicated in terms of a clean shutdown.

internal/impl/redpanda/processor_data_transforms.go
mihaitodor: I think it's covered in the docs: https://pkg.go.dev/github.com/redpanda-data/benthos/v4@v4.31.0/public/service#BatchProcessor

> // Process a batch of messages into one or more resulting batches, or return
> // an error if the entire batch could not be processed. If zero messages are
> // returned and the error is nil then all messages are filtered.
> //
> // The provided MessageBatch should NOT be modified, in order to return a
> // mutated batch a copy of the slice should be created instead.
> //
> // When an error is returned all of the input messages will continue down
> // the pipeline but will be marked with the error with *message.SetError,
> // and metrics and logs will be emitted.
> //
> // In order to add errors to individual messages of the batch for downstream
> // handling use *message.SetError(err) and return it in the resulting batch
> // with a nil error.
> //
> // The Message types returned MUST be derived from the provided messages,
> // and CANNOT be custom instantiations of Message. In order to copy the
> // provided messages use the Copy method.

Sometimes, I find it useful to set a breakpoint here and follow the call stack. There's some logic here: https://github.com/redpanda-data/benthos/blob/d940a3e6243d9c6e615fa57f99bc90f55d360c22/public/service/processor.go#L112-L115 and then here https://github.com/redpanda-data/benthos/blob/c209d989cf506fcb25c5ac0b0aa1ca8e90f54c09/internal/component/processor/auto_observed.go#L223 which shows what's going on. If you return an error, the returned batch is discarded in the `airGapBatchProcessor` and then the `AutoObservedBatched` processor will take the original input batch, mark all the messages in it with the received error and then return this batch. Not sure what you mean by stopping the pipeline, but you can filter out messages if you don't include them in the returned batch (and also return a nil error).

rockwotj: Ah thank you for the detailed explaination. I believe applying an error to the entire batch is the correct approach, the ABI works and checkpoints per batch.

internal/impl/redpanda/processor_data_transforms.go
```diff
@@ -0,0 +1,414 @@
...
+			Example(`${! metadata("kafka_timestamp_unix") }`).
+			Optional().
+			Advanced()).
+		Field(service.NewDurationField("timeout").
+			Description("The maximum period of time for a message to be processed").
+			Default("10s").
+			Advanced()).
+		Field(service.NewIntField("max_memory_pages").
+			Description("The maximum amount of wasm memory pages (64KiB) that an individual wasm module instance can use").
+			Default("1600"). // 100MiB
```
mihaitodor, line 65: Does 1600 correspond to 100MiB?

internal/impl/redpanda/processor_data_transforms.go
```diff
@@ -0,0 +1,414 @@
...
+type dataTransformEnginePool struct {
+	log           *service.Logger
+	wasmBinary    wazero.CompiledModule
+	runtimeConfig wazero.RuntimeConfig
+	modulePool    sync.Pool
+	cfg           dataTransformConfig
+}
+
+func newDataTransformProcessorFromConfig(conf *service.ParsedConfig, mgr *service.Resources) (*dataTransformEnginePool, error) {
+	pathStr, err := conf.FieldString("module")
```
mihaitodor, line 105: Small nit: Not all components have been cleaned up yet, but the new pattern is to have all these field names defined as consts similar to here: https://github.com/redpanda-data/connect/blob/968786d30ca4859bda3f3c8099096d9e86feb439/internal/impl/redis/input_streams.go#L31-L41
In this case, I guess the prefix should be `dtp` for "data transform processor".

internal/impl/redpanda/processor_data_transforms.go
```diff
@@ -0,0 +1,414 @@
+// Copyright 2024 Redpanda Data, Inc.
```
mihaitodor, line 1: Minor nit: I'd consider renaming this file to `processor_data_transform.go` and the tests to `processor_data_transform_test.go` since that's the name of the processor.

internal/impl/redpanda/processor_data_transforms.go
```diff
@@ -0,0 +1,414 @@
...
+
+func newDataTransformProcessor(wasmBinary []byte, cfg dataTransformConfig, mgr *service.Resources) (*dataTransformEnginePool, error) {
+	ctx := context.Background()
+	runtimeCfg := wazero.NewRuntimeConfig().
+		WithCloseOnContextDone(true).
+		WithCompilationCache(wazero.NewCompilationCache()).
+		WithMemoryLimitPages(uint32(cfg.maxMemoryPages))
+	r := wazero.NewRuntimeWithConfig(ctx, runtimeCfg)
+	cm, err := r.CompileModule(ctx, wasmBinary)
+	_ = r.Close(ctx)
```
mihaitodor, line 180: Why swallow the error here instead of returning it if it's non-nil? If this is the right thing to do, I'd consider adding a comment to explain why.

internal/impl/redpanda/processor_data_transforms.go
```diff
@@ -0,0 +1,414 @@
...
+		}
+		engine.procErr = err
+		close(engine.hostChan)
+	}()
+
+	// Wait for the engine to start
+	select {
+	case <-engine.hostChan:
+	case <-time.After(p.cfg.timeout):
+		_ = engine.mod.Close(ctx)
```
mihaitodor, line 263: Nitpick: If the goroutine is quick enough, `engine.mod.Close()` might run inside it, then get pre-empted while `<-time.After(p.cfg.timeout)` triggers and this line is executed. Could that lead to a panic?

internal/impl/redpanda/processor_data_transforms.go
```diff
@@ -0,0 +1,414 @@
...
+func (p *dataTransformEnginePool) ProcessBatch(ctx context.Context, batch service.MessageBatch) ([]service.MessageBatch, error) {
+	var modRunner *dataTransformEngine
+	var err error
+	if modRunnerPtr := p.modulePool.Get(); modRunnerPtr != nil {
+		modRunner = modRunnerPtr.(*dataTransformEngine)
+	} else {
+		if modRunner, err = p.newModule(); err != nil {
+			return nil, err
+		}
+	}
```
mihaitodor, line 278: I wonder if this could be simplified by setting the `New` func in `sync.Pool`: https://pkg.go.dev/sync#Pool

internal/impl/redpanda/processor_data_transforms.go
```diff
@@ -0,0 +1,414 @@
...
+	p.modulePool.Put(modRunner)
+	return []service.MessageBatch{res}, nil
+}
+
+func (p *dataTransformEnginePool) Close(ctx context.Context) error {
+	for {
+		mr := p.modulePool.Get()
+		if mr == nil {
+			return nil
+		}
```
mihaitodor, line 294: I could be completely off here, but I wonder if this could loop for a long time if `ProcessBatch()` keeps getting invoked. I'm not sure if that's a realistic scenario, though.

internal/impl/redpanda/functions.go
```diff
@@ -0,0 +1,156 @@
...
+
+	"github.com/tetratelabs/wazero/api"
+)
+
+const (
+	noActiveTransform = int32(-1)
+	invalidBuffer     = int32(-2)
+)
+
+var transformHostFunctions = map[string]func(r *dataTransformEngine) interface{}{}
```
mihaitodor, line 28: The codebase switched to using `any` instead of `interface{}` a while back. Not sure why the linter didn't catch this... I'll look into re-enabling it.

internal/impl/redpanda/functions.go
```diff
@@ -0,0 +1,156 @@
...
+		// Wait for new batch to be submitted for processing.
+		select {
+		case _, ok := <-r.guestChan:
+			if !ok {
+				return noActiveTransform
+			}
+		case <-ctx.Done():
+			return noActiveTransform
+		}
+		if !m.Memory().WriteUint32Le(recordCount, uint32(len(r.inputBatch))) {
```
mihaitodor, line 71: Could `m.Memory()` ever be nil? (same in the rest of the functions)

internal/impl/redpanda/processor_data_transforms_test.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/redpanda-data/benthos/v4/public/service"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func defaultConfig() dataTransformConfig {
+	var cfg dataTransformConfig
+	cfg.maxMemoryPages = 100
+	cfg.timeout = 10 * time.Second
```
mihaitodor, line 35: Nit: Would it be possible to reduce this timeout? The tests run in under a second on my machine.

mihaitodor (commented):
Thank you for all the work you put into this @rockwotj! I don't have much experience with wasm to confidently review the implementation. I left some non-critical comments which may be worth addressing and some questions on stuff that I didn't understand. I'll defer to @Jeffail for extra eyes on this.

internal/impl/wasm/build.sh
```diff
@@ -1,2 +1,3 @@
 #!/bin/sh
-tinygo build -scheduler=none -target=wasi -o uppercase.wasm ../../../public/wasm/examples/tinygo
+#TODO
+#tinygo build -scheduler=none -target=wasi -o uppercase.wasm ../../../public/wasm/examples/tinygo
```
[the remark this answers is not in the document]
rockwotj: It is actually! We don't do this in Redpanda because wasip1 requires too much memory, but there are no restrictions like that here. I will not switch the old wasm processor in this PR.

resources/testdata/redpanda_data_transforms/uppercase/transform.yaml
```diff
@@ -0,0 +1,4 @@
+name: uppercase
```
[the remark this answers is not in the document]
rockwotj: It's used by `rpk` to generate a wasm file (to know what language to support), but since we are going to switch to "BigGo" I can remove.

internal/impl/redpanda/build.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
rockwotj: I will switch to using go generate for this and switch to wasip1 and not check in the binary. I am happy to do other suggestions too 😄

internal/impl/redpanda/processor_data_transforms.go
rockwotj: Yes. 1600 pages and each page is 64KiB so it's 100MiB of memory. It's enough to be dangerous in Wasm but not too much to overwhelm the system (hopefully).

internal/impl/redpanda/processor_data_transforms.go
rockwotj: Ack

internal/impl/redpanda/processor_data_transforms.go
rockwotj: SGTM

internal/impl/redpanda/processor_data_transforms.go
rockwotj: sync.Pool#New does not allow for error reporting, which is why this pattern is here.

internal/impl/redpanda/processor_data_transforms.go
rockwotj: I would expect you stop calling `ProcessBatch` if you're closing. This is what the old wasm module does. We could use Read/Write locks to prevent calling ProcessBatch and Close at the same time (read locks so ProcessBatch can be called in parallel). Althorugh I am not sure if that's worth it.

internal/impl/redpanda/functions.go
rockwotj: Ah oops, this is copied from the old wasm module - will update!

internal/impl/redpanda/functions.go
rockwotj: Ah good catch - I will validate that in newDataTransformProcessor :)

internal/impl/redpanda/processor_data_transforms_test.go
rockwotj: Done

internal/impl/redpanda/processor_data_transforms.go
rockwotj: Done

internal/impl/redpanda/processor_data_transforms.go
rockwotj: Ended up handling the error properly

internal/impl/redpanda/processor_data_transforms.go
rockwotj: No multiple calls to Close is safe: https://github.com/tetratelabs/wazero/blob/9a19a04962881134301ddedf7dc84c50004dbedf/runtime.go#L367

Is there another panic you're referring to?

internal/impl/redpanda/processor_data_transforms.go
rockwotj: Done!

internal/impl/redpanda/processor_data_transforms.go
mihaitodor: Fair enough. I haven't looked in detail at the original `wasm` processor. Thanks for the details!

internal/impl/redpanda/processor_data_transforms.go
mihaitodor: Good point, thank you for the explanation!

internal/impl/redpanda/processor_data_transforms.go
mihaitodor: Nope, that should be fine. Just thought that a double close might panic. Thanks for looking into it!

internal/impl/redpanda/processor_data_transform.go
```diff
@@ -0,0 +1,441 @@
...
+	dtpFieldModulePath     = "module_path"
+	dtpFieldInputKey       = "input_key"
+	dtpFieldOutputKey      = "output_key"
+	dtpFieldInputHeaders   = "input_headers"
+	dtpFieldOutputMetadata = "output_metadata"
+	dtpFieldTimestamp      = "timestamp"
+	dtpFieldTimeout        = "timeout"
+	dtpFieldMaxMemoryPages = "max_memory_pages"
+	wasmPageSize           = 64 * humanize.KiByte
+	dtpDefaultMaxMemory    = 100 * humanize.MiByte
```
mihaitodor, line 45: Nice 🥇

internal/impl/redpanda/processor_data_transform_test.go
```diff
@@ -0,0 +1,183 @@
...
+	var cfg dataTransformConfig
+	cfg.maxMemoryPages = 1000
+	cfg.timeout = time.Second
+	return cfg
+}
+
+func TestDataTransformProcessorSerial(t *testing.T) {
+	wasm, err := os.ReadFile("./uppercase.wasm")
+	if os.IsNotExist(err) {
+		t.Skip("skipping as wasm example not compiled, run build.sh to remedy")
```
mihaitodor, line 44: I guess I'd rather fail the tests if this file is missing instead of a silent skip now that we can `go generate` it.

mihaitodor (approved):
Thanks a lot Tyler! I think that covers all the stuff I was able to spot. I'll defer to Ash for any extra feedback.

PS: I left one other small nit on the tests.

internal/impl/redpanda/processor_data_transform_test.go
rockwotj: I don't feel strongly. Ash do you have opinions? Happy to change them to fail. I'll update the message here either way 😄

Jeffail: I've refactored the tests slightly so that we actually build the WASM on the fly. The advantage is that we don't end up with the file sitting stagnant in the repo after the tests run, which can be annoying for some repo activities when it's not automatically cleaned up. It has the disadvantage of slowing the tests down by about a second, but unless this dramatically changes CI test times (I doubt it'll change it) then I'm happy as is.

PR #2702, opened 2024-07-16
introduce ollama processor

rockwotj:
The ollama_completion and ollama_embeddings processor can complete text
and generate embeddings respectively. The use ollama.com server to
perform these operations and with the x_benthos_extra flags will also
embed a server and launch it locally.

I tried to embed the ollama server library directly instead of embedding
the entire server, but it's not quite setup for that at the moment (go
generate is needed in the dependency and I couldn't find a good way to
do that).

Otherwise it's a fairly standard barebones processor, there are more
options for requests and models that can be exposed but I figured that I
would start with the MVP and iterate from there.


internal/impl/ollama/embed_ollama_linux_arm64.go
```diff
@@ -0,0 +1,18 @@
+// Copyright 2024 Redpanda Data, Inc.
+//
+// Licensed as a Redpanda Enterprise file under the Redpanda Community
+// License (the "License"); you may not use this file except in compliance with
+// the License. You may obtain a copy of the License at
+//
+// https://github.com/redpanda-data/connect/blob/main/licenses/rcl.md
+
+//go:build x_benthos_extra
```
rockwotj, line 9: I re-used this existing extra, but I am happy to either remove the embedded server feature, or use a different tag

internal/impl/ollama/embeddings_processor_ollama.go
```diff
@@ -0,0 +1,116 @@
...
+		p, err := o.computeText(msg)
+		if err != nil {
+			return nil, err
+		}
+		e, err := o.generateEmbedding(ctx, p)
+		if err != nil {
+			return nil, err
+		}
+		m := service.NewMessage(nil)
+		m.SetStructuredMut(e)
```
mihaitodor, line 84: This is a bit annoying, but if you set `[]float64` in here, Bloblang won't be able to operate on it. You'll have to convert it to `[]any`. See docs here: https://pkg.go.dev/github.com/redpanda-data/benthos/v4@v4.32.0/public/service#Message.SetStructured

internal/impl/ollama/completion_processor_ollama.go
```diff
@@ -0,0 +1,117 @@
...
+	return o.baseOllamaProcessor.ProcessBatch(ctx, batch, func(ctx context.Context, msg *service.Message) (*service.Message, error) {
+		p, err := o.computePrompt(msg)
+		if err != nil {
+			return nil, err
+		}
+		g, err := o.generateCompletion(ctx, p)
+		if err != nil {
+			return nil, err
+		}
+		return service.NewMessage([]byte(g)), nil
```
mihaitodor, line 83: I'm not entirely sure what's the right thing to do here, but my impression is that if you just return a `service.NewMessage` here, you'll loose any metadata that's set on the origin message and I think most processors don't do that (should look that up, I'm not 100% sure).

mihaitodor (commented):
Thanks Tyler! I spotted a few minor issues, but it looks OK otherwise.

internal/impl/ollama/base_processor.go
```diff
@@ -0,0 +1,189 @@
...
+		serverPath = path.Join(os.TempDir(), "ollama")
+		if err = extractEmbeddedServer(fs, serverPath); err != nil {
+			return nil, err
+		}
+	} else if err != nil {
+		return nil, err
+	}
+	proc := exec.Command(serverPath, "serve")
+	// proc.Stdout = os.Stdout
+	// proc.Stderr = os.Stderr
```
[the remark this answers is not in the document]
rockwotj: If you have ideas on how to hook it up I'm all for doing it - it's actually pretty helpful for troubleshooting.

internal/impl/ollama/base_processor.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (o *baseOllamaProcessor) waitForServer(ctx context.Context) error {
+	timeout := time.After(5 * time.Second)
+	tick := time.NewTicker(500 * time.Millisecond)
+	for {
+		select {
+		case <-timeout:
+			return errors.New("timed out waiting for server to start")
+		case <-tick.C:
+			if err := o.client.Heartbeat(ctx); err == nil {
```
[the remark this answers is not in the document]
rockwotj: Good question but no - it's an http.Client under the hood and still subject to network timeouts.

internal/impl/ollama/embeddings_processor_ollama.go
rockwotj: Thank you! 🥇 I just ran into this today with the pinecone output and it was driving me bonkers why I couldn't get the length of this vector

internal/impl/ollama/base_processor.go
```diff
@@ -0,0 +1,189 @@
...
+			// create the client
+			_ = p.Close(context.Background())
+			return nil, err
+		}
+		p.client = c
+	}
+	if err := p.waitForServer(context.Background()); err != nil {
+		// We failed to start, return the real error not any additional errors if
+		// we fail to close, which would distract from the real reason startup failed.
+		_ = p.Close(context.Background())
```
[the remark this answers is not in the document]
rockwotj: Done

.github/workflows/test.yml
```diff
@@ -26,7 +26,9 @@ jobs:
         check-latest: true
 
     - name: Install dependencies for x_benthos_extra
-      run: sudo apt install -y --no-install-recommends libzmq3-dev
+      run: |
+        sudo apt install -y --no-install-recommends libzmq3-dev
+        go generate -tags=x_benthos_extra ./internal/impl/ollama
```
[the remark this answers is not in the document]
rockwotj: Yes done.

internal/impl/ollama/completion_processor_ollama.go
rockwotj: Thanks I looks at a couple other processors (redis, awk) and it seems they preserve metadata as well. I'm good with that - done!

On a side note - should the wasm processor be doing that too? Maybe not because the user is uploading code there and there are mappings for metadata?

.github/workflows/test.yml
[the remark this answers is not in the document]
rockwotj: I undid this change, redownloading this every `make docs` seems broken, so I've made it more explicit. Maybe there is makefile magic that can prevent running this if the file is already there? Honestly I'm pretty terrible at raw make. Let me know if there a strong preference here!

rockwotj:
FYI you can hold off on reviewing this one, after the batching comments from @mihaitodor I am going to make this a non batch processor and let users specify the batching and concurrently themselves.

rockwotj:
Okay, I've simplified it and we're ready for another looksee here. Also confirmed that we're going to start this one as enterprise for now.

internal/impl/ollama/base_processor.go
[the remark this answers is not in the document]
mihaitodor: Both `proc.Stderr` and `proc.Stdout` are of type [`io.Writer`](https://pkg.go.dev/io#Writer). I'd create a struct which holds a reference to the logger and implements this interface (it just needs a `Write(p []byte) (n int, err error)` method which does a write to the logger member variable from the receiver if that makes sense). Let me know if you need an example.

internal/impl/ollama/completion_processor_ollama.go
mihaitodor: It's a tricky question, but if you look [here](https://github.com/redpanda-data/connect/blob/96f6b395e5e2cee71c0bced15ffeb1a575e015ab/internal/impl/wasm/functions.go#L74) and [here](https://github.com/redpanda-data/connect/blob/96f6b395e5e2cee71c0bced15ffeb1a575e015ab/internal/impl/wasm/functions.go#L97), you'll see utilities for reading and setting the metadata from the wasm processor. Users could run a wasm model which overwrites certain metadata fields in the current message, but they might want to preserve the rest. I can imagine that being relevant when processing Kafka messages for example and you may want at the output level to send them to different topics based on the `kafka_topic` metadata field. It would be annoying if the processor pipeline erased the metadata (although you can work around it in various ways, for example by using a `branch`).

mihaitodor (approved):
Thanks Tyler! I had another look and left some clarifications. Otherwise, it should be OK. I haven't tried to run the code myself, but let me know if you need any help with it.

internal/impl/ollama/base_processor.go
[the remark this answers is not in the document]
mihaitodor: Indeed, it might be more complicated. My initial thinking was that wazero is smart enough to do a single write / log entry, so, even if it contains new lines, we could emit those as-is and leave these as debug logs or, we could also escape the newline characters so we don't end up with multiline logs (I think there's some other stuff which emits multiline logs in debug mode).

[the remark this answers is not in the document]
rockwotj: Took a stab at this 😄

internal/impl/ollama/base_processor.go
```diff
@@ -0,0 +1,182 @@
...
+	return len(b), nil
+}
+
+func startServer(logger *service.Logger, fs *service.FS) (*exec.Cmd, error) {
+	serverPath, err := exec.LookPath("ollama")
+	if errors.Is(err, exec.ErrNotFound) {
+		if embeddedServer == nil {
+			return nil, errors.New("embedded ollama not supported for this platform - please specify a server_address")
+		}
+		serverPath = path.Join(os.TempDir(), "ollama")
```
mihaitodor, line 83: Small nit: I think we might want this path to be configurable in case people deal with readonly file systems.

internal/impl/ollama/base_processor.go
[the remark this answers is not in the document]
mihaitodor: Awesome, thanks! That looks fine as long as there aren't any concurrent calls to `Write`. Might be worth protecting it with a mutex just in case.

.github/workflows/test.yml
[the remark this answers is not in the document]
rockwotj: I cobbled something together - PTAL I'm not good with Makefiles 😄

internal/impl/ollama/base_processor.go
rockwotj: Good call - done!

.github/workflows/test.yml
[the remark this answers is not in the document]
mihaitodor: Nice 🥇 That's what I was expecting to work. Thanks for digging into it! I think it's probably not that relevant to have both `ollama-linux-amd64` and `ollama-linux-arm64` or they could be placed in a folder called `ollama` or `ollama-resources` or smth like that so then you won't really need that `if` statement in the Makefile because you can check for the generic name. Guess you can remove the `go generate` stuff from the CGo Dockerfile.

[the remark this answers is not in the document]
rockwotj: Just named them both `ollama` that cleaned things up thanks!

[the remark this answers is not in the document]
mihaitodor: You're welcome! I think you can actually use `./internal/impl/ollama/ollama` as the dependency of the target and then you don't even need the if inside, but that's just a nitpick :)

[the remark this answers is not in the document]
rockwotj: Hrm but it's not working in CI as it can't find the file to embed - it seems to work locally... Any tips?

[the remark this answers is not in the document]
mihaitodor: So this is what I had in mind:
```
./internal/impl/ollama/ollama:
	go generate -tags "$(TAGS)" ./internal/impl/ollama

generate: ./internal/impl/ollama/ollama
```
And then you'd need to make sure `generate` is called first for the `$(APPS)` target like so:

```
$(APPS): %: generate $(PATHINSTBIN)/%
```

That should do the trick I think. If you get stuck let me know and I'll send you the full code.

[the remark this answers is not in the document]
rockwotj: Really appreciate that! That seems to work for me locally, let's see how CI feels about it.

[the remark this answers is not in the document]
rockwotj: Seems to work - thank you! 🙏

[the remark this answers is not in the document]
mihaitodor: You're welcome!

mihaitodor (approved):
Thanks again Tyler! I don't see any other issues, but I'll let @Jeffail have a look through it as well.

docs/modules/components/pages/processors/ollama_chat.adoc
```diff
@@ -0,0 +1,123 @@
...
+     And:
+
+     https://github.com/redpanda-data/connect/tree/main/cmd/tools/docs_gen/templates/plugin.adoc.tmpl
+////
+
+
+component_type_dropdown::[]
+
+
+Processor that uses the Ollama API to generate text.
```
asimms41, line 24:
```suggestion
Sends prompts to your chosen Ollama large language model (LLM) and generates text from the responses, using the Ollama API.
```
I wondered if we could give the user a more detailed description of what the processor does. This is just a suggestion.

docs/modules/components/pages/processors/ollama_chat.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+Introduced in version 4.31.0.
```
asimms41, line 26:
```suggestion
Introduced in version 4.32.0.
```

docs/modules/components/pages/processors/ollama_chat.adoc
```diff
@@ -0,0 +1,123 @@
...
+model: gemma2
+
+model: qwen2
+
+model: phi3
+```
+
+=== `prompt`
+
+The prompt to submit to the model. By default the entire payload as a string is submitted.
```
asimms41, line 107:
```suggestion
The prompt you want to generate a response for. By default, the processor submits the entire payload as a string. This field supports [interpolation functions](http://localhost:8000/configuration/interpolation/#bloblang-queries).
```
Just a suggestion to make this description distinct from the system prompt?

docs/modules/components/pages/processors/ollama_embeddings.adoc
```diff
@@ -0,0 +1,113 @@
...
+     And:
+
+     https://github.com/redpanda-data/connect/tree/main/cmd/tools/docs_gen/templates/plugin.adoc.tmpl
+////
+
+
+component_type_dropdown::[]
+
+
+Processor that uses the Ollama API to create vector embeddings.
```
asimms41, line 24:
```suggestion
Sends text to your chosen Ollama large language model (LLM) and creates vector embeddings, using the Ollama API. Vector embeddings are long arrays of numbers that represent values or objects, in this case text. 
```
I've suggested a more detailed description using a similar format to the chat processor.

docs/modules/components/pages/processors/ollama_embeddings.adoc
```diff
@@ -0,0 +1,113 @@
...
+
+
+[tabs]
+======
+Common::
++
+--
+
+```yml
+# Common config fields, showing default values
```
asimms41, line 36:
```suggestion
# Common configuration fields, showing default values
```
Brings into line with the style guide.

docs/modules/components/pages/processors/ollama_embeddings.adoc
```diff
@@ -0,0 +1,113 @@
...
+label: ""
+ollama_embeddings:
+  server_address: "" # No default (optional)
```
asimms41, line 39: Isn't the local server address the default server address?

docs/modules/components/pages/processors/ollama_embeddings.adoc
```diff
@@ -0,0 +1,113 @@
...
+  text: "" # No default (optional)
+```
+
+--
+Advanced::
++
+--
+
+```yml
+# All config fields, showing default values
```
asimms41, line 50:
```suggestion
# All configuration fields, showing default values
```
Brings into line with the style guide.

docs/modules/components/pages/processors/ollama_embeddings.adoc
```diff
@@ -0,0 +1,113 @@
...
+  server_address: "" # No default (optional)
+  ollama_directory: "" # No default (optional)
+  model: nomic-embed-text # No default (required)
+  text: "" # No default (optional)
+```
+
+--
+======
+
+Processor that uses the Ollama API to create vector embeddings.
```
asimms41, line 62: Do we need this line?

docs/modules/components/pages/processors/ollama_embeddings.adoc
```diff
@@ -0,0 +1,113 @@
...
+
+See https://ollama.com/[https://ollama.com/^] for more information.
```
asimms41, line 64:
```suggestion
For more information, see the https://ollama.com/[Ollama website^].
```
Just to bring into line with the style guide.

docs/modules/components/pages/processors/ollama_embeddings.adoc
```diff
@@ -0,0 +1,113 @@
...
+
+== Fields
+
+=== `server_address`
+
+The address of a remote ollama server to use. If omitted a local ollama server will be started and ran.
```
asimms41, line 70:
```suggestion
The address of the Ollama server to use. By default, a local Ollama server starts and runs unless you specify the address of a remote server.
```

docs/modules/components/pages/processors/ollama_embeddings.adoc
```diff
@@ -0,0 +1,113 @@
...
+
+
+*Type*: `string`
+
+
+=== `ollama_directory`
+
+The directory to unpack the ollama binary into. Defaults to /tmp
```
asimms41, line 78:
```suggestion
Unpack the Ollama binary to this directory. Default is `/tmp`.

```

docs/modules/components/pages/processors/ollama_embeddings.adoc
```diff
@@ -0,0 +1,113 @@
...
+
+
+*Type*: `string`
+
+
+=== `model`
+
+The ollama model to use. A full list is available at ollama.com/models
```
asimms41, line 86:
```suggestion
The name of the Ollama LLM to use. For a full list of models, see the https://ollama.com/models[Ollama website].
```

docs/modules/components/pages/processors/ollama_embeddings.adoc
```diff
@@ -0,0 +1,113 @@
...
+model: mxbai-embed-large
+
+model: snowflake-artic-embed
+
+model: all-minilm
+```
+
+=== `text`
+
+The text to create embeddings for. By default the entire payload as a string is submitted.
```
asimms41, line 106: The text you want to create vector embeddings for. By default, the processor submits the entire payload as a string.

asimms41 (changes requested):
Sorry, I've just realised that I've added my comments to the .adoc versions of the processors. I'll add them to the Go files for the next set of reviews.

docs/modules/components/pages/processors/ollama_chat.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+
+[tabs]
+======
+Common::
++
+--
+
+```yml
+# Common config fields, showing default values
```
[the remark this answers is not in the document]
rockwotj: This is common code in another place of the codebase, I can open another PR to update.

docs/modules/components/pages/processors/ollama_chat.adoc
```diff
@@ -0,0 +1,123 @@
...
+label: ""
+ollama_chat:
+  server_address: "" # No default (optional)
```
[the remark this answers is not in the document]
rockwotj: Writing `http://127.0.0.1:11434` is different because then the processor will not start a locally running ollama instance but use an already running server. Any ideas on how to make that clearer?

docs/modules/components/pages/processors/ollama_chat.adoc
```diff
@@ -0,0 +1,123 @@
...
+  prompt: "" # No default (optional)
+```
+
+--
+Advanced::
++
+--
+
+```yml
+# All config fields, showing default values
```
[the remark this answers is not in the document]
rockwotj: I will open another PR to fix

docs/modules/components/pages/processors/ollama_chat.adoc
```diff
@@ -0,0 +1,123 @@
...
+  prompt: "" # No default (optional)
+  system_prompt: "" # No default (optional)
+```
+
+--
+======
+
+Processor that uses the Ollama API to generate text.
+
+See https://ollama.com/[https://ollama.com/^] for more information.
```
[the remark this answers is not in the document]
rockwotj: Thank you!

docs/modules/components/pages/processors/ollama_chat.adoc
```diff
@@ -0,0 +1,123 @@
...
+  ollama_directory: "" # No default (optional)
+  model: llama3 # No default (required)
+  prompt: "" # No default (optional)
+  system_prompt: "" # No default (optional)
+```
+
+--
+======
+
+Processor that uses the Ollama API to generate text.
```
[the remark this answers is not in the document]
rockwotj: I think it will be better with another one of your suggestions

docs/modules/components/pages/processors/ollama_chat.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+See https://ollama.com/[https://ollama.com/^] for more information.
+
+== Fields
+
+=== `server_address`
+
+The address of a remote ollama server to use. If omitted a local ollama server will be started and ran.
```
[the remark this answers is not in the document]
rockwotj: Thanks, done!

docs/modules/components/pages/processors/ollama_chat.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+
+*Type*: `string`
+
+
+=== `ollama_directory`
+
+The directory to unpack the ollama binary into. Defaults to /tmp
```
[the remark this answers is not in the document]
rockwotj: Done, thanks!

docs/modules/components/pages/processors/ollama_chat.adoc
rockwotj: Thanks, much better!

rockwotj:
golangci-lint failed due to out of space in GHA 😞

docs/modules/components/pages/processors/ollama_chat.adoc
rockwotj: > Also, both processors in the PR need adding to the nav.adoc to show up in the left-hand navigation.

Where is nav.adoc? I don't know where that is

asimms41: I'll need to do this in the other repo but it's just a reminder because needs to go out at the same time. This caught me out on Friday.

docs/modules/components/pages/processors/ollama_chat.adoc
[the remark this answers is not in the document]
asimms41: You could say this?

The address of the Ollama server to use. The processor starts and runs a local Ollama server unless you specify the address of a local or remote server.

docs/modules/components/pages/processors/ollama_chat.adoc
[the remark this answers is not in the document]
asimms41: Is there any more details we need to give, otherwise this will just be a duplication of the short description at the top of the page?

internal/impl/ollama/chat_processor.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		panic(err)
+	}
+}
+
+func ollamaChatProcessorConfig() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Categories("AI").
+		Summary("Processor that uses the Ollama API to generate text.").
+		Description(`Sends prompts to your chosen Ollama large language model (LLM) and generates text from the responses, using the Ollama API.
```
asimms41, line 41:
```suggestion
		Description(`This processor ends prompts to your chosen Ollama large language model (LLM) and generates text from the responses, using the Ollama API.
```

internal/impl/ollama/chat_processor.go
```diff
@@ -0,0 +1,159 @@
...
+
+For more information, see the https://ollama.com/[Ollama website^].`).
+		Version("4.32.0").
+		Fields(
+			service.NewStringField(bopFieldServerAddress).
+				Description("The address of the Ollama server to use. By default, a local Ollama server starts and runs unless you specify the address of a remote server.").
```
asimms41, line 47:
```suggestion
				Description("The address of the Ollama server to use. By default, a local Ollama server starts and runs unless you specify the address of a local or remote server.").
```

internal/impl/ollama/chat_processor.go
```diff
@@ -0,0 +1,159 @@
...
+				Optional(),
+			service.NewStringField(bopFieldOllamaDir).
+				Description("Unpack the Ollama binary to this directory. Default is `"+os.TempDir()+"`.").
+				Advanced().
+				Optional(),
+			service.NewStringField(bopFieldModel).
+				Description("The name of the Ollama LLM to use. For a full list of models, see the https://ollama.com/models[Ollama website].").
+				Examples("llama3", "gemma2", "qwen2", "phi3"),
+			service.NewInterpolatedStringField(ocpFieldUserPrompt).
+				Description("The prompt you want to generate a response for. By default, the processor submits the entire payload as a string. This field supports xref:configuration:interpolation.adoc#bloblang-queries[interpolation functions].").
```
asimms41, line 58: The sentence 'This field supports....' is duplicated in the docs.

internal/impl/ollama/embeddings_processor.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		panic(err)
+	}
+}
+
+func ollamaEmbeddingProcessorConfig() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Categories("AI").
+		Summary("Processor that uses the Ollama API to create vector embeddings.").
+		Description(`Sends text to your chosen Ollama large language model (LLM) and creates vector embeddings, using the Ollama API. Vector embeddings are long arrays of numbers that represent values or objects, in this case text.
```
asimms41, line 40:
```suggestion
		Description(`This processor sends text to your chosen Ollama large language model (LLM) and creates vector embeddings, using the Ollama API. Vector embeddings are long arrays of numbers that represent values or objects, in this case text. 
```

internal/impl/ollama/embeddings_processor.go
```diff
@@ -0,0 +1,131 @@
...
+		Categories("AI").
+		Summary("Processor that uses the Ollama API to create vector embeddings.").
+		Description(`Sends text to your chosen Ollama large language model (LLM) and creates vector embeddings, using the Ollama API. Vector embeddings are long arrays of numbers that represent values or objects, in this case text. 
+
+
+See https://ollama.com/[https://ollama.com/^] for more information.`).
+		Version("4.32.0").
+		Fields(
+			service.NewStringField(bopFieldServerAddress).
+				Description("The address of the Ollama server to use. By default, a local Ollama server starts and runs unless you specify the address of a remote server.").
```
asimms41, line 47:
```suggestion
				Description("The address of the Ollama server to use. By default, a local Ollama server starts and runs unless you specify the address of a local or remote server.").
```

internal/impl/ollama/embeddings_processor.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func ollamaEmbeddingProcessorConfig() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Categories("AI").
+		Summary("Processor that uses the Ollama API to create vector embeddings.").
+		Description(`Sends text to your chosen Ollama large language model (LLM) and creates vector embeddings, using the Ollama API. Vector embeddings are long arrays of numbers that represent values or objects, in this case text. 
+
+
+See https://ollama.com/[https://ollama.com/^] for more information.`).
```
asimms41, line 43:
```suggestion
For more information, see the https://ollama.com/[Ollama website^]`).
```

asimms41 (changes requested):
Just minor changes.

internal/impl/ollama/chat_processor.go
rockwotj: Fixed! Thanks

docs/modules/components/pages/processors/ollama_chat.adoc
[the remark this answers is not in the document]
rockwotj: Done!

docs/modules/components/pages/processors/ollama_chat.adoc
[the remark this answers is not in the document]
rockwotj: I added more details

Jeffail (approved):
LGTM, thanks @rockwotj.

PR #2703, opened 2024-07-17
Add cloud and community build artifacts

Jeffail:
This PR adds two new builds of Redpanda Connect, which will be added to our release artifacts:

1. Cloud edition, which has enterprise features and restricts the suite of connectors available to only those we're explicitly allowing within our cloud product.
2. Community edition, which doesn't have any enterprise features but includes all other connectors.


mihaitodor (approved):
LGTM :shipit:

PR #2706, opened 2024-07-17
Pinecone output

rockwotj:
Support an output to pinecone vector database. We support three
different kind of operations: upsert (default), update, delete.

There are mappings for metadata and vectors in Pinecone, as well as
interpolations for ID and namespaces.

The tests here are rather slim as there isn't a ton of logic here, it's
just hooking stuff up, and there is no way to write an isolated
integration test (pinecone is only a hosted service).


internal/impl/pinecone/output.go
```diff
@@ -0,0 +1,369 @@
...
+	default:
+		return nil, fmt.Errorf("invalid operation: %s", rawOp)
+	}
+	host, err := conf.FieldString(poFieldHost)
+	if err != nil {
+		return nil, err
+	}
+	if strings.HasPrefix(host, "https://") {
+		return nil, fmt.Errorf("host field must be a FQDN not a URL: %q (remove the https:// prefix)", host)
+	}
```
rockwotj, line 146: I ran into this and I'm assuming others might. Pinecone's UI is a little confusing in this aspect, and the SDK just hangs if the https is added

Screenshot 2024-07-17 at 3 40 49 PM

internal/impl/pinecone/output.go
```diff
@@ -0,0 +1,369 @@
...
+			service.NewOutputMaxInFlightField(),
+			service.NewBatchPolicyField(poFieldBatching),
+			service.NewStringField(poFieldHost).
+				Description("The host for the Pinecone index."),
+			service.NewStringField(poFieldAPIKey).
+				Secret().
+				Description("The Pinecone api key."),
+			service.NewStringEnumField(poFieldOp, string(operationUpdate), string(operationUpsert), string(operationDelete)).
+				Default(string(operationUpsert)).
+				Description("The operation to perform against the Pinecone index."),
```
rockwotj, line 55: This follows a similar pattern that the mongodb output does.

internal/impl/pinecone/output.go
```diff
@@ -0,0 +1,369 @@
...
+			service.NewInterpolatedStringField(poFieldNamespace).
+				Default("").
+				Advanced().
+				Description("The namespace to write to - writes to the default namespace by default."),
+			service.NewInterpolatedStringField(poFieldID).
+				Description("The ID for the index entry in Pinecone."),
+			service.NewBloblangField(poFieldVectorMapping).
+				Optional().
+				Description("The mapping to extract out the vector from the document. The result must be a floating point array. Required if not a delete operation."),
```
rockwotj, line 64: I don't think we have anywhere where we need a single type as the output. Happy to take suggestions for other ways to let folks map this.

Jeffail: Yeah I think this is probably the best overall UX. Ideally we'd add a few examples here to the field of mappings like might make sense. Something like `root = this.foos.map_each(foo -> foo.blah)`, but we can put together some typical examples later.

Jeffail (approved):
Awesome, thanks @rockwotj!

PR #2716, opened 2024-07-19
Fix pinecone sink

rockwotj:
Two issues I found:

1. We need to close connections on errors
2. Bloblang needs []any not primative arrays (sigh!)

Fix them and add a test for the mappings.


Jeffail (approved):
LGTM, thanks @rockwotj

PR #2724, opened 2024-07-22
Fix wasm processor docs

mihaitodor:
Should this be templated out somehow similar to [`project-github`](https://github.com/redpanda-data/rp-connect-docs/blob/78fa9ddaa59e57250217e37e54d9db36141c95c3/antora.yml#L9)?


Jeffail (approved):
LGTM, ideally it'd be templated one day but for now this is fine, it'll always point to the same repo containing the WASM proc itself.

PR #2729, opened 2024-07-23
Change cockroachdb to service category

rockwotj:
Integration is a one off category in the docs and seems wrongly placed,
so move it to where the rest of the services are.


Jeffail (approved):
LGTM, thanks @rockwotj

PR #2732, opened 2024-07-24
Fix openai embeddings processor

rockwotj:
Forgot to attach the model name.

Also correct some config constants copy/pasta issues, even if the
constants are the same values today.


mihaitodor (approved):
:shipit:

PR #2735, opened 2024-07-24
Add Content-MD5 to output_s3

### LINKED ISSUE #2687: S3 output missing header since aws-sdk-go-v2 upgrade
Hi, I've recently attempted to upgrade from v4.24.0, but encountered an issue with how the S3 output behaves when writing to a bucket with object lock enabled.

Using Benthos with versions greater than 4.24.0, AWS rejects the PutObject request with error: "Content-MD5 OR x-amz-checksum- HTTP header is required for Put Object requests with Object Lock parameters".

I believe this is due to the switch to `aws-sdk-go-v2` [which occurred in v4.25.0](https://github.com/redpanda-data/connect/blob/main/CHANGELOG.md#4250---2024-03-01). I'm not very familiar with Go, but I've had a look through the code for `aws-sdk-go`, and it looks like the original version of the SDK would automatically set the `Content-MD5` header for each `PutObject` request. See [here](https://github.com/aws/aws-sdk-go/blob/main/service/s3/customizations.go#L56) and [here](https://github.com/aws/aws-sdk-go/blob/main/service/s3/body_hash.go#L80).

I don't expect the Benthos connector to derive the MD5 and apply it automatically like the old `aws-sdk-go` did, but would like a way to provide it to the output e.g.:

```
      aws_s3:
          bucket: ingest-prod-data
          path: ${! meta("write_path") }
          tags: {}
          region: eu-west-2
          content_type: application/octet-stream
          content_md5: $!{ meta("file_hash") }
``` 

I'd be grateful if someone could check my assumptions regarding the `aws-sdk-go` to make sure I'm reading it right, and for alternative suggestions for how to fix this, whether it needs to be fixed or we can work around it somehow.

Thanks

mihaitodor:
Fixes #2687.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,13 @@ Changelog
 
 All notable changes to this project will be documented in this file.
 
-## 4.32.1 - 2024-07-24
+## 4.33.0 - TBD
 
 ### Added
 
 - Field `content_md5` added to the `aws_s3` output. (@dom-lee-naimuri)
+
+## 4.32.1 - 2024-07-24
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 
 ## 4.32.1 - 2024-07-24
 
+### Added
+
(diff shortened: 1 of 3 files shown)


Jeffail (approved):
LGTM, thanks @mihaitodor

PR #2737, opened 2024-07-25
#2601 Review OpenAI connector docs

asimms41:
Review comments for the following OpenAI connector docs:

- open_chat_completion
- open_embeddings
- openai_image_generation
- openai_speech
- openai_transcription
- openai_translation

Relates to the following open docs issue: [https://github.com/redpanda-data/documentation-private/issues/2601](https://github.com/redpanda-data/documentation-private/issues/2601)

### DIFF
--- a/internal/impl/openai/base_processor.go
+++ b/internal/impl/openai/base_processor.go
@@ -26,7 +26,6 @@ func baseConfigFieldsWithModels(modelExamples ...any) []*service.ConfigField {
 	return []*service.ConfigField{
 		service.NewStringField(opFieldServerAddress).
 			Description("The Open API endpoint that the processor sends requests to. Update the default value to use another OpenAI compatible service.").
-			Advanced().
 			Default("https://api.openai.com/v1"),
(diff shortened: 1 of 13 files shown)


internal/impl/openai/base_processor.go
```diff
@@ -26,7 +26,6 @@ func baseConfigFieldsWithModels(modelExamples ...any) []*service.ConfigField {
 	return []*service.ConfigField{
 		service.NewStringField(opFieldServerAddress).
 			Description("The Open API endpoint that the processor sends requests to. Update the default value to use another OpenAI compatible service.").
-			Advanced().
```
asimms41, line 29: Would it be OK to remove the Advanced tab where only the server_address is different from the common configuration, especially as the server_address is the first field listed in the docs?

rockwotj: Seems fine to me!

rockwotj (approved):
LGTM

PR #2738, opened 2024-07-26
Topic written logs and status events to set pipeline id as key

Jeffail:
Sets the key for topic written logs and status events to the configured pipeline identifier.


mihaitodor (approved):
:shipit:

PR #2740, opened 2024-07-26
input/nats - provide boolean switch for sending ACK packages

### LINKED ISSUE #2739: Nats Input - Sending ACK messages is not expected in non-Jetstream topics
- we interpret the protocol specification that the ACK package send as response is not part of the specification for plain (non JetStream) topics.
- we suggest to add a boolean configuration flag to the Nats input (default TRUE) that allows us to control the behavior

plejd-sebman:
- default value is TRUE
- when set to FALSE, the ACK package isn't sent
- this would comply with Nats protocol specification
- closes #2739

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 - Field `content_md5` added to the `aws_s3` output. (@dom-lee-naimuri)
+- Field `send_ack` added to the `nats` input. (@plejd-sebman)
 
 ## 4.32.1 - 2024-07-24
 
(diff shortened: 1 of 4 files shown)


Jeffail (approved):
LGTM, thanks @plejd-sebman. I've pushed a gofmt commit that's unrelated to your work, our CI has been a bit funky the last couple of days.

plejd-sebman:
@Jeffail thanks for considering this change. We're so grateful that we could contribute as company to this exiting software product.

heikkilamarko:
Hi. Should the `send_ack` cover NACKs as well?

PR #2752, opened 2024-07-31
Support multiple ollama processors with processes

rockwotj:
Currently if one tries to spin up multiple ollama processors in the same
processor you get errors about port conflicts on the second processor
that starts. Fix this by using a singleton pattern with the ollama
servers we start up automatically.

Tested manually.

### DIFF
--- a/internal/impl/ollama/base_processor.go
+++ b/internal/impl/ollama/base_processor.go
@@ -21,6 +21,7 @@ import (
 	"github.com/dustin/go-humanize"
 	"github.com/ollama/ollama/api"
 	"github.com/redpanda-data/benthos/v4/public/service"
+	"github.com/redpanda-data/connect/v4/internal/singleton"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


mihaitodor (commented):
Thanks Tyler! I think this approach makes sense if the underlying process can handle multiple types of models simultaneously. I left a small comment which is mostly a nitpick.

While looking at the implementation, I recalled that we now have some APIs for registering arbitrary resources that Ash added recently: https://pkg.go.dev/github.com/redpanda-data/benthos/v4@v4.33.0/public/service#Resources.GetGeneric and https://pkg.go.dev/github.com/redpanda-data/benthos/v4@v4.33.0/public/service#Resources.SetGeneric. There's no API to remove them, so might not fit the design here, but thought I'd mention them just in case.

internal/impl/ollama/base_processor.go
```diff
@@ -144,11 +166,12 @@ func (o *baseOllamaProcessor) pullModel(ctx context.Context) error {
...
+	defer o.cleanupMu.Unlock()
+	if o.cmd == nil {
 		return nil
 	}
-	if err := o.cmd.Process.Kill(); err != nil {
-		return err
-	}
-	return o.cmd.Wait()
+	err := ollamaProcess.Close(ctx)
+	o.cmd = nil
```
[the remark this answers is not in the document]
rockwotj: > maybe that should be a no-op

Thanks - I thought more about this and agree about the noop and avoiding panics. I switched to a ticket based mechanism for allowing noops. It seems the Close for a single processor can be called concurrently from multiple goroutines, so I still need a mechanism to detect noops.

rockwotj:
> While looking at the implementation, I recalled that we now have some APIs for registering arbitrary resources that Ash added recently: https://pkg.go.dev/github.com/redpanda-data/benthos/v4@v4.33.0/public/service#Resources.GetGeneric and https://pkg.go.dev/github.com/redpanda-data/benthos/v4@v4.33.0/public/service#Resources.SetGeneric and https://pkg.go.dev/github.com/redpanda-data/benthos/v4@v4.33.0/public/service#Resources.GetOrSetGeneric. There's no API to remove them, so might not fit the design here, but thought I'd mention them just in case.

Thanks, will keep this in mind! That doesn't quite fit because we really want a lock in this case - we can only start a single process and we also need to ensure that it's cleaned up and not leaked. Storing the Singleton itself in the Resources doesn't seem to offer an advantage over the global variable.

internal/singleton/singleton.go
```diff
@@ -0,0 +1,109 @@
...
+type Singleton[T any] struct {
+	mu         sync.Mutex
+	tickets    map[Ticket]struct{}
+	nextTicket Ticket
+	cfg        Config[T]
+	value      T
+}
+
+// Ticket is an opaque type signifying that a singleton's resource is acquired.
+type Ticket = int
```
mihaitodor, line 50: Why are we using a type alias? In this code I guess it's the same thing as having `type Ticket int`, but the latter would prevent someone from doing `var test int = s.nextTicket` for example. Still doesn't prevent `Close(context.Background(), 42)`, but, alas :)

mihaitodor (approved):
LGTM :shipit:

internal/singleton/singleton.go
rockwotj: Ah yes I was meaning to define a new type, thanks for catching 😄

PR #2753, opened 2024-07-31
Automatically download ollama if not installed

rockwotj:
Stacked on #2752

To improve the 0 to 60second experience we want to auto download ollama for devs if it's not installed locally.

### DIFF
--- a/internal/impl/ollama/base_processor.go
+++ b/internal/impl/ollama/base_processor.go
@@ -12,12 +12,20 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
+	"io"
 	"net/http"
 	"net/url"
+	"os"
 	"os/exec"
+	"path"
+	"path/filepath"
+	"runtime"
 	"sync"
+	"syscall"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/dustin/go-humanize"
 	"github.com/ollama/ollama/api"
 	"github.com/redpanda-data/benthos/v4/public/service"
@@ -25,8 +33,10 @@ import (
 )
 
 const (
-	bopFieldServerAddress = "server_address"
-	bopFieldModel         = "model"
+	bopFieldServerAddress  = "server_address"
+	bopFieldModel          = "model"
+	bopFieldCacheDirectory = "cache_directory"
+	bopFieldDownloadURL    = "download_url"
 )
 
 type commandOutput struct {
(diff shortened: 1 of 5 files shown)


emaxerrno:
SUPER SUPER COOL.

internal/impl/ollama/base_processor.go
```diff
@@ -86,6 +116,39 @@ var ollamaProcess = singleton.New(singleton.Config[*exec.Cmd]{
 		return cmd.Wait()
 	}})
 
+func downloadOllama(ctx context.Context, path string) error {
+	var url string
+	const baseURL string = "https://github.com/ollama/ollama/releases/download/v0.3.1/ollama"
+	switch runtime.GOOS {
```
emaxerrno, line 122: should this be configurable.

emaxerrno: with a default.

rockwotj: Good point. This will be the default, and I'll add a string to support overriding.

mihaitodor (commented):
Thanks Tyler! Just left one comment on how to do file stuff

internal/impl/ollama/base_processor.go
```diff
@@ -86,6 +116,39 @@ var ollamaProcess = singleton.New(singleton.Config[*exec.Cmd]{
...
+	}
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to download ollama binary: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("failed to download ollama binary: status_code=%d", resp.StatusCode)
+	}
+	ollama, err := os.OpenFile(path, os.O_CREATE|os.O_TRUNC|os.O_RDWR, 0777)
```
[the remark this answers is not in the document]
rockwotj: the `service.FS` interface doesn't allow for writing to the filesystem. How do I work around that? Do I need to add that to Benthos core?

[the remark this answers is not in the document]
rockwotj: I just am casting to `io.Writer` for now

internal/impl/ollama/base_processor.go
rockwotj: Done

internal/impl/ollama/base_processor.go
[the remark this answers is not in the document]
mihaitodor: Oh, good point... I'm actually not sure if there's any plan to expand that interface. Will defer to @Jeffail on this stuff.

internal/impl/ollama/base_processor.go
```diff
@@ -55,22 +65,109 @@ func (co *commandOutput) Write(b []byte) (int, error) {
...
+		p, err := c.fs.Stat(path)
+		if err != nil || p.IsDir() {
+			continue
+		}
+		// Check that the file is executable
+		if err := syscall.Access(path, 0x1); err == nil {
+			return path, nil
+		}
+	}
+	return "", exec.ErrNotFound
```
mihaitodor, line 98: If possible, can we have an explanation here on why we have to traverse `$PATH` manually and check if the file is executable? I just tested https://pkg.go.dev/os/exec#LookPath and I'm not sure how that function differs.

internal/impl/ollama/base_processor.go
```diff
@@ -55,22 +65,109 @@ func (co *commandOutput) Write(b []byte) (int, error) {
...
+}
+
+func (c *runOllamaConfig) downloadOllama(ctx context.Context, path string) error {
+	var url string
+	const baseURL string = "https://github.com/ollama/ollama/releases/download/v0.3.1/ollama"
```
mihaitodor, line 103: Should we consider exposing the version as a config parameter?

mihaitodor (approved):
Thanks Tyler! I left a few small comments, but it should be good to go otherwise.

internal/impl/ollama/base_processor.go
rockwotj: Yeah, the difference is that we're using our filesystem abstractions - will add a comment

internal/impl/ollama/base_processor.go
rockwotj: Oh probably since this is already old 😄

rockwotj: Well you can already provide a override URL to download from whereever, so I think that is a good enough workaround for now

internal/impl/ollama/base_processor.go
mihaitodor: Might be a good idea to have it in Benthos. Smth to check with @Jeffail for a future refactor

internal/impl/ollama/base_processor.go
```diff
@@ -55,22 +65,111 @@ func (co *commandOutput) Write(b []byte) (int, error) {
...
+	switch runtime.GOOS {
+	case "darwin":
+		// They ship an universal executable for darwin
+		url = baseURL + "-darwin"
+	case "linux":
+		url = fmt.Sprintf("%s-%s-%s", baseURL, runtime.GOOS, runtime.GOARCH)
+	default:
+		return fmt.Errorf("automatic download of ollama is not supported on %s, please download ollama manually", runtime.GOOS)
+	}
+	if c.downloadURL != "" {
```
mihaitodor, line 115: Should we maybe move this check above the switch just in case people have a download link for Windows?

rockwotj: Oh good catch! Fixed.

mihaitodor (approved):
LGTM :shipit:

PR #2760, opened 2024-08-06
Clarify queue and durable fields conflict in NATS input configuration.

BenBasilTomy:
**Changes Made**

- Updated documentation to explain the conflict between queue and durable fields in NATS input.

- Added a linting rule to detect and report this configuration issue at startup.

**Description**

- This PR addresses an issue where the queue and durable fields in the NATS input configuration cannot be used together, as this leads to a conflict between push and pull consumer configurations. The documentation has been updated to clarify this.

- A linting rule has been added to Benthos to automatically detect and report this configuration issue on startup, providing early feedback and preventing runtime errors.


mihaitodor:
Hey @BenBasilTomy, thank you for the contribution! It looks good, but we should also add a check somewhere here https://github.com/redpanda-data/connect/blob/0fd9100b70cb1f00a00ef9ac3b1454efb1d96856/internal/impl/nats/input_jetstream.go#L171-L180 just in case someone skips the linter by running Connect with the `--chilled` flag.

BenBasilTomy:
Hi @mihaitodor,

Thank you for the feedback! I've added the check to ensure that both durable and queue cannot be set simultaneously, even if someone skips the linter by running Connect with the --chilled flag. The check has been added in the **newJetStreamReaderFromConfig** function.

Could you please review the updated PR?

mihaitodor (approved):
LGTM :shipit: Also, thank you for the contribution! ❤️

PR #2761, opened 2024-08-07
Add cloud images to docker artefacts

Jeffail:
This PR adds cloud build artefacts to our published docker images. Also fixes an issue where the allowlist for components was too broad and so we couldn't have a `csv` scanner without allowing a `csv` input. The new system executes the allowlist per-component type.

### DIFF
--- a/resources/docker/Dockerfile.cloud
+++ b/resources/docker/Dockerfile.cloud
@@ -34,5 +34,4 @@ EXPOSE 4195
 
 ENTRYPOINT ["/redpanda-connect"]
 
-CMD ["-c", "/connect.yaml"]
-
+CMD ["run", "/connect.yaml"]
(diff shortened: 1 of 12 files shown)


resources/docker/Dockerfile.cloud
```diff
@@ -0,0 +1,38 @@
...
+COPY --from=build /go/src/github.com/redpanda-data/connect/target/bin/redpanda-connect-cloud ./redpanda-connect
+COPY ./config/docker.yaml /connect.yaml
+
+USER connect
+
+EXPOSE 4195
+
+ENTRYPOINT ["/redpanda-connect"]
+
+CMD ["-c", "/connect.yaml"]
```
mihaitodor, line 37: Should this be changed to use the new `run` action?

mihaitodor (approved):
LGTM 🐑 🚀

resources/docker/Dockerfile.cloud
Jeffail: Yesssss, nice find

PR #2765, opened 2024-08-07
Separate cloud build imports into a public package

Jeffail:
Moves the cloud schema and component imports to public packages so that other tools can utilise them. Also adds a couple other bits:
- License headers to dockerfiles
- Changed the default service name in logs to `redpanda-connect`


mihaitodor (approved):
LGTM :shipit:

PR #2766, opened 2024-08-07
gcp: add a vertex AI processor

rockwotj:
In similar form to our other AI processors add a processor for vertex
AI.


internal/impl/gcp/processor_vertex_ai.go
```diff
@@ -0,0 +1,316 @@
...
+	vaiFieldPrompt           = "prompt"
+	vaiFieldSystemPrompt     = "system_prompt"
+	vaiFieldTemp             = "temperature"
+	vaiFieldTopP             = "top_p"
+	vaiFieldTopK             = "top_k"
+	vaiFieldMaxTokens        = "max_tokens"
+	vaiFieldStop             = "stop"
+	vaiFieldPresencePenalty  = "presence_penalty"
+	vaiFieldFrequencyPenalty = "frequency_penalty"
+	vaiFieldResponseFormat   = "response_format"
```
mihaitodor, line 36: Nit: The prefix should be `vaip`

internal/impl/gcp/processor_vertex_ai.go
```diff
@@ -0,0 +1,316 @@
...
+}
+
+func newVertexAIProcessorConfig() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Categories("AI").
+		Summary("Generates responses to messages in a chat conversation, using the Vertex AI API.").
+		Description(`This processor sends prompts to your chosen large language model (LLM) and generates text from the responses, using the Vertex AI API.
+
+For more information, see the https://cloud.google.com/vertex-ai/docs[Vertex AI documentation^].`).
+		Version("4.33.0").
```
mihaitodor, line 57: Guess we'll have to bump this to `4.34.0` or not sure what will be the next release.

internal/impl/gcp/processor_vertex_ai.go
```diff
@@ -0,0 +1,316 @@
...
+				Description("The response format of generated type, the model must also be prompted to output the appropriate response type.").
+				Default("text"),
+			service.NewFloatField(vaiFieldTopP).
+				Advanced().
+				Description("If specified, nucleus sampling will be used.").
+				Optional(),
+			service.NewIntField(vaiFieldTopK).
+				Advanced().
+				Description("If specified top-k sampling will be used.").
+				Optional(),
```
mihaitodor, line 95: Would be nice to add linting for these like in the other PR

internal/impl/gcp/processor_vertex_ai.go
```diff
@@ -0,0 +1,316 @@
...
+		if err != nil {
+			return
+		}
+		proc.frequencyPenalty = genai.Ptr(float32(fp))
+	}
+	var format string
+	format, err = conf.FieldString(vaiFieldResponseFormat)
+	if format == "json" {
+		proc.responseMIMEType = "application/json"
+	} else {
```
mihaitodor, line 221: Maybe return an error if the else branch is not `text`. Users could try to run the config with `--chilled` which I think lets them place any value they want in this field.

internal/impl/gcp/processor_vertex_ai.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+	chat := m.StartChat()
+	prompt, err := p.computePrompt(msg)
+	if err != nil {
+		return nil, err
+	}
+	resp, err := chat.SendMessage(ctx, genai.Text(prompt))
+	if err != nil {
+		return nil, err
+	}
```
mihaitodor, line 272: Maybe consider enriching these errors

mihaitodor (approved):
Thanks Tyler, LGTM! Just left a few nitpics. Feel free to :shipit:

internal/impl/gcp/processor_vertex_ai.go
rockwotj: Thanks! Done.

internal/impl/gcp/processor_vertex_ai.go
rockwotj: I've bumped it! I wish there was a way to automate this somehow... 🤔

internal/impl/gcp/processor_vertex_ai.go
rockwotj: Done - some of these values differ based on the model used, but I've just used the biggest range for each.

internal/impl/gcp/processor_vertex_ai.go
rockwotj: Good catch!

internal/impl/gcp/processor_vertex_ai.go
mihaitodor: Yeah, indeed. It would be a good investment to build some automation. Maybe we can use `redpanda-connect list --format jsonschema` to figure out which components / fields are new. Should think about that and chat with Ash when he's around.

PR #2767, opened 2024-08-08
Add `schema_registry` input and output

mihaitodor:


emaxerrno:
awesome.

internal/impl/kafka/enterprise/schema_registry_input.go
```diff
@@ -0,0 +1,265 @@
...
+		}
+	}
+
+	return
+}
+
+//------------------------------------------------------------------------------
+
+func (i *input) Connect(ctx context.Context) error {
+	if i.connected.Load() {
```
rockwotj, line 146: Shouldn't this be a lock otherwise we'll get races writing to `i.subjects`?

internal/impl/kafka/enterprise/schema_registry_input.go
```diff
@@ -0,0 +1,265 @@
...
+	var body []byte
+	if body, err = io.ReadAll(resp.Body); err != nil {
+		return nil, fmt.Errorf("failed to read response body: %s", err)
+	}
+
+	return body, nil
+}
+
+func (i *input) Close(ctx context.Context) error {
+	return nil
```
rockwotj, line 264: Should we set connected false?

internal/impl/kafka/enterprise/schema_registry_input.go
```diff
@@ -0,0 +1,265 @@
...
+			i.subjects = append(i.subjects, s)
+		}
+	}
+
+	i.connected.Store(true)
+
+	return nil
+}
+
+func (i *input) Read(ctx context.Context) (*service.Message, service.AckFunc, error) {
```
rockwotj, line 172: Is this only going to have a single goroutine calling it?

internal/impl/kafka/enterprise/schema_registry_output.go
```diff
@@ -0,0 +1,195 @@
...
+		Categories("Integration").
+		Summary(`Publishes schemas to SchemaRegistry.`).
+		Description(service.OutputPerformanceDocs(true, true)).
+		Fields(
+			service.NewStringField(sroFieldURL).Description("The base URL of the schema registry service."),
+			service.NewInterpolatedStringField(sroFieldSubject).Description("Subject."),
+			service.NewStringField(sroFieldURL).Description("The base URL of the schema registry service."),
+			service.NewTLSToggledField(sroFieldTLS),
+			service.NewOutputMaxInFlightField(),
+		)
```
rockwotj, line 47: Consider adding your example from the PR description here

internal/impl/kafka/enterprise/schema_registry_output.go
```diff
@@ -0,0 +1,195 @@
...
+
+//------------------------------------------------------------------------------
+
+func outputSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Beta().
+		Version("4.32.2").
+		Categories("Integration").
+		Summary(`Publishes schemas to SchemaRegistry.`).
+		Description(service.OutputPerformanceDocs(true, true)).
```
rockwotj, line 40: benefitsFromBatching should be false

internal/impl/kafka/enterprise/schema_registry_output.go
```diff
@@ -0,0 +1,195 @@
...
+			o.log.Debugf("Failed to push data to SchemaRegistry with status %d: %s", resp.StatusCode, string(respData))
+		}
+
+		return fmt.Errorf("request returned status: %d", resp.StatusCode)
+	}
+
+	return nil
+}
+
+func (o *output) Close(_ context.Context) error { return nil }
```
rockwotj, line 195: connected to false?

internal/impl/kafka/enterprise/schema_registry_input.go
mihaitodor: I just noticed there's a new `atomic.Bool` thing since Go 1.19 and thought it would be fun to play with it, but you're right and I should've put more thought into this. Oh well... Locks it is :) Thank you for catching this! ❤️

internal/impl/kafka/enterprise/schema_registry_input.go
mihaitodor: It's a good discussion topic for the next Blobtalk. We really should document somehow what the state machine is. Like, can `Connect()` be called in parallel by multiple goroutines and why? For `Read()`, I guess one can define a `schema_registry` input component and reference it in multiple streams or use a `broker` input with `copies > 1`, so all bets are off...

internal/impl/kafka/enterprise/schema_registry_output.go
mihaitodor: The check in the `Connect()` method is a bit superfluous because `Write()` will error out anyway if the sink Schema Registry isn't writable so I thought it should only run once, but, then again, someone could flip SR to `READWRITE` while Connect is running, so yeah, good suggestion! Also, it's less surprising for a reviewer.

Jeffail:
@mihaitodor am I right in concluding that the globally unique identifier for each schema version when migrating this way wouldn't be preserved? I'm looking at https://docs.confluent.io/platform/current/schema-registry/develop/api.html#post--subjects-(string-%20subject)-versions and the implication is that we can't reliably enforce what the ID is going to be and it's globally incremented across all subjects.

Jeffail (approved):
LGTM, thanks @mihaitodor! I can imagine we're going to be asked to add a refresh interval field at some point so that we can continuously refresh the downstream schemas. However, given the various ways that a live replication can go wrong I think it's right to lean against that until we have time to document all of the caveats.

PR #2772, opened 2024-08-12
feat: Qdrant integration

Anush008:
## Description

This PR adds support for [Qdrant](https://qdrant.tech) to be used as a sink destination.

Supports dense/sparse/multi vectors.


internal/impl/qdrant/output.go
```diff
@@ -0,0 +1,243 @@
...
+	return service.NewConfigSpec().
+		Version("4.32.2").
+		Categories("AI").
+		Summary("Adds items to a [Qdrant](https://qdrant.tech/) collection").
+		Description(service.OutputPerformanceDocs(true, true)).
+		Fields(
+			service.NewOutputMaxInFlightField(),
+			service.NewBatchPolicyField(poFieldBatching),
+			service.NewStringField(poFieldGrpcHost).
+				Description("The gRPC host of the Qdrant collection. Defaults to 'localhost:6334'.").Default("localhost:6334"),
```
rockwotj, line 47: Is this a good default? Maybe we should make it an example instead and make it required.

internal/impl/qdrant/integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+				// Is is possible to test output only without a `getFn GetMessageFunc` arg?
+				integration.StreamTestOutputOnlySendBatch(10, defaultQuery),
+			)
+			suite.Run(
+				t, template, integration.StreamTestOptPort(containerPort.Port()),
+				integration.StreamTestOptVarSet("POINT_ID", tc.pointID),
+				integration.StreamTestOptVarSet("COLLECTION_NAME", collectionName),
+				integration.StreamTestOptVarSet("VECTOR", tc.vector),
+				integration.StreamTestOptVarSet("PAYLOAD", tc.payload),
+			)
```
rockwotj, line 112: This is fine, but it would be good to query qdrant after to ensure the data is written properly. Maybe it's also simpler to directly construct a writer via `newOutputWriter` then call `WriteBatch` explicitly? An example to take inspiration from is elasticsearch's integration test: https://github.com/redpanda-data/connect/blob/main/internal/impl/elasticsearch/writer_integration_test.go

internal/impl/qdrant/integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+func TestIntegrationQdrant(t *testing.T) {
+	integration.CheckSkip(t)
+
+	t.Parallel()
+
+	ctx := context.Background()
+	qdrantContainer, err := qdrant.Run(ctx, "qdrant/qdrant:v1.10.1")
+	if err != nil {
+		log.Fatalf("failed to start container: %s", err)
+	}
```
rockwotj, line 54: nit: use the require testing package like so
```
require.NoError(t, err, "failed to start container")
```

internal/impl/qdrant/integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+			},
+		},
+		SparseVectorsConfig: &pb.SparseVectorConfig{
+			Map: map[string]*pb.SparseVectorParams{
+				"some_sparse": {},
+			},
+		},
+	})
+
+	if err != nil {
```
rockwotj, line 165: Can we have this helper return an error and using `require.NoError` in the test itself?

internal/impl/qdrant/integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+			name:    "Test With dense and sparse vector",
+			pointID: `"465213dd-3f11-4534-8daf-9fedf203549a"`,
+			vector:  `{"some_dense": [0.352,0.532,0.532],"some_sparse": {"indices": [23,325,532],"values": [0.352,0.532,0.532]}}`,
+			payload: `{"content":"test payload"}`,
+		},
+	}
+
+	containerPort, err := qdrantContainer.MappedPort(ctx, "6334/tcp")
+	if err != nil {
+		log.Fatalf("failed to get container port: %s", err)
```
rockwotj, line 90: `require.NoError`

internal/impl/qdrant/integration_test.go
```diff
@@ -0,0 +1,168 @@
...
+				integration.StreamTestOptVarSet("POINT_ID", tc.pointID),
+				integration.StreamTestOptVarSet("COLLECTION_NAME", collectionName),
+				integration.StreamTestOptVarSet("VECTOR", tc.vector),
+				integration.StreamTestOptVarSet("PAYLOAD", tc.payload),
+			)
+		})
+	}
+
+	if err := qdrantContainer.Terminate(ctx); err != nil {
+		log.Fatalf("failed to terminate container: %s", err)
```
rockwotj, line 117: require.NoError

rockwotj (commented):
Thank you @Anush008, this looks *awesome*!

The only thing here outside of nits is that I would like to strengthen the integration test by asserting the output is expected. Does that sound do-able? Happy to help if needed.

internal/impl/qdrant/output.go
```diff
@@ -0,0 +1,243 @@
...
+	"context"
+	"fmt"
+
+	pb "github.com/qdrant/go-client/qdrant"
+	"github.com/redpanda-data/benthos/v4/public/bloblang"
+	"github.com/redpanda-data/benthos/v4/public/service"
+)
+
+const (
+	poFieldBatching       = "batching"
```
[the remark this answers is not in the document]
Anush008: Oh. I kept wondering what `po` meant.
Thanks.

internal/impl/qdrant/client.go
```diff
@@ -0,0 +1,103 @@
...
+	return &qdrantClient{
+		serviceClient: pb.NewQdrantClient(conn),
+		pointsClient:  pb.NewPointsClient(conn),
+		connection:    conn,
+		logger:        logger,
+	}, nil
+}
+
+func (c *qdrantClient) Upsert(ctx context.Context, collectionName string, points []*pb.PointStruct) error {
+	c.logger.Infof("Upserting %d points to collection %s", len(points), collectionName)
```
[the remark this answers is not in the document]
Anush008: Any specific benefits with trace?

internal/impl/qdrant/output.go
```diff
@@ -0,0 +1,243 @@
...
+	poFieldGrpcHost       = "grpc_host"
```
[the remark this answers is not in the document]
Anush008: That's what I used to call.
But Qdrant's REST port,  6333 is more commonly used.

Most people will mistakenly specify 6333 here. So this clearly states `grpc`.

internal/impl/qdrant/output.go
```diff
@@ -0,0 +1,243 @@
...
+			service.NewOutputMaxInFlightField(),
+			service.NewBatchPolicyField(poFieldBatching),
+			service.NewStringField(poFieldGrpcHost).
+				Description("The gRPC host of the Qdrant collection. Defaults to 'localhost:6334'.").Default("localhost:6334"),
+			service.NewStringField(poFieldAPIToken).
+				Secret().
+				Description("The Qdrant API token for authentication. Defaults to an empty string.").Default(""),
+			service.NewTLSToggledField(poFieldUseTLS).Default("TLS(HTTPS) config to use when connecting"),
+			service.NewInterpolatedStringField(poFieldCollectionName).
+				Description("The name of the collection in Qdrant. REQUIRED"),
```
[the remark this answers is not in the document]
Anush008: Will do.

internal/impl/qdrant/output.go
Anush008: That should be fine.

internal/impl/qdrant/client.go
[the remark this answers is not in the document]
rockwotj: Logging volume, in higher throughput systems this could generate a ton of logging information. Other outputs (like kafka_franz) output this information at `Debug` level, we should adhere to that standard here.

internal/impl/qdrant/value.go
```diff
@@ -0,0 +1,178 @@
...
+//	╔════════════════════════╤════════════════════════════════════════════╗
+//	║ Go type                │ Conversion                                 ║
+//	╠════════════════════════╪════════════════════════════════════════════╣
+//	║ nil                    │ stored as NullValue                        ║
+//	║ bool                   │ stored as BoolValue                        ║
+//	║ int, int32, int64      │ stored as IntegerValue                     ║
+//	║ uint, uint32, uint64   │ stored as IntegerValue                     ║
+//	║ float32, float64       │ stored as DoubleValue                      ║
+//	║ string                 │ stored as StringValue; must be valid UTF-8 ║
+//	║ []byte                 │ stored as StringValue; base64-encoded      ║
```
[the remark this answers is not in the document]
Anush008: The mapping is pretty intuitive. So I think it's fine.

internal/impl/qdrant/client.go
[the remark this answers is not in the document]
Anush008: Will do.

emaxerrno:
This is dope.

internal/impl/qdrant/output.go
```diff
@@ -42,32 +42,32 @@ func outputSpec() *service.ConfigSpec {
...
+				Description("The mapping to extract the vector from the document.").
 				Example(`root = this.vector`).
 				Example(`root = [1.2, 0.5, 0.76]`).
 				Example(`root = [[0.352,0.532,0.532,0.234],[0.352,0.532,0.532,0.234]]`).
 				Example(`root = {"some_sparse": {"indices":[23,325,532],"values":[0.352,0.532,0.532]}}`).
 				Example(`root = {"some_multi": [[0.352,0.532,0.532,0.234],[0.352,0.532,0.532,0.234]]}`).
 				Example(`root = {"some_dense": [0.352,0.532,0.532,0.234],"some_sparse": {"indices": [23,325,532],"values": [0.352,0.532,0.532]}}`),
-			service.NewBloblangField(poFieldPayloadMapping).
-				Optional().
+			service.NewBloblangField(qoFieldPayloadMapping).
```
rockwotj, line 67: Why is this no longer optional? Can we keep it optional?

internal/impl/qdrant/output.go
```diff
@@ -103,36 +103,36 @@ type outputWriter struct {
...
+	vectorMapping, err := conf.FieldBloblang(qoFieldVectorMapping)
 	if err != nil {
 		return nil, err
 	}
 
-	payloadMapping, err := conf.FieldBloblang(poFieldPayloadMapping)
+	payloadMapping, err := conf.FieldBloblang(qoFieldPayloadMapping)
 	if err != nil {
 		return nil, err
 	}
```
rockwotj, line 138: If we go back to making this optional then you need:

```
var payloadMapping service.BloblangExecutor*
if conf.Contains(qoFieldPayloadMapping) {
  payloadMapping, err = conf.FieldBloblang(qoFieldPayloadMapping)
  if err != nil {
    return nil, err
  }
}
```

Then you just need to handle this being `nil` while writing outputs and omit the payload (or write nil or something). You can also default to `root = {}` so you write an empty object by default.

internal/impl/qdrant/output.go
```diff
@@ -42,32 +42,32 @@ func outputSpec() *service.ConfigSpec {
...
-			service.NewInterpolatedStringField(poFieldCollectionName).
-				Description("The name of the collection in Qdrant. REQUIRED"),
-			service.NewBloblangField(poFieldID).
-				Description("The ID of the point to insert. REQUIRED").
+			service.NewTLSToggledField(qoFieldUseTLS).Default("TLS(HTTPS) config to use when connecting"),
+			service.NewInterpolatedStringField(qoFieldCollectionName).
+				Description("The name of the collection in Qdrant."),
+			service.NewBloblangField(qoFieldID).
+				Description("The ID of the point to insert. Can be a UUID string or positive integer.").
 				Example(`root.id = this.id`).
```
rockwotj, line 56: this should be `root = this.id`

internal/impl/qdrant/output.go
```diff
@@ -42,32 +42,32 @@ func outputSpec() *service.ConfigSpec {
...
 				Example(`root.id = 832`).
```
rockwotj, line 57: `root = 832`

internal/impl/qdrant/output.go
```diff
@@ -42,32 +42,32 @@ func outputSpec() *service.ConfigSpec {
...
 				Example(`root.id = "dc88c126-679f-49f5-ab85-04b77e8c2791"`),
```
rockwotj, line 58: `root = "..."`

rockwotj (commented):
Looks awesome just one small thing about making payload optional.

internal/impl/qdrant/output.go
Anush008: We're reading from the "id" key.
https://github.com/Anush008/connect/blob/main/internal/impl/qdrant/point_id.go

Setting `root` directly to a string, kept failing with some cryptic errors.

Anush008:
```
root = "dc88c126-679f-49f5-ab85-04b77e8c2791"
```
Kept failing.

internal/impl/qdrant/output.go
Anush008: See https://github.com/redpanda-data/connect/pull/2772#discussion_r1714089591

internal/impl/qdrant/output.go
Anush008: Default would be simpler, yeah.
Is that
```
.Default(`root = {}`)
```
for a `BloblangField`?

Anush008: [`1ff5a1c` (#2772)](https://github.com/redpanda-data/connect/pull/2772/commits/1ff5a1c0c889a8e0c8a2a803c571b342381ad1b7)

Anush008:
Hey @rockwotj. Added assertions for the payload after the writes.
If any of the vectors is invalid, the upload process itself will fail.

internal/impl/qdrant/output.go
rockwotj: It needed `QueryValue` instead of just `Query`. I pushed a fix

internal/impl/qdrant/output.go
rockwotj: Yes the default is right

internal/impl/qdrant/output.go
Anush008: Awesome. Thank you 🙏.

rockwotj (approved):
Nice work! Thank you for your contribution @Anush008

Anush008:
Thanks for the save in https://github.com/redpanda-data/connect/commit/60d60a550f4e4c323a62af1753144aaaacf24d02 Tyler 🙏.

internal/impl/qdrant/point_id.go
```diff
@@ -0,0 +1,33 @@
+package qdrant
```
mihaitodor, line 1: Missing license header

internal/impl/qdrant/value.go
```diff
@@ -0,0 +1,178 @@
+package qdrant
```
mihaitodor, line 1: Same

internal/impl/qdrant/vectors.go
```diff
@@ -0,0 +1,189 @@
+package qdrant
```
mihaitodor, line 1: Same

internal/impl/qdrant/integration_test.go
```diff
@@ -0,0 +1,227 @@
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+
+			defaultQuery := func(ctx context.Context, testID, messageID string) (string, []string, error) {
+				return fmt.Sprintf(`{"content":"%v","id":%v}`, "hello world", messageID), nil, err
+			}
+
+			suite := integration.StreamTests(
+				// Is it possible to test output only without a `getFn GetMessageFunc` arg?
```
mihaitodor, line 114: The idea of the `getFn` is to have a way to manually fetch whatever data the output received and then have the test suite check that whatever data was sent to the output matches what was written to it, but the assumption was that the messages are kind of stored as-is, which isn't the case here as far as I can tell.

Since your test doesn't really fit into this pattern, you can also ditch the `integration.StreamTests` utility and just spin up a stream yourself which lets you push arbitrary data into it and then again read whatever the output received and assert it. See for example [this one](https://github.com/redpanda-data/connect/blob/2bc199978b94a012fbe1422052b3524839148f71/internal/impl/gcp/integration_pubsub_test.go#L132-L231) where we just check if the output rejects the message. In your case, you can also run `assertPoints()` after the stream terminates if you follow this pattern.

internal/impl/qdrant/output.go
```diff
@@ -0,0 +1,237 @@
...
+				Description("The gRPC host of the Qdrant collection. Defaults to 'localhost:6334'.").Default("localhost:6334"),
+			service.NewStringField(qoFieldAPIToken).
+				Secret().
+				Description("The Qdrant API token for authentication. Defaults to an empty string.").Default(""),
+			service.NewTLSToggledField(qoFieldUseTLS).Description("TLS(HTTPS) config to use when connecting"),
+			service.NewInterpolatedStringField(qoFieldCollectionName).
+				Description("The name of the collection in Qdrant."),
+			service.NewBloblangField(qoFieldID).
+				Description("The ID of the point to insert. Can be a UUID string or positive integer.").
+				Example(`root = "dc88c126-679f-49f5-ab85-04b77e8c2791"`).
```
mihaitodor, line 56: In terms of ergonomics for fields which need to just resolve to a single value, would it not make more sense to use interpolation here so the users could do `${! uuid_v4() }` or `832`?

internal/impl/qdrant/output.go
```diff
@@ -0,0 +1,237 @@
...
+			Id:      id,
+			Vectors: vec,
+			Payload: payload,
+		})
+	}
+	return batches, nil
+}
+
+func (w *outputWriter) Close(ctx context.Context) error {
+	if w.client != nil {
```
mihaitodor, line 233: The nil check here is probably superfluous unless `w.client.Connect(ctx)` can make it nil. Do we need any locking to make sure `Connect()` has executed successfully before attempting to write / close?

internal/impl/kafka/enterprise/integration_test.go
```diff
@@ -35,6 +35,7 @@ import (
 	_ "github.com/redpanda-data/benthos/v4/public/components/pure"
 	"github.com/redpanda-data/benthos/v4/public/service"
 	"github.com/redpanda-data/benthos/v4/public/service/integration"
+
```
mihaitodor, line 38: Guess we can get rid of this change

internal/impl/qdrant/value.go
```diff
@@ -0,0 +1,178 @@
...
+//	║ uint, uint32, uint64   │ stored as IntegerValue                     ║
+//	║ float32, float64       │ stored as DoubleValue                      ║
+//	║ string                 │ stored as StringValue; must be valid UTF-8 ║
+//	║ []byte                 │ stored as StringValue; base64-encoded      ║
+//	║ map[string]interface{} │ stored as StructValue                      ║
+//	║ []interface{}          │ stored as ListValue                        ║
+//	╚════════════════════════╧════════════════════════════════════════════╝
+
+func newValueMap(input any) (map[string]*pb.Value, error) {
+	valueMap := make(map[string]*pb.Value)
```
mihaitodor, line 49: Small nit: `make(map[string]*pb.Value, len(inputMap))` and place it after the type assertion.

mihaitodor (commented):
LGTM, thank you for all this work! Just left a bunch of nitpicks.

internal/impl/kafka/enterprise/integration_test.go
Anush008: `make fmt` keeps adding that.

internal/impl/qdrant/output.go
Anush008: I tried that. Since it's interpolated _string_ and evaluates to a string in the code, it felt off trying to parse it into a number. Where `bloblang` fields evaluated to `any`.

internal/impl/qdrant/point_id.go
Anush008: [`2bf4f7e` (#2772)](https://github.com/redpanda-data/connect/pull/2772/commits/2bf4f7e0ee0eff8c273718f110a894f2dcb3ba4d)

internal/impl/qdrant/integration_test.go
Anush008: Refactor the `getFn` to assert value after write.
[`c02af75` (#2772)](https://github.com/redpanda-data/connect/pull/2772/commits/c02af75475c7fa572b4e6b40b1744b282d3eafbe)

internal/impl/qdrant/output.go
Jeffail: This should be an example not a default, it's a small amount of work from the users perspective but having it explicit is good practice for any fields that represent a connection.

Anush008: Done in [`b25a024` (#2772)](https://github.com/redpanda-data/connect/pull/2772/commits/b25a0245b16588a51b481b1d3529d695190a2f73). Thank you.

Jeffail:
Awesome, thanks @Anush008!

internal/impl/kafka/enterprise/integration_test.go
mihaitodor: Oh, oops… Sounds like our linter is missing some stuff. Thanks for catching it!

Anush008:
Hey guys. Is there an ETA on when https://docs.redpanda.com/redpanda-connect/components/outputs/qdrant will be available?
We'll link to it in the announcements and our docs.

Anush008:
Would it be possible to get this doc link fix PR in
https://github.com/redpanda-data/connect/pull/2787
?

rockwotj:
Hey, yes at the moment it's a manual process to update the docs as I know it. Our docs team is on it and I will send you an email once it's live 😄 Thank you again for your work here - this is awesome.

PR #2777, opened 2024-08-12
Cloud AI image

rockwotj:
If we want to run ollama processors in cloud, then split out a separate AI image to do so.


public/schema/ai_allow_list.txt
```diff
@@ -0,0 +1,3 @@
+ollama_chat
+ollama_embeddings
```
Jeffail, line 2: It should be possible to add these to the general `cloud_allow_list` since it doesn't crash when the plugins aren't recognised, it'll just ignore them. Since the cloud build doesn't import the ollama plugins they won't be present.

If I have time I'm going to rework the allow list so that it's a csv file with columns for each type of distribution we might make, at which point it'll all be in one place even if we end up with a joined suite of imports.

rockwotj: Ah yes good point, fixed - I've kept the single allow list

Jeffail (approved):
LGTM, thanks @rockwotj

PR #2814, opened 2024-08-28
Flush topic logs and add missing cloud components

Jeffail:
This PR adds a ref counter to dispatched log messages so that we can wait for them to flush before shutting down. It's quite crude, especially within the close blocking loop, but it's a lot simpler than adding some form of lock.


mihaitodor (approved):
LGTM :shipit:

PR #2815, opened 2024-08-28
Add QuestDB Output Component

sklarsa:
Adds an output component for writing to [QuestDB](https://questdb.io/) tables using ILP (Influx Line Protocol) as implemented by the official [QuestDB Go Client](https://github.com/questdb/go-questdb-client).


emaxerrno:
This is cool

internal/impl/questdb/output.go
```diff
@@ -0,0 +1,526 @@
...
+			service.NewBoolField("tlsEnabled").
+				Description("Use TLS to secure the connection to the server").
+				Optional(),
+			service.NewStringField("tlsVerify").
+				Description("Whether to verify the server's certificate. This should only be used for testing as a "+
+					"last resort and never used in production as it makes the connection vulnerable to "+
+					"man-in-the-middle attacks. Options are 'on' or 'unsafe_off'.").
+				Optional().
+				Default("on").
+				LintRule(`root = if ["on","unsafe_off"].contains(this != true { ["valid options are \"on\" or \"unsafe_off\"" ] }`),
```
rockwotj, line 62: We have a service.NewTLSToggledField("tls") that has default fields supported everywhere - can we please us that?

https://github.com/redpanda-data/connect/blob/1c664da8ab51b796bb5e11189fdd41f2b41d9125/internal/impl/kafka/input_kafka_franz.go#L123

internal/impl/questdb/output.go
```diff
@@ -0,0 +1,526 @@
...
+			service.NewDurationField("retryTimeout").
```
rockwotj, line 63: We use snake case naming: `retry_timeout`. This also applies to other fields.

internal/impl/questdb/output.go
sklarsa: Yep, how's this? Our sender uses the tlsConfig contained in the unmarshalled conf.

rockwotj (approved):
I'm not too familiar with QuestDB, but the code LGTM. Will let other team members review as well.

Jeffail:
LGTM, thanks @sklarsa!

PR #2825, opened 2024-08-30
[CI] Distribute `connect` as a rpk managed plugin

jackietung-redpanda:
Project context is [here](https://docs.google.com/document/d/1ubeclJ9mwF5q2E0HzwOkRLOx2o5X8Ghnmm_kWQfnCSg/edit#heading=h.mvlqwsjv79vf).

This PR tackles the workflow of uploading `connect` to a designed S3 repo, which acts as the "rpk plugin" repo, used only by `connect` for now.

**plugin_uploader.py**
To be run after `goreleaser` has built the binaries.  `upload-archives` creates and uploads the archives.  `upload-manifest` creates and uploads the final `manifest.json` file.  RPK uses `manifest.json` to know what plugins are available and how to download them (download URL, etc.).  Comes with unit tests, basic linters, type checkers, setup in CI.

**Additional release workflow**
`.github/workflows/upload_plugin.yml`.  High level:

* `goreleaser build`
* Run `plugin_uploader.py upload-archives ...`
* Run `plugin_uploader.py upload-manifest ...`

Only uploads for real when triggered from a release tag (inc. RC).  Otherwise, runs in `--dry-run` mode - no writes to S3.

**Why not use `goreleaser` built in [blob publish support](https://goreleaser.com/customization/blob/)?**

Goreleaser has "blob" publish support, which kind of can do `upload-archives` in spirit.  However, our use case requires additional flexibility in how we set S3 object tags.  Our `manifest.json` must contain SHA256 checksums of the binary payload WITHIN the `tar.gz` (`rpk` requirement).   It would not work gracefully with said blob publishing within Goreleaser.

`manifest.json` construction logic (`upload-archives`) would be needed in any case and most cleanly implemented separately from goreleaser.

Generally the design of goreleaser is not conducive to running, testing, debugging, retrying specific parts of the `goreleaser` steps.  E.g. Publishing steps are all or nothing.  To test blob publishing, one would also be publishing to Github!

Overall the monolithic design of goreleaser's runtime makes things tricky for more complex use cases.

Thankfully `goreleaser` provides a clean interface for extension scripts (`artifacts.json`, `metadata.json`)

**Why not use `goreleaser` [custom publisher support](https://goreleaser.com/customization/publishers/)?**

Similar reasoning to above.

Note the calculus may be different for other use cases.  E.g. next task is to build Linux packages for `redpanda-connect` (deb / rpm).  The functionality may fit much more closely such that we can accept the drawbacks we describe here.  We do that for [`redpanda-console`](https://github.com/redpanda-data/console-enterprise/blob/master/.goreleaser.yml) (see [`nfpm`](https://goreleaser.com/customization/nfpm/) usage there).

**Why not extend existing `release.yml` Github action workflow?**

This could be an option in the long term.  However this functionality is new and may introduce instability in the main `release.yml` pipeline in the meantime.  It will be much cleaner to test, debug, run (and rerun!) as an independent workflow.  Some observations:
 * The S3 upload flow does not need a `goreleaser release`, only `goreleaser build`!
 * `release.yml` pushes real Github releases and may not be idempotent!

Once we gain confidence that both `.github/workflows/upload_plugin.yml` AND `release.yml` are *very* reliable, we can consider merging into `release.yml`.


resources/plugin_uploader/test_data/dist/metadata_v4_34_0.json
```diff
@@ -0,0 +1,12 @@
+{
```
Jeffail, line 1: Are these files a necessary addition to the repo? And if so do they need to be generated and checked in for each tagged release?

jackietung-redpanda: This is static test data (`test_data`), used by unit tests, no changes needed w.r.t. new release cuts.

ivotron (approved):
🚀  lgtm, this is great!

.github/workflows/upload_plugin.yml
```diff
@@ -0,0 +1,84 @@
...
+  workflow_dispatch: {}
+
+env:
+  # Do dry run in most cases, UNLESS the triggering event was a "tag".
+  DRY_RUN: ${{ ! github.ref_type == 'tag' }}
+
+jobs:
+  upload_rpk_connect_plugin:
+    # Let's make this fast by using a beefy runner.
+    runs-on: ubuntu-latest-32
```
andrewhsu, line 28:
```suggestion
    if: ${{ github.event.pull_request.head.repo.full_name == 'redpanda-data/connect' }}
    runs-on: ubuntu-latest-32
```
skip this job if the PR is from a forked repo or else you will get an error on configuring aws credentials because it is trying to access secrets

.github/workflows/upload_plugin.yml
```diff
@@ -0,0 +1,84 @@
...
+
+      - name: Get secrets from AWS Secrets Manager (for read/writing S3-backed plugins repo)
+        uses: aws-actions/aws-secretsmanager-get-secrets@v2
+        with:
+          secret-ids: |
+            ,sdlc/prod/github/rpk_plugin_publisher
+          parse-json-secrets: true
+
+      - name: Check out code repo
+        uses: actions/checkout@v4
```
andrewhsu, line 47:
```suggestion
      - uses: actions/checkout@v4
```
nit: you can reduce this to just 'uses:' because it will be a step that says "Run actions/checkout@v4"

andrewhsu: nit: same goes for all other steps that is just a one-line 'uses:' or 'run:'

.github/workflows/test_plugin_uploader.yml
```diff
@@ -0,0 +1,68 @@
...
+        uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - uses: chartboost/ruff-action@v1
+        with:
+          version: 0.4.10
+          src: './resources/plugin_uploader'
```
andrewhsu, line 47:
```suggestion
      - name: Lint with ruff
        run: |
          pip install ruff
          ruff check --output=github ./resources/plugin_uploader
```
i feel the github action `chartboost/ruff-action` does not bring a lot to the table and may be risky if they push a bad release. instead, i suggest using github's [example](https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-python#using-ruff-to-lint-code).

andrewhsu (changes requested):
needs guard in `upload_plugin.yml`

.github/workflows/upload_plugin.yml
jackietung-redpanda: I just added a check on the repo name.   Won't check for PR-ness here, that is handled in the trigger filters.

Jeffail (approved):
LGTM

Jeffail:
Hey @jackietung-redpanda, I'm aiming to tag a release later today, is this PR required before we do it or can it wait until the one afterwards?

.github/workflows/upload_plugin.yml
```diff
@@ -0,0 +1,83 @@
...
+
+env:
+  # Do dry run in most cases, UNLESS the triggering event was a "tag".
+  DRY_RUN: ${{ ! github.ref_type == 'tag' }}
+
+jobs:
+  upload_rpk_connect_plugin:
+    # Let's make this fast by using a beefy runner.
+    runs-on: ubuntu-latest-32
+    if: ${{ github.repository == 'redpanda-data/connect' }}
```
andrewhsu, line 29:
```suggestion
    if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == 'redpanda-data/connect' }}
```
if i create a PR to this repo from a forked repo `andrewhsu/connect`, the `github.repository` will still have value `redpanda-data/connect`. and the guard will let the control flow into step `Configure AWS credentials` where it will fail. unfortunately, it is poorly [documented](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) about that specific.

you won't see that situation in this PR because this PR was created from a branch on `redpanda-data/connect` which means you have write permissions to the repo and therefore are allowed to access the `${{ secrets.RP_AWS_CRED_ACCOUNT_ID }}`

andrewhsu (commented):
i think PRs from forks will still have errors in `upload_plugin.yml`, but if you only create PRs from branches on `redpanda-data/connect` you will be fine

.github/workflows/upload_plugin.yml
jackietung-redpanda: Appreciate suggestion again.  After doing a bit of research myself, I accept that this somewhat ugly check is the right way!  Updated.

andrewhsu (approved):
LGTM

PR #2827, opened 2024-09-03
Add components csv and formatting command

Jeffail:
This PR moves the allow list for cloud into a CSV file. The new file has more information (support level, more granular product checks, etc) and also:

- Formatted with `make docs` so that the file remains sorted and also new components added to the build are added with default values so that they're less likely to be forgotten
- Tested to ensure that all listed plugins exist and are present in the build of RPCN, this ensures we aren't "enabling" plugins that aren't actually compiled into the product


internal/plugins/info.csv
```diff
@@ -0,0 +1,248 @@
+name                    ,type      ,commercial_name         ,support    ,deprecated ,cloud ,cloud_ai
```
mihaitodor, line 1: What about `zmq4`?

mihaitodor (approved):
LGTM, just one small question and :shipit:

internal/plugins/info.csv
Jeffail: Fdfrhaskldjghyjklsdahfgjkedhfkjladfhjv, yep. CI caught it which is good.

Deflaimun (approved):
LGTM. We will prob have to find a way in the future to fill the commercial_name with our drivers. For now, we can do it manually

internal/plugins/info.csv
```diff
@@ -0,0 +1,250 @@
...
+socket                  ,input     ,socket                  ,certified  ,n          ,n     ,n
+socket                  ,output    ,socket                  ,certified  ,n          ,n     ,n
+socket_server           ,input     ,socket_server           ,certified  ,n          ,n     ,n
+split                   ,processor ,split                   ,certified  ,n          ,y     ,y
+splunk                  ,input     ,splunk                  ,enterprise ,n          ,y     ,y
+splunk_hec              ,output    ,splunk_hec              ,enterprise ,n          ,y     ,y
+sql                     ,cache     ,sql                     ,certified  ,n          ,n     ,n
+sql                     ,output    ,sql                     ,community  ,y          ,n     ,n
+sql                     ,processor ,sql                     ,community  ,y          ,n     ,n
+sql_insert              ,output    ,sql_insert              ,certified  ,n          ,y     ,y
```
JakeSCahill, line 217: We need some more data in this table.

`sql_*` connectors have different support levels for different drivers. For example the ClickHouse driver is community whereas the PostgreSQL driver is certified.

Similarly the `cache` outputs have different support levels depending on the target cache.

Jeffail: The individual cache output support levels are covered by the `cache` type, those are the underlying components being used.

For `sql` drivers I think maybe the way forward would be to add a specific `sql_driver` type as it's redundant to have each driver represented for each sql type. Also, having a given plugin/type combo like `sql_insert` represented multiple times in the table will be effectively seen as duplicates on the back end.

JakeSCahill: > The individual cache output support levels are covered by the cache type, those are the underlying components being used.

Thanks!

> For sql drivers I think maybe the way forward would be to add a specific sql_driver type as it's redundant to have each driver represented for each sql type. Also, having a given plugin/type combo like sql_insert represented multiple times in the table will be effectively seen as duplicates on the back end.

Will this be handled as part of this PR?

Jeffail: @JakeSCahill yep, done. Going to merge as is but if you take a look and want changes just let me know and I'll follow up.

PR #2829, opened 2024-09-03
openai: use interpolated string instead of bloblang for prompt

rockwotj:
This is for consistency between all the other AI processors.


mihaitodor (approved):
LGTM :shipit:

PR #2837, opened 2024-09-05
remove unused MacOS quill signing logic

jackietung-redpanda:
This code path is unused.

Currently only place we do any signing is when we statically bundle `connect` with `rpk` on the `rpk` homebrew tap.  That path does not actually use this `.goreleaser.yml` file.  Note also that `quill` is not installed and none of the signing secrets are ever present when using this `.goreleaser.yml` in any case.

In very near future, we will do away with static bundling altogether.


Jeffail (approved):
LGTM

jackietung-redpanda:
I spoke too soon I think.  We need to revert this - otherwise next release cut of RP will fail at this point.  I was confused about the goreleaser files of connect vs rpk's.

I will prep the revert and merge it.

PR #2838, opened 2024-09-05
Parallelize docker releases using matrix

rockwotj:
See documentation here: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,6 +42,30 @@ jobs:
 
   docker:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        flavor: [default, cgo, cloud, ai]
+        platform: ["linux/amd64", "linux/arm64"]
+        exclude:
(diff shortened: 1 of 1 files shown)


jackietung-redpanda:
This is awesome!  I also just noticed the multiple, slow serial steps.  In future, if we want it even faster, we can use larger instances, or use something like depot.dev.

rockwotj:
I think after we'll probably be good for a while, but larger instances might be nice. It only takes be a couple minutes to build an image locally with 12 cores.

rockwotj:
BTW I tested that this expands to the right set of variables here: https://github.com/rockwotj/redpanda-connect/actions/runs/10728166709

(The job fails because there are secrets not set, but that's expected)

Jeffail (approved):
Awesome, thanks @rockwotj!

PR #2839, opened 2024-09-05
plugin upload release workflow - fix dry-run determination

jackietung-redpanda:
Issue was that DRY_RUN always evaluated to true, due to misunderstanding about GHA expressions syntax and how that can be used as env vars under different contexts.  At least this is probably the least bad way to fail.

We fix it all up here.


jackietung-redpanda:
Tested manually:
* PR CI itself - confirmed that it resolves to `--dry-run`.
* Tagged a dummy RC release tag on the PR commit, indeed, NOT `--dry-run`.

Build error shown as headline is only because I cancelled the full release build (docker, goreleaser jobs) from the dummy RC tag test.

ivotron (approved):

PR #2843, opened 2024-09-06
openai: use BloblangQuery instead of BloblangQueryValue

rockwotj:
Looking at QueryValue I'm quite confused what it's for - it doesn't bind
any variables, so stuff like `this.text` fails with `context was undefined,
unable to reference "text"`.

So switch to normal Query and extract the result using AsBytes


mihaitodor (approved):
:shipit:

PR #2844, opened 2024-09-06
Release Linux packages

jackietung-redpanda:
Publishing `connect` via deb/rpm packages is part of the new `rpk connect` distribution strategy.  More context [here](https://docs.google.com/document/d/1ubeclJ9mwF5q2E0HzwOkRLOx2o5X8Ghnmm_kWQfnCSg/edit#heading=h.mvlqwsjv79vf).

This PR adds said Linux packaging.  Packages will be created and published to Redpanda's production Linux package repo on Cloudsmith.

Solution is modeled off [`redpanda-console`](https://github.com/redpanda-data/console-enterprise/blob/fd1e0d089f4b1969224494ec275f219cd338b222/.goreleaser.yml#L52), which also publishes packages easily via goreleaser's [nfpm](https://goreleaser.com/customization/nfpm/) integration.

This PR must merge after `CLOUDSMITH_API_KEY` access has been granted.  This means after [this PR](https://github.com/redpanda-data/qa-infra/pull/116) gets merged and TF applied.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -94,3 +94,32 @@ release:
     name: connect
   prerelease: auto
   disable: false
+
+nfpms:
+  - id: connect-linux-pkgs
+    description: Redpanda Connect is a high performance and resilient stream processor.
+    package_name: redpanda-connect
+    file_name_template: "{{ .ConventionalFileName }}"
+    # this is the default value, but specifying explicitly it relates to the symlink creation below
+    bindir: /usr/bin
+    contents:
+      - src: /usr/bin/redpanda-connect
+        dst: /usr/bin/.rpk-ac.connect
+        type: symlink
+    builds:
+      - connect
+    vendor: Redpanda Data, Inc.
+    license: "https://github.com/redpanda-data/connect/blob/main/licenses/README.md"
+    homepage: redpanda.com
+    maintainer: Redpanda Data <support@redpanda.com>
+    formats:
+      - deb
+      - rpm
+publishers:
+  # Gets run once per artifact (deb or rpm)
+  - name: Publish Linux packages to Cloudsmith
+    ids:
+      - connect-linux-pkgs
+    cmd: ./resources/scripts/push_pkg_to_cloudsmith.sh {{ .ArtifactPath }}
+    env:
+      - CLOUDSMITH_API_KEY={{ .Env.CLOUDSMITH_API_KEY }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,10 +8,26 @@ on:
 jobs:
   goreleaser:
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
     steps:
 
     - name: Check Out Repo
       uses: actions/checkout@v4
+
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v4
+      with:
+        aws-region: ${{ vars.RP_AWS_CRED_REGION }}
+        role-to-assume: arn:aws:iam::${{ secrets.RP_AWS_CRED_ACCOUNT_ID }}:role/${{ vars.RP_AWS_CRED_BASE_ROLE_NAME }}${{ github.event.repository.name }}
+
+    - name: Get secrets from AWS Secrets Manager
+      uses: aws-actions/aws-secretsmanager-get-secrets@v2
+      with:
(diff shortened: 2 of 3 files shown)


jackietung-redpanda:
Linux package publishing was tested by running `goreleaser release` locally, after commenting out certain parts of the config:
* Disable GitHub release publishing
* Keep only connect build (for speed)
* Keep only Linux / amd64 build (for speed)

jackietung-redpanda:
cc @r-vasquez  in case interested to see.

.goreleaser.yml
```diff
@@ -84,3 +84,26 @@ release:
...
+    file_name_template: "{{ .ConventionalFileName }}"
+    builds:
+      - connect
+    vendor: Redpanda Data, Inc.
+    license: "https://github.com/redpanda-data/connect/blob/main/licenses/README.md"
+    homepage: redpanda.com
+    maintainer: Redpanda Data <support@redpanda.com>
+    formats:
+      - deb
+      - rpm
```
ivotron, line 101: shouldn't there be a dependency with `redpanda-rpk`?

jackietung-redpanda: AFAIK, the intent is not to pull in `rpk` automatically when someone does `apt install redpanda-connect`.

For almost ALL users, they would be installing `apt install redpanda-rpk` as the entrypoint.  `rpk` at runtime takes care of the rest.

For **airgapped** users, they would be installing `apt install redpanda-rpk`.  They would then ALSO do `apt install redpanda-connect` IF they want the `connect` plugin.  I think it keeps the mental model simple this way, without adding complexity:
* You want to run `rpk`?  `apt install redpanda-rpk`.
* You want to run `rpk connect`?  Add `apt install redpanda-connect`

Note this is not a one-way decision - adding the dep is much easier than removing it (if folks start depending on it, rightly or wrongly!).  Mechanics wise, there will be some complications (we can figure something out):
* Depend on `redpanda-rpk` which version of RPK?  Some 24.1.x, some 24.2.x?  How might that be expressed in the deps?  How might that play with the existing pins folks have in their env for redpanda itself?
* Would we need to wait till we have `redpanda-rpk` packages that DO NOT bundle `.rpk-ac.connect`?  I think yes.  We would need to think of some time line of a sequential update schedule.  Something like:
   * have `redpanda-connect` (no deps) - _we are here_.
   * have new `redpanda-rpk` packages A that do not have connect inside
   * have `redpanda-connect` start depending on A (or newer, etc)

Don't feel very strongly.  I think not adding the dep is slightly more sensible right now.

As an aside - `redpanda-connect` is actually fully usable as an independent binary (just as we distribute it that way on our `redpanda-data/connect` releases on GitHub.

ivotron: thanks a lot for the complete explanation! I agree that this route (not adding the dependency) is preferable. This raises a question though: what is the expectation when the following happens?

  * install `redpanda-rpk`
  * run `rpk connect ..` which will fetch the plugin
  * install `redpanda-connect`

I think we discussed this in one of our meetings but I forgot what was the answer.

jackietung-redpanda: This will be handled by new rpk (that can do fetches).  My understanding is that `rpk` maintains it's own set of managed plugin binaries, separated by user.  So the initially fetched plugin will live in some user specific folder, and that trumps everything.  I.e. even if `redpanda-connect` is subsequently installed by the admin (root owned), it will not have an effect as long as the user's own fetched plugin remains installed.

`rpk` even provides a command to list plugins, show the effective plugin binary being used, and any background ones that are being "shadowed".

jackietung-redpanda:
Prerequisite[ PR](This means after https://github.com/redpanda-data/qa-infra/pull/116 gets merged) has been merged and applied.  Linux publishing step here should have sufficient perms to push to Cloudsmith.

.goreleaser.yml
```diff
@@ -93,4 +93,33 @@ release:
     owner: redpanda-data
     name: connect
   prerelease: auto
-  disable: false
+  disable: true
```
Jeffail, line 96:
```suggestion
  disable: false
```

I'm assuming this is a left over from testing.

Jeffail (changes requested):

.goreleaser.yml
jackietung-redpanda: yes great catch - serious one.

Jeffail (approved):

PR #2855, opened 2024-09-09
Adds functionality to `schema_registry_decode` to emit defaults or not

hendoxc:
# What

This PR allows users to specify different behaviour for `schema_registry_decode` when decoding `protobuf` messages; allowing them to chose to Emit defaults.

# Why

if a protobuf message like `{"a": 0, "b": 1}` is encoded to protobuf, when we decode with `schema_registry_decode` the resulting message is `{"b": 1}` and the keyval `"a": 0` is lost. 

This behaviour is controlled from `protojson.MarshalerOpts` with the field `EmitDefaultValues`, which by fault is `false`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,12 +3,16 @@ Changelog
 
 All notable changes to this project will be documented in this file.
 
-## 4.58.0 - TBD
+## 4.58.0
 
 ### Added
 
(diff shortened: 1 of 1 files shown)


ntx-ben:
Currently facing this issue. Any updates on this?

PR #2856, opened 2024-09-09
ollama: bump automatically downloaded version

rockwotj:
They also changed to a tarball format instead of a plain binary


mihaitodor (approved):
LGTM 🚀 🐑

PR #2861, opened 2024-09-10
ollama: fix subprocess shutdown

rockwotj:
Currently when starting a subprocess there is a hang on shutdown. You
can't call Wait after Kill I guess. So instead use gracefully
shutdown with SIGINT, but if there


mihaitodor (approved):
Nice :shipit: Just not sure if the description was supposed to contain some more details after "but if there"

rockwotj:
Oops, thanks! Fixed the commit message 😅

PR #2865, opened 2024-09-11
release: re-enable arm ai images

rockwotj:
We were once only downloading the x86_64 version of ollama, but now we
use the correct platform, so we can build both.


bojand (approved):

Jeffail (approved):
LGTM

PR #2868, opened 2024-09-11
cohere: introduce cohere processors

rockwotj:
support for chat and embed processors


Jeffail (approved):
LGMT, thanks @rockwotj. I also owe the import test a few slices of my own bacon already 😅

PR #2872, opened 2024-09-13
cohere: use api_key instead of auth_token

rockwotj:
The code/dev docs uses auth token, but their console and product docs
use api_key. Let's use api key for consistency with OpenAI


Jeffail (approved):
LGTM

PR #2886, opened 2024-09-20
couchbase: add output

rockwotj:
Very much based on the processor.


internal/plugins/info.csv
```diff
@@ -53,6 +53,7 @@ cohere_embeddings       ,processor ,cohere_embeddings       ,4.37.0  ,enterprise
 command                 ,processor ,command                 ,4.21.0  ,certified  ,n          ,n     ,n
 compress                ,processor ,compress                ,0.0.0   ,certified  ,n          ,y     ,y
 couchbase               ,cache     ,Couchbase               ,4.12.0  ,community  ,n          ,n     ,n
+couchbase               ,output    ,couchbase               ,0.0.0   ,community  ,n          ,n     ,n
```
Jeffail, line 56: Version needs setting. I'm aware that this is annoying and awkward as we also have the redundant `.Version` method on the docs type. I think going forwards I'm going to suggest we don't bother filling `.Version` and just focus on this CSV file, as we can use the CSV file to hydrate the actual plugin docs.

internal/impl/couchbase/output.go
```diff
@@ -0,0 +1,179 @@
...
+			content, err = res.AsBytes()
+			if err != nil {
+				return err
+			}
+		}
+
+		ops[index] = o.op(k, content)
+	}
+
+	return o.client.collection.Do(ops, &gocb.BulkOpOptions{})
```
mihaitodor, line 170: I wonder if we'll need some locking for this since the client gets closed in `Close()`.

mihaitodor (approved):
LGTM, just one small nit and :shipit:

internal/plugins/info.csv
rockwotj: Done

internal/impl/couchbase/output.go
rockwotj: Looks like the driver does it for us!

https://github.com/neo4j/neo4j-go-driver/blob/67137658d9757ba6a89868ed8e1a22a0e0bf9bd0/neo4j/driver_with_context.go#L387

rockwotj: Ooops wrong PR, I still think it's done for us: https://github.com/couchbase/gocb/blob/5c2640b2b5f2b17fd24918f22a6738394f66b7a2/client_core.go#L686

PR #2894, opened 2024-09-24
Fix redpanda-migrator topic creation logic

mihaitodor:
I forgot to look at the logs when testing #2878. My bad.

This should avoid log spam as well. It's handy to see which topics get created and when, but it's less useful to print logs when they already exist.


Jeffail (approved):
LGTM, thanks @mihaitodor!

PR #2895, opened 2024-09-24
feat: added nkey field to nats

ye11ow:
Currently in order to use nkey authentication, the user has to provide a local file that contains the nkey seed. This is OK if the user has full access of the deployment environment of redpanda-connect.
However, for managed redpanda-connect environment, the user doesn't have directly control of the file system. Thus, we'd like to introduce a new secret field called `nkey` to allow the user to specify the nkey seed directly in the config yaml.

### DIFF
--- a/internal/impl/nats/auth.go
+++ b/internal/impl/nats/auth.go
@@ -71,9 +71,9 @@ func authFieldSpec() *service.ConfigField {
 			Optional(),
 		service.NewStringField("nkey").
 			Description("The NKey seed.").
-			Default("").
 			Secret().
-			Optional(),
+			Optional().
+			Example("UDXU4RCSJNZOIQHZNWXHXORDPRTGNJAHAHFRGZNEEJCPQTT2M7NLCNF4"), // don't worry, this sample seed is from Nats offical doc
 		service.NewStringField("user_credentials_file").
 			Description("An optional file containing user credentials which consist of an user JWT and corresponding NKey seed.").
 			Example("./user.creds").
@@ -286,8 +286,7 @@ func saveTempNkeyFile(content string) (string, error) {
 	}
 	defer file.Close()
 
-	_, err = file.WriteString(content)
-	if err != nil {
+	if _, err = file.WriteString(content); err != nil {
 		return "", err
 	}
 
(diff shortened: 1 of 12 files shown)


internal/impl/nats/auth.go
```diff
@@ -248,3 +272,23 @@ func loadFileContents(filename string, fs *service.FS) ([]byte, error) {
 
 	return io.ReadAll(f)
 }
+
+func saveTempNkeyFile(content string) (string, error) {
```
Jeffail, line 276: I think writing this file to disk is acceptible but we need to have it documented in the field description so that users are aware. Maybe add something like "Specify an NKey seed directly in config, this seed will be written to (and loaded from) a temporary file." to the description.

The other issue is that a new temporary file would be written upon each instantiation, which could potentially result in unbounded numbers of files being created during crashloops, batch jobs, dynamically run pipelines, and so on. We'd need to solve that either by having a consistent temporary file path, or just require that `nkey_file` is specified but the file is created when it doesn't already exist and this new field is populated, that allows users to choose where the file is written.

Jeffail (commented):
Hey @ye11ow, is it possible to use https://pkg.go.dev/github.com/nats-io/nats.go#Nkey instead of writing a temporary file?

internal/impl/nats/auth.go
mihaitodor: Instead of creating a temp file, maybe try this instead:
```go
kp, err := nkeys.ParseDecoratedNKey(auth.NKey)
if err != nil {
    return nil, fmt.Errorf("failed to parse nkey")
}

pub, err := kp.PublicKey()
if err != nil {
    return nil, fmt.Errorf("failed to extract public key from nkey")
}
if !nkeys.IsValidPublicUserKey(pub) {
    return nil, fmt.Errorf("invalid nkey user seed")
}
sigCB := func(nonce []byte) ([]byte, error) {
	sig, _ := kp.Sign(nonce)
	return sig, nil
}

opts = append(opts, nats.Nkey(string(pub), sigCB))
```

I took a peek at their code: https://github.com/nats-io/nats.go/blob/278f9f188bca4d7bdee283a0e98ab66b82530c60/nats.go#L5782

ye11ow: Thanks @mihaitodor ! I will give it a try.

Again, I'm just surprised that they don't have a `nats.NkeyOptionFromString` 😄 I thought this might be missing for a good reason.

mihaitodor: You’re welcome! I think they’d be open to adding it if you open an issue on their repo.

ye11ow: Hi @mihaitodor , I've updated the code based on your comment so that it no longer stores anything in temp file.

I did setup local nats server and tested it, all work well.

Thanks!

mihaitodor (approved):
LGTM :shipit:

PR #2898, opened 2024-09-25
Add `fetch_in_order` field to `schema_registry` input

mihaitodor:


internal/impl/kafka/enterprise/integration_test.go
```diff
@@ -376,6 +395,19 @@ func TestSchemaRegistryIntegration(t *testing.T) {
 				resp, err = http.DefaultClient.Do(req)
 				require.NoError(t, err)
 				require.NoError(t, resp.Body.Close())
+				require.Equal(t, http.StatusOK, resp.StatusCode)
+			}
+
+			if test.schemaWithReference != "" {
+				body, err := json.Marshal(payload{Schema: test.schemaWithReference, References: []map[string]any{{"name": "foo", "subject": subject, "version": 1}}})
```
mihaitodor, line 402: A bit janky, but I can clean it up to use the name from `t.schema` if need be...

internal/impl/kafka/enterprise/redpanda_migrator_bundle_output.tmpl.yaml
```diff
@@ -33,24 +33,33 @@ fields:
...
-  if ["topic", "key", "partition", "partitioner", "timestamp"].any(f -> this.redpanda_migrator.keys().contains(f)) {
-    root = throw("The topic, key, partition, partitioner and timestamp fields of the redpanda_migrator output must be left empty")
+
+  let srMaxInFlight = this.schema_registry.max_in_flight.or(1)
+  let schemaRegistry = if this.schema_registry.length() > 0 {
+    this.schema_registry.assign({
+      "subject": "${! @schema_registry_subject }",
+      "max_in_flight": $srMaxInFlight
+    })
   }
```
mihaitodor, line 63: Just making sure that we set `max_in_flight` to `1` unless users set it to something else explicitly (and hopefully are aware they'll get out-of-order messages in that case).

docs/modules/components/pages/inputs/schema_registry.adoc
```diff
@@ -148,6 +149,15 @@ Include only subjects which match the regular expression filter. All subjects ar
...
 
+=== `fetch_all_schemas_on_connect`
+
+Fetch all schemas on connect. Set this to `true` when schema references are used.
+
+
+*Type*: `bool`
+
+*Default*: `true`
+
```
rockwotj, line 160: This name is not very intuitive for me.

Can we name it:
```
enforce_ordering: bool

If true all schemas will be fetched and sorted, so that schemas are inserted in the same order. This is required to correctly resolve schema references.

default: true
```

On another note: do we even need this option? Is this only false if we would OOM reading all the schemas?

rockwotj: Or maybe `fetch_in_order` since they can still mess up the order on the output with max_in_flight

mihaitodor: Naming things is hard 😅 I like `fetch_in_order`, thanks!

mihaitodor: > Is this only false if we would OOM reading all the schemas?

I'm not super-concerned about OOM unless the schemas are huge, but it could take a long time to fetch them all if there are many of them, so it might be handy to stream them if people wish to stick a cache in the middle. Or maybe I'm overthinking it... The tests cover both flows now, so I'm happy to leave it there.

internal/impl/kafka/enterprise/schema_registry_input.go
```diff
@@ -167,23 +174,57 @@ func (i *schemaRegistryInput) Connect(ctx context.Context) error {
...
+				var schema sr.SchemaInfo
+				if schema, err = i.client.GetSchemaBySubjectAndVersion(ctx, subject, &version); err != nil {
+					return fmt.Errorf("failed to fetch schema version %d for subject %q: %s", version, subject, err)
+				}
+
+				schemas[schema.ID] = schemaInfo{
+					SchemaInfo: schema,
+					subject:    subject,
+					version:    version,
+				}
```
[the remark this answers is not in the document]
mihaitodor: Thanks for spotting that! Indeed, this is a corner case that I was trying to get my head around and it looks like you're right and the schemas will get duplicate IDs :( The trouble is that I can't get Redpanda to import schemas with a fixed ID but different subjects:

```
$ curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" --data '{"id": 1, "version": 1, "schema": "{\"name\": \"Foo\", \"type\": \"record\", \"fields\": [{\"name\": \"data\", \"type\": \"int\"}]}"}' http://localhost:8081/subjects/foo/versions
{"id":1}%                                                                                                                                                                       
$ curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" --data '{"id": 1, "version": 1, "schema": "{\"name\": \"Foo\", \"type\": \"record\", \"fields\": [{\"name\": \"data\", \"type\": \"int\"}]}"}' http://localhost:8081/subjects/bar/versions
{"error_code":42207,"message":"Overwrite new schema with id 1 is not permitted."}%                                                                                              
```

It does work, however, if I drop the `id` field or both the `id and `version` fields:
```shell
$ curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" --data '{"schema": "{\"name\": \"Foo\", \"type\": \"record\", \"fields\": [{\"name\": \"data\", \"type\": \"int\"}]}"}' http://localhost:8081/subjects/bar/versions
{"id":1}%
$ curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" --data '{"version": 1, "schema": "{\"name\": \"Foo\", \"type\": \"record\", \"fields\": [{\"name\": \"data\", \"type\": \"int\"}]}"}' http://localhost:8081/subjects/baz/versions
{"id":1}%
```
However, if I do this, then it feels kind of brittle because I'd have to make sure I first send the schema which does have the `id` field set...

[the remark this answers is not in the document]
mihaitodor: This should be fixed now. I'll defer adding a test until Redpanda lifts the above constraint as part of 
CORE-7734.

internal/impl/kafka/enterprise/schema_registry_input.go
```diff
@@ -145,17 +158,11 @@ func inputFromParsed(pConf *service.ParsedConfig, mgr *service.Resources) (i *sc
...
-			i.client.Transport = cloned
-		} else {
-			i.client.Transport = &http.Transport{
-				TLSClientConfig: tlsConf,
-			}
-		}
+	if !tlsEnabled {
+		tlsConf = nil
+	}
+	if i.client, err = sr.NewClient(srURL.String(), reqSigner, tlsConf, mgr); err != nil {
```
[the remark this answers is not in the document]
mihaitodor: Thank you for that suggestion! The code is much cleaner indeed! ❤️

rockwotj (approved):
LGTM

Jeffail (approved):

internal/impl/kafka/enterprise/schema_registry_input.go
```diff
@@ -196,60 +239,78 @@ func (i *schemaRegistryInput) Read(ctx context.Context) (*service.Message, servi
...
+		}()
 	}
 
-	var structured map[string]any
-	if err := json.Unmarshal(data, &structured); err != nil {
-		return nil, nil, fmt.Errorf("failed to unmarshal response body: %s", err)
+	// Omit `subject` from the schema since the `/subjects/<subject>/versions` endpoint doesn't allow it as part of
+	// the payload we pass it along as metadata.
+	schema := map[string]any{
+		"id":      si.ID,
```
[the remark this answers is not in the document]
mihaitodor: Hopefully CORE-7734 gets addressed so we won’t have to make this logic even more convoluted.

PR #2906, opened 2024-09-30
PostgreSQL CDC

Jeffail:
I've merged the PgSQL repos for CDC into `./internal/impl/postgresql` and have done some tidy up to get it (almost) compiling. There's a bunch of TODOs where we have panics and other things we wouldn't want to carry into a production plugin, but we're most of the way. There's also placeholders for `*service.Logger` which aren't hooked up yet.


Jeffail:
@le-vlad It's probably easiest for you to kick off future work on a fork of this branch and then merge it in via a new PR.

PR #2908, opened 2024-09-30
Update bundled packages to the latest Connect and Benthos versions

### LINKED ISSUE #2905: Bump bundled version to latest release (v4.47.0)
Currently both [free](https://github.com/redpanda-data/connect/blob/main/public/bundle/free/go.mod) and [enterprise](https://github.com/redpanda-data/connect/blob/main/public/bundle/enterprise/go.mod) are on v4.31.0, requiring the use for `replace` when writing plugins that target the newer versions (https://github.com/united-manufacturing-hub/benthos-umh/pull/63/files).

mihaitodor:
Fixes #2905.

### DIFF
--- a/public/bundle/enterprise/go.mod
+++ b/public/bundle/enterprise/go.mod
@@ -1,40 +1,43 @@
 module github.com/redpanda-data/connect/public/bundle/enterprise/v4
 
-go 1.21
+go 1.22.2
 
-require github.com/redpanda-data/connect/v4 v4.31.0
+require github.com/redpanda-data/connect/v4 v4.37.0
 
 require (
-	cloud.google.com/go v0.112.2 // indirect
-	cloud.google.com/go/auth v0.2.2 // indirect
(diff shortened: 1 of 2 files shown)


rockwotj (approved):
We should really automate this one - can dependabot or renovate do this?

mihaitodor:
@rockwotj Good question... I guess we can do a preflight check before tagging a new release. It's a bit like a chicken before the egg issue, since you can't both bump the version and, also, modify these files as part of the same commit. Curious if you have any ideas on how to approach it.

rockwotj:
Ah these are incremented in the same "tag", they are separate tags/release: https://github.com/redpanda-data/connect/releases/tag/public%2Fbundle%2Fenterprise%2Fv4.31.0

So we can: release, open PR to bump, then separately release the bundles

rockwotj:
cc: @Jeffail

PR #2913, opened 2024-10-01
support images in various LLM processors

rockwotj:
- **ollama: add ability to specify image**
- **gcp/vertex: add support for attachments**
- **openai: support images**


Jeffail (approved):
LGTM, is it worth adding some `.Example`'s to these fields?

PR #2915, opened 2024-10-02
[azure-blob-storage] Make `container` interpolated

mfamador:
Support interpolation on `azure_blob_storage` input `container` parameter.

Example:

```yaml
input:
  azure_blob_storage:
    storage_connection_string: UseDevelopmentStorage=true;
    container: container-${! now().ts_format("2006") }
```

Fix output `container` parameter interpolation. It now interpolates correctly from message contents not only from bloblang functions.

Example:

```yaml
output:
  azure_blob_storage:
    storage_connection_string: UseDevelopmentStorage=true;
    container: container-${! this.event_timestamp.ts_format("2006") }
```


Jeffail (approved):
LGTM, thanks @mfamador!

PR #2916, opened 2024-10-03
Certify MQTT plugin

ooesili:
I went through the MQTT plugin and kicked the tires and it seems solid.


Jeffail (approved):
LGTM, thanks @ooesili! I'm also going to enable these components in the cloud builds.

PR #2917, opened 2024-10-07
PostgreSQL CDC Plugin

le-vlad:
Added:
- Support for PostgreSQL versions (17,....10)
  - Implemented a different approach for taking snapshots in earlier versions (<15)
  - Using `pg_create_logical_replication_slot` instead of `CREATE_REPLICATION_SLOT`
  - Note: Snapshot name not provided by default with this method
  - Investigating discrepancy with documentation (https://www.postgresql.org/docs/12/protocol-replication.html)
  - Aim to use newer API instead of old one once resolved

- Replication monitoring
  - Added progress tracking for snapshot tables (%)
  - Added WAL lag monitoring (bytes) when streaming is started

- Stream uncommitted changes
  - New `stream_uncommitted` flag for use with `pgoutput` plugin
  - Allows receiving updates mid-transaction
  - Contrasts with `wal2json`, which sends updates with LSN of committed transactions

- Configurable decoding plugin
  - Added `decoding_plugin` option to specify plugin for streaming data
  - Supports `wal2json` or `pgoutput`

- Temporary replication slots
  - New `temporary_slot` setting to create temporary slots
  - Instructs PostgreSQL to drop replication slot on connection close
  - Useful for testing or when persistent replication slots are not needed

- Comprehensive integration tests for various scenarios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,16 @@ Changelog
 
 All notable changes to this project will be documented in this file.
 
+## 4.40.0 - TBD
+
+### Added
+
+- New `pg_stream` input supporting change data capture (CDC) from PostgreSQL (@le-vlad)
+
+### Changed
+
+- `snowflake_streaming` with `schema_evolution.enabled` set to true can now autocreate tables.
+
 ## 4.39.0 - 2024-11-07
 
 ### Added


internal/impl/postgresql/input_postgrecdc.go
```diff
@@ -0,0 +1,562 @@
...
+	Field(service.NewFloatField("snapshot_memory_safety_factor").
+		Description("Determines the fraction of available memory that can be used for streaming the snapshot. Values between 0 and 1 represent the percentage of memory to use. Lower values make initial streaming slower but help prevent out-of-memory errors.").
+		Example(0.2).
+		Default(1)).
+	Field(service.NewIntField("snapshot_batch_size").
+		Description("The number of rows to fetch in each batch when querying the snapshot. A value of 0 lets the plugin determine the batch size based on `snapshot_memory_safety_factor` property.").
+		Example(10000).
+		Default(0)).
+	Field(service.NewStringEnumField("decoding_plugin", "pgoutput", "wal2json").
+		Description("Specifies the logical decoding plugin to use for streaming changes from PostgreSQL. 'pgoutput' is the native logical replication protocol, while 'wal2json' provides change data as JSON.").
```
rockwotj, line 89: We parse the native protocol, so maybe it's useful to mention that the data is still converted into JSON for the pipeline?

internal/impl/postgresql/input_postgrecdc.go
```diff
@@ -0,0 +1,562 @@
...
+		Description("If set to true, the plugin will stream uncommitted transactions before receiving a commit message from PostgreSQL. This may result in duplicate records if the connector is restarted.").
+		Default(false)).
+	Field(service.NewStringField("pg_conn_options").
+		Description("Additional PostgreSQL connection options as a string. Refer to PostgreSQL documentation for available options.").
+		Default(""),
+	).
+	Field(service.NewBoolField("stream_snapshot").
+		Description("When set to true, the plugin will first stream a snapshot of all existing data in the database before streaming changes.").
+		Example(true).
+		Default(false)).
```
rockwotj, line 79: @Jeffail did we want to decouple the snapshot from the change stream as two different plugins?

internal/impl/postgresql/input_postgrecdc.go
```diff
@@ -0,0 +1,562 @@
...
+		Description("Password for the specified PostgreSQL user.")).
+	Field(service.NewStringField("schema").
+		Description("The PostgreSQL schema from which to replicate data.")).
+	Field(service.NewStringField("database").
+		Description("The name of the PostgreSQL database to connect to.")).
+	Field(service.NewStringEnumField("tls", "require", "none").
+		Description("Specifies whether to use TLS for the database connection. Set to 'require' to enforce TLS, or 'none' to disable it.").
+		Example("none").
+		Default("none")).
+	Field(service.NewBoolField("stream_uncomited").
```
rockwotj, line 69: `stream_uncommitted`

internal/impl/postgresql/input_postgrecdc.go
```diff
@@ -0,0 +1,562 @@
...
+	Version("0.0.1").
+	Summary(`Creates a PostgreSQL replication slot for Change Data Capture (CDC)
+		== Metadata
+
+This input adds the following metadata fields to each message:
+- streaming (Indicates whether the message is part of a streaming operation or snapshot processing)
+- table (Name of the table that the message originated from)
+- operation (Type of operation that generated the message, such as INSERT, UPDATE, or DELETE)
+		`).
+	Field(service.NewStringField("host").
```
rockwotj, line 48: nit: we've been making all of these values constants instead of inlining in two places.

internal/impl/postgresql/input_postgrecdc.go
```diff
@@ -0,0 +1,562 @@
...
+	Field(service.NewBoolField("stream_snapshot").
+		Description("When set to true, the plugin will first stream a snapshot of all existing data in the database before streaming changes.").
+		Example(true).
+		Default(false)).
+	Field(service.NewFloatField("snapshot_memory_safety_factor").
+		Description("Determines the fraction of available memory that can be used for streaming the snapshot. Values between 0 and 1 represent the percentage of memory to use. Lower values make initial streaming slower but help prevent out-of-memory errors.").
+		Example(0.2).
+		Default(1)).
+	Field(service.NewIntField("snapshot_batch_size").
+		Description("The number of rows to fetch in each batch when querying the snapshot. A value of 0 lets the plugin determine the batch size based on `snapshot_memory_safety_factor` property.").
```
rockwotj, line 85: We should instead of this make them unset so we can easily check for mutually exclusive as a lint check:

https://github.com/redpanda-data/connect/blob/754fd8a7e54edc44e22affca29af0fee41b52235/internal/impl/snowflake/output_snowflake_streaming.go#L98-L100

You can check if it's set using `.Contains` and default to the appropriate value if neither is set when extracting the config.

internal/impl/postgresql/input_postgrecdc.go
```diff
@@ -0,0 +1,562 @@
...
+	Field(service.NewStringEnumField("decoding_plugin", "pgoutput", "wal2json").
+		Description("Specifies the logical decoding plugin to use for streaming changes from PostgreSQL. 'pgoutput' is the native logical replication protocol, while 'wal2json' provides change data as JSON.").
+		Example("pgoutput").
+		Default("pgoutput")).
+	Field(service.NewStringListField("tables").
+		Description("A list of table names to include in the logical replication. Each table should be specified as a separate item.").
+		Example(`
+			- my_table
+			- my_table_2
+		`)).
```
rockwotj, line 97: I think you want `[]string{"my_table", "my_table_2"}` instead.

internal/impl/postgresql/input_postgrecdc.go
```diff
@@ -0,0 +1,562 @@
...
+	Field(service.NewIntField("checkpoint_limit").
+		Description("The maximum number of messages of the same topic and partition that can be processed at a given time. Increasing this limit enables parallel processing and batching at the output level to work on individual partitions. Any given offset will not be committed unless all messages under that offset are delivered in order to preserve at least once delivery guarantees.").
```
rockwotj, line 99: This description mentions partitions which is I assume copied from the kafka input?

internal/impl/postgresql/input_postgrecdc.go
```diff
@@ -0,0 +1,562 @@
...
+		Version("3.33.0").Default(1024)).
+	Field(service.NewBoolField("temporary_slot").
+		Description("If set to true, creates a temporary replication slot that is automatically dropped when the connection is closed.").
+		Default(false)).
+	Field(service.NewStringField("slot_name").
+		Description("The name of the PostgreSQL logical replication slot to use. If not provided, a random name will be generated. You can create this slot manually before starting replication if desired.").
+		Example("my_test_slot").
+		Default(randomSlotName)).
+	Field(service.NewAutoRetryNacksToggleField()).
+	Field(service.NewBatchPolicyField("batching").Advanced())
```
rockwotj, line 109: Do we normally mark batching as advanced for inputs?

internal/impl/postgresql/input_postgrecdc.go
```diff
@@ -0,0 +1,562 @@
...
+	Field(service.NewStringField("password").
+		Description("Password for the specified PostgreSQL user.")).
+	Field(service.NewStringField("schema").
+		Description("The PostgreSQL schema from which to replicate data.")).
+	Field(service.NewStringField("database").
+		Description("The name of the PostgreSQL database to connect to.")).
+	Field(service.NewStringEnumField("tls", "require", "none").
+		Description("Specifies whether to use TLS for the database connection. Set to 'require' to enforce TLS, or 'none' to disable it.").
+		Example("none").
+		Default("none")).
```
rockwotj, line 68: Let's use NewTLSToggledField instead: https://github.com/redpanda-data/benthos/blob/219c3a411197867e97648e5ae26eb864ac1235c5/public/service/config_tls.go#L59

that gives more control over custom certs, etc.

internal/impl/postgresql/input_postgrecdc.go
```diff
@@ -0,0 +1,562 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return conf.WrapBatchInputExtractTracingSpanMapping("pg_stream", r)
+}
+
+func init() {
+	rng, _ := codename.DefaultRNG()
+	randomSlotName = strings.ReplaceAll(codename.Generate(rng, 5), "-", "_")
```
rockwotj, line 289: Can we do this in `newPgStreamInput`? Otherwise we will use the same slot name if this is running in streams mode or there is a `broker` input that is reading from multiple databases.

internal/impl/postgresql/integration_test.go
```diff
@@ -0,0 +1,939 @@
...
+}
+
+func TestIntegrationPgCDC(t *testing.T) {
+	tmpDir := t.TempDir()
+	pool, err := dockertest.NewPool("")
+	require.NoError(t, err)
+
+	// Use custom PostgreSQL image with wal2json plugin compiled in
+	resource, err := pool.RunWithOptions(&dockertest.RunOptions{
+		Repository: "usedatabrew/pgwal2json",
```
rockwotj, line 109: Can we use https://hub.docker.com/r/debezium/postgres/dockerfile instead? I'm assuming you're not going to maintain writing new images as new version of postgres come out.

internal/impl/postgresql/input_postgrecdc.go
```diff
@@ -0,0 +1,562 @@
...
+		Password: dbPassword,
+	}
+
+	if tlsSetting == "none" {
+		pgconnConfig.TLSConfig = nil
+	}
+
+	snapsotMetrics := mgr.Metrics().NewGauge("snapshot_progress", "table")
+	replicationLag := mgr.Metrics().NewGauge("replication_lag_bytes")
+	snapshotMessageRate := mgr.Metrics().NewGauge("snapshot_message_rate")
```
rockwotj, line 247: Can we prefix these with `postgres_cdc` or something? So it's more clear which input it's for if you're looking at stats of many different instances.

rockwotj: Also for the message rate - I think it's preferred to just use a counter and let a stats engine like influxdb or prometheus compute the rate.

internal/impl/postgresql/input_postgrecdc.go
```diff
@@ -0,0 +1,562 @@
...
+				if err != nil {
+					p.mgr.Logger().Debugf("Timed flush batch error: %w", err)
+					break
+				}
+
+				if !flushBatch(ctx, p.msgChan, flushedBatch, latestOffset, nil) {
+					break
+				}
+
+				// TrxCommit LSN must be acked when all the bessages in the batch are processed
```
rockwotj, line 392: nit: `bessages` typo

internal/impl/postgresql/pglogicalstream/logical_stream.go
```diff
@@ -39,53 +35,72 @@ type Stream struct {
...
-		config.DbPort,
-		config.DbName,
+	connectionParams := ""
+	if config.PgConnRuntimeParam != "" {
+		connectionParams = "&" + config.PgConnRuntimeParam
+	}
+
+	q := fmt.Sprintf("postgres://%s:%s@%s:%d/%s?replication=database%s%s",
+		config.DBUser,
+		config.DBPassword,
```
rockwotj, line 88: You mentioned something about special chars in the database password being an issue - did that get fixed?

internal/impl/postgresql/pglogicalstream/logical_stream.go
```diff
@@ -96,85 +111,120 @@ func NewPgStream(config Config) (*Stream, error) {
...
+			"\"add-tables\"" + " " + fmt.Sprintf("'%s'", tablesFilterRule),
+		}
+	}
+
+	stream.decodingPluginArguments = pluginArguments
+
+	// create snapshot transaction before creating a slot for older PostgreSQL versions to ensure consistency
+
+	pubName := "pglog_stream_" + config.ReplicationSlotName
+	fmt.Println("Creating publication", pubName, "for tables", tableNames)
```
rockwotj, line 186: Can we replace all `fmt.Println` with `*service.Logger`?

internal/impl/postgresql/pglogicalstream/logical_stream.go
```diff
@@ -96,85 +111,120 @@ func NewPgStream(config Config) (*Stream, error) {
...
 			freshlyCreatedSlot = true
 		} else {
 			slotCheckRow := slotCheckResults[0].Rows[0]
 			confirmedLSNFromDB = string(slotCheckRow[0])
-			stream.logger.Infof("Replication slot restart LSN extracted from DB: LSN %v", confirmedLSNFromDB)
 		}
 	}
 
-	var lsnrestart pglogrepl.LSN
+	// TODO:: check decoding plugin and replication slot plugin should match
```
rockwotj, line 221: Still plans to do this?

internal/impl/postgresql/pglogicalstream/logical_stream.go
```diff
@@ -96,85 +111,120 @@ func NewPgStream(config Config) (*Stream, error) {
...
-	stream.logger.Infof("System identification result SystemID: %v Timeline: %v XLogPos: %v DBName: %v", sysident.SystemID, sysident.Timeline, sysident.XLogPos, sysident.DBName)
+	sysident, err := IdentifySystem(context.Background(), stream.pgConn)
+	if err != nil {
+		return nil, err
+	}
 
 	var freshlyCreatedSlot = false
 	var confirmedLSNFromDB string
 	// check is replication slot exist to get last restart SLN
 	connExecResult := stream.pgConn.Exec(context.TODO(), fmt.Sprintf("SELECT confirmed_flush_lsn FROM pg_replication_slots WHERE slot_name = '%s'", config.ReplicationSlotName))
```
rockwotj, line 199: Can we pass in a context here? There should be one from the input, same for the ones above.

internal/impl/postgresql/pglogicalstream/pglogrepl.go
```diff
@@ -289,6 +337,28 @@ func DropReplicationSlot(ctx context.Context, conn *pgconn.PgConn, slotName stri
 	return err
 }
 
+// CreatePublication creates a new PostgreSQL publication with the given name for a list of tables and drop if exists flag
+func CreatePublication(ctx context.Context, conn *pgconn.PgConn, publicationName string, tables []string, dropIfExist bool) error {
+	result := conn.Exec(context.Background(), fmt.Sprintf("DROP PUBLICATION IF EXISTS %s;", publicationName))
```
rockwotj, line 342: use the passed in context here and below.

rockwotj (commented):
This is a first pass, will do another tomorrow

rockwotj:
@le-vlad can we also rebase this off main instead of the other PR? I think we can just merge this directly into main

rockwotj (commented):
Okay thanks for rebasing the PR - I see a lot more of the diff and have been digging in more... High level feedback:

- We need to remove all the os.Exit calls
- Snapshotting assumes that tables have a primary key we can work on. We need to fail gracefully if there is no primary key and also mention this is required if we're processing snapshots in the documentation.
- More than passwords, there are other places where we have this special character issue because we're fmt.Sprintf creating DSN strings for config. We should just have the user give us a DSN string like our normal sql plugins and we can add the runtime parameter for replication=database after we parse. We'll need to Copy the parsed configuration because a lot of places delete the replication runtime parameter.
- We need to remove the string interpolation calls to create SQL statements. That's a SQL injection attack. Yes a CDC configuration file would be an odd place to dynamically populate, but we don't get to control what our users do.
    - Okay it looks like the replication setting only allows the simple query protocol (no extended protocol with parameters) so we need validation that the string there is only alphanumeric with - or _
- Ideally we only create contexts in the root input, and the pglogicalstream package never uses context.Background directly, but is always passed in a context.

Here are some commits that have some examples of addressing the above feedback or fixes bugs:

- https://github.com/redpanda-data/connect/commit/9ff7079b29089b216868e2b536c38c8464c60c63
- https://github.com/redpanda-data/connect/commit/2517627c355d716564488139bb0e4744fa9915fb
- https://github.com/redpanda-data/connect/commit/79d9f1faaac6c0c733b35206b159bd7cb757fb11

internal/impl/postgresql/input_postgrecdc.go
```diff
@@ -0,0 +1,566 @@
...
+					p.logger.Errorf("Failed to stop pglogical stream: %v", err)
+				}
+			}
+		}
+	}()
+
+	return err
+}
+
+func (p *pgStreamInput) asyncCheckpointer() func(context.Context, chan<- asyncMessage, service.MessageBatch, *int64, *chan bool) bool {
```
mihaitodor, line 491: Since this `asyncCheckpointer` is only used as the `flushBatch` func in `Connect()` which is always preceded by `callbackChan := make(chan bool)` and then followed by `<-callbackChan`, I wonder if we can get rid of this `callbackChan`.

internal/impl/postgresql/pglogicalstream/util.go
```diff
@@ -0,0 +1,56 @@
+// Copyright 2024 Redpanda Data, Inc.
```
mihaitodor, line 1: Nit: Guess this could be called `connection.go`

internal/impl/postgresql/pglogicalstream/util.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	var versionString string
+	err = conn.QueryRow("SHOW server_version").Scan(&versionString)
+	if err != nil {
+		return 0, fmt.Errorf("failed to execute query: %w", err)
+	}
+
+	// Extract the major version number
+	re := regexp.MustCompile(`^(\d+)`)
```
mihaitodor, line 44: Nit: I'd make this a global so it panics on start if the regex is invalid and it runs only once.

internal/impl/postgresql/input_postgrecdc.go
```diff
@@ -0,0 +1,566 @@
...
+	logger                  *service.Logger
+	mgr                     *service.Resources
+	metrics                 *service.Metrics
+	cMut                    sync.Mutex
+	msgChan                 chan asyncMessage
+	batching                service.BatchPolicy
+	checkpointLimit         int
+
+	snapshotRateCounter *RateCounter
+	snapshotMessageRate *service.MetricGauge
```
mihaitodor, line 331: Unused fields

internal/impl/postgresql/input_postgrecdc.go
```diff
@@ -0,0 +1,566 @@
...
+			case <-nextTimedBatchChan:
+				nextTimedBatchChan = nil
+				flushedBatch, err := batchPolicy.Flush(ctx)
+				if err != nil {
+					p.mgr.Logger().Debugf("Timed flush batch error: %w", err)
+					break
+				}
+
+				if !flushBatch(ctx, p.msgChan, flushedBatch, latestOffset, nil) {
+					break
```
mihaitodor, line 396: Ineffective `break` statement

internal/impl/postgresql/pglogicalstream/logical_stream.go
```diff
@@ -0,0 +1,802 @@
...
+
+	if err = dbConn.Ping(context.Background()); err != nil {
+		return nil, err
+	}
+
+	var tableNames []string
+	tableNames = append(tableNames, config.DBTables...)
+
+	if config.ReplicationSlotName == "" {
+		rng, _ := codename.DefaultRNG()
```
mihaitodor, line 113: Why are we ignoring the error?

go.mod
```diff
@@ -70,9 +70,12 @@ require (
 	github.com/gosimple/slug v1.14.0
 	github.com/influxdata/influxdb1-client v0.0.0-20220302092344-a9ab5670611c
 	github.com/jackc/pgx/v4 v4.18.3
+	github.com/jackc/pgx/v5 v5.6.0
+	github.com/jaswdr/faker v1.19.1
```
mihaitodor, line 74: We already include `github.com/go-faker/faker/v4`. Would be best to use that one instead.

internal/impl/postgresql/pglogicalstream/logical_stream.go
```diff
@@ -0,0 +1,802 @@
...
+
+	stream.lsnrestart = lsnrestart
+
+	if freshlyCreatedSlot {
+		stream.clientXLogPos = sysident.XLogPos
+	} else {
+		stream.clientXLogPos = lsnrestart
+	}
+
+	stream.standbyMessageTimeout = time.Second * 10
```
mihaitodor, line 236: Should this be bubbled up as a config parameter?

internal/impl/postgresql/pglogicalstream/monitor.go
```diff
@@ -0,0 +1,174 @@
...
+
+	if err = m.readTablesStat(tables); err != nil {
+		return nil, err
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	m.ctx = ctx
+	m.cancelTicker = cancel
+	// hardocded duration to monitor slot lag
+	m.ticker = time.NewTicker(time.Second * 3)
```
mihaitodor, line 73: Should this be bubbled up as a config parameter?

internal/impl/postgresql/pglogicalstream/consts.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// TLSVerify is a type for the TLS verification mode
+type TLSVerify string
+
+// TLSNoVerify is the value for no TLS verification
+const TLSNoVerify TLSVerify = "none"
+
+// TLSRequireVerify is the value for TLS verification with a CA
+const TLSRequireVerify TLSVerify = "require"
```
mihaitodor, line 43: Unused

mihaitodor (changes requested):
Hey @le-vlad, thank you so much for all the hard work! I finally got a chance to take a closer look at the code and left some comments, mostly nits. I'll have to take another pass through this once you get a chance to address them.

internal/impl/postgresql/input_postgrecdc.go
```diff
@@ -0,0 +1,566 @@
...
+	msgChan                 chan asyncMessage
+	batching                service.BatchPolicy
+	checkpointLimit         int
+
+	snapshotRateCounter *RateCounter
+	snapshotMessageRate *service.MetricGauge
+	snapshotMetrics     *service.MetricGauge
+	replicationLag      *service.MetricGauge
+
+	releaseTrxChan chan bool
```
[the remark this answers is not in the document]
mihaitodor: Oh, OK. Guess it's a Postgres abbreviation. Thanks!

internal/impl/postgresql/input_postgrecdc.go
mihaitodor: Sure, but it can be cleaned up a bit. The caller shouldn't have to manage that channel. Here's what I had in mind: https://github.com/redpanda-data/connect/commit/4e05248557f30c5df6b2b0ffd7cfc37a91612d41. I'd have to spend more time on it to see if it can be cleaned up more, but this should be a bit more readable I think.

internal/impl/postgresql/integration_test.go
```diff
@@ -0,0 +1,942 @@
...
+
+	template := fmt.Sprintf(`
+pg_stream:
+    host: %s
+    slot_name: test_slot_native_decoder
+    user: user_name
+    password: %s
+    port: %s
+    schema: public
+    tls: none
```
mihaitodor, line 327: This test fails because of this. Might be worth rechecking the other tests as well.

internal/impl/postgresql/utils.go
```diff
@@ -0,0 +1,81 @@
+package pgstream
```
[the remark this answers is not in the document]
mihaitodor: Sure, that's even better, thanks!

rockwotj:
@le-vlad please revert the commenting out of the integration test checks here: https://github.com/redpanda-data/connect/pull/2917/commits/fe18543338d19eaf5db1e3d3451893ecc4973ab5

internal/impl/postgresql/pglogicalstream/logical_stream.go
```diff
@@ -503,10 +503,12 @@ func (s *Stream) processSnapshot(ctx context.Context) error {
 				return err
 			}
 
+			var lastPkVal interface{}
```
rockwotj, line 506: nit: we prefer `any`

internal/impl/postgresql/input_pg_stream.go
```diff
@@ -0,0 +1,505 @@
...
+
+	go func() {
+		batchPolicy, err := p.batching.NewBatcher(p.mgr)
+		if err != nil {
+			p.logger.Errorf("Failed to initialise batch policy: %v, falling back to no policy.", err)
+			conf := service.BatchPolicy{Count: 1}
+			if batchPolicy, err = conf.NewBatcher(p.mgr); err != nil {
+				panic(err)
+			}
+		}
```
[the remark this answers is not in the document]
rockwotj: I addressed this old comment - can ignore

internal/impl/postgresql/pglogicalstream/debouncer.go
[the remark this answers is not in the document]
rockwotj: nice catch!

internal/impl/postgresql/pglogicalstream/connection.go
```diff
@@ -0,0 +1,47 @@
...
+
+func openPgConnectionFromConfig(dbDSN string) (*sql.DB, error) {
+	return sql.Open("postgres", dbDSN)
+}
+
+func getPostgresVersion(dbDSN string) (int, error) {
+	conn, err := openPgConnectionFromConfig(dbDSN)
+	if err != nil {
+		return 0, fmt.Errorf("failed to connect to the database: %w", err)
+	}
```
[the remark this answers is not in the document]
rockwotj: Very good spot! 🦅 👀

internal/impl/postgresql/pglogicalstream/logical_stream.go
```diff
@@ -0,0 +1,627 @@
...
+		lsnrestart = sysident.XLogPos
+	} else {
+		lsnrestart, _ = ParseLSN(confirmedLSNFromDB)
+	}
+
+	if freshlyCreatedSlot {
+		stream.clientXLogPos = watermark.New(sysident.XLogPos)
+	} else {
+		stream.clientXLogPos = watermark.New(lsnrestart)
+	}
```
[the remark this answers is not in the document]
rockwotj: Done

internal/impl/postgresql/input_pg_stream.go
```diff
@@ -0,0 +1,485 @@
...
+	}
+	return nil
+}
+
+func init() {
+	err := service.RegisterBatchInput(
+		"pg_stream", pgStreamConfigSpec,
+		func(conf *service.ParsedConfig, mgr *service.Resources) (service.BatchInput, error) {
+			return newPgStreamInput(conf, mgr)
+		})
```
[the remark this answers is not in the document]
rockwotj: Done

internal/impl/postgresql/pglogicalstream/monitor.go
```diff
@@ -0,0 +1,173 @@
...
+	// replicationLagInBytes is the replication lag in bytes measured by
+	// finding the difference between the latest LSN and the last confirmed LSN for the replication slot
+	replicationLagInBytes int64
+
+	dbConn       *sql.DB
+	slotName     string
+	logger       *service.Logger
+	ticker       *time.Ticker
+	cancelTicker context.CancelFunc
+	ctx          context.Context
```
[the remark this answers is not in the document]
rockwotj: Ah nice I have a little `periodic` utility for this I can swap out -it's doing the same thing here 🔥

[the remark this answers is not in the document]
rockwotj: I hadn't taken a close look at monitor before - thanks this is way cleaner and simpler now!

internal/impl/postgresql/pglogicalstream/sanitize/sanitize.go
```diff
@@ -358,3 +362,29 @@ func SQLQuery(sql string, args ...any) (string, error) {
...
+	}
+
+	// First character must be a letter or underscore
+	if !unicode.IsLetter(rune(name[0])) && name[0] != '_' {
+		return errors.New("identifier must start with a letter or underscore")
+	}
+
+	// Subsequent characters must be letters, numbers, underscores, or dots
+	for i, char := range name {
+		if !unicode.IsLetter(char) && !unicode.IsDigit(char) && char != '_' && char != '.' {
```
rockwotj, line 384: Isn't a dot the delimiter for schema.table? Any concerns there?

rockwotj (approved):

Jeffail (approved):
LGTM, can one of you stick an entry in the CHANGELOG.md pls?

rockwotj:
Done! I also snuck in another changelog entry

PR #2918, opened 2024-10-07
Refactor kafka components to reduce duplication

Jeffail:
I've done a broad refactor of the various kafka input/output components in order to:

1. Reduce the amount of duplicate field definitions and parsers
2. Reduce the amount of duplicate client code around applying those field definitions
3. Improve our mechanisms for sharing client and connection info across components

There's quite a lot going on here and it touches lots of the existing connectors. The integration tests run fine but we'll need to do some manual testing of the migrator components before merging/releasing.

One facet of this work is the idea of bringing out an implementation of a "consumer" and "producer" into a reusable type, this is important as we will have multiple plugins essentially doing the same thing (reading or writing benthos messages from/to kafka records) where the implementation is non-trivial (ordering, checkpointing, batching).

Eventually I want to have a solid mechanism for instantiating inputs and outputs that will be more user friendly for guaranteeing things like strict ordering. However, this is a complicated body of work and so it's important to first ensure that it will only need to be implemeted in one place, and that the codebase is tidy enough to reason about.


internal/impl/kafka/enterprise/redpanda_migrator_output.go
```diff
@@ -220,156 +158,36 @@ func NewRedpandaMigratorWriterFromConfig(conf *service.ParsedConfig, mgr *servic
...
-		}
-	}
-
-	if w.clientID, err = conf.FieldString("client_id"); err != nil {
-		return nil, err
-	}
-
-	if w.rackID, err = conf.FieldString("rack_id"); err != nil {
+	var err error
+	if w.recordConverter, err = kafka.NewFranzWriterFromConfig(conf, nil, nil); err != nil {
```
mihaitodor, line 162: The docs for `NewFranzWriterFromConfig` say:
> A closure function must be provided that is responsible for granting access to a connected client.

Might be worth updating that to mention it's optional.

internal/impl/kafka/franz_writer.go
```diff
@@ -0,0 +1,348 @@
...
+
+//------------------------------------------------------------------------------
+
+const (
+	rwFieldTopic     = "topic"
+	rwFieldKey       = "key"
+	rwFieldPartition = "partition"
+	rwFieldMetadata  = "metadata"
+	rwFieldTimestamp = "timestamp"
+)
```
mihaitodor, line 186: Should these be prefixed with `fw` (or `kfw` like the ones above) instead of `rw`?

internal/impl/kafka/output_kafka_franz.go
```diff
@@ -55,76 +56,20 @@ if this.partition.or("") == "" {
 // FranzKafkaOutputConfigFields returns the full suite of config fields for a
```
mihaitodor, line 56: Nit: The lint rule indentation above is a bit off.

mihaitodor (approved):
Stellar job on this refactor 🏆 Thank you for taking the time to clean up all of this! I left a few nits in the comments, but it looks great otherwise. Feel free to :shipit:

internal/impl/kafka/franz_client.go
```diff
@@ -0,0 +1,131 @@
...
+			Example([]string{"localhost:9092"}).
+			Example([]string{"foo:9092", "bar:9092"}).
+			Example([]string{"foo:9092,bar:9092"}),
+		service.NewStringField(kfcFieldClientID).
+			Description("An identifier for the client connection.").
+			Default("benthos").
+			Advanced(),
+		service.NewTLSToggledField(kfcFieldTLS),
+		SASLFields(),
+		service.NewDurationField(kfcFieldMetadataMaxAge).
```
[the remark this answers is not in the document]
rockwotj: Technically producers use metadata too. Metadata contains which brokers host which partitions, so if there are leadership changes or partitions move to partitions that are rarely written too it would prevent latency by discovering the change in the write. Most errors fallback to reading metadata and metadata refresh is mostly a tradeoff around extra traffic vs stale info. Anyways I doubt anyone would ever need to change this for an output practically.

internal/impl/kafka/enterprise/redpanda_output.go
```diff
@@ -0,0 +1,94 @@
...
+func redpandaOutputConfig() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Beta().
+		Categories("Services").
+		Summary("TODO").
+		Fields(kafka.FranzWriterConfigFields()...).
+		Fields(
+			service.NewIntField(roFieldMaxInFlight).
+				Description("The maximum number of batches to be sending in parallel at any given time.").
+				Default(10),
```
rockwotj, line 28: nit: could use `service.NewOutputMaxInFlightField()`

internal/impl/kafka/enterprise/redpanda_output.go
```diff
@@ -0,0 +1,94 @@
...
+
+func init() {
+	err := service.RegisterBatchOutput("redpanda", redpandaOutputConfig(),
+		func(conf *service.ParsedConfig, mgr *service.Resources) (
+			output service.BatchOutput,
+			batchPolicy service.BatchPolicy,
+			maxInFlight int,
+			err error,
+		) {
+			if maxInFlight, err = conf.FieldInt(roFieldMaxInFlight); err != nil {
```
rockwotj, line 80: nit: `conf.FieldMaxInFlight()`

internal/impl/kafka/enterprise/topic_logger.go
```diff
@@ -31,51 +28,34 @@ const (
...
+				Default("").
+				Example("__redpanda.connect.logs"),
+			service.NewStringEnumField("logs_level", "debug", "info", "warn", "error").
+				Default("info"),
+			service.NewStringField("status_topic").
+				Description("A topic to send status updates to.").
+				Default("").
+				Example("__redpanda.connect.status"),
+
+			// Deprecated
```
rockwotj, line 54: Should we mark it deprecated in the config too?

rockwotj (approved):
Legendary - thanks for cleaning this up! A couple nits for you.

internal/impl/kafka/franz_shared_client.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+//------------------------------------------------------------------------------
+
+type franzSharedClientKeyType int
+
+var franzSharedClientKey franzSharedClientKeyType
+
+func getSharedClientRegister(res *service.Resources) *franzSharedClientRegister {
+	reg, _ := res.GetOrSetGeneric(franzSharedClientKey, &franzSharedClientRegister{})
```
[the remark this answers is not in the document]
Jeffail: It's to avoid a map allocation every time we get a shared client handle. I'm not actually sure what the real life impact would be, it's probably negligible but I'd rather keep it as is for now, will add a comment though.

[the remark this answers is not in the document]
mihaitodor: Oh, I didn't think of that. Good point! ❤️

internal/impl/kafka/franz_reader_ordered.go
```diff
@@ -0,0 +1,511 @@
...
+		for _, p := range v {
+			if b := p.pop(); b != nil {
+				return b
+			}
+		}
+	}
+	return nil
+}
+
+func (c *partitionState) addRecords(ctx context.Context, topic string, partition int32, batch *batchWithRecords, bufferSize uint64) (pauseFetch bool) {
```
rockwotj, line 253: context unused

internal/impl/kafka/franz_reader_ordered.go
```diff
@@ -0,0 +1,511 @@
...
+	}
+	partTracker := topicTracker[partition]
+	if partTracker == nil {
+		return false
+	}
+
+	return partTracker.pauseFetch(limit)
+}
+
+func (c *partitionState) removeTopicPartitions(ctx context.Context, m map[string][]int32) {
```
rockwotj, line 291: context unused

internal/impl/kafka/franz_reader_unordered.go
```diff
@@ -0,0 +1,619 @@
...
+	batchPolicy     service.BatchPolicy
+
+	batchChan atomic.Value
+	res       *service.Resources
+	log       *service.Logger
+	shutSig   *shutdown.Signaller
+}
+
+func (f *FranzReaderUnordered) getBatchChan() chan batchWithAckFn {
+	c, _ := f.batchChan.Load().(chan batchWithAckFn)
```
rockwotj, line 96: I made a `typed.Atomic` package in `impl` if you want to hide the casts.

rockwotj (commented):
Made it though just the inputs - will go over the outputs tomorrow. Looks pretty sweeeettttt. I have a few small comments over small details

internal/dispatch/detect.go
```diff
@@ -0,0 +1,55 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dispatch
+
+import "context"
+
+// CtxOnTriggerSignal creates a context that is enriched with a closure function
+// which may be called by downstream components once any batch or transaction
+// associated with the context has been dispatched to an output.
```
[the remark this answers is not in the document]
rockwotj: Reading the input - yes exactly that's what this is (amazing!)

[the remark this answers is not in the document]
rockwotj: Is the message context used for anything else? I've mostly been ignoring that so far

internal/impl/kafka/output_redpanda.go
```diff
@@ -0,0 +1,115 @@
...
+	roFieldMaxInFlight = "max_in_flight"
+)
+
+func redpandaOutputConfig() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Beta().
+		Categories("Services").
+		Summary("A Kafka output using the https://github.com/twmb/franz-go[Franz Kafka client library^].").
+		Description(`
+Writes a batch of messages to Kafka brokers and waits for acknowledgement before propagating it back to the input.
```
rockwotj, line 36: The franz output talks about how it out performs the normal kafka output... Should we give guidance here on using Redpanda vs Franz outputs?

internal/impl/kafka/output_redpanda.go
```diff
@@ -0,0 +1,115 @@
...
+					Client:      client,
+					ConnDetails: connDetails,
+				})
+			}, func(context.Context) error {
+				if client == nil {
+					return nil
+				}
+				client.Close()
+				client = nil
+				return nil
```
rockwotj, line 108: This isn't safe because close can be called concurrently.

rockwotj: https://redpandadata.slack.com/archives/C076AV2M5MH/p1730827236296859

internal/impl/kafka/franz_writer.go
```diff
@@ -0,0 +1,384 @@
...
+		}
+		wg.Wait()
+
+		// TODO: This is very cool and allows us to easily return granular errors,
+		// so we should honor travis by doing it.
+		return results.FirstErr()
+	})
+}
+
+// Close does nothing as this component doesn't own the client.
```
rockwotj, line 381: Not quite true, it delegates to the provided yield fn

rockwotj (commented):
I think I found one sneaky thing - otherwise this LGTM

internal/impl/kafka/output_redpanda.go
Jeffail: I think once we're happy we deprecate the franz components entirely. It's not just performance it's also the general user experience that's improved (pending user feedback).

internal/impl/kafka/enterprise/redpanda_common_input.go
```diff
@@ -0,0 +1,107 @@
...
+    "this input does not support both regular expression topics and explicit topic partitions"
+  }
+}
+`)
+}
+
+func init() {
+	err := service.RegisterBatchInput("redpanda_common", redpandaCommonInputConfig(),
+		func(conf *service.ParsedConfig, mgr *service.Resources) (service.BatchInput, error) {
+			tmpOpts, err := kafka.FranzConnectionOptsFromConfig(conf, mgr.Logger())
```
mihaitodor, line 86: This fails if the config doesn't have `seed_brokers`

internal/impl/kafka/franz_client.go
```diff
@@ -0,0 +1,131 @@
...
+		return nil, err
+	}
+
+	if d.ClientID, err = conf.FieldString(kfcFieldClientID); err != nil {
+		return nil, err
+	}
+
+	if d.MetaMaxAge, err = conf.FieldDuration(kfcFieldMetadataMaxAge); err != nil {
+		return nil, err
+	}
```
mihaitodor, line 100: Should these not be optional? The `redpanda_common` input errors out if they're not set in the current implementation when running with `--chilled`.

internal/impl/kafka/output_redpanda.go
rockwotj: Oh you don't think the ordering will be a performance difference between the Redpanda and kafka_franz components? I thought we'd end up keeping both of those (but removing the old samara based one)

internal/impl/kafka/franz_reader_ordered.go
```diff
@@ -0,0 +1,511 @@
...
+
+		for {
+			// Using a stall prevention context here because I've realised we
+			// might end up disabling literally all the partitions and topics
+			// we're allocated.
+			//
+			// In this case we don't want to actually resume any of them yet so
+			// I add a forced timeout to deal with it.
+			stallCtx, pollDone := context.WithTimeout(closeCtx, time.Second)
+			fetches := cl.PollFetches(stallCtx)
```
mihaitodor, line 401: Not sure if this is an issue, but if we get a lot of backpressure, would it be possible to end up with a giant batch here? That was my reasoning when I used `PollRecords()` in the `redpanda_migrator` input, so I can pass in that `batch_size` parameter, which I probably should rename to `max_batch_size`.

internal/impl/kafka/enterprise/redpanda_common_output.go
```diff
@@ -0,0 +1,92 @@
...
+        root.id = uuid_v4()
+        root.loud_name = this.name.uppercase()
+
+output:
+  redpanda_common:
+    topic: bar
+    key: ${! @id }
+
+redpanda:
+  seed_brokers: [ "127.0.0.1:9093" ]
```
mihaitodor, line 56: `127.0.0.1:9092`?

mihaitodor (commented):
Thanks for all this work @Jeffail! I left a few small config issues and some nitpicks.

internal/impl/kafka/output_redpanda.go
rockwotj: nvm thats for a processor, outputs are fine

internal/impl/kafka/franz_client.go
Jeffail: Whoops, yeah the redpanda_common input was totally broken, should be fixed now

internal/dispatch/detect.go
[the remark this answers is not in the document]
Jeffail: Only a few other niche use cases, there's a couple of places where its used to take a scrambled batch and reorder it according to the original order.

internal/impl/kafka/franz_reader.go
```diff
@@ -0,0 +1,263 @@
...
+
+// FranzRecordToMessageV1 converts a record into a service.Message, adding
+// metadata and other relevant information.
+func FranzRecordToMessageV1(record *kgo.Record) *service.Message {
+	msg := service.NewMessage(record.Value)
+	msg.MetaSetMut("kafka_key", record.Key)
+	msg.MetaSetMut("kafka_topic", record.Topic)
+	msg.MetaSetMut("kafka_partition", int(record.Partition))
+	msg.MetaSetMut("kafka_offset", int(record.Offset))
+	msg.MetaSetMut("kafka_timestamp_unix", record.Timestamp.Unix())
```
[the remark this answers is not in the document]
Jeffail: I've added `kafka_timestamp` which is going to be an RFC3339 formatted string with nanosecond precision seconds. It's a bit of overkill but then users can do whatever they want with it.

internal/impl/kafka/franz_reader_ordered.go
Jeffail: I'm not fully clued in to the mechanisms at play here, but I'm relying on the fetch max bytes, and fetch max partition bytes, fields for tuning the batch size. I'm definitely seeing huge batches in my test pipelines but benthos is actually handling it better than smaller batches.

rockwotj: https://github.com/twmb/franz-go/blob/cea7aa5d803781e5f0162187795482ba1990c729/pkg/kgo/consumer.go#L401

`PollFetches` is just `PollRecords` but you get everything that is currently buffered. Relying on the configs for franz-go internal buffering is the way to go.

internal/impl/kafka/franz_reader.go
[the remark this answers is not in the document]
mihaitodor: Sweet, thanks!

internal/impl/kafka/franz_reader_ordered.go
mihaitodor: Sounds good to me! Thank you!

mihaitodor (approved):
🐑 🚀

PR #2921, opened 2024-10-08
feat: new configuration parameters for input: kafka_franz

darkxeno:
This PR enables the following configuration parameters to be used for the input kafka_franz:

**session_timeout**: "Sets how long a member in the group can go between heartbeats. If a member does not heartbeat in this timeout, the broker will remove the member from the group and initiate a rebalance." 
defaults to 45 secs

**heartbeat_period** "Sets how long a group member goes between heartbeats to Kafka. Kafka uses heartbeats to ensure that a group member's session stays active. This value can be any value lower than the session timeout, but should be no higher than 1/3rd the session timeout." 
defaults to 3 secs

**rebalance_timeout** "Sets how long group members are allowed to take when a rebalance has begun. This timeout is how long all members are allowed to complete work and commit offsets, minus the time it took to detect the rebalance (from a heartbeat)."
defaults to 60 secs

Default values are based on frank-go defaults:
https://github.com/twmb/franz-go/blob/b77dd13e2bfaee7f5181df27b40ee4a4f6a73b09/pkg/kgo/config.go#L1448
https://github.com/twmb/franz-go/blob/b77dd13e2bfaee7f5181df27b40ee4a4f6a73b09/pkg/kgo/config.go#L1467
https://github.com/twmb/franz-go/blob/b77dd13e2bfaee7f5181df27b40ee4a4f6a73b09/pkg/kgo/config.go#L1481

Supporting these parameters helps to support workloads where processing 1 single message might take longer than 1 or 2 minutes, as longer session timeouts with longer heartbeat intervals helps to avoid or reduce rebalances.

### DIFF
--- a/internal/impl/kafka/input_kafka_franz.go
+++ b/internal/impl/kafka/input_kafka_franz.go
@@ -116,6 +116,18 @@ Finally, it's also possible to specify an explicit offset to consume from by add
 			Description("The period of time between each commit of the current partition offsets. Offsets are always committed during shutdown.").
 			Default("5s").
 			Advanced(),
+		service.NewDurationField("session_timeout").
+			Description("Sets how long a member in the group can go between heartbeats. If a member does not heartbeat in this timeout, the broker will remove the member from the group and initiate a rebalance.").
+			Default("45s").
+			Advanced(),
+		service.NewDurationField("heartbeat_period").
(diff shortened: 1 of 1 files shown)


rockwotj:
Hey @darkxeno thanks for this! I ended up doing this in a separate PR. The next release will have this functionality available.

PR #2924, opened 2024-10-09
feat: add Azure Data Lake Gen2 Output

ooesili:
This PR adds an output plugin for [Azure's Data Lake Storage Gen2](https://learn.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-introduction).

## Notes
- Unfortunately at the time of this change, [Azurite does not yet support the Data Lake Gen2 APIs](https://github.com/Azure/Azurite/issues/553), so we can't use the integration test pattern that the blob storage plugin also uses. I could write an integration test that interacts with Azure, but I wasn't sure if that would be worth it.

### DIFF
--- a/internal/impl/azure/output_adls.go
+++ b/internal/impl/azure/output_adls.go
@@ -1,0 +1,158 @@
+// Copyright 2024 Redpanda Data, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package azure
+
+import (
+	"context"
+	"fmt"
+
+	dlservice "github.com/Azure/azure-sdk-for-go/sdk/storage/azdatalake/service"
+	"github.com/redpanda-data/benthos/v4/public/service"
+)
+
+const (
+	// Azure Data Lake Storage Output Fields
+	adlsFieldFilesystem = "filesystem"
+	adlsFieldPath       = "path"
+)
+
+type adlsConfig struct {
+	client     *dlservice.Client
+	path       *service.InterpolatedString
+	filesystem *service.InterpolatedString
+}
+
+func init() {
+	err := service.RegisterOutput("azure_data_lake_gen2", adlsSpec(),
+		func(conf *service.ParsedConfig, mgr *service.Resources) (out service.Output, mif int, err error) {
+			var pConf *adlsConfig
+			if pConf, err = adlsConfigFromParsed(conf); err != nil {
+				return
+			}
+			if mif, err = conf.FieldMaxInFlight(); err != nil {
+				return
+			}
+			if out, err = newAzureADLSWriter(pConf, mgr.Logger()); err != nil {
+				return
+			}
+			return
+		})
+	if err != nil {
+		panic(err)
+	}
+}
+
+func adlsConfigFromParsed(pConf *service.ParsedConfig) (*adlsConfig, error) {
+	var conf adlsConfig
+	var err error
+	conf.filesystem, err = pConf.FieldInterpolatedString(adlsFieldFilesystem)
+	if err != nil {
+		return nil, err
+	}
+	conf.path, err = pConf.FieldInterpolatedString(adlsFieldPath)
+	if err != nil {
+		return nil, err
+	}
+	var isFilesystemSASToken bool
+	conf.client, isFilesystemSASToken, err = adlsClientFromParsed(pConf, conf.filesystem)
+	if err != nil {
+		return nil, err
+	}
+	if isFilesystemSASToken {
+		conf.filesystem, _ = service.NewInterpolatedString("")
+	}
+	return &conf, nil
+}
+
+func newAzureADLSWriter(conf *adlsConfig, log *service.Logger) (*azureADLSWriter, error) {
+	return &azureADLSWriter{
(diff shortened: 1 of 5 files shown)


internal/plugins/info.csv
```diff
@@ -28,6 +28,7 @@ azure_blob_storage        ,output    ,azure_blob_storage        ,3.36.0  ,certif
 azure_cosmosdb            ,input     ,azure_cosmosdb            ,4.25.0  ,certified  ,n          ,y     ,y
 azure_cosmosdb            ,output    ,azure_cosmosdb            ,4.25.0  ,certified  ,n          ,y     ,y
 azure_cosmosdb            ,processor ,azure_cosmosdb            ,4.25.0  ,certified  ,n          ,y     ,y
+azure_data_lake_gen2      ,output    ,azure_data_lake_gen2      ,4.37.0  ,certified  ,n          ,n     ,n
```
Jeffail, line 31:
```suggestion
azure_data_lake_gen2      ,output    ,azure_data_lake_gen2      ,4.38.0  ,certified  ,n          ,y     ,y
```

Jeffail (approved):
LGTM, thanks @ooesili! Do you think it's worth more time to try and expand this with tests or are you ready for a merge?

internal/impl/azure/output_adls.go
```diff
@@ -0,0 +1,158 @@
...
+	"fmt"
+
+	dlservice "github.com/Azure/azure-sdk-for-go/sdk/storage/azdatalake/service"
+	"github.com/redpanda-data/benthos/v4/public/service"
+)
+
+const (
+	// Azure Data Lake Storage Output Fields
+	adlsFieldFilesystem = "filesystem"
+	adlsFieldPath       = "path"
```
mihaitodor, line 28: Not sure what `s` in `adls` stands for. The prefix here should be something like `adlo`, where `o` stands for output in case we'll want to add a matching input in the future.

internal/impl/azure/output_adls.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		return nil, err
+	}
+	var isFilesystemSASToken bool
+	conf.client, isFilesystemSASToken, err = adlsClientFromParsed(pConf, conf.filesystem)
+	if err != nil {
+		return nil, err
+	}
+	if isFilesystemSASToken {
+		conf.filesystem, _ = service.NewInterpolatedString("")
```
mihaitodor, line 74: Shouldn't this be `service.NewInterpolatedString(adlsFieldFilesystem)` and have error handling like the rest?

internal/impl/azure/output_adls.go
```diff
@@ -0,0 +1,158 @@
...
+		return fmt.Errorf("uploading message body: %w", err)
+	}
+	return nil
+}
+
+func (a *azureADLSWriter) Close(ctx context.Context) error {
+	return nil
+}
+
+func adlsSpec() *service.ConfigSpec {
```
mihaitodor, line 125: Small nit: We usually place the spec towards the upper part of the file, but no worries if you wish to deviate from that pattern.

internal/impl/azure/output_adls.go
```diff
@@ -0,0 +1,158 @@
...
+	return azureComponentSpec(true).
+		Beta().
+		Version("4.37.0").
```
mihaitodor, line 128: `4.38.0`

mihaitodor (changes requested):
Nice job @ooesili! Thanks for putting this together! I left a few nits, but it looks great overall. Don't forget to update the [Changelog](https://github.com/redpanda-data/connect/blob/main/CHANGELOG.md) :)

internal/impl/azure/output_adls.go
ooesili: It stands for "Azure Data Lake Storage"

internal/impl/azure/output_adls.go
ooesili: I cargo culted this from the `azure_blob_storage` output.

https://github.com/redpanda-data/connect/blob/9e67ce7a8b85eb150cab6afa81eb9db0dc06f541/internal/impl/azure/output_blob_storage.go#L56-L57

mihaitodor: Thanks for looking into it! I forgot about that... If it works the same way for Data Lake, then it's fine.

internal/impl/azure/output_adls.go
mihaitodor: Ohhh, OK. `dlo` is fine, thanks! It's not super-consistent yet :)

mihaitodor (approved):
LGTM 🐑 🚀

internal/impl/azure/output_data_lake.go
```diff
@@ -0,0 +1,159 @@
...
+
+If the `+"`storage_connection_string`"+` does not contain the `+"`AccountName`"+` parameter, please specify it in the
+`+"`storage_account`"+` field.`+service.OutputPerformanceDocs(true, false)).
+		Fields(
+			service.NewInterpolatedStringField(dloFieldFilesystem).
+				Description("The data lake storage filesystem name for uploading the messages to.").
+				Example(`messages-${!timestamp("2006")}`),
+			service.NewInterpolatedStringField(dloFieldPath).
+				Description("The path of each message to upload within the filesystem.").
+				Example(`${!count("files")}-${!timestamp_unix_nano()}.json`).
```
Jeffail, line 52:
```suggestion
				Example(`${!counter()}-${!timestamp_unix_nano()}.json`).
```

internal/impl/azure/output_data_lake.go
```diff
@@ -0,0 +1,159 @@
...
+				Example(`${!meta("kafka_key")}.json`).
+				Example(`${!json("doc.namespace")}/${!json("doc.id")}.json`).
+				Default(`${!count("files")}-${!timestamp_unix_nano()}.txt`),
```
Jeffail, line 55:
```suggestion
				Default(`${!counter()}-${!timestamp_unix_nano()}.txt`),
```

PR #2936, opened 2024-10-15
Remove count references in favour of counter

Jeffail:
This is _technically_ a breaking change, as the behaviour of a default path previously using `count` would potentially conflict with another `count` under the same name and they would cascade their counts. However, given that `count` has been long deprecated, and that the "ideal" behaviour here would be to have no default `path` value at all, this is a good middle ground before we can eventually (in a V5 release) remove all the path defaults entirely.

### DIFF
--- a/internal/impl/azure/integration_test.go
+++ b/internal/impl/azure/integration_test.go
@@ -95,7 +95,7 @@ output:
     blob_type: BLOCK
     container: $VAR1-$ID
     max_in_flight: 1
-    path: $VAR2/${!count("$ID")}.txt
+    path: $VAR2/${!counter()}.txt
     public_access_level: PRIVATE
     storage_connection_string: $VAR3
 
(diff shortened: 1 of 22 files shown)


internal/impl/azure/integration_test.go
```diff
@@ -123,7 +123,7 @@ output:
     blob_type: BLOCK
     container: $VAR1-$ID
     max_in_flight: 1
-    path: $VAR2/${!count("$ID")}.txt
+    path: $VAR2/${!counter()}.txt
```
mihaitodor, line 126: It probably doesn't make a difference here, but I guess in such cases where one wants a new counter for every new value of `$ID`, they should use the old `count()`.

mihaitodor (approved):
LGTM :shipit: 🏆

internal/impl/azure/integration_test.go
Jeffail: It actually doesn't change anything (I think) because the config is instantiated once per test anyway, so each counter will still be unique to the ID. I can't really remember when these tests were written though, but they pass so I'm over it 😅

PR #2943, opened 2024-10-17
cloud: enable http client

nicolaferraro:
Cloud will have the right level of fencing for this to be enabled in next release.


Jeffail (approved):
LGTM, thanks @nicolaferraro!

PR #2947, opened 2024-10-23
feat: added timeplus input

ye11ow:
Similar to #2900, this PR added Timeplus as an input. It supports both HTTP and TCP protocol.

There is no test in this PR yet. I'll try to prepare a ready-to-use docker image later so that I can add integration test like other input/output.


Jeffail (approved):
LGTM, thanks @ye11ow!

PR #2949, opened 2024-10-23
Add schema id translation

mihaitodor:
~This requires merging #2918 first.~ Sorted!

TODO:
- [x] Unit tests
- [x] Integration tests


internal/impl/confluent/sr/serde.go
```diff
@@ -0,0 +1,43 @@
...
+	return int(binary.BigEndian.Uint32(msg[1:5])), nil
+}
+
+// UpdateID updates the schema ID in a raw message.
+func UpdateID(msg []byte, id int) error {
+	if len(msg) < 5 {
+		return errors.New("message is empty or too small")
+	}
+
+	binary.BigEndian.PutUint32(msg, uint32(id))
```
rockwotj, line 40: off by one :)

mihaitodor: Well spotted, thank you! ❤️ I fixed it now.

internal/impl/confluent/sr/client.go
```diff
@@ -275,68 +247,3 @@ func (c *Client) walkReferencesTracked(ctx context.Context, seen map[string]int,
...
-	headerKey := "Accept"
-	if verb == http.MethodPost {
-		headerKey = "Content-Type"
-	}
-	req.Header.Add(headerKey, "application/vnd.schemaregistry.v1+json")
-	if err = c.requestSigner(c.mgr.FS(), req); err != nil {
-		return
-	}
-
-	for i := 0; i < 3; i++ {
```
mihaitodor, line 313: Note that the `github.com/twmb/franz-go/pkg/sr` client doesn't do 3 retries under the hood. I can try to shoehorn this behaviour into the custom `roundTripper` if you think it's important to preserve it.

internal/impl/confluent/sr/client.go
```diff
@@ -15,33 +15,45 @@
...
 	mgr                   *service.Resources
 }
 
+type roundTripper struct {
+	reqSigner func(req *http.Request) error
+	*http.Transport
+}
+
+func (rt *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	// This is naughty, but it's probably fine...
```
rockwotj, line 47: Is Travis up for allowing some middleware option that intercepts the http request properly?

internal/impl/kafka/enterprise/schema_registry_input.go
```diff
@@ -166,6 +160,12 @@ func inputFromParsed(pConf *service.ParsedConfig, mgr *service.Resources) (i *sc
 		return nil, fmt.Errorf("failed to create Schema Registry client: %s", err)
 	}
 
+	if label := mgr.Label(); label != "" {
+		mgr.SetGeneric(mgr.Label(), i)
```
rockwotj, line 164: instead of a raw string can we use some typed wrapper so it's not possible for another component to clobber this? Then we don't need to do they empty string special case.

internal/impl/kafka/enterprise/schema_registry_output.go
```diff
@@ -183,3 +234,129 @@ func (o *schemaRegistryOutput) Close(_ context.Context) error {
...
+}
+
+// createSchema creates and caches the provided schema.
+func (o *schemaRegistryOutput) createSchema(ctx context.Context, key schemaLineageCacheKey, ss franz_sr.SubjectSchema) (int, error) {
+	if destinationID, ok := o.schemaLineageCache.Load(key); ok {
+		return destinationID.(int), nil
+	}
+
+	// TODO: Use `CreateSchemaWithID()` when `translate_ids: false` after https://github.com/twmb/franz-go/pull/849
+	// is merged.
```
rockwotj, line 350: Are we waiting on this before we merge?

internal/impl/confluent/sr/client.go
mihaitodor: I did ask, but he hasn't gotten around to it yet... Not sure if we should block this PR until that happens, but yeah, it would be great to avoid doing this.

internal/impl/kafka/enterprise/schema_registry_output.go
mihaitodor: I think it's fine to leave it as a TODO for now. The main use case is to do the ID translation.

rockwotj (approved):
outside of my comment about the typed key in resources this LGTM

CHANGELOG.md
```diff
@@ -17,6 +28,7 @@ All notable changes to this project will be documented in this file.
 - New `redpanda_common` input and output. (@Jeffail)
 - New `redpanda` input and output. (@Jeffail)
 - New `snowflake_streaming` output. (@rockwotj)
+- Field `metadata_max_age` added to the `redpanda_migrator_offsets` output. (@mihaitodor)
```
Jeffail, line 31: Is this correctly placed under v4.39?

internal/impl/confluent/sr/client.go
Jeffail: Nah I think this is fine, the concept of attempts is abstract and users don't have any existing documentation or control so I think it's okay to change the behaviour slightly here.

CHANGELOG.md
mihaitodor: Oops, thank you! ❤️

PR #2956, opened 2024-10-25
Add allow/deny list support

Jeffail:
This PR adds the ability for connect instances to check a file at the path `/etc/redpanda/connector_list.yaml`, and if so it'll use it to determine which connectors the instance will run. The format must look like this:

```yaml
allow:
  - foo
  - bar
```

Or this:

```yaml
deny:
  - foo
  - bar
```

You're not allowed to specify both an `allow` and `deny` list.

### DIFF
--- a/internal/cli/components_list.go
+++ b/internal/cli/components_list.go
@@ -1,0 +1,59 @@
+// Copyright 2024 Redpanda Data, Inc.
+//
+// Licensed as a Redpanda Enterprise file under the Redpanda Community
+// License (the "License"); you may not use this file except in compliance with
+// the License. You may obtain a copy of the License at
+//
+// https://github.com/redpanda-data/connect/blob/main/licenses/rcl.md
+
+package cli
+
+import (
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/redpanda-data/benthos/v4/public/service"
+	"gopkg.in/yaml.v2"
+)
+
+const connectorListPath = "/etc/redpanda/connector_list.yaml"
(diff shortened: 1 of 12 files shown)


mihaitodor (approved):
LGTM :shipit: Just a small nit

internal/cli/components_list.go
```diff
@@ -0,0 +1,59 @@
...
+	"fmt"
+	"os"
+
+	"github.com/redpanda-data/benthos/v4/public/service"
+	"gopkg.in/yaml.v2"
+)
+
+const connectorListPath = "/etc/redpanda/connector_list.yaml"
+
+type connectorsList struct {
```
emaxerrno, line 22: For ai models, companies need specific approved models. Last is there a way to remove all. So deny all except some. Like a deny star or if file only contains allow, rest is deny? I think ip table rules here figure out the right ux as example

Jeffail: Yeah if they specify anything within `allow` then everything else is denied.

internal/cli/components_list.go
```diff
@@ -0,0 +1,59 @@
...
+
+package cli
+
+import (
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/redpanda-data/benthos/v4/public/service"
+	"gopkg.in/yaml.v2"
```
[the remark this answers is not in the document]
Jeffail: Well spotted!

Jeffail:
Added some unit tests.

PR #2961, opened 2024-10-25
fix(mqtt): improve docs for `urls` field

ooesili:


rockwotj (commented):
LGTM - you need to run make docs

PR #2964, opened 2024-10-28
Add cli flag for secrets backends

Jeffail:
This PRs adds the ability for us to add backends for accessing secrets during config parse time, where interpolation functions of the form `${FOO}` will prompt a secret lookup.

A new cli flag `--secrets` is being added, which can be specified any number of times. Each entry adds a mechanism for resolving interpolation lookups via a secrets management solution (defined entirely as a single URN, e.g. `redis://foo:bar@example.com/`). Each specified entry will be attempted until the lookup is successful.

By default there is a single entry that enables environment variable lookups with the urn `env:`, this preserves the current behaviour. A user can disable all lookups by specifying a single entry with the urn `none:`.

If none of the specified secrets backends yield a given key lookup then we will exit the config parse as usual, with an error indicating that an environment variable is referenced in the config but has not been found.

### DIFF
--- a/internal/cli/enterprise.go
+++ b/internal/cli/enterprise.go
@@ -107,6 +107,11 @@ func InitEnterpriseCLI(binaryName, version, dateBuilt string, schema *service.Co
 				secretLookupFn, err = secrets.ParseLookupURNs(c.Context, slog.New(rpLogger), disableEnvLookup, secretsURNs...)
 				return err
 			}
+			if disableEnvLookup {
+				secretLookupFn = func(context.Context, string) (string, bool) {
+					return "", false
+				}
+			}
 			return nil
 		}),
 		service.CLIOptSetEnvVarLookup(func(ctx context.Context, key string) (string, bool) {
--- a/internal/cli/enterprise.go
+++ b/internal/cli/enterprise.go
@@ -106,8 +106,7 @@ func InitEnterpriseCLI(binaryName, version, dateBuilt string, schema *service.Co
 				var err error
 				secretLookupFn, err = secrets.ParseLookupURNs(c.Context, slog.New(rpLogger), disableEnvLookup, secretsURNs...)
 				return err
-			}
-			if disableEnvLookup {
+			} else if disableEnvLookup {
 				secretLookupFn = func(context.Context, string) (string, bool) {
 					return "", false
 				}
--- a/internal/cli/enterprise.go
+++ b/internal/cli/enterprise.go
@@ -16,8 +16,10 @@ import (
 
 	"github.com/redpanda-data/benthos/v4/public/service"
 	"github.com/rs/xid"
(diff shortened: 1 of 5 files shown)


internal/cli/enterprise.go
```diff
@@ -82,6 +88,35 @@ func InitEnterpriseCLI(binaryName, version, dateBuilt string, schema *service.Co
...
+		// Secrets management
+		service.CLIOptCustomRunFlags([]cli.Flag{
+			&cli.StringSliceFlag{
+				Name:  "secrets",
+				Usage: "Attempt to load secrets from a provided URN. If a referenced secret isn't identified using the provided method then environment variables will also be looked up, you can disable this behaviour with the --disable-env-lookup flag.",
+			},
+			&cli.BoolFlag{
+				Name:  "disable-env-lookup",
+				Usage: "Disable the ability for configs to interpolate environment variables with ${FOO} syntax",
+			},
```
rockwotj, line 101: Should the env lookup just be the default for the above flag?

internal/cli/enterprise.go
```diff
@@ -82,6 +88,35 @@ func InitEnterpriseCLI(binaryName, version, dateBuilt string, schema *service.Co
...
+		}, func(c *cli.Context) error {
+			disableEnvLookup := c.Bool("disable-env-lookup")
+			secretsURNs := c.StringSlice("secrets")
+			if len(secretsURNs) > 0 {
+				var err error
+				secretLookupFn, err = secrets.ParseLookupURNs(c.Context, slog.New(rpLogger), disableEnvLookup, secretsURNs...)
+				return err
+			}
+			if disableEnvLookup {
```
rockwotj, line 110: should this be an else if?

Jeffail: Whoops, yep, thanks 😅

internal/cli/enterprise.go
rockwotj: You could have a "disabled" backend that does nothing for an easy fallback. Also support a base scheme or the empty scheme so it's a little more natural and you can do --secrets=disabled (or --secrets=none)

This is also fine too, I just generally prefer less knobs (it's harder to forget one that way, where a missing env var will explode noisily).

rockwotj (approved):
LGTM, happy to take another look if I convinced you to only add one flag

internal/cli/enterprise.go
Jeffail: Got a branch to try it out. The only awkwardness seems to be that if you want to override the default behaviour and disable env lookups you need to do something like `rpk connect run --secrets none: ./foo.yaml`. I think that's probably fine though as I doubt many people would want to do this.

Jeffail: Yeah I think this works better, pushed the change and updated the description to match.

rockwotj (approved):

Issue #2955, opened 2024-10-25
Enhance the SFTP input component to provide modification time as metadata


prashant-patil-tb:
Enhance the SFTP input component to provide modification time as metadata.

Right now we just get file_path as metadata.

danriedl:
@mihaitodor please assign me to the issue.

anthonyvitale:
@danriedl what’s left to do on this? Mind if I take it over?

mihaitodor:
@anthonyvitale I'm happy to review a new PR. There was a comment in @danriedl's PR which I don't think got addressed.

anthonyvitale:
@mihaitodor when you get a chance to review, I threw something together in #3358

PR #2970, opened 2024-10-30
feat(sftp): add modtime as metadata. [#2955]

danriedl:
This PR adds:
- feat(sftp): add modtime as metadata. [#2955]

### DIFF
--- a/internal/impl/sftp/input.go
+++ b/internal/impl/sftp/input.go
@@ -53,6 +53,7 @@ func sftpInputSpec() *service.ConfigSpec {
 This input adds the following metadata fields to each message:
 
 - sftp_path
+- sftp_modification_time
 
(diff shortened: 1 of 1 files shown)


Jeffail (commented):
Hey @danriedl, this could end up with quite a large performance impact when there's many messages coming from individual files (unless there's some underlying caching during the stat calls). It's probably better to extract this info in the same place where we capture the `currentPath`, so that it's only performed once per file.

If we add more information over time it'll start looking a bit messy having all of these fields added to the `sftpReader` type so it'd also be good to add some sort of `currentFileInfo` struct that carries all of this info.

mihaitodor:
This metadata will be added via #3498 which should get merged soon.

PR #2972, opened 2024-10-30
linter: fix common mistakes at lint time

nicolaferraro:
This to give feedback before the pipeline is run and crashes at runtime.

From Kafka and migrator, the config without consumer group was not always working (unless the consumer group was explicitly set to "").


Jeffail (approved):
LGTM, thanks @nicolaferraro!

PR #2973, opened 2024-10-31
kafka: output perf improvement with large batches

rockwotj:
I saw TryInterpolatedString showing up in profiles even if the topic was
a static string... Then I realized we should be using the new batch
executor to prevent N^2 allocations in batches. This fixes that for both
kafka outputs.


Jeffail (approved):
Hero, thanks @rockwotj!

PR #2975, opened 2024-10-31
Port @rockwotj's fix from #2973.

mihaitodor:
Thanks Tyler, nice catch! <3

### DIFF
--- a/internal/impl/kafka/enterprise/redpanda_migrator_output.go
+++ b/internal/impl/kafka/enterprise/redpanda_migrator_output.go
@@ -433,10 +433,24 @@ func (w *RedpandaMigratorWriter) WriteBatch(ctx context.Context, b service.Messa
 		return service.ErrNotConnected
 	}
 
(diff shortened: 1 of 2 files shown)


internal/impl/kafka/enterprise/redpanda_migrator_output.go
```diff
@@ -471,12 +485,12 @@ func (w *RedpandaMigratorWriter) WriteBatch(ctx context.Context, b service.Messa
 			return
 		}
 		if w.key != nil {
```
rockwotj, line 487: Is it better to change this check to look at `keyExecutor != nil`? Same in other places.

rockwotj (approved):

internal/impl/kafka/enterprise/redpanda_migrator_output.go
mihaitodor: Thank you! Yeah... Guess this doesn't matter much since it will get dropped once #2918 is merged.

rockwotj (approved):

PR #2978, opened 2024-11-01
snowflake: optimize performance of common data paths

rockwotj:
Found from benchmarking some example workloads. This change allowed me to push about 2.5K records/sec per core *more*.

### DIFF
--- a/internal/impl/snowflake/streaming/parquet.go
+++ b/internal/impl/snowflake/streaming/parquet.go
@@ -56,7 +56,7 @@ func constructRowGroup(
 	batch service.MessageBatch,
 	schema *parquet.Schema,
 	transformers []*dataTransformer,
-) ([]parquet.Row, error) {
+) ([]parquet.Row, []*statsBuffer, error) {
 	// We write all of our data in a columnar fashion, but need to pivot that data so that we can feed it into
 	// out parquet library (which sadly will redo the pivot - maybe we need a lower level abstraction...).
 	// So create a massive matrix that we will write stuff in columnar form, but then we don't need to move any
@@ -66,13 +66,14 @@ func constructRowGroup(
 	rowWidth := len(schema.Fields())
 	matrix := make([]parquet.Value, len(batch)*rowWidth)
 	nameToPosition := make(map[string]int, rowWidth)
+	stats := make([]*statsBuffer, rowWidth)
 	for idx, t := range transformers {
 		leaf, ok := schema.Lookup(t.name)
 		if !ok {
-			return nil, fmt.Errorf("invariant failed: unable to find column %q", t.name)
+			return nil, nil, fmt.Errorf("invariant failed: unable to find column %q", t.name)
 		}
 		t.buf.Prepare(matrix, leaf.ColumnIndex, rowWidth)
-		t.stats.Reset()
+		stats[idx] = &statsBuffer{}
 		nameToPosition[t.name] = idx
 	}
 	// First we need to shred our record into columns, snowflake's data model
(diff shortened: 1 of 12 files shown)


internal/impl/snowflake/streaming/parquet.go
```diff
@@ -21,20 +21,19 @@ import (
...
 	}
 	row, ok := v.(map[string]any)
 	if !ok {
-		return nil, fmt.Errorf("expected object, got: %T", v)
+		return fmt.Errorf("expected object, got: %T", v)
 	}
-	mapped := make(map[string]any, len(row))
 	for k, v := range row {
-		mapped[normalizeColumnName(k)] = v
+		out[normalizeColumnName(k)] = v
```
rockwotj, line 34: use a map to write to an array instead

internal/impl/snowflake/streaming/parquet.go
```diff
@@ -11,76 +11,86 @@
...
 	}
 	row, ok := v.(map[string]any)
 	if !ok {
-		return nil, fmt.Errorf("expected object, got: %T", v)
+		return fmt.Errorf("expected object, got: %T", v)
 	}
-	mapped := make(map[string]any, len(row))
 	for k, v := range row {
-		mapped[normalizeColumnName(k)] = v
+		idx, ok := nameToPosition[normalizeColumnName(k)]
```
emaxerrno, line 37: this is cool. wondeirng if normalizing to integer and doing integer comparison of a column name would be faster. 

`k=key in the row` would "hash" or whatev to a memoized column name. but probably not where the perf is.

rockwotj: `normalizeColumnName` is fairly expensive. I'm thinking about writing a microbench for it and optimizing it, but I think the greater perf at this point would be parallelizing zstd compression in the parquet writing bit. This conversion is already parallelized.

internal/impl/snowflake/streaming/streaming.go
```diff
@@ -272,72 +271,132 @@ func (c *SnowflakeIngestionChannel) nextRequestID() string {
...
+		end := min(maxRowGroupSize, len(batch[i:]))
+		j := len(rowGroups)
+		rowGroups = append(rowGroups, rowGroup{})
+		chunk := batch[i : i+end]
+		wg.Go(func() error {
+			rows, stats, err := constructRowGroup(chunk, c.schema, c.transformers)
+			rowGroups[j] = rowGroup{rows, stats}
+			return err
+		})
+	}
```
mihaitodor, line 308: I wonder if this combined with the default `max_in_flight` value (64) can lead to too much concurrency issues.

internal/impl/snowflake/streaming/streaming.go
```diff
@@ -272,72 +271,132 @@ func (c *SnowflakeIngestionChannel) nextRequestID() string {
...
-		return stats, err
+		return insertStats, err
 	}
+
 	uploadStartTime := time.Now()
-	fileMD5Hash := md5.Sum(encrypted)
 	uploaderResult := c.uploader.Load()
 	if uploaderResult.err != nil {
-		return stats, fmt.Errorf("failed to acquire stage uploader: %w", uploaderResult.err)
+		return insertStats, fmt.Errorf("failed to acquire stage uploader: %w", uploaderResult.err)
```
mihaitodor, line 381: Nit: might be worth decorating the other returned errors as well.

mihaitodor (approved):
LGTM, just a nit and a small question. Feel free to :shipit:

internal/impl/snowflake/streaming/streaming.go
rockwotj: Yeah I guess I haven't seen past parallelism issues. In theory this is purely compute bound work and I was not seeing full utilization without this (actually I don't see full utilization with this I need to parallelize compression too, but I have to switch to arrow for that).

I will add a limit to the err group and plumb that down as an advanced config option called build_parallelism.

mihaitodor: I haven't really done much benchmarking, so dunno how it might behave. I was just thinking that the scheduler might end up doing a bit of extra context switching if the number of goroutines is much bigger than the number of cores for CPU-intensive tasks, but maybe it's smart enough to not do that. I might toy around with an example when I get a bit of time.

rockwotj: The runtime can preempt long running goroutines and context switch so limiting this is better. I will just surface this as a config

internal/impl/snowflake/streaming/streaming.go
rockwotj: Done

internal/impl/snowflake/streaming/streaming.go
rockwotj: Done - thanks for this suggestion!

internal/impl/snowflake/streaming/parquet.go
rockwotj: Doubled the common case for `normalizeColumnName` here: https://github.com/redpanda-data/connect/pull/2979

PR #2983, opened 2024-11-05
feat(benchmark): add benchmark processor

ooesili:
This processor emits rough throughput metrics which is useful for ad-hoc speed measurements of different combinations of input and output plugins.


ooesili:
Moving to benthos repo instead

PR #2984, opened 2024-11-06
Add headers support to the redpanda_migrator_bundle output

mihaitodor:
This is a quick hack which we can use, but I'd rather have a more robust fix. Also, this approach is problematic if users have custom headers which are prefixed with `kafka_`.

Would you consider emitting Kafka headers from the input as an array using structured metadata and maybe call the field `kafka_header`? This would be a breaking change for existing components, but maybe we can just do it for as part of `FranzRecordToMessageV1()` in `internal/impl/kafka/franz_reader.go` in #2918.

### DIFF
--- a/internal/impl/kafka/enterprise/redpanda_migrator_bundle_output.tmpl.yaml
+++ b/internal/impl/kafka/enterprise/redpanda_migrator_bundle_output.tmpl.yaml
@@ -44,7 +44,13 @@ mapping: |
       "partition": "${! metadata(\"kafka_partition\").or(throw(\"missing kafka_partition metadata\")) }",
       "partitioner": "manual",
       "timestamp": "${! metadata(\"kafka_timestamp_unix\").or(timestamp_unix()) }",
-      "max_in_flight": $rpMigratorMaxInFlight
+      "max_in_flight": $rpMigratorMaxInFlight,
+      "metadata": {
+        "include_patterns": [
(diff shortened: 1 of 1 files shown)


internal/impl/kafka/enterprise/redpanda_migrator_bundle_output.tmpl.yaml
```diff
@@ -44,7 +44,13 @@ mapping: |
       "partition": "${! metadata(\"kafka_partition\").or(throw(\"missing kafka_partition metadata\")) }",
       "partitioner": "manual",
       "timestamp": "${! metadata(\"kafka_timestamp_unix\").or(timestamp_unix()) }",
-      "max_in_flight": $rpMigratorMaxInFlight
+      "max_in_flight": $rpMigratorMaxInFlight,
+      "metadata": {
+        "include_patterns": [
+          # Exclude metadata fields which start with `kafka_`
+          "^(?:[^k].*|k[^a].*|ka[^f].*|kaf[^k].*|kafk[^a].*|kafka[^_].*)"
```
rockwotj, line 51: Ugh that's annoying!

rockwotj (approved):

rockwotj:
This seems better as a quick fix - but yea it would be better to have some way to be more robust about this for the migrator as you mention in the PR description. Maybe can happen after Ash's Kafka refactor PR

PR #2988, opened 2024-11-07
snowflake: uppercase db, schema and roles

rockwotj:
It seems that these fields are case sensitive in the API calls and by
default everything is uppercased, so just upper case these fields. If
someone is using weird quoted database/schema/roles then we can handle
that as needed.


mihaitodor (approved):
:shipit:

PR #2999, opened 2024-11-13
sql_insert: support keyword options to add before the INTO clause of the query

pedroluislopez:
Add support for including optionally a list of keyword options before the INTO clause of the query in `sql_insert` components.

It's useful when you want to insert into a MySQL ignoring duplicate key errors.


pedroluislopez:
- Remains check integration tests.

rockwotj (approved):
LGTM

PR #3008, opened 2024-11-14
snowpipe: batch schema migration

rockwotj:
When first running a pipeline with heterogeneous data, you might need to
do a lot of column additions. Support a batch path for updating columns
so that we don't have to fallback to retrying the whole batch - also
it makes the experience much faster.


mihaitodor (approved):
LGTM :shipit:

PR #3014, opened 2024-11-18
MySQL CDC Plugin

le-vlad:
Adds support for MySQL CDC using golang-canal lib

## Features supported:
- BinLog Streaming
- Initial snapshot streaming
- Using composite primary keys for snapshot
- Storing binlog position using cache

## Versions tested:
`"8.0", "9.0", "9.1"`


internal/impl/mysql/input_mysql_stream.go
```diff
@@ -0,0 +1,645 @@
...
+			Description("The DSN of the MySQL database to connect to.").
+			Example("user:password@tcp(localhost:3306)/database"),
+		service.NewStringListField(fieldMySQLTables).
+			Description("A list of tables to stream from the database.").
+			Example([]string{"table1", "table2"}),
+		service.NewStringField(fieldCheckpointKey).
+			Description("The key to store the last processed binlog position."),
+		service.NewStringField(fieldFlavor).
+			Description("The flavor of MySQL to connect to.").
+			Example("mysql"),
```
rockwotj, line 57: What is the alternative? Should this be the default?

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -0,0 +1,645 @@
...
+		service.NewBoolField(fieldMaxSnapshotParallelTables).
+			Description("Int specifies a number of tables to be streamed in parallel when taking a snapshot. If set to true, the connector will stream all tables in parallel. Otherwise, it will stream tables one by one.").
```
rockwotj, line 59: This declared as a bool field

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -0,0 +1,645 @@
...
+	Summary("Enables MySQL streaming for RedPanda Connect.").
+	Fields(
+		service.NewStringField(fieldMySQLDSN).
+			Description("The DSN of the MySQL database to connect to.").
+			Example("user:password@tcp(localhost:3306)/database"),
+		service.NewStringListField(fieldMySQLTables).
+			Description("A list of tables to stream from the database.").
+			Example([]string{"table1", "table2"}),
+		service.NewStringField(fieldCheckpointKey).
+			Description("The key to store the last processed binlog position."),
```
rockwotj, line 54: Don't we also need a cache field? Then this is the key for the cache

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -0,0 +1,645 @@
...
+			Description("Int specifies a number of tables to be streamed in parallel when taking a snapshot. If set to true, the connector will stream all tables in parallel. Otherwise, it will stream tables one by one.").
+			Default(1),
+		service.NewIntField(fieldSnapshotMaxBatchSize).
+			Description("The maximum number of rows to be streamed in a single batch when taking a snapshot.").
+			Default(1000),
+		service.NewBoolField(fieldStreamSnapshot).
+			Description("If set to true, the connector will query all the existing data as a part of snapshot procerss. Otherwise, it will start from the current binlog position."),
+		service.NewAutoRetryNacksToggleField(),
+		service.NewIntField(fieldCheckpointLimit).
+			Description("The maximum number of messages that can be processed at a given time. Increasing this limit enables parallel processing and batching at the output level. Any given LSN will not be acknowledged unless all messages under that offset are delivered in order to preserve at least once delivery guarantees.").
```
rockwotj, line 68: LSN is a postgres concept not MySQL

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -0,0 +1,645 @@
...
+
+	if streamInput.fieldSnapshotMaxBatchSize, err = conf.FieldInt(fieldSnapshotMaxBatchSize); err != nil {
+		return nil, err
+	}
+
+	if streamInput.checkPointLimit, err = conf.FieldInt(fieldCheckpointLimit); err != nil {
+		return nil, err
+	}
+
+	if streamInput.binLogCache, err = conf.FieldString(fieldCheckpointKey); err != nil {
```
rockwotj, line 164: Based on the docs this isn't quite right.

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -0,0 +1,645 @@
...
+			if err = json.Unmarshal(binLogPositionBytes, &storedMySQLBinLogPosition); err != nil {
+				res.Logger().With("error", err.Error()).Error("Failed to unmarshal stored binlog position.")
+				return
+			}
+
+			streamInput.startBinLogPosition = &storedMySQLBinLogPosition
+		}); err != nil {
+
+			res.Logger().With("error", err.Error()).Error("Failed to access cursor cache.")
+		}
```
rockwotj, line 186: We should delay doing this until `Connect`.

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -0,0 +1,645 @@
...
+	}
+}
+
+func (i *mysqlStreamInput) startMySQLSync(ctx context.Context) {
+	i.canal.SetEventHandler(i)
+	go i.readMessages(ctx)
+	// If we require snapshot streaming && we don't have a binlog position cache
+	// initiate default run for Canal to process snapshot and start incremental sync of binlog
+	if i.streamSnapshot && i.startBinLogPosition == nil {
+		// Doesn't work at the moment
```
rockwotj, line 339: This still true?

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -0,0 +1,645 @@
...
+			Description("The flavor of MySQL to connect to.").
+			Example("mysql"),
+		service.NewBoolField(fieldMaxSnapshotParallelTables).
+			Description("Int specifies a number of tables to be streamed in parallel when taking a snapshot. If set to true, the connector will stream all tables in parallel. Otherwise, it will stream tables one by one.").
+			Default(1),
+		service.NewIntField(fieldSnapshotMaxBatchSize).
+			Description("The maximum number of rows to be streamed in a single batch when taking a snapshot.").
+			Default(1000),
+		service.NewBoolField(fieldStreamSnapshot).
+			Description("If set to true, the connector will query all the existing data as a part of snapshot procerss. Otherwise, it will start from the current binlog position."),
```
rockwotj, line 65: typos, and we need to talk about primary key requirement

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -0,0 +1,645 @@
...
+
+	go i.startMySQLSync(ctx)
+	return nil
+}
+
+func (i *mysqlStreamInput) readMessages(ctx context.Context) {
+	var nextTimedBatchChan <-chan time.Time
+	var latestPos *mysqlReplications.Position
+
+	for !i.shutSig.IsHasStoppedSignalled() {
```
rockwotj, line 279: This is the wrong thing to be checking, we should `IsSoftStopSignalled`

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -0,0 +1,645 @@
...
+	db, err := sql.Open("mysql", i.dsn)
+	if err != nil {
+		return err
+	}
+
+	// create snapshot instance
+	snapshot := NewSnapshot(ctx, i.logger, db)
+	i.snapshot = snapshot
+
+	go i.startMySQLSync(ctx)
```
rockwotj, line 271: We should not be launching goroutines holding onto contexts - that's bad practice. We should be using the context from the shutdown signaller here.

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -0,0 +1,645 @@
...
+			maxOffset := resolveFn()
+			if maxOffset == nil {
+				return nil
+			}
+
+			if msgType, ok := lastMsg.MetaGet("type"); ok && msgType == "snapshot" {
+				return nil
+			}
+
+			if err := i.syncBinlogPosition(context.Background()); err != nil {
```
rockwotj, line 490: use the passed in context

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -0,0 +1,645 @@
...
+	select {
+	case i.msgChan <- asyncMessage{
+		msg: batch,
+		ackFn: func(ctx context.Context, res error) error {
+			maxOffset := resolveFn()
+			if maxOffset == nil {
+				return nil
+			}
+
+			if msgType, ok := lastMsg.MetaGet("type"); ok && msgType == "snapshot" {
```
rockwotj, line 486: This is wrong. Remember the resolveFn may give back an offset that isn't tied to this specific message. So it's possible to process 100 stream messages *then* the snapshot message is ack'd which means that we should save that offset from the 100th message not this snapshot message.

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -0,0 +1,645 @@
...
+	snapshot := NewSnapshot(ctx, i.logger, db)
+	i.snapshot = snapshot
+
+	go i.startMySQLSync(ctx)
+	return nil
+}
+
+func (i *mysqlStreamInput) readMessages(ctx context.Context) {
+	var nextTimedBatchChan <-chan time.Time
+	var latestPos *mysqlReplications.Position
```
rockwotj, line 277: It's confusing to me that sometimes we use this variable and sometimes we use `i.currentLogPosition` to use for acking. Can we always use `i.currentLogPosition`?

internal/impl/mysql/snapshot.go
```diff
@@ -0,0 +1,216 @@
...
+	if _, err := s.tx.ExecContext(ctx, "START TRANSACTION WITH CONSISTENT SNAPSHOT"); err != nil {
+		if rErr := s.tx.Rollback(); rErr != nil {
+			return nil, rErr
+		}
+
+		return nil, fmt.Errorf("failed to start consistent snapshot: %v", err)
+	}
+
+	// 2. Acquire global read lock (minimizing lock time)
+	if _, err := s.lockConn.ExecContext(ctx, "FLUSH TABLES WITH READ LOCK"); err != nil {
```
rockwotj, line 75: This `WITH READ LOCK` is only held with the context of this statement or the connection?

internal/impl/mysql/snapshot.go
```diff
@@ -0,0 +1,216 @@
...
+	// Execute START TRANSACTION WITH CONSISTENT SNAPSHOT
+	if _, err := s.tx.ExecContext(ctx, "START TRANSACTION WITH CONSISTENT SNAPSHOT"); err != nil {
+		if rErr := s.tx.Rollback(); rErr != nil {
+			return nil, rErr
+		}
+
+		return nil, fmt.Errorf("failed to start consistent snapshot: %v", err)
+	}
+
+	// 2. Acquire global read lock (minimizing lock time)
```
rockwotj, line 74: Can we comment as to *why* we need to FLUSH TABLES?

internal/impl/mysql/snapshot.go
```diff
@@ -0,0 +1,216 @@
...
+
+	// 1. Start a consistent snapshot transaction
+	s.tx, err = s.snapshotConn.BeginTx(ctx, &sql.TxOptions{
+		Isolation: sql.LevelRepeatableRead,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to start transaction: %v", err)
+	}
+
+	// Execute START TRANSACTION WITH CONSISTENT SNAPSHOT
```
rockwotj, line 65: This is not a helpful comment. Maybe better would be to explain why we need to use this transaction option?

internal/impl/mysql/snapshot.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+		return nil, fmt.Errorf("failed to release global read lock: %v", err)
+	}
+
+	return &pos, nil
+}
+
+func (s *Snapshot) getRowsCount(table string) (int, error) {
+	var count int
+	if err := s.tx.QueryRowContext(s.ctx, "SELECT COUNT(*) FROM "+table).Scan(&count); err != nil {
```
rockwotj, line 109: Is this actually fast? Should we be querying the table stats instead? https://stackoverflow.com/a/61548683

internal/impl/mysql/snapshot.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	return count, nil
+}
+
+func (s *Snapshot) getTablePrimaryKeys(table string) ([]string, error) {
+	// Get primary key columns for the table
+	rows, err := s.tx.QueryContext(s.ctx, fmt.Sprintf(`
+SELECT COLUMN_NAME
+FROM information_schema.KEY_COLUMN_USAGE
+WHERE TABLE_NAME = '%s' AND CONSTRAINT_NAME = 'PRIMARY';
```
rockwotj, line 120: Will this yield them in the right order?

rockwotj (commented):
Initial pass - thanks!

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -0,0 +1,637 @@
...
+	case canal.DeleteAction:
+		return i.onMessage(e, 0, 1)
+	case canal.UpdateAction:
+		return i.onMessage(e, 1, 2)
+	default:
+		return errors.New("invalid rows action")
+	}
+}
+
+func (i *mysqlStreamInput) OnPosSynced(eh *replication.EventHeader, pos mysqlReplications.Position, gtid mysqlReplications.GTIDSet, synced bool) error {
```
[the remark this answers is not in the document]
rockwotj: Can you explain when this is called? Because right now we seem to blindly overwrite `currentLogPosition` in multiple places and I don't know if that's always OK to do.

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -0,0 +1,637 @@
...
+		i.canal.SyncedPosition()
+		i.canal.Close()
+	}
+
+	return i.snapshot.releaseSnapshot(ctx)
+}
+
+// ---- Redpanda Connect specific methods end----
+
+// --- MySQL Canal handler methods ----
```
rockwotj, line 591: The other handler methods we don't need those? Can you explain why?

internal/impl/mysql/input_mysql_stream.go
rockwotj: This is still incorrect.

internal/impl/mysql/snapshot.go
rockwotj: Why do we need a count anyways? In postgres we just query until we get no rows back, meaning we've reached the end of a table (because the pk is sorted). Can we do the same here? I think this will be slow and expensive for large tables.

internal/impl/mysql/snapshot.go
```diff
@@ -0,0 +1,231 @@
...
+// NewSnapshot creates new snapshot instance
+func NewSnapshot(ctx context.Context, logger *service.Logger, db *sql.DB) *Snapshot {
+	return &Snapshot{
+		db:     db,
+		ctx:    ctx,
+		logger: logger,
+	}
+}
+
+func (s *Snapshot) prepareSnapshot(ctx context.Context) (*mysql.Position, error) {
```
rockwotj, line 43: @toddfarmer have any concerns about how we take a snapshot of the DB?

rockwotj (commented):
Round two! Looking better, thanks for your work here 😄

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -0,0 +1,637 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return conf.WrapBatchInputExtractTracingSpanMapping("mysql_stream", r)
+}
+
+func init() {
+	err := service.RegisterBatchInput(
+		"mysql_stream", mysqlStreamConfigSpec,
```
rockwotj, line 195: Based on some internal conversations, we should rename this to be more explicit it's a CDC connector

```suggestion
		"mysql_cdc", mysqlStreamConfigSpec,
```

rockwotj (approved):

mihaitodor (approved):
Nice one! I fixed a few small nits, added the package to the public components to generate docs and re-enabled the `rowserrcheck` linter since it seems to work fine now.

Looking through the code, I think it's good and feel free to :shipit: Just left one small question.

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -0,0 +1,765 @@
...
+					Row:       row,
+					Operation: MessageOperationRead,
+					Table:     table,
+					Position:  nil,
+				}:
+				case <-ctx.Done():
+					return ctx.Err()
+				}
+			}
+			// TODO(cdc): Save checkpoint
```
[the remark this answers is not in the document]
rockwotj: I'll drop it, it's a note for if we want to be able to resume from the initial snapshot phase or not, but I don't think we'll do that anytime soon 👍

PR #3020, opened 2024-11-20
snowflake: fix channel creation leak

rockwotj:
- **snowflake: lower default max in flight**
- **snowflake: prevent more than `max_in_flight` channels from being created**


internal/impl/snowflake/output_snowflake_streaming.go
```diff
@@ -38,7 +39,10 @@ const (
 	ssoFieldBatching                            = "batching"
 	ssoFieldChannelPrefix                       = "channel_prefix"
 	ssoFieldMapping                             = "mapping"
-	ssoFieldBuildParallelism                    = "build_parallelism"
+	ssoFieldBuildOpts                           = "build_options"
+	ssoFieldBuildParallelismLegacy              = "build_parallelism"
+	ssoFieldBuildParallelism                    = "parallelism"
```
rockwotj, line 44: intra_file_parallelism

Jeffail (approved):
LG to the M

PR #3021, opened 2024-11-20
pgcdc: remove stats passing as part of message

rockwotj:
In reality we should just be periodically reporting these stats, and not
tying them to data being emitted on the stream


Jeffail (approved):
LGTM

PR #3022, opened 2024-11-20
emit default values for non populated fields in protobuf

ealonsodb:
When you use schema_registry_decode processor, the protobuf default values (0 for int32, "" for string, false for bool, 0.0 for float, etc...) are not populated in the row. 
With this new parameter you can have all those default values populated, no need to check for nulls after that processor


rockwotj:
Thanks for this! I will review after Thanksgiving.

One quick note is it would be good if the configuration option was either nested in a protobuf_options object or if the field was prefixed with protobuf to make it more clear what it was for (also for alignment with avro). I personally prefer the former option

internal/impl/confluent/processor_schema_registry_decode.go
```diff
@@ -65,7 +65,11 @@ This processor decodes protobuf messages to JSON documents, you can read more ab
 		Field(service.NewBoolField("avro_raw_json").
 			Description("Whether Avro messages should be decoded into normal JSON (\"json that meets the expectations of regular internet json\") rather than https://avro.apache.org/docs/current/specification/_print/#json-encoding[Avro JSON^]. If `true` the schema returned from the subject should be decoded as https://pkg.go.dev/github.com/linkedin/goavro/v2#NewCodecForStandardJSONFull[standard json^] instead of as https://pkg.go.dev/github.com/linkedin/goavro/v2#NewCodec[avro json^]. There is a https://github.com/linkedin/goavro/blob/5ec5a5ee7ec82e16e6e2b438d610e1cab2588393/union.go#L224-L249[comment in goavro^], the https://github.com/linkedin/goavro[underlining library used for avro serialization^], that explains in more detail the difference between the standard json and avro json.").
 			Advanced().Default(false)).
-		Field(service.NewURLField("url").Description("The base URL of the schema registry service."))
+		Field(service.NewURLField("url").
+			Description("The base URL of the schema registry service.")).
+		Field(service.NewBoolField("emit_default_values").
```
rockwotj, line 70: Can we instead create a nested config object and add a boolean field for `emit_defaults`? I'd like to namespace these as over time I expect to have more protobuf knobs available.

That will look something like this:

```
service.NewObjectField("protobuf", service.NewBoolField("emit_defaults"))
```

rockwotj (commented):
Looks great thank you for adding tests! I have one request for the configuration to namespace the option under a `protobuf` field so that it's clear this only works for protobuf and we have a place for more protobuf related configs as we inevitably add more options here 😄

rockwotj:
This functionality has landed in another PR and is released

PR #3023, opened 2024-11-20
pgcdc: remove batch_transactions

rockwotj:
Just support emitting begin and commit txn markers, which then a custom
batch policy for batching transactions can be created. This means that
each message is a single row, which is a very simple mental model.

This still has the slow performance when not emitting commit txn markers
so that we don't reprocess the last txn, and I can fix that in another
PR by using a layer of indirection with the LSN that the plugin is
acking and what we actually send to postgres.

NOTE: this is a draft because I still need to fix the integration tests


Jeffail (approved):
LGTM

rockwotj:
Thanks @Jeffail! I'm sneaking in the docs updates from the benthos upgrade into this PR

PR #3026, opened 2024-11-21
Refactor Redpanda Migrator components

mihaitodor:
I hijacked this PR to address several issues:

### Fixed

- The `redpanda_migrator` output no longer rejects messages if it can't perform schema ID translation.
- The `redpanda_migrator` input no longer converts the kafka key to string.

### Added

- New `redpanda_migrator_offsets` input.
- Fields `offset_topic`, `offset_group`, `offset_partition`, `offset_commit_timestamp` and `offset_metadata` added to the `redpanda_migrator_offsets` output.
- Field `topic_lag_refresh_period` added to the `redpanda` and `redpanda_common` inputs.
- Metric `redpanda_lag` now emitted by the `redpanda` and `redpanda_common` inputs.
- Metadata `kafka_lag` now emitted by the `redpanda` and `redpanda_common` inputs.
- The `redpanda_migrator_bundle` input and output now set labels for their subcomponents.

### Changed

- The `kafka_key` and `max_in_flight` fields of the `redpanda_migrator_offsets` output are now deprecated.
- Fields `batch_size`, `multi_header`, `replication_factor`, `replication_factor_override` and `output_resource` for the `redpanda_migrator` input are now deprecated.
- Fields `kafka_key` and `max_in_flight` for the `redpanda_migrator_offsets` output are now deprecated.
- Field `batching` for the `redpanda_migrator` output is now deprecated.
- The `redpanda_migrator` input no longer emits tombstone messages.

## Redpanda Migrator offset metadata

One quick way to test this is via the following config. Note how I overwrite `kafka_offset_metadata` to `foobar` in a `mapping` processor.

```yaml
input:
  redpanda_migrator_bundle:
    redpanda_migrator:
      seed_brokers: [ "localhost:9092" ]
      topics:
        - '^[^_]' # Skip internal topics which start with `_`
      regexp_topics: true
      consumer_group: migrator_bundle
      start_from_oldest: true
      replication_factor_override: true
      replication_factor: -1

    schema_registry:
      url: http://localhost:8081
      include_deleted: true
      subject_filter: ""

output:
  processors:
    - switch:
        - check: metadata("input_label") == "redpanda_migrator_offsets_input"
          processors:
            - mapping: |
                meta kafka_offset_metadata = "foobar"
  redpanda_migrator_bundle:
    redpanda_migrator:
      seed_brokers: [ "localhost:9093" ]
      max_in_flight: 1
      replication_factor_override: true
      replication_factor: -1

    schema_registry:
      url: http://localhost:8082
```


internal/impl/kafka/franz_reader_ordered.go
```diff
@@ -76,18 +105,25 @@ type FranzReaderOrdered struct {
 
 // NewFranzReaderOrderedFromConfig attempts to instantiate a new
 // FranzReaderOrdered reader from a parsed config.
-func NewFranzReaderOrderedFromConfig(conf *service.ParsedConfig, res *service.Resources, optsFn func() ([]kgo.Opt, error)) (*FranzReaderOrdered, error) {
+func NewFranzReaderOrderedFromConfig(conf *service.ParsedConfig, res *service.Resources, clientOptsFn clientOptsFn, recordToMessageFn recordToMessageFn, preflightHookFn preflightHookFn, closeHookFn closeHookFn) (*FranzReaderOrdered, error) {
```
mihaitodor, line 108: This constructor got a bit messy to use... It can be hard to tell which of these funcs is set to nil at the call site and one can easily mix them up. I'm thinking to maybe introduce functional options for it or maybe a struct which contain all the parameters. WDYT?

internal/impl/kafka/enterprise/redpanda_migrator_offsets_input.go
```diff
@@ -0,0 +1,207 @@
...
+			}
+			if d.FetchMinBytes, err = kafka.BytesFromStrFieldAsInt32(rmoiFieldFetchMinBytes, conf); err != nil {
+				return nil, err
+			}
+			if d.FetchMaxPartitionBytes, err = kafka.BytesFromStrFieldAsInt32(rmoiFieldFetchMaxPartitionBytes, conf); err != nil {
+				return nil, err
+			}
+
+			// Consume messages from the `__consumer_offsets` topic
+			d.Topics = []string{`__consumer_offsets`}
```
rockwotj: What's the plan for Redpanda Serverless, which doesn't support this?

internal/impl/kafka/enterprise/redpanda_migrator_bundle_input.tmpl.yaml
```diff
@@ -137,9 +137,9 @@ tests:
               consumer_group: "migrator"
             processors:
               - mapping: meta input_label = "redpanda_migrator"
-          - kafka_franz:
+          - redpanda_migrator_offsets:
               seed_brokers: [ "127.0.0.1:9092" ]
-              topics: [ "__consumer_offsets" ]
+              topics: [ "foobar" ]
```
rockwotj, line 158: Why is this `foobar`?

internal/impl/kafka/enterprise/redpanda_migrator_offsets_input.go
```diff
@@ -0,0 +1,207 @@
...
+					isExpectedTopic = slices.ContainsFunc(topicPatterns, func(tp *regexp.Regexp) bool {
+						return tp.MatchString(key.Topic)
+					})
+				} else {
+					isExpectedTopic = slices.ContainsFunc(topics, func(t string) bool {
+						return t == key.Topic
+					})
+				}
+				if !isExpectedTopic {
+					return nil, fmt.Errorf("skipping updates for the %q topic: %s", key.Topic, err)
```
rockwotj: should we just `return nil, nil` for skipping records? I guess that's why in the previous commit you have the error case for this being to skip the record instead of raising an error.

internal/impl/kafka/enterprise/redpanda_migrator_offsets_input.go
```diff
@@ -0,0 +1,207 @@
...
+				msg := kafka.FranzRecordToMessageV1(record)
+
+				// Check the version to ensure that we process only offset commit keys
+				key := kmsg.NewOffsetCommitKey()
+				if err := key.ReadFrom(record.Key); err != nil || (key.Version != 0 && key.Version != 1) {
+					return nil, fmt.Errorf("failed to decode record key: %s", err)
+				}
+
+				isExpectedTopic := false
+				if len(topicPatterns) > 0 {
```
rockwotj, line 151: Can we instead pull a predicate out as a local variable so this becomes: `isExpectedTopic := matchesTopic(key.Topic)` or something like that?

internal/impl/kafka/enterprise/redpanda_migrator_offsets_output.go
```diff
@@ -162,54 +200,79 @@ func (w *RedpandaMigratorOffsetsWriter) Connect(ctx context.Context) error {
...
+	var offsetMetadata string
+	if w.offsetMetadata != nil {
+		if offsetMetadata, err = w.offsetMetadata.TryString(msg); err != nil {
+			return fmt.Errorf("failed to extract offset metadata: %w", err)
+		}
 	}
 
 	updateConsumerOffsets := func() error {
-		listedOffsets, err := w.client.ListOffsetsAfterMilli(ctx, val.CommitTimestamp, key.Topic)
+		listedOffsets, err := w.client.ListOffsetsAfterMilli(ctx, offsetCommitTimestamp, topic)
```
rockwotj, line 251: Ideally this should be `ListOffsetsAtOrBeforeMilli` right? Can we add a TODO and file a task to improve this?

internal/impl/kafka/franz_reader_ordered.go
```diff
@@ -76,18 +99,25 @@ type FranzReaderOrdered struct {
...
-		res:         res,
-		log:         res.Logger(),
-		shutSig:     shutdown.NewSignaller(),
-		clientOpts:  optsFn,
+		readBackOff:       readBackOff,
+		res:               res,
+		log:               res.Logger(),
+		shutSig:           shutdown.NewSignaller(),
+		clientOpts:        optsFn,
+		topicLagGauge:     res.Metrics().NewGauge("redpanda_lag", "topic", "partition"),
```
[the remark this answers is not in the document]
rockwotj: I think the label is enough. Do we really want this lag metric for all these inputs? Probably I would assume...

internal/impl/kafka/franz_reader_ordered.go
```diff
@@ -80,7 +99,7 @@ type FranzReaderOrdered struct {
 
 // NewFranzReaderOrderedFromConfig attempts to instantiate a new
 // FranzReaderOrdered reader from a parsed config.
-func NewFranzReaderOrderedFromConfig(conf *service.ParsedConfig, res *service.Resources, optsFn func() ([]kgo.Opt, error), recordToMessageFn RecordToMessageFn) (*FranzReaderOrdered, error) {
+func NewFranzReaderOrderedFromConfig(conf *service.ParsedConfig, res *service.Resources, optsFn func() ([]kgo.Opt, error), recordToMessageFn recordToMessageFn, preflightHookFn preflightHookFn, closeHookFn func(res *service.Resources)) (*FranzReaderOrdered, error) {
```
rockwotj: I'm not really loving all these extra hooks into this method... It would be great to wrap this method/reader instead (ala decorator pattern style) - can we do that? I think the net result would be cleaner.

internal/impl/kafka/franz_reader_ordered.go
```diff
@@ -389,13 +421,50 @@ func (f *FranzReaderOrdered) Connect(ctx context.Context) error {
...
+	// Check connectivity to cluster
+	if err = f.client.Ping(ctx); err != nil {
+		return fmt.Errorf("failed to connect to cluster: %s", err)
+	}
+
+	if f.lagUpdater != nil {
+		f.lagUpdater.Stop()
+	}
+	adminClient := kadm.NewClient(f.client)
+	f.lagUpdater = asyncroutine.NewPeriodic(f.topicLagRefreshPeriod, func() {
```
rockwotj, line 443: I think you want `asyncroutine.NewPeriodicWithContext`

internal/impl/kafka/franz_reader_ordered.go
```diff
@@ -449,6 +518,7 @@ func (f *FranzReaderOrdered) Connect(ctx context.Context) error {
 			fetches.EachPartition(func(p kgo.FetchTopicPartition) {
 				if len(p.Records) > 0 {
 					batch := f.recordsToBatch(p.Records)
+					// TODO: do we have to ack?
```
rockwotj, line 527: We had talked earlier and we don't have to ack.

internal/impl/kafka/franz_writer.go
```diff
@@ -262,17 +271,21 @@ type FranzWriter struct {
...
+	yieldClientFn   yieldClientFn
+	onConnectHookFn func(client *kgo.Client)
+	onWriteHookFn   onWriteHookFn
 }
 
 // NewFranzWriterFromConfig uses a parsed config to extract customisation for
 // writing data to a Kafka broker. A closure function must be provided that is
 // responsible for granting access to a connected client.
-func NewFranzWriterFromConfig(conf *service.ParsedConfig, accessClientFn func(FranzSharedClientUseFn) error, yieldClientFn func(context.Context) error) (*FranzWriter, error) {
+func NewFranzWriterFromConfig(conf *service.ParsedConfig, accessClientFn accessClientFn, yieldClientFn yieldClientFn, onConnectHookFn func(client *kgo.Client), onWriteHookFn onWriteHookFn) (*FranzWriter, error) {
```
rockwotj, line 283: Same here with these hooks, can we wrap the writer that is returned instead?

internal/impl/kafka/enterprise/redpanda_migrator_offsets_input.go
```diff
@@ -139,23 +123,20 @@ func init() {
...
-				return nil, err
-			}
-			if d.FetchMinBytes, err = kafka.BytesFromStrFieldAsInt32(rmoiFieldFetchMinBytes, conf); err != nil {
-				return nil, err
-			}
-			if d.FetchMaxPartitionBytes, err = kafka.BytesFromStrFieldAsInt32(rmoiFieldFetchMaxPartitionBytes, conf); err != nil {
-				return nil, err
-			}
+			// Configure `start_from_oldest: true`
+			// This is probably not necessary since `__consumer_offsets` is a compacted topic
```
rockwotj, line 131: I don't really follow this comment.

internal/impl/kafka/franz_reader_ordered.go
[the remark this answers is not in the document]
mihaitodor: I also think it's a bit overkill and I don't recall now which conversation led to this pattern. I also emit the `kafka_lag` metadata field with each message, so one could add a `metric` processor in the pipeline which creates a gauge for topics as needed. One downside with this approach is if messages stop flowing completely, then this gauge wouldn't get any updates. I think the main idea was to make it easier for people to discover this metric, but it's not clear what the perf impact might be if we consume from thousands of topics, each having multiple partitions. Should I remove it? (cc @Jeffail)

internal/impl/kafka/franz_reader_ordered.go
```diff
@@ -114,8 +121,13 @@ func (f *FranzReaderOrdered) recordsToBatch(records []*kgo.Record) *batchWithRec
 	var length uint64
 	var batch service.MessageBatch
 	for _, r := range records {
+		record, err := f.recordToMessageFn(r)
+		if err != nil {
+			f.log.Debugf("Failed to convert kafka record to message: %s", err)
```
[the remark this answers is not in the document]
mihaitodor: We're intentionally dropping tombstone records. Otherwise, `kafka.FranzRecordToMessageV1()` doesn't return any errors, but yeah, it's easy to misuse. Alternatively, I could set an error on the message, but I'm not aware of any other input which does that and, if the idea is to drop those messages anyway, it would be wasteful to need an extra processor for that. Not sure what's the best approach here.

internal/impl/kafka/franz_reader_ordered.go
```diff
@@ -76,18 +79,22 @@ type FranzReaderOrdered struct {
 
 // NewFranzReaderOrderedFromConfig attempts to instantiate a new
 // FranzReaderOrdered reader from a parsed config.
-func NewFranzReaderOrderedFromConfig(conf *service.ParsedConfig, res *service.Resources, optsFn func() ([]kgo.Opt, error)) (*FranzReaderOrdered, error) {
+func NewFranzReaderOrderedFromConfig(conf *service.ParsedConfig, res *service.Resources, optsFn func() ([]kgo.Opt, error), recordToMessageFn RecordToMessageFn) (*FranzReaderOrdered, error) {
```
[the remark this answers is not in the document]
mihaitodor: > Can we expand the docs to explain what `recordToMessageFn` is and that it can be `nil`?
Good idea, sorted! I am still thinking if I should change the signature of this constructor to make it easier to use. Maybe a separate refactor...

> Also is there no way to do this transform after the fact? IE a mapping on the V1 message type?
For the `redpanda_migrator` input, I'd lose the ability to drop tombstone messages directly here, so I'd need to use a `mapping` processor like I explained in the other comment. I guess that's fine even if it adds a bit of perf overhead, but may be a bit annoying if someone wants to use this `redpanda_migrator` input directly instead of the `redpanda_migrator_bundle`.
For the `redpanda_migrator_offsets` input it gets a bit awkward, because then we'd need a custom processor which can decode both the `record.Key` and `record.Value` using franz-go APIs. We'd also lose the ability to filter out messages which contain consumer group updates about topics we don't care about, so I'd have to move the topic filter to the processor instead. It would be confusing for people who don't know they always need to attach this processor to the `redpanda_migrator_offsets` input.

internal/impl/kafka/enterprise/redpanda_migrator_offsets_input.go
mihaitodor: I have a ticket for that and for now it's still something which needs to be discussed. I don't know what we can do there, but both the `redpanda_migrator_offsets` input and output don't currently work with RP Serverless.

internal/impl/kafka/enterprise/redpanda_migrator_offsets_input.go
mihaitodor: This way, I can `log.Debugf` higher up the reason for skipping messages. In general, I don't think the input should raise errors if it can't decode a message. Either drop it or send it as-is. We can, however, consider setting an error on the message itself, but I don't think any input does that currently and it might be a bit weird to do it. Curious if @Jeffail has any thoughts on this.

internal/impl/kafka/franz_reader_ordered.go
mihaitodor: Yep, thank you! I forgot to remove the TODO 😁

internal/impl/kafka/enterprise/redpanda_migrator_offsets_input.go
mihaitodor: I'm actually not sure... My thought was that `kgo.ConsumeResetOffset(kgo.NewOffset().AtStart())` doesn't do anything when used with a compacted topic, but maybe that's just a silly assumption since I don't really know such topics are read. I removed the comment and left the code as is. Don't think it breaks anything.

internal/impl/kafka/franz_reader_ordered.go
mihaitodor: Nice catch!

internal/impl/kafka/enterprise/redpanda_migrator_offsets_output.go
mihaitodor: I think so too, but, unfortunately we don't have such an API and I'm guessing the Kafka protocol might not be flexible enough to add it, but I'll have to check again with @twmb. 

If that's the case, maybe we should just decrement the timestamp that's passed to `ListOffsetsAfterMilli()`, but not sure what would be a reasonable value in practice... I can make it configurable, but it's definitely tricky to come up with a value which will guarantee that the consumer group offset always gets set to the current offset (or a slightly earlier offset) in the destination. Setting it to a slightly earlier offset will just lead to a few duplicates, which is *fine*, but setting it to a newer offset would be a problem for consumers switching from the source to the destination cluster...

internal/impl/kafka/enterprise/redpanda_migrator_offsets_input.go
mihaitodor: I took a shot at it, but it I'm not sure if that's what you had in mind. Let me know if this looks OK.

I think it would be nice to introduce a generic helper similar to https://github.com/redpanda-data/benthos/blob/9fdcbd09cc5c8b8276728e14a18a4612b7e0460c/internal/metadata/include_filter.go#L75, but I'll leave that for when we'll have more use cases.

internal/impl/kafka/franz_reader_ordered.go
mihaitodor: Yep, I hate it too (see [above](https://github.com/redpanda-data/connect/pull/3026#discussion_r1900156426)), but I wasn't sure how to go about it without making this PR huge.

Which of the following would you prefer:
- Have exported fields on the reader and writer and set them after calling the `NewFranzReaderOrderedFromConfig()` and `NewFranzWriterFromConfig()` constructors
- Use the functional options pattern
- Pass a struct which contains all these parameters to the constructors
- Something else
I guess you're thinking of the first approach which I haven't considered initially and the functional options approach feels overkill. Embedding the `FranzReaderOrdered` and `FranzWriter` structs in new structs and overriding the relevant methods would be nice, but then I have to introduce some more exported helper methods... Let me know if you'd like me to explore this approach.

internal/impl/kafka/enterprise/redpanda_migrator_offsets_input.go
rockwotj: Compacted topics are not really special from a reader prospective, except they can have gaps in the offsets, because the retention is per key instead of just prefix truncating the log. AtStart will still start from the beginning of the log, but all the old keys might be truncated away. I think it's the right thing to do.

internal/impl/kafka/franz_reader_ordered.go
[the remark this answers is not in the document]
rockwotj: Here is an example of what I mean by changing this via a mapping: https://gist.github.com/rockwotj/48bddf4557a5210d6fa415a1cc736090

Basically just wrap the FranzReader in another struct that implements service.BatchReader and in it's ReadBatch function do what ever filtering/transforming you need to do. No hooks required.

internal/impl/kafka/enterprise/redpanda_migrator_bundle_input.tmpl.yaml
rockwotj: But we don't want a filter by default? Or this should be configurable? I don't understand why this is hardcoded the way it is.

internal/impl/kafka/franz_reader_ordered.go
rockwotj: This is what I was thinking as a diff you can apply on top of this PR: https://gist.github.com/rockwotj/48bddf4557a5210d6fa415a1cc736090. It requires exporting the kgo.Client, but I think that's an okay thing. I haven't seen any hooks yet that fundamentally cannot be implemented in a wrapper layer ([decorator pattern](https://refactoring.guru/design-patterns/decorator)).

internal/impl/kafka/enterprise/redpanda_migrator_offsets_input.go
```diff
@@ -147,16 +146,18 @@ func init() {
...
-					isExpectedTopic = slices.ContainsFunc(topics, func(t string) bool {
-						return t == key.Topic
+				matchesTopic := func(topic string) bool {
+					if len(topicPatterns) > 0 {
+						return slices.ContainsFunc(topicPatterns, func(tp *regexp.Regexp) bool {
+							return tp.MatchString(topic)
+						})
+					}
+					return slices.ContainsFunc(topics, func(t string) bool {
+						return t == topic
```
rockwotj, line 156: good! Now I think this can be lifted out of this closure so it's declared before `NewFranzReaderOrderedFromConfig` is called. It helps remove clutter from what this function is doing.

internal/impl/kafka/franz_reader_ordered.go
Jeffail: Yeah I think I'd go with something like a `FranzReaderHooks` type, where you can do stuff like `NewFranzReaderHooks().WithRecordToMessageFn(foo).WithCloseHookFn(bar)`.

internal/impl/kafka/franz_reader_ordered.go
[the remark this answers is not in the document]
Jeffail: I like having the metric emitted here, it's relatively cheap, and extracting from meta is awkward enough no one is going to do it willingly.

internal/impl/kafka/franz_reader_ordered.go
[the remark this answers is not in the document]
Jeffail: Maybe we should change the signature of the record to message function so that it's clearer messages are dropped in this case, so have it return `(*service.Message, bool)`, where `nil, false` is an explicit instruction to discard the record as if it didn't exist.

internal/impl/kafka/enterprise/redpanda_migrator_offsets_input.go
mihaitodor: Good point! Sorted ❤️

internal/impl/kafka/enterprise/redpanda_migrator_bundle_input.tmpl.yaml
mihaitodor: Answered on Slack. TL; DR: This is the test section of the template where I'm asserting that the topic filter for `redpanda_migrator_offsets` is set to a specific topic. We don't want to send consumer group updates for all topics to the `redpanda_migrator_offsets` output.

internal/impl/kafka/franz_reader_ordered.go
mihaitodor: Very cool! Indeed, that does do the trick for the reader. I didn't think I can leverage this pattern initially, but, with enough patience, it does work out nicely :)

For the writer I'll have to do some refactoring to get rid of those hooks, but not sure if there's a clean way to avoid using a hook in `WriteBatch` unless I add a separate API method that's called in `WriteBatch` and then I can override that one. I'll think about it.

mihaitodor: Update: I refactored the writer as well in line with @Jeffail's comment [here](https://github.com/redpanda-data/connect/pull/3026#discussion_r1906982310). Guess it could be prettier and I'm happy to adjust it if needed.

internal/impl/kafka/enterprise/redpanda_migrator_input.go
```diff
@@ -143,363 +128,94 @@ func init() {
...
-						r.mgr.Logger().Errorf("Failed to create ACLs for topic %q: %s", topic, err)
-					}
-				}
-			}
-			r.outputTopicsCreated = true
-			return nil
-		}); err != nil {
-			r.mgr.Logger().With("error", err, "resource", r.outputResource).Warn("Failed to access shared client for given resource identifier")
+		if len(batch) == 0 {
+			_ = ack(ctx, nil) // TODO: Log this error?
```
mihaitodor, line 209: I wonder if this can cause trouble... I occasionally seeing a panic when Ctrl+C-ing Migrator:

```
panic: send on closed channel

goroutine 76 [running]:
github.com/redpanda-data/benthos/v4/internal/impl/pure.(*fallbackBroker).loop.func3({0x10971c488, 0x140017c8050}, {0x10968d038?, 0x10ce6ae40?})
	/Users/mihaitodor/go/pkg/mod/github.com/redpanda-data/benthos/v4@v4.43.0/internal/impl/pure/output_fallback.go:232 +0x134
github.com/redpanda-data/benthos/v4/internal/message.(*Transaction).Ack(0x10971c488?, {0x10971c488?, 0x140017c8050?}, {0x10968d038?, 0x10ce6ae40?})
	/Users/mihaitodor/go/pkg/mod/github.com/redpanda-data/benthos/v4@v4.43.0/internal/message/transaction.go:90 +0xd8
github.com/redpanda-data/benthos/v4/internal/pipeline.(*Processor).loop(0x14001bdcf00)
	/Users/mihaitodor/go/pkg/mod/github.com/redpanda-data/benthos/v4@v4.43.0/internal/pipeline/processor.go:76 +0x36c
created by github.com/redpanda-data/benthos/v4/internal/pipeline.(*Processor).Consume in goroutine 1
	/Users/mihaitodor/go/pkg/mod/github.com/redpanda-data/benthos/v4@v4.43.0/internal/pipeline/processor.go:154 +0x84
panic: send on closed channel

goroutine 80 [running]:
github.com/redpanda-data/benthos/v4/internal/impl/pure.(*fallbackBroker).loop.func3({0x10971c488, 0x14001e0a640}, {0x10968d2c0?, 0x14001e5e000?})
	/Users/mihaitodor/go/pkg/mod/github.com/redpanda-data/benthos/v4@v4.43.0/internal/impl/pure/output_fallback.go:232 +0x134
github.com/redpanda-data/benthos/v4/internal/message.(*Transaction).Ack(0x10968d2c0?, {0x10971c488?, 0x14001e0a640?}, {0x10968d2c0?, 0x14001e5e000?})
	/Users/mihaitodor/go/pkg/mod/github.com/redpanda-data/benthos/v4@v4.43.0/internal/message/transaction.go:90 +0xd8
github.com/redpanda-data/benthos/v4/internal/component/output.(*notBatchedOutput).loop(0x140022dfc50)
	/Users/mihaitodor/go/pkg/mod/github.com/redpanda-data/benthos/v4@v4.43.0/internal/component/output/not_batched.go:134 +0x1f4
created by github.com/redpanda-data/benthos/v4/internal/component/output.(*notBatchedOutput).Consume in goroutine 1
	/Users/mihaitodor/go/pkg/mod/github.com/redpanda-data/benthos/v4@v4.43.0/internal/component/output/not_batched.go:149 +0xb8
```

rockwotj (approved):
LGTM, since this just incremental improvements (besides maybe the panic), lets :shipit: and continue on

Jeffail (approved):

internal/impl/kafka/enterprise/redpanda_migrator_input.go
mihaitodor: I didn't think through it well enough after pulling some hairs out trying to figure out what I broke. I'm pretty sure now it's likely because I have `shutdown_timeout: 2s` in my test configs to avoid having to `kill -9` when the process gets stuck and I think this leads the input to shut down to quickly. I raised a bug here: https://github.com/redpanda-data/benthos/issues/159. Sorry for the red herring!

CHANGELOG.md
```diff
@@ -15,25 +15,51 @@ All notable changes to this project will be documented in this file.
...
+- (Benthos) Metadata field `label` can now be utilized within a template's `mapping` field to access the label that is associated with the template instantiation in a config. (@mihaitodor)
+- (Benthos) `bloblang` scalar type added to template fields. (@mihaitodor)
+- (Benthos) Go API: Method `SetOutputBrokerPattern` added to the `StreamBuilder` type. (@mihaitodor)
+- (Benthos) New `error_source_name`, `error_source_label` and `error_source_path` bloblang functions. (@mihaitodor)
+- (Benthos) Flag `--verbose` added to the `benthos lint` and `benthos template lint` commands. (@mihaitodor)
 
-### Fixed
+### Changed
 
 - The `code` and `file` fields on the `javascript` processor docs no longer erroneously mention interpolation support. (@mihaitodor)
```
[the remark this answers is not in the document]
mihaitodor: Awwww, thanks for spotting that 🦅 👀

internal/impl/kafka/enterprise/redpanda_migrator_output.go
```diff
@@ -127,217 +127,210 @@ func init() {
...
+						if err := fn(&kafka.FranzSharedClientInfo{Client: client, ConnDetails: connDetails}); err != nil {
+							return err
+						}
+
+						if createdTopicsOnConnect {
+							return nil
+						}
+
+						// Make multiple attempts until the input connects in the background.
+						// TODO: It would be nicer to somehow get notified when the input is ready.
```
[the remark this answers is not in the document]
mihaitodor: Good callout! I should've revised all these interaction points again after the last round of changes. It makes a lot more sense now (see 11a3078b5). I think I'll end up putting together a Mermaid diagram with all the interactions for Migrator after this.

internal/impl/kafka/enterprise/redpanda_migrator_offsets_output.go
mihaitodor: > Do we always ensure the data is committed before writing offsets?
Nope. The two flows are decoupled, although I guess the broker output does apply some throttling if one of the child outputs is slow... Need to check this assumption, but anyway, the `redpanda_migrator` input could be slow and emit the the data sometime after the `redpanda_migrator_offsets` input emits offset updates for that data...

> I think this api does the right thing and give you the end of the log if you ask for stuff past it?
Yep, I believe so... Need to check again

> Anyways it could lead to consumers reprocessing the whole log
Worst case, yeah. We don't have any guarantees. If, for whatever reason the `redpanda_migrator_offsets` input emits updates very, very late, then destination consumers can end up starting to process the topic from scratch in the current design.

PR #3030, opened 2024-11-22
Remove cgo builds

Jeffail:
This PR removes the `-cgo` suffixed docker images from our build and release process. The reason for this change is that security vulnerabilities are being reported on the base of this image (Debian) and the release cadence for security updates is unacceptable to us.

The only major distinction these builds have is that they support `zmq4` as an input and output. However, based on the metrics that we have this image is rarely used and these connectors even less so. Therefore, the few users that are impacted are probably better served on a case by case basis than by trying to build a new image with a different base.

When reviewing please take extra note of the changelog entry as the important thing here is to give users a clear and concise description of why we've made this change.


CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ Changelog
 
 All notable changes to this project will be documented in this file.
 
+## 4.41.0 - TBD
+
+### Changed
+
+- The `-cgo` suffixed docker images are no longer built and pushed along with the regular images. This decision was made due to low demand, and the unacceptable cadence with which the image base (Debian) receives security updates. It is still possible to create your own builds using the same methods as before.
```
rockwotj, line 10: Do we want to specifically add the build command for users? Just to remind them they need to add the tag?

rockwotj (approved):
LGTM

If we do ever want to bring this back, I have had success with [distroless](https://github.com/GoogleContainerTools/distroless) in the past and just adding the packages required instead of a bunch of extra stuff that has these random security vulnerabilities

CHANGELOG.md
Jeffail: Yeah that makes sense, added.

PR #3037, opened 2024-11-26
fix(sftp): make sure to delete last file when `watch` and `delete_on_finish` are enabled

### LINKED ISSUE #2435: SFTP input - last file not deleted
Hi 👋 

I have discovered an issue when the sftp input component is configured with watcher enabled and delete_on_finish enabled. 

Steps to reproduce:
- Upload 10 files to the SFTP
- Run the benthos pipeline (see below for config)
- After the files have been consumed by the pipeline, check the SFTP => one file is still there even if it has been consumed by benthos.

Here is the configuration file I am using
```yaml
cache_resources:
  - label: memory
    memory:
      default_ttl: 5m
      compaction_interval: 60s
      init_values: {}

input:
  sftp:
    address: ${SFTP_HOST}
    credentials:
      username: ${SFTP_USERNAME}
      password: ${SFTP_PASSWORD}
    paths:
      - "folder/**"
    scanner:
      to_the_end: {}
    delete_on_finish: true
    watcher:
      enabled: true
      minimum_age: 1s
      poll_interval: 20ms
      cache: memory

pipeline:
  processors:
    - log:
        level: INFO
        message: "message received"

output:
  label: ""
  stdout:
    codec: lines
```

ooesili:
Fixes #2435 

## Questions

I believe I have fixed the underlying issue, but I am not sure how to write an integration test to verify the fix. I have created a new integration test function with a `TODO` comment on where I got stuck. The questions I have around this are:

- My plan was to start a pipeline with `watch` and `delete_on_finished` enabled, the use an SFTP client directly to inspect which files exist on the server to make sure they are all deleted after the pipeline runs. However, I'm not sure how to actually run the pipeline. Is too specific of a test to run using `integration.StreamTests()`, and if not, could you point me in the right direction?
- The other pattern I've seen would be to call `newSFTPReaderFromParsed()` directly from the tests then use `Connect()`, and `ReadBatch()` to interact with the plugin. However this plugin appears to be unusually structured in the way that it progresses through the input files. What it does is finds the first file in `Connect()` and sets up the scanner for the file. In `ReadBatch()`, when the file is exhausted, `ReadBatch()` returns `service.ErrNotConnected` which will cause the engine to re-run `Connect()` which advances to the next file. If the plugin only required `Connect()` to be called once, I would be happy to drive the plugin directly in the tests, but because of the reconnection logic required, I was hesitant to reimplement the reconnection loop in the tests. Is there a utility somewhere that I can use from a test that implements the reconnect logic?

### DIFF
--- a/internal/impl/sftp/input.go
+++ b/internal/impl/sftp/input.go
@@ -187,7 +187,6 @@ func (s *sftpReader) Connect(ctx context.Context) (err error) {
 	if s.scanner, err = s.scannerCtor.Create(file, func(ctx context.Context, aErr error) (outErr error) {
 		_ = s.pathProvider.Ack(ctx, nextPath, aErr)
 		if aErr != nil {
-			s.log.Errorf("skipping delete on finish: %s", aErr)
 			return nil
 		}
 		if s.deleteOnFinish {
@@ -223,7 +222,7 @@ func (s *sftpReader) Connect(ctx context.Context) (err error) {
 	return
 }
 
-func (s *sftpReader) seekNextPath(ctx context.Context) (file *sftp.File, nextPath string, skip bool, err error) {
+func (s *sftpReader) initState(ctx context.Context) (client *sftp.Client, pathProvider pathProvider, skip bool, err error) {
 	s.scannerMut.Lock()
 	defer s.scannerMut.Unlock()
 
@@ -242,11 +241,22 @@ func (s *sftpReader) seekNextPath(ctx context.Context) (file *sftp.File, nextPat
 		s.pathProvider = s.getFilePathProvider(ctx)
 	}
 
+	return s.client, s.pathProvider, false, nil
+}
+
+func (s *sftpReader) seekNextPath(ctx context.Context) (file *sftp.File, nextPath string, skip bool, err error) {
+	client, pathProvider, skip, err := s.initState(ctx)
+	if err != nil || skip {
+		return
+	}
+
 	for {
-		if nextPath, err = s.pathProvider.Next(ctx, s.client); err != nil {
+		if nextPath, err = pathProvider.Next(ctx, client); err != nil {
 			if errors.Is(err, sftp.ErrSshFxConnectionLost) {
-				_ = s.client.Close()
+				_ = client.Close()
+				s.scannerMut.Lock()
 				s.client = nil
+				s.scannerMut.Unlock()
 				return
 			}
 			if errors.Is(err, errEndOfPaths) {
@@ -255,21 +265,23 @@ func (s *sftpReader) seekNextPath(ctx context.Context) (file *sftp.File, nextPat
 			return
 		}
 
-		if file, err = s.client.Open(nextPath); err != nil {
+		if file, err = client.Open(nextPath); err != nil {
 			if errors.Is(err, sftp.ErrSshFxConnectionLost) {
-				_ = s.client.Close()
+				_ = client.Close()
+				s.scannerMut.Lock()
 				s.client = nil
+				s.scannerMut.Unlock()
 			}
 
 			s.log.With("path", nextPath, "err", err.Error()).Warn("Unable to open previously identified file")
 			if os.IsNotExist(err) {
 				// If we failed to open the file because it no longer exists
 				// then we can "ack" the path as we're done with it.
-				_ = s.pathProvider.Ack(ctx, nextPath, nil)
+				_ = pathProvider.Ack(ctx, nextPath, nil)
 			} else {
 				// Otherwise we "nack" it with the error as we'll want to
 				// reprocess it again later.
-				_ = s.pathProvider.Ack(ctx, nextPath, err)
+				_ = pathProvider.Ack(ctx, nextPath, err)
 			}
 		} else {
 			return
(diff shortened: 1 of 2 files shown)


rockwotj:
I don't think there is a utility so either you need to do option 1 or implement the retry logic - which I don't think should be too bad? 

Here's the code that drives this in benthos AFAIK: https://github.com/redpanda-data/benthos/blob/dad70374cd8fb323f0c7f47452498ea94c2ed7aa/internal/component/input/async_reader.go#L115

The pipeline option (number 1) might be the best route, but I'm not too familiar with that test helper myself.

internal/impl/sftp/integration_test.go
```diff
@@ -129,6 +109,133 @@ cache_resources:
...
+`
+	config = strings.NewReplacer(
+		"$PORT", resource.GetPort("22/tcp"),
+	).Replace(config)
+
+	var receivedPathsMut sync.Mutex
+	var receivedPaths []string
+
+	builder := service.NewStreamBuilder()
+	require.NoError(t, builder.SetYAML(config))
```
ooesili, line 156: @rockwotj 
> I don't think there is a utility so either you need to do option 1 or implement the retry logic - which I don't think should be too bad?
> 
> Here's the code that drives this in benthos AFAIK: https://github.com/redpanda-data/benthos/blob/dad70374cd8fb323f0c7f47452498ea94c2ed7aa/internal/component/input/async_reader.go#L115
> 
> The pipeline option (number 1) might be the best route, but I'm not too familiar with that test helper myself.

Knowledge from the great and powerful @mihaitodor

internal/impl/sftp/input.go
```diff
@@ -363,61 +386,62 @@ type watcherPathProvider struct {
...
+
+					// We process it if the marker is a pending symbol (!) and we're
+					// polling for the first time, or if the path isn't found in the
+					// cache.
+					//
+					// If we got an unexpected error obtaining a marker for this
+					// path from the cache then we skip that path because the
+					// watcher will eventually poll again, and the cache.Get
+					// operation will re-run.
+					if v, err := cache.Get(ctx, path); errors.Is(err, service.ErrKeyNotFound) || (!w.followUpPoll && string(v) == "!") {
```
rockwotj, line 431: nit: separately it would be nice to use constants for the pending symbol and other cache values :)

internal/impl/sftp/input.go
```diff
@@ -174,11 +174,61 @@ func newSFTPReaderFromParsed(conf *service.ParsedConfig, mgr *service.Resources)
...
+	return
+}
+
+func (s *sftpReader) initState(ctx context.Context) (client *sftp.Client, pathProvider pathProvider, skip bool, err error) {
 	s.scannerMut.Lock()
 	defer s.scannerMut.Unlock()
 
 	if s.scanner != nil {
-		return nil
+		skip = true
```
rockwotj, line 230: Why do we skip if there is a scanner? Can you add a comment?

internal/impl/sftp/input.go
```diff
@@ -174,11 +174,62 @@ func newSFTPReaderFromParsed(conf *service.ParsedConfig, mgr *service.Resources)
...
+				defer func() {
+					_ = client.Close()
+				}()
+			}
+			if outErr == nil {
+				if outErr = client.Remove(nextPath); outErr != nil {
+					outErr = fmt.Errorf("remove %v: %w", nextPath, outErr)
+				}
+			}
+			s.scannerMut.Unlock()
```
rockwotj, line 208: since we always return after this block we could defer this right? I just get worried if the unlock is not right after.

internal/impl/sftp/input.go
```diff
@@ -174,11 +174,62 @@ func newSFTPReaderFromParsed(conf *service.ParsedConfig, mgr *service.Resources)
...
+	if err != nil {
+		return err
+	}
+	if skip {
+		return nil
+	}
+
+	details := service.NewScannerSourceDetails()
+	details.SetName(nextPath)
+	if s.scanner, err = s.scannerCtor.Create(file, func(ctx context.Context, aErr error) (outErr error) {
```
rockwotj, line 187: I think the assignment of `s.scanner` needs to be under the mutex as well based on the `ReadBatch` function right?

internal/impl/sftp/input.go
```diff
@@ -242,11 +241,22 @@ func (s *sftpReader) seekNextPath(ctx context.Context) (file *sftp.File, nextPat
 		s.pathProvider = s.getFilePathProvider(ctx)
 	}
 
+	return s.client, s.pathProvider, false, nil
```
rockwotj, line 244: As a newcomer to this code this feels unsafe.

If the only important part of this code that all these variables are accessed/set together atomically, I do wonder if an atomic is better suited. You can use `Swap` to set the new value and destroy in `Close`, `Store` in Connect and `Load` in ReadBatch. I don't quite understand the higher level contract here of **why** it's only required that they are accessed concurrently and we don't have to worry about `Close` clobbering something ongoing in `Connect` or `ReadBatch`.

internal/impl/sftp/integration_test.go
rockwotj: Great tests thanks :)

internal/impl/sftp/input.go
rockwotj: And when I talk about using atomics I mean using `typed.AtomicValue` in our typed package in `internal/typed` and wrapping all this state into a struct.

ooesili:
Closed in favor of #3073

PR #3039, opened 2024-11-27
bigquery: suppport parquet file loading

rockwotj:


internal/impl/gcp/output_bigquery.go
```diff
@@ -276,7 +283,9 @@ type gcpBigQueryOutput struct {
 
 	fieldDelimiterBytes []byte
 	csvHeaderBytes      []byte
-	newLineBytes        []byte
+	// if nil, then this is a format that we expect to be created downstream in a processor and each
```
mihaitodor, line 286: I might be misreading this, but are the processors not upstream?

mihaitodor (approved):
LGTM, just one small question then 🐑 🚀

internal/impl/gcp/output_bigquery.go
rockwotj: Lol 🙃

PR #3040, opened 2024-11-27
Update templating.adoc

peczenyj:
remove `-c` option from connect run since it is no longer supported


mihaitodor (approved):
Nice catch! ❤️

Issue #3038, opened 2024-11-27
pg_stream: UUIDs are not handled as strings when handling changes


ntx-ben:
I'm using the newly released `v1.41.0` version.

I'm trying the `pg_stream` input connector and am receiving an initial snapshot (`stream_snapshot: true`) of a particular table to pump data into kafka. My table contains various columns, the primary key being `machine_id` of type UUID. 

After making a change in a specific record in that table, I receive the updated record in redpanda-connect. However, I noticed that the primary key is now an array instead of a string (which is the native type for UUID in postgres I assume):

![image](https://github.com/user-attachments/assets/07f3cb0b-34ef-4ec2-9fd6-3e38977bc754)

How can I stream the same data structure consistently using `pg_stream` input connector? Thanks!

mihaitodor:
Hey @ntx-ben 👋 I guess we're not doing anything special on our end (maybe @rockwotj has more details), but a quick workaround would be to use a `mapping` processor to fix it with some Bloblang code like this:

```coffee
root = this
root.machine_id = this.machine_id.map_each(i -> i.string()).join("")
```

I'm actually not sure what each number in that array represents or how to convert them into a proper UUID. If you can figure out what the algorithm is, please let me know.

ntx-ben:
Thanks for the quick turn around!

rockwotj:
This will be out in the next release as the fix has landed on main - thanks for the report and giving the output a spin!

ntx-ben:
I tested with `main` and confirm the issue is fixed! JSONB is also handled very nicely now!

Thanks again for your swift support 👍

PR #3041, opened 2024-11-28
Pg cdc types

le-vlad:
PR Fixes issue #3038 with incorrect parsing UUID types during the replication.
Also adds support for complex PG datatypes such as: 

- JSONB
- TEXT[]
- INET
- TSVECTOR
- TSRANGE,
- POINT
- UUID
- INTEGER[]

Added integration tests to make sure types are parsed equally for Snapshot && CDC stages


internal/impl/postgresql/pglogicalstream/snapshotter.go
```diff
@@ -156,6 +159,69 @@ func (s *Snapshotter) prepareScannersAndGetters(columnTypes []*sql.ColumnType) (
...
+		case "JSONB":
+			scanArgs[i] = new(sql.NullString)
+			valueGetters[i] = func(v interface{}) interface{} {
+				payload := v.(*sql.NullString).String
+				if payload == "" {
+					return nil
+				}
+				var dst any
+				if err := json.Unmarshal([]byte(v.(*sql.NullString).String), &dst); err != nil {
+					s.logger.Warnf("Failed to unmarshal JSONB value: %v", err)
```
rockwotj, line 171: Let's return an error instead of logging

internal/impl/postgresql/pglogicalstream/snapshotter.go
```diff
@@ -156,6 +159,69 @@ func (s *Snapshotter) prepareScannersAndGetters(columnTypes []*sql.ColumnType) (
...
+
+				return dst
+			}
+		case "INET":
+			scanArgs[i] = new(sql.NullString)
+			valueGetters[i] = func(v interface{}) interface{} {
+				inet := pgtype.Inet{}
+				val := v.(*sql.NullString).String
+				if err := inet.Scan(val); err != nil {
+					s.logger.Warnf("Failed to scan array of INT4 values: %v", err)
```
rockwotj, line 182: Let's change the return type to support errors

internal/impl/postgresql/pglogicalstream/replication_message_decoders.go
```diff
@@ -148,7 +150,28 @@ func decodePgOutput(WALData []byte, relations map[uint32]*RelationMessage, typeM
...
 func decodeTextColumnData(mi *pgtype.Map, data []byte, dataType uint32) (interface{}, error) {
 	if dt, ok := mi.TypeForOID(dataType); ok {
-		return dt.Codec.DecodeValue(mi, dataType, pgtype.TextFormatCode, data)
+		val, err := dt.Codec.DecodeValue(mi, dataType, pgtype.TextFormatCode, data)
+		if err != nil {
+			return val, err
+		}
+
+		if dt.Name == "uuid" {
+			typesValueForUUID := val.([16]uint8)
```
rockwotj, line 159: This will panic if the cast fails. Can we also check if the type is correct and return an error if not?

internal/impl/postgresql/pglogicalstream/replication_message_decoders.go
```diff
@@ -148,7 +150,28 @@ func decodePgOutput(WALData []byte, relations map[uint32]*RelationMessage, typeM
...
+			return uuid.UUID(typesValueForUUID).String(), err
```
rockwotj, line 160: nit just use nil error directly

rockwotj (approved):

PR #3044, opened 2024-11-28
Fix `avro` scanner

mihaitodor:


rockwotj (approved):
Thank you for the test!

PR #3049, opened 2024-11-29
Add license checks for enterprise features

Jeffail:
The behavior is that in the absence of a valid enterprise license redpanda connect will refuse to run configs using enterprise features.

How exactly a license is provided is as follows:

- If the cli run flag `--redpanda-license` is provided then its contents will be parsed as a redpanda license
- If the environment variable `REDPANDA_LICENSE_FILEPATH` is defined then the file it points to will be read and parsed as a license
- If a file in the path `/etc/redpanda/redpanda.license` exists then it will be read and parsed as a license file

If any of the above is attempted and fails then an error log will be provided but the service will continue to run as though no license were provided. This means that a failed license parse will only cause the service to fail to run if enterprise features are being used.


internal/license/shared_service.go
```diff
@@ -0,0 +1,62 @@
...
+
+// CheckRunningEnterprise returns a non-nil error if the instance of Redpanda
+// Connect is not operating with a valid enterprise license.
+func CheckRunningEnterprise(res *service.Resources) error {
+	l, err := LoadFromResources(res)
+	if err != nil {
+		return err
+	}
+	if !l.AllowsEnterpriseFeatures() {
+		return errors.New("this feature requires a valid enterprise license https://docs.redpanda.com/current/get-started/licenses/")
```
BenPope, line 43: In core we have something more like:
```suggestion
		return errors.New("this feature requires a valid enterprise license. 
To request an Enterprise license, please visit https://redpanda.com/upgrade. To try Redpanda Enterprise for 30 days, visit https://redpanda.com/try-enterprise. For more information, see https://docs.redpanda.com/current/get-started/licenses")

**Comment by rockwotj** [comment]
I wonder if it would be nice to have an additional environment variable that took license file contents, incase of environments where it's hard to modify a container filesystem or its existing practice to use only env vars for configuration/secrets


Down the road it would be awesome if secrets stuff could load licenses too

File: internal/license/service.go [hunk 2]
```diff
@@ -0,0 +1,215 @@
...
+
+	// Followed by the default file path.
+	if licenseFileContents, err = os.ReadFile(s.conf.defaultLicenseFilepath()); err != nil {
+		if !os.IsNotExist(err) {
+			return nil, fmt.Errorf("failed to read default path license file: %w", err)
+		}
+		return nil, nil
+	} else {
+		s.logger.Debug("Loaded Redpanda Enterprise license from default file path")
+	}
```
mihaitodor, line 162: Nit: Redundant else

mihaitodor (approved):
LGTM 🐑 🚀

PR #3050, opened 2024-11-29
release: push RCs to unstable repo of cloudsmith

gousteris:
only GA versions should be pushed to the `redpanda` repo. all others should go to `redpanda-unstable`


Jeffail (approved):
LGTM, I'm about to trigger an RC so we can test this today.

PR #3053, opened 2024-12-02
Add error log when trial licenses are being used

Jeffail:
Most of our docs and pages for enterprise licenses are heavily pushing for generating trial licenses. However, we cannot support trial licenses in connect, and therefore these licenses are considered invalid for the purposes of running enterprise connect features.

This leaves us vulnerable to a poor user experience where a user tries to run enterprise features, is told by RPCN to visit https://cloud.redpanda.com/try-enterprise, where we then encourage them to create a trial (for Redpanda proper). When a user attempts to run the trial we will tell them it's invalid from Connect.

In order to at least clarify what is happening to the user I've added an explicit error log that notes the license is a trial, not a full license that connect will recognize.


mihaitodor (approved):
:shipit:

PR #3056, opened 2024-12-03
Fix ga tag pattern and add tests

gousteris:
This PR:
* fixes GA tag pattern
goreleaser runs on an ubuntu VM managed by gh actions. ubuntu doesn't
understand `\d` for digits. Instead, using the equivalent pattern
`[0-9]` that works fine on ubuntu:

```
$ version=0.0.0
$ pattern='^\d+\.\d+\.\d+$'
$ if [[ $version =~ $pattern ]]; then echo match; fi
$ pattern='^[0-9]+\.[0-9]+\.[0-9]+$'
$ if [[ $version =~ $pattern ]]; then echo match; fi
match
```
* adds gh action step to test push_pkg_to_cloudsmith.sh

### DIFF
--- a/resources/scripts/push_pkg_to_cloudsmith.sh
+++ b/resources/scripts/push_pkg_to_cloudsmith.sh
@@ -40,7 +40,7 @@ else
     version=$PKG_VERSION
 fi
 
-GA_VERSION_PATTERN='^\d+\.\d+\.\d+$'
+GA_VERSION_PATTERN='^[0-9]+\.[0-9]+\.[0-9]+$'
 if [[ $version =~ $GA_VERSION_PATTERN ]]; then
   repo="redpanda"
 else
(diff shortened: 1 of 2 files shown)


resources/scripts/push_pkg_to_cloudsmith.sh
```diff
@@ -40,7 +40,7 @@ else
     version=$PKG_VERSION
 fi
 
-GA_VERSION_PATTERN='^\d+\.\d+\.\d+$'
+GA_VERSION_PATTERN='^[0-9]+\.[0-9]+\.[0-9]+$'
```
ivotron, line 43: would help if the commit message explains why this change is needed

ivotron (changes requested):

ivotron (approved):

peczenyj:
It os already done but you can also use posix classes like `[[:digit:]]`, if you think it is more readable

PR #3059, opened 2024-12-03
pgcdc: fix concurrent connection usage

rockwotj:
I tried to be good about writing commit messages to explain everything, but TL;DR is that we move all connection usage to postgres to a single thread to prevent concurrent usage (which is unsafe), and also removed the performance bottleneck when not emitting txn markers

- **pgcdc: rename Lsn to LSN**
- **pgcdc: make sure LSNs are lexicographical**
- **pgcdc: remove extra roundtrip for LSNs**
- **pgcdc: remove plugin handlers**
- **pgcdc: remove watermark**
- **pgcdc: don't regress the status update**
- **pgcdc: move all pg conn usage to a single goroutine**


emaxerrno:
Very cool fix

ooesili (approved):

PR #3063, opened 2024-12-05
Fix panic in redpanda_migrator output

mihaitodor:

PR #3065, opened 2024-12-06
Fix schema ID translation in `redpanda_migrator_bundle` output

mihaitodor:
We want to skip this when `translate_schema_ids: false` even if there's `schema_registry` component configured under `redpanda_migrator_bundle`.

This is a follow-up to #3063.


Jeffail (approved):
LGTM, thanks @mihaitodor

rockwotj (approved):
Looks good to me! Thanks for the fix

PR #3073, opened 2024-12-11
Sftp refactor

ooesili:
Before this commit, when a file was exuasted the `ReadBatch` method
returned ErrNotConnected which cause the engine to call `Connect` again.
Aside from being awkward, this causes the connection status to
incorrectly be reported as disconnected during normal operation.

This commit moves the logic to advance to the next file when the current
file is exhuasted into a the ReadBatch method.

Builds on top of #3037

### DIFF
--- a/internal/impl/sftp/input.go
+++ b/internal/impl/sftp/input.go
@@ -213,7 +213,7 @@ func (s *sftpReader) ReadBatch(ctx context.Context) (service.MessageBatch, servi
 	s.stateLock.Lock()
 	defer s.stateLock.Unlock()
 
-	parts, codecAckFn, err := s.tryReadBatchV2(ctx)
+	parts, codecAckFn, err := s.tryReadBatch(ctx)
 	if errors.Is(err, sftp.ErrSSHFxConnectionLost) {
 		s.closeScanner(ctx)
 		s.closeClient()
@@ -222,7 +222,7 @@ func (s *sftpReader) ReadBatch(ctx context.Context) (service.MessageBatch, servi
 	return parts, codecAckFn, nil
 }
 
-func (s *sftpReader) tryReadBatchV2(ctx context.Context) (service.MessageBatch, service.AckFunc, error) {
+func (s *sftpReader) tryReadBatch(ctx context.Context) (service.MessageBatch, service.AckFunc, error) {
 	client, cleanup, err := s.initClient()
 	if err != nil {
 		return nil, nil, err
@@ -363,11 +363,11 @@ type pathProvider interface {
 	Ack(context.Context, string, error) error
 }
 
-type staticPathProviderV2 struct {
+type staticPathProvider struct {
 	expandedPaths []string
 }
 
-func (s *staticPathProviderV2) Next(context.Context, *sftp.Client) (string, bool, error) {
+func (s *staticPathProvider) Next(context.Context, *sftp.Client) (string, bool, error) {
 	if len(s.expandedPaths) == 0 {
 		return "", false, nil
 	}
@@ -376,7 +376,7 @@ func (s *staticPathProviderV2) Next(context.Context, *sftp.Client) (string, bool
 	return path, true, nil
 }
 
-func (s *staticPathProviderV2) Ack(context.Context, string, error) error {
+func (s *staticPathProvider) Ack(context.Context, string, error) error {
 	return nil
 }
 
@@ -466,7 +466,7 @@ func (w *watcherPathProvider) Ack(ctx context.Context, name string, err error) (
 
 func (s *sftpReader) getFilePathProvider(client *sftp.Client) pathProvider {
 	if !s.watcherEnabled {
-		provider := &staticPathProviderV2{}
+		provider := &staticPathProvider{}
 		for _, path := range s.paths {
 			expandedPaths, err := client.Glob(path)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


internal/impl/sftp/input.go
```diff
@@ -173,180 +174,209 @@ func newSFTPReaderFromParsed(conf *service.ParsedConfig, mgr *service.Resources)
...
+			s.log.With("error", err).Error("Failed to close scanner")
+		}
+		s.scanner = nil
+		s.currentPath = ""
+	}
+}
 
-var errEndOfPaths = errors.New("end of paths")
+func (s *sftpReader) closeClient() {
+	if s.client == nil {
```
mihaitodor, line 353: Not sure if it's an issue, but when `s.closed == true`, can we get multiple concurrent calls to this `closeClient()` method via the `initClient()` `cleanup` which is called in the ackFn callback. If yes, I'm thinking that one goroutine could close the client while the other one is just preparing to remove the file.

mihaitodor (commented):
Hey @ooesili, thanks for bearing with me for a review on this! Really awesome job man 🏆 This does make it easier to go through the code. I think I saw a few potential issues, but should be good otherwise.

I think @rockwotj also left a few comments in #3037. Please have a look and then I'm happy to merge both PRs.

internal/impl/sftp/input.go
```diff
@@ -173,177 +176,243 @@ func newSFTPReaderFromParsed(conf *service.ParsedConfig, mgr *service.Resources)
...
+func (s *sftpReader) tryReadBatch(ctx context.Context) (service.MessageBatch, service.AckFunc, error) {
+	if err := s.initScanner(ctx); err != nil {
+		return nil, nil, err
 	}
 
-	parts, codecAckFn, err := scanner.NextBatch(ctx)
+	s.stateLock.Lock()
+	defer s.stateLock.Unlock()
+
+	parts, codecAckFn, err := s.scanner.NextBatch(ctx)
```
Jeffail, line 258: I'm not sure it's actually possible to be hit in practice but there's a race condition here, as the lock around `s.scanner` is yielded within `initScanner` before the lock is re-acquired, so there's opportunity for another goroutine to set `s.scanner` to `nil` in that gap. If you want to avoid this then have `initScanner` return the scanner pointer and use that reference.

internal/impl/sftp/input.go
```diff
@@ -173,177 +176,243 @@ func newSFTPReaderFromParsed(conf *service.ParsedConfig, mgr *service.Resources)
...
+
+		details := service.NewScannerSourceDetails()
+		details.SetName(path)
+		scanner, err := s.scannerCtor.Create(file, s.newCodecAckFn(path), details)
+		if err != nil {
+			_ = file.Close()
+			return fmt.Errorf("creating scanner: %w", err)
+		}
+
+		s.stateLock.Lock()
```
Jeffail, line 328: Re-acquiring the lock here means there's a period of time where another goroutine could potentially have used the client to open a file, create a scanner and assign it, and now this goroutine is going to overwrite `s.scanner`, which means it's lost. I think in terms of performance you should be fine holding the lock for the entire method call.

Jeffail (commented):
Found a couple of concurrency issues, I don't think either would actually be encountered in practice, but it's better to hold the locks for longer and be sure.

internal/impl/sftp/input.go
ooesili: Wow good catch. I'll do this

internal/impl/sftp/input.go
ooesili: The problem holding the lock for the entire call is that in watch mode it will hold the lock open while `s.pathProvider.Next()` blocks and waits for a new file, which will cause AckFns to block because they need to grab the lock for a bit too to access `s.pathProvider`

internal/impl/sftp/input.go
```diff
@@ -173,184 +173,190 @@ func newSFTPReaderFromParsed(conf *service.ParsedConfig, mgr *service.Resources)
...
+			return nil
 		}
+
+		if s.deleteOnFinish {
+			if err := s.client.Remove(path); err != nil {
+				return fmt.Errorf("remove %v: %w", path, err)
+			}
+		}
+
+		time.Sleep(time.Millisecond * 100)
```
mihaitodor, line 310: Is this intentional? (If yes, a comment would be handy)

internal/impl/sftp/input.go
```diff
@@ -433,21 +448,22 @@ func (w *watcherPathProvider) Ack(ctx context.Context, name string, err error) (
...
+		provider := &staticPathProvider{}
+		for _, path := range s.paths {
+			expandedPaths, err := client.Glob(path)
 			if err != nil {
-				s.log.Warnf("Failed to scan files from path %v: %v", p, err)
+				s.log.Warnf("Failed to scan files from path %v: %v", path, err)
 				continue
 			}
-			filepaths = append(filepaths, paths...)
+			provider.expandedPaths = append(provider.expandedPaths, expandedPaths...)
```
mihaitodor, line 460: This `append` looks redundant since `provider.expandedPaths` is initialised on line 453

mihaitodor (approved):
Nice job @ooesili! 🏆 Thanks for bearing with us on this review! The pool implementation looks like a great idea!

I left a small comment and please don't forget to add a note in the Changelog.

Feel free to merge if the stuff Ash mentioned is sorted.

internal/impl/sftp/input.go
ooesili: Nope, removed

internal/impl/sftp/input.go
Jeffail: Ah I see 👍

Jeffail (approved):
LGTM 🎉

PR #3075, opened 2024-12-12
Fix `postgres_cdc` input

mihaitodor:
Allow quoted identifiers for the table names.

Note: This is a very naive implementation and we might want to instead just quote the identifier like @rockwotj suggested (and implemented [here](https://github.com/redpanda-data/connect/blob/main/internal/impl/snowflake/streaming/compat.go#L189) for Snowflake).

If we always quote, then it's a bit unclear how to validate the identifiers because we can't tell when it genuinely needs to be quoted to apply the alternate validation rules: https://www.postgresql.org/docs/current/sql-syntax-lexical.html

> Quoted identifiers can contain any character, except the character with code zero. (To include a double quote, write two double quotes.) This allows constructing table or column names that would otherwise not be possible, such as ones containing spaces or ampersands. The length limitation still applies.

Also, users might get confused if they specify the table name as upper case in the config for a non-quoted table:

```sql
> CREATE TABLE MIHAI(ID INT NOT NULL, CATEGORY VARCHAR);
> INSERT INTO "MIHAI"(ID, CATEGORY) VALUES(13, 'FOOBAR');
relation "MIHAI" does not exist
LINE 1: INSERT INTO "MIHAI"(ID, CATEGORY) VALUES(13, 'FOOBAR')
```

## How to test

Run Postgres and connect to it via `pgcli`:

```shell
$ docker run --rm -it -e POSTGRES_USER=testuser -e "POSTGRES_PASSWORD=testpass" -e POSTGRES_DB=testdb -p5432:5432 postgres -c wal_level=logical
$ pgcli postgres://testuser:testpass@localhost:5432/testdb
```

Create a quoted table:

```sql
> CREATE SCHEMA "TestABC";
> CREATE TABLE "TestABC"."FooBar"(ID INT NOT NULL, CATEGORY VARCHAR);
> CREATE TABLE "TestABC"."BarFoo"(ID INT NOT NULL, CATEGORY VARCHAR);
```

Run Connect:

```yaml
input:
  postgres_cdc:
    dsn: postgres://testuser:testpass@localhost:5432/testdb?sslmode=disable
    include_transaction_markers: true
    stream_snapshot: false
    snapshot_memory_safety_factor: 1
    snapshot_batch_size: 0
    schema: public # No default (required)
    tables:
      - '"FooBar"'
      - '"BarFoo"'
    checkpoint_limit: 1024
    temporary_slot: false
    slot_name: "mihaifoo" # No default (optional)
    pg_standby_timeout: 10s
    pg_wal_monitor_interval: 3s
    max_parallel_snapshot_tables: 1
    auto_replay_nacks: true

output:
  stdout: {}
```

Insert rows:

```sql
> INSERT INTO "TestABC"."FooBar"(ID, CATEGORY) VALUES(13, 'FOOBAR');
> INSERT INTO "TestABC"."BarFoo"(ID, CATEGORY) VALUES(13, 'FOOBAR');
```

Comment out the `"BarFoo"` table in the Connect config, restart it and insert another row in `"FooBar"`.

### DIFF
--- a/internal/impl/postgresql/pglogicalstream/pglogrepl.go
+++ b/internal/impl/postgresql/pglogicalstream/pglogrepl.go
@@ -21,7 +21,6 @@ import (
 	"context"
 	"database/sql/driver"
 	"encoding/binary"
-	"errors"
 	"fmt"
 	"slices"
 	"strconv"
@@ -343,7 +342,7 @@ func DropReplicationSlot(ctx context.Context, conn *pgconn.PgConn, slotName stri
 }
 
 // CreatePublication creates a new PostgreSQL publication with the given name for a list of tables and drop if exists flag
-func CreatePublication(ctx context.Context, conn *pgconn.PgConn, publicationName string, schema string, tables []string) error {
+func CreatePublication(ctx context.Context, conn *pgconn.PgConn, publicationName string, tables []TableFQN) error {
 	// Check if publication exists
 	pubQuery, err := sanitize.SQLQuery(`
 			SELECT pubname, puballtables
@@ -352,17 +351,6 @@ func CreatePublication(ctx context.Context, conn *pgconn.PgConn, publicationName
 		`, publicationName)
 	if err != nil {
 		return fmt.Errorf("failed to sanitize publication query: %w", err)
-	}
-
-	// Since we need to pass table names without quoting, we need to validate it
-	for _, table := range tables {
-		if err := sanitize.ValidatePostgresIdentifier(table); err != nil {
-			return errors.New("invalid table name")
-		}
-	}
-	// the same for publication name
-	if err := sanitize.ValidatePostgresIdentifier(publicationName); err != nil {
-		return errors.New("invalid publication name")
 	}
 
 	result := conn.Exec(ctx, pubQuery)
@@ -374,16 +362,13 @@ func CreatePublication(ctx context.Context, conn *pgconn.PgConn, publicationName
 
 	tablesClause := "FOR ALL TABLES"
 	if len(tables) > 0 {
-		// quotedTables := make([]string, len(tables))
-		// for i, table := range tables {
-		// 	// Use sanitize.SQLIdentifier to properly quote and escape table names
-		// 	quoted, err := sanitize.SQLIdentifier(table)
-		// 	if err != nil {
-		// 		return fmt.Errorf("invalid table name %q: %w", table, err)
-		// 	}
-		// 	quotedTables[i] = quoted
-		// }
-		tablesClause = "FOR TABLE " + strings.Join(tables, ", ")
+		tablesClause = "FOR TABLE "
+		for i, table := range tables {
+			if i > 0 {
+				tablesClause += ", "
+			}
+			tablesClause += table.String()
+		}
 	}
 
 	if len(rows) == 0 || len(rows[0].Rows) == 0 {
@@ -403,7 +388,7 @@ func CreatePublication(ctx context.Context, conn *pgconn.PgConn, publicationName
 
 	// assuming publication already exists
 	// get a list of tables in the publication
-	pubTables, forAllTables, err := GetPublicationTables(ctx, conn, publicationName, schema)
+	pubTables, forAllTables, err := GetPublicationTables(ctx, conn, publicationName)
 	if err != nil {
 		return fmt.Errorf("failed to get publication tables: %w", err)
 	}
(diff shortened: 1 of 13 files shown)


internal/impl/postgresql/pglogicalstream/pglogrepl.go
```diff
@@ -457,14 +457,15 @@ func CreatePublication(ctx context.Context, conn *pgconn.PgConn, publicationName
...
 // GetPublicationTables returns a list of tables currently in the publication
 // Arguments, in order: list of the tables, exist for all tables, errror
-func GetPublicationTables(ctx context.Context, conn *pgconn.PgConn, publicationName string) ([]string, bool, error) {
+func GetPublicationTables(ctx context.Context, conn *pgconn.PgConn, publicationName, schema string) ([]string, bool, error) {
 	query, err := sanitize.SQLQuery(`
 		SELECT DISTINCT
 			tablename as table_name
 		FROM pg_publication_tables
 		WHERE pubname = $1
+		AND schemaname = $2
```
ligfx, line 466: This should _return_ the schema names rather than filtering on them, right?

internal/impl/postgresql/pglogicalstream/sanitize/sanitize.go
```diff
@@ -366,6 +366,18 @@ func SQLQuery(sql string, args ...any) (string, error) {
 // ValidatePostgresIdentifier checks if a string is a valid PostgreSQL identifier
 // This follows PostgreSQL's standard naming rules
 func ValidatePostgresIdentifier(name string) error {
+	if parts := strings.Split(name, "."); len(parts) == 2 {
```
ligfx, line 369: Incorrect if the identifier is quoted. Also not sure of the logic here, if a schema is passed won’t you end up with `schema.schema.table`?

internal/impl/postgresql/pglogicalstream/sanitize/sanitize.go
```diff
@@ -366,6 +366,18 @@ func SQLQuery(sql string, args ...any) (string, error) {
...
+		if err := ValidatePostgresIdentifier(parts[0]); err != nil {
+			return fmt.Errorf("invalid schema identifier: %s", err)
+		}
+		name = parts[1]
+	}
+
+	// Strip quotes if they are present
+	if strings.HasPrefix(name, "\"") && strings.HasSuffix(name, "\"") {
+		name = strings.Trim(name, "\"")
```
ligfx, line 378: Quoted identifiers don’t need the below verification, they follow different rules (no NULs, any double-quote characters must be doubled).

rockwotj: Yes will fixup in a bit

internal/impl/postgresql/pglogicalstream/pglogrepl.go
rockwotj: Yes, done.

internal/impl/postgresql/pglogicalstream/sanitize/sanitize.go
rockwotj: This is gone now

rockwotj:
Okay updates here:
* Added a new type to clarify we always are working with table FQN (with schema)
* We only validate identifiers once - at the package entrypoint
* We validate quoted identifiers

rockwotj (approved):

internal/impl/postgresql/pglogicalstream/pglogrepl.go
mihaitodor: Oh, I see. Thanks for catching that!

internal/impl/postgresql/pglogicalstream/sanitize/sanitize.go
```diff
@@ -363,28 +363,72 @@ func SQLQuery(sql string, args ...any) (string, error) {
 	return query.Sanitize(args...)
 }
 
-// ValidatePostgresIdentifier checks if a string is a valid PostgreSQL identifier
+// QuotePostgresIdentifier returns the valid escaped identifier.
+func QuotePostgresIdentifier(name string) string {
```
ligfx, line 367:
```go
return `"` + name.Replace(`"`, `""`) + `"`
```
?

ligfx (approved):
Looks good to me and works for me locally. Love the normalization of passed in identifiers.

internal/impl/postgresql/pglogicalstream/sanitize/sanitize.go
rockwotj: Yeah that's simpler. This happens to be faster, but this isn't on a critical path. I'll fix it next time I hack on this input

PR #3080, opened 2024-12-13
examples: add an example for polling a database

rockwotj:
As requested by popular demand.

### DIFF
--- a/config/examples/stateful_polling.yaml
+++ b/config/examples/stateful_polling.yaml
@@ -1,0 +1,101 @@
+# This example shows how to periodically poll postgres and fetch a series of records
+# saving a cursor in postgres of the last poll.
+input:
+  generate:
+    interval: '@every 5s'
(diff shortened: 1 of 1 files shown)


config/examples/stateful_polling.yaml
```diff
@@ -0,0 +1,101 @@
...
+      - stdout: {} # TODO: Do your actual output
+      # It's important that the last thing we do is save our state
+      # This allows at-least-once delivery semantics.
+      - processors:
+          # *Assuming* your processing is in order and the output is atomic
+          # then we only need to save the ID of the last message in the cache.
+          - select_parts:
+              parts: [-1]
+          - mapping: |
+              root.id = this.fetched.id
```
rockwotj, line 49: This should change to just looking for a `max` in the batch - it's probably simpler, and I certainly need an ORDER BY in the query above.

PR #3081, opened 2024-12-13
examples: add a postgres replication example using CDC

rockwotj:
Note that we can't dynamically do anything with the schema at the
moment, we don't emit schema changes and creating the migration
statements would require a complicated sql_raw output.

### DIFF
--- a/config/examples/cdc_replication.yaml
+++ b/config/examples/cdc_replication.yaml
@@ -1,0 +1,54 @@
+input:
+  postgres_cdc:
+    dsn: postgres://me:foobar@localhost:5432?sslmode=disable
+    include_transaction_markers: true
+    slot_name: test_slot_native_decoder
+    stream_snapshot: true
+    schema: public
+    tables: [my_src_table]
+    # Group by transaction, each message batch is all rows changed in a transaction
+    # this might be massive, but might be required for foreign key constraints
+    batching:
+      check: '@operation == "commit"'
+      period: 10s
+      processors:
+        # But drop the placeholder messages for start/end transaction
+        - mapping: |
+            root = if @operation == "begin" || @operation == "commit" {
+              deleted()
+            } else {
+              this
+            }
+output:
+  # Dispatch the write based on the operation metadata
+  switch:
+    strict_mode: true
+    cases:
+      - check: '@operation == "read" || @operation == "insert"'
+        output:
+          sql_insert:
+            driver: postgres
+            dsn: postgres://me:foobar@localhost:5432?sslmode=disable
+            table: my_dst_table
+            columns: [id, foo, bar]
+            args_mapping: root = [this.id, this.foo, this.bar]
+            init_statement: |
+              CREATE TABLE IF NOT EXISTS my_dst_table (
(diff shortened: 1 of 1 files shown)


config/examples/cdc_replication.yaml
```diff
@@ -0,0 +1,54 @@
...
+    include_transaction_markers: true
+    slot_name: test_slot_native_decoder
+    stream_snapshot: true
+    schema: public
+    tables: [my_src_table]
+    # Group by transaction, each message batch is all rows changed in a transaction
+    # this might be massive, but might be required for foreign key constraints
+    batching:
+      check: '@operation == "commit"'
+      period: 10s
```
mihaitodor, line 13: This might need a disclaimer that it needs to be large enough to account for massive transactions... Or maybe the `mapping` processor below should use batch-wide queries (`from_all().<insert sanity check here>` when `batch_index() == 0`) to determine if a transaction looks incomplete and throw an error?

mihaitodor (approved):
LGTM, just one small nitpick. Feel free to :shipit:

ligfx:
Two thoughts:

- If `@operation == "begin" || @operation == "commit"` isn't handled explicitly (i.e. they don't care about restraints, as many analytical warehouses don't), what do those messages look like? Will they mess up the following `SQL` code or are they just empty messages?

- Relying on `@operation` to choose `INSERT` and `UPDATE` is in general not safe and will do weird things if messages come out of order or get replayed. Usually, people will either do `INSERT ... ON CONFLICT` or `MERGE` to perform an upsert (and that's what the Kafka Connect JDBC Sink connector does in upsert mode).

  Possible approach using `MERGE` (requires Postgres >= 15):
```sql
sql_raw:
  driver: postgres
  dsn: postgres://michaelmaltese@localhost:5432/michaelmaltese?sslmode=disable
  args_mapping: |
    root = [ this.id, this.name, this.updated_at ]
  query: |
    MERGE INTO journey_apps3_cdc AS old
    USING (SELECT
      $1::integer id,
      $2 name,
      $3::timestamptz updated_at,
    ) AS new
    ON new.id = old.id
    WHEN MATCHED THEN
      UPDATE SET
        name = case when new.updated_at > old.updated_at OR old.updated_at is null THEN new.name ELSE old.name END,
        updated_at = greatest(new.updated_at, old.updated_at)
    WHEN NOT MATCHED THEN
      INSERT (id, name, updated_at) VALUES (
        new.id,
        new.name,
        new.updated_at
      )
    ;
```

ligfx (changes requested):
Upsert comment above. Sorry, I'm a bit rusty with GitHub!

config/examples/cdc_replication.yaml
rockwotj: Added a disclaimer! Thanks ❤️

rockwotj:
> Will they mess up the following SQL code or are they just empty messages?

Yes they are empty and will mess up stuff downstream.

rockwotj:
If you don't have foreign key restraints than batching by transaction is quite useless indeed.

PR #3082, opened 2024-12-14
avro: add metadata of the original schema

rockwotj:


mihaitodor (approved):
Nice, thank you Tyler! 🐑 🚀 

I'm also thinking to have a new config field (not sure what to call it) which allows users to tell the scanner to set the `datum` in the message without passing it through `TextualFromNative`. I'll raise a separate PR for it.

PR #3096, opened 2024-12-19
#3095 AMQP 1.0 Output - Support for message body type

timo102:
See github issue for an explanation: #3095

The word "opaque" is mentioned in the documentation (https://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-data); that is why I chose this word for the configuration option.

## How to test

You need a broker with AMQP 1 support. Example: Apache ActiveMQ with docker-compose:
```yaml
  activemq:
    container_name: activemq
    image: apache/activemq-classic:5.18.3
    restart: always
    ports:
      - 8161:8161
      - 5673:5672
```

Configuration to insert string messages:
```yaml
input:
  generate:
    interval: '@every 2s'
    mapping: root = {"id":uuid_v4(),"a":"а, е, ё, и, о, у, ы, э, ю, я","c":"幸福","b":"äöüÄÖÜßabc"}

output:
  amqp_1:
    urls: [ amqp://localhost:5673/ ]
    target_address: DEMOQUEUE
    max_in_flight: 64
    content_type: string
    sasl:
      mechanism: anonymous
      user: ""
      password: ""
    metadata:
      exclude_prefixes: [ "demo" ]
```
Reading the messages again:
```yaml
input:
  amqp_1:
    urls: [ amqp://localhost:5673/ ]
    source_address: "DEMOQUEUE"
    sasl:
      mechanism: anonymous
      user: ""
      password: ""

output:
  stdout:
    codec: lines
```

PR #3098, opened 2024-12-26
AMQP 1.0 Output - Support for message body type

### LINKED ISSUE #3095: AMQP 1.0 Output - Support for message body type
In AMQP 0.9 the message format was simple, and the message body was always an opaque byte array. In AMQP 1.0, the message body can be 1 of 3 types: data (opaque byte array), AMQP sequence (list of polymorphic data), or AMQP value (a single instance of polymorphic data). In the current implementation, the message is always sent as an opaque byte array. If the use case is to process JSON messages (UTF-8 String), then it is better to use AMQP as it was intended and send the data as AMQP value of type string. This increases compatibility and simplifies the usage of such a message.

My suggestion is to introduce an advanced configuration option called `content_type`. The allowed values are:

- `opaque_binary`: Same behavior as today. This is the default to keep compatibility and prevent a breaking change.
- `string`: New option which sends the message as string content.

The `amqp_1` input module implementation already supports the string message type, so there is no change needed.

I will provide a pull request for this enhancement.

mihaitodor:
This is a cleaned up version of #3096 which should be good to go.

Fixes #3095.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
-- `avro` scanner now emits metadata for the Avro schema it used along with the schema fingerprint (@rockwotj)
+- `avro` scanner now emits metadata for the Avro schema it used along with the schema fingerprint. (@rockwotj)
+- Field `content_type` added to the `amqp_1` output. (@timo102)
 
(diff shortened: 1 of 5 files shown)


rockwotj (approved):
LGTM

Merry Christmas! 🎄

PR #3103, opened 2025-01-04
aws/sqs: prevent ackLoop starvation

rockwotj:
The ackLoop can be starved, we add a max outstanding limit to ensure that stuff gets acked properly.


internal/impl/aws/input_sqs.go
```diff
@@ -181,99 +199,116 @@ func (a *awsSQSReader) Connect(ctx context.Context) error {
...
-			timeoutSeconds: 30,
-			receiptHandle:  *m.ReceiptHandle,
-			addedAt:        time.Now(),
-		}
-		if timeoutStr, exists := m.Attributes[sqsiAttributeNameVisibilityTimeout]; exists {
-			// Might as well keep the queue timeout setting refreshed as we
-			// consume new data.
-			if tmpTimeoutSeconds, err := strconv.Atoi(timeoutStr); err == nil {
-				handle.timeoutSeconds = tmpTimeoutSeconds
-			}
```
rockwotj, line 254: FYI this attribute is not mentioned in the docs and doesn't seem to be present in my testing...

internal/impl/aws/input_sqs.go
```diff
@@ -412,24 +502,35 @@ func (a *awsSQSReader) readLoop(wg *sync.WaitGroup, inFlightTracker *sqsInFlight
...
+	// The timestamp of when the message expires
+	deadline time.Time
 }
 
-func (a *awsSQSReader) deleteMessages(ctx context.Context, msgs ...sqsMessageHandle) error {
+func (a *awsSQSReader) deleteMessages(ctx context.Context, msgs ...*sqsMessageHandle) error {
+	if !a.conf.DeleteMessage {
+		return nil
+	}
+	const maxBatchSize = 10
```
rockwotj, line 520: The API limit here on the amazon side is 10, so currently if you MaxNumberOfMessages > 10 then you can't delete or reset visibility for anything...

internal/impl/aws/input_sqs.go
```diff
@@ -440,48 +541,79 @@ func (a *awsSQSReader) deleteMessages(ctx context.Context, msgs ...sqsMessageHan
 			return err
```
mihaitodor, line 541: Just wondering, if there's some transitive error and `DeleteMessageBatch` fails, we return an error here. However, further below, we only log an error when `fail.Message != nil`. Can that lead to failed messages being retried over and over?

internal/impl/aws/input_sqs.go
```diff
@@ -283,47 +318,44 @@ func (a *awsSQSReader) ackLoop(wg *sync.WaitGroup, inFlightTracker *sqsInFlightT
...
-	}
-
 	flushTimer := time.NewTicker(time.Second)
 	defer flushTimer.Stop()
 
-	// Both maps are of the message ID to the receipt handle
-	pendingAcks := map[string]string{}
-	pendingNacks := map[string]string{}
+	pendingAcks := []*sqsMessageHandle{}
+	pendingNacks := []*sqsMessageHandle{}
```
mihaitodor, line 335: Since these are lists of pointers now, do `deleteMessages()` and `resetMessages()` always get non-nil message handles? Might be worth adding some defensive nil checks in those methods.

internal/impl/aws/input_sqs.go
```diff
@@ -333,21 +365,65 @@ ackLoop:
...
+			}
+			err := a.updateVisibilityMessages(closeNowCtx, int(a.conf.MessageTimeout.Seconds()), currentHandles...)
+			if err == nil {
+				continue
+			}
+			partialErr := &batchUpdateVisibilityError{}
+			if errors.As(err, &partialErr) {
+				for _, fail := range partialErr.entries {
+					// Mitigate erroneous log statements due to the race described above by making sure we're still tracking the message
+					if !inFlightTracker.IsTracking(*fail.Id) {
```
mihaitodor, line 390: Can the `Id` be nil?

internal/impl/aws/input_sqs.go
```diff
@@ -440,48 +541,79 @@ func (a *awsSQSReader) deleteMessages(ctx context.Context, msgs ...sqsMessageHan
...
+	if len(err.entries) == 0 {
+		return "(no failures)"
+	}
+	var msg strings.Builder
+	msg.WriteString("failed to update visibility for messages: [")
+	for i, fail := range err.entries {
+		if i > 0 {
+			msg.WriteByte(',')
+		}
+		msg.WriteString(fmt.Sprintf("'%v'", *fail.Id))
```
mihaitodor, line 575: Can `Id` be nil? Also, we could use `%q` instead of `'%v'`.

mihaitodor (approved):
Nice job! 🏆 Just left some small nitpicky questions, but feel free to :shipit:

internal/impl/aws/input_sqs.go
rockwotj: never nil 😆

mihaitodor: Hahah, yeah :)) I figured ❤️

PR #3107, opened 2025-01-09
Add third party licenses list

Jeffail:


mihaitodor (approved):
:shipit:

rockwotj:
Should we hook this up to CI so we don't forget to update it?

Jeffail:
> Should we hook this up to CI so we don't forget to update it?

Probably but I'd need to fix some false detections first, best to get this in now and we can iterate later.

PR #3108, opened 2025-01-10
chore(mongodb): upgrade driver to version 2.0.0

ooesili:
- `SetSocketTimeout` was removed, and its usage has been replaced with `SetTimeout`
- mongo.Connect` no longer accepts a context because it was never used
- The `WriteConcern` struct no longer contains a timeout field, context is used instead, so I added a wrapper struct with this field to minimize code changes
- The `options` packages has been overhauled so there are a bunch of code changes around setting options

For more information, see this doc https://github.com/mongodb/mongo-go-driver/blob/v2.0.0/docs/migration-2.0.md


internal/impl/mongodb/common.go
```diff
@@ -106,10 +105,7 @@ func getClient(parsedConf *service.ParsedConfig) (client *mongo.Client, database
 		opt.SetAuth(creds)
 	}
 
-	ctx, done := context.WithTimeout(context.Background(), time.Minute)
-	defer done()
-
-	if client, err = mongo.Connect(ctx, opt); err != nil {
+	if client, err = mongo.Connect(opt); err != nil {
```
rockwotj, line 108: Does this no longer do IO?

internal/impl/mongodb/processor.go
```diff
@@ -188,10 +188,13 @@ func (m *Processor) ProcessBatch(ctx context.Context, batch service.MessageBatch
 				Hint:   hintJSON,
 			}
 		case OperationFindOne:
-			collection := m.database.Collection(collectionStr, m.writeConcernCollectionOption)
+			ctx, cancel := context.WithTimeout(context.Background(), m.writeConcernSpec.wTimeout)
```
rockwotj, line 191: Read timeout? Seems wrong to use a write timeout here...

rockwotj:
Looks good thanks! Looks like the linter is :blobmad:

internal/impl/mongodb/common.go
rockwotj: duh you mentioned in pr description

mihaitodor (approved):
Nice job @ooesili! I just left one small comment, but feel free to ignore it if it would cause trouble for users. Otherwise feel free to :shipit:

internal/impl/mongodb/processor.go
ooesili: Ah, I was trying to keep the existing behavior but I misread the code. I saw `writeConcernCollectionOption` passed in here and assumed that `FindOne` would use the timeout configured there but it is not the case. I'm just going to use `context.Background()` here which should keep the existing behavior.

ooesili: Since `wtimeout` seems to still be [part of the write concern API](https://www.mongodb.com/docs/manual/reference/write-concern/#write-concern-specification), and we can still use the value of the field to configure writes (e.g. [here](https://github.com/redpanda-data/connect/pull/3108/files#diff-2b6bd5e79a424a19c8b47ee8461c85609b536e8a88864196898549ab7d78d4ceR217)] I don't think we need to deprecate the field.

mihaitodor: Oh, OK! That's cool. Then we can leave it as is. I got confused by their migration docs. Thanks for checking!

rockwotj (approved):

Issue #3106, opened 2025-01-07
postgres_cdc input does not include TOAST column in update message when a different column is changed


ligfx:

ligfx:
[Debezium handles this by either](https://debezium.io/documentation/reference/stable/connectors/postgresql.html#postgresql-toasted-values):
- Including TOAST values when the table's REPLICA IDENTITY is set to FULL (doesn't seem to change anything with postgres_cdc)
- Replacing the TOAST column with the value of the `unavailable.value.placeholder` property

ligfx:
I see [our documentation](https://docs.redpanda.com/redpanda-connect/components/inputs/postgres_cdc/#receive-toast-and-deleted-values) mentions that setting REPLICA IDENTITY to FULL that should work — doesn't seem to do anything in my testing 😞

rockwotj:
erg... okay let me look into this

rockwotj:
Good docs from others on this subject: https://xata.io/blog/replica-identity-full-performance https://docs.peerdb.io/bestpractices/clickhouse_toast_columns

rockwotj:
Replicated this in a test: https://github.com/redpanda-data/connect/pull/3111


Will figure out how to fix it...

rockwotj:
Found the issue. Will go out in the next release. Thanks @ligfx

PR #3111, opened 2025-01-10
pgcdc: fix TOAST values in updates

rockwotj:
Fixes: https://github.com/redpanda-data/connect/issues/3106

### DIFF
--- a/internal/impl/postgresql/pglogicalstream/logical_stream.go
+++ b/internal/impl/postgresql/pglogicalstream/logical_stream.go
@@ -86,8 +86,6 @@ func NewPgStream(ctx context.Context, config *Config) (*Stream, error) {
 		return nil, err
 	}
 
-	println("main conn pid:", dbConn.PID())
-
 	schema, err := sanitize.NormalizePostgresIdentifier(config.DBSchema)
 	if err != nil {
 		return nil, fmt.Errorf("invalid schema name %q: %w", config.DBSchema, err)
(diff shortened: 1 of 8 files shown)


internal/impl/postgresql/pglogicalstream/logical_stream.go
mihaitodor, line 1: Nice cleanup! 🏆

internal/impl/postgresql/pglogicalstream/snapshotter.go
```diff
@@ -96,51 +93,37 @@ func (s *Snapshotter) setTransactionSnapshotName(snapshotName string) {
...
 	}
-
-	if _, err := s.pgConnection.Exec("BEGIN TRANSACTION ISOLATION LEVEL REPEATABLE READ;"); err != nil {
-		return err
+	if s.readerTxn != nil {
+		return errors.New("reader txn already open")
 	}
-
+	// Use a background context because we explicitly want the Tx to be long lived, we explicitly close it in the close method
+	tx, err := s.pool.BeginTx(context.Background(), &sql.TxOptions{ReadOnly: true, Isolation: sql.LevelRepeatableRead})
```
mihaitodor, line 104: Should we use `ctx` here?

mihaitodor (approved):
Looks great! Just one small nit and feel free to 🐑 🚀

internal/impl/postgresql/pglogicalstream/logical_stream.go
rockwotj: Thank you!

internal/impl/postgresql/pglogicalstream/snapshotter.go
rockwotj: I have a comment above but using the passed in context can cancel the transaction early, we want to explicitly control the lifetime of the transaction instead

mihaitodor: Ugh... Please ignore me 🙈 Sounds good!

PR #3112, opened 2025-01-13
Delay mysql_cdc for a release

rockwotj:
We've got a big honking release for v4.45.0, so let's delay mysql to limit what's needed for that release


mihaitodor (approved):
:shipit:

PR #3116, opened 2025-01-15
snowflake: support base64 encoded key

rockwotj:
This is the same format that Snowflake's Kafka Connect driver takes and
I have seen snowflake savvy folks get tripped up, so let's fix that
papercut and support both formats.


Jeffail (approved):
LGTM 👍

PR #3117, opened 2025-01-15
Remove busy allocations of paused partitions

Jeffail:
During a read loop of Kafka topics there's the potential to enter a scenario where there are no active partitions currently being consumed. This can be either because we haven't been allocated any partitions, or because our allocated partitions have fully saturated our in-memory buffer and we temporarily must apply back pressure.

When we enter this state it's important that we can recover quickly in order for our feed to be consistent. The buffers for each partition can in theory be saturated and cleared in a matter of microseconds, and this can happen continuously, and so any delay in reacting to these events will increase latencies and lower throughput.

Previously I had an almost busy loop here where we were continuously obtaining the full series of paused partitions, and then comparing that against the live state. However, the memory allocated from obtaining that series was significant enough that users were seeing high memory churn with standard use cases. In order to avoid this I've added an exponential back off, starting at 50 micro seconds.


rockwotj (approved):
LGTM - thanks!

mihaitodor (approved):
Nice one! 🐑 🚀

PR #3125, opened 2025-01-17
aws/s3: skip empty files with a scanner

rockwotj:
If a scanner reports EOF on construction (due to empty file for
example), then skip the file (ack it) and retry getting a file instead of
returning an error and nacking the message.

Fixes: https://github.com/redpanda-data/benthos/issues/137


mihaitodor (approved):
Nice one 🐑 🚀

PR #3126, opened 2025-01-17
kafka: add instance ID to all kafka inputs

rockwotj:
This will allow static group membership and also identifying consumers
in the group.


mihaitodor (approved):
LGTM 🐑 🚀

PR #3135, opened 2025-01-22
Update Schema Registry client

mihaitodor:
This is a follow-up to https://github.com/twmb/franz-go/pull/867.

Since `sr.URLs()` from franz-go doesn't validate URLs during client initialisation, I decided to leave our `url.Parse()` in there because this way we can catch malformed URLs on startup, before any HTTP requests are attempted.


rockwotj (approved):
🎉

PR #3136, opened 2025-01-22
packaging: rename typo in plugin name for linux pkgs

r-vasquez:
When installing `redpanda-connect` through our
linux package, we also add a symlink to a directory where rpk can pick up the binary and use it as a
plugin.

However, there was a typo and it was not possible.


clee (approved):
LGTM

rockwotj (approved):

PR #3142, opened 2025-01-23
schema_reg/decode: add schema_id metadata

rockwotj:
- **schema_registry/decode: add schema ID metadata**
- **update change log**


mihaitodor (approved):
LGTM 🐑 🚀

PR #3143, opened 2025-01-24
docker: fix ai dockerfile

rockwotj:
Instead of trying to package ollama ourselves, just use the upstream
docker image. I've tested this works and uses the local GPU on my
machine.


mihaitodor (approved):
🐑 🚀

PR #3145, opened 2025-01-26
mycdc: fix snapshot consistency issue

rockwotj:
We need to take the global lock, then we can create our consistent
snapshot. The lock is needed to prevent writes from sneaking in between.

The process is actually explained very nicely be debezium documentation.

https://debezium.io/documentation/reference/stable/connectors/mysql.html#mysql-initial-snapshot-workflow-with-global-read-lock

### DIFF
--- a/internal/impl/mysql/input_mysql_stream.go
+++ b/internal/impl/mysql/input_mysql_stream.go
@@ -226,14 +226,16 @@ func (i *mysqlStreamInput) Connect(ctx context.Context) error {
 	// canalConfig.Logger
 
 	for _, table := range i.tables {
-		canalConfig.IncludeTableRegex = append(canalConfig.IncludeTableRegex, regexp.QuoteMeta(table))
+		canalConfig.IncludeTableRegex = append(
+			canalConfig.IncludeTableRegex,
+			"^"+regexp.QuoteMeta(i.mysqlConfig.DBName+"."+table)+"$",
+		)
 	}
 
 	c, err := canal.NewCanal(canalConfig)
 	if err != nil {
 		return err
 	}
-	c.AddDumpTables(i.mysqlConfig.DBName, i.tables...)
 
 	i.canal = c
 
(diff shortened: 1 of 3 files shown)


rockwotj:
This is basically the same as https://github.com/redpanda-data/connect/pull/3144 but for MySQL.

MySQL was much closer, we just had one operation flipped to get the sequencing correct. There is a good test for this and I've ran it a bunch without issues.

There might be further work to do table level locks later on, but for now we'll stick with the global lock.

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -226,14 +226,16 @@ func (i *mysqlStreamInput) Connect(ctx context.Context) error {
...
+			canalConfig.IncludeTableRegex,
+			"^"+regexp.QuoteMeta(i.mysqlConfig.DBName+"."+table)+"$",
+		)
 	}
 
 	c, err := canal.NewCanal(canalConfig)
 	if err != nil {
 		return err
 	}
-	c.AddDumpTables(i.mysqlConfig.DBName, i.tables...)
```
rockwotj, line 236: This isn't used, we disable the built in dumping because it's just a subprocess of `mysqldump`

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -226,14 +226,16 @@ func (i *mysqlStreamInput) Connect(ctx context.Context) error {
 	// canalConfig.Logger
 
 	for _, table := range i.tables {
-		canalConfig.IncludeTableRegex = append(canalConfig.IncludeTableRegex, regexp.QuoteMeta(table))
+		canalConfig.IncludeTableRegex = append(
+			canalConfig.IncludeTableRegex,
+			"^"+regexp.QuoteMeta(i.mysqlConfig.DBName+"."+table)+"$",
```
rockwotj, line 231: Another bug that was caught, this is database and table, not just table. We only support a single database right now however.

someday we might support regex here...

mihaitodor (approved):
That's cool, thanks for adding a test! :shipit:

Issue #3146, opened 2025-01-27
Add support for retry_on_conflict in elasticsearch output module.


defgenx:
The`retry_on_conflict` option is missing in the Elasticsearch output module from Redpanda-Connect.

This option is necessary to handle version conflicts that can occur when multiple clients attempt to update the same document simultaneously. By implementing the`retry_on_conflict` option, we can instruct Elasticsearch to retry the update or upsert operation a certain number of times before failing, ensuring more robust and reliable data updates without making the complete bulk operation to fail.

Would you be ok to add this option to the module ? I could create a PR for this if it's ok for you.

defgenx:
@mihaitodor thanks for putting some labels on it. Could you tell me when a maintainer will review my PR ?

mihaitodor:
@defgenx Sorted. It should be in the next release (`v4.47.0`). Please note that we're also adding an [`elasticsearch_v8` output](https://github.com/redpanda-data/connect/pull/3160) which you should consider switching to. The existing `elasticsearch` output is based on the deprecated https://github.com/olivere/elastic library.

defgenx:
@mihaitodor great ! I'll migrate to the new output at the same time. Thanks for the information.

PR #3147, opened 2025-01-27
feat(elasticsearch): add retry_on_conflict option on output

defgenx:
## Description
This introduces the `RetryOnConflict` option to the Elasticsearch output configuration. This option allows users to specify the number of times an update or upsert should be retried in the case of a version conflict.
This feature is useful for scenarios where multiple clients might be updating the same document concurrently, and version conflicts are expected.

## Change(s)
* Added `RetryOnConflict` field to the Elasticsearch output configuration.

## Testing
* Ensured existing tests pass without modification.
* I didn't see any unit tests for this connector so I'm not sure how we could test it as it is.

## Issue(s)

* https://github.com/redpanda-data/connect/issues/3146

### DIFF
--- a/internal/impl/elasticsearch/output.go
+++ b/internal/impl/elasticsearch/output.go
@@ -33,23 +33,24 @@ import (
 )
 
 const (
-	esoFieldURLs         = "urls"
-	esoFieldSniff        = "sniff"
-	esoFieldHealthcheck  = "healthcheck"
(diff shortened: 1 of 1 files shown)


nept:
Great addition for the elasticsearch configuration! 👌

mihaitodor (approved):
LGTM, thank you!

PR #3160, opened 2025-01-31
feat(elasticsearch) add new elasticsearch_v8 output

ooesili:
This new output uses the latest official [Elasticsearch Go library](https://github.com/elastic/go-elasticsearch).

- Although the docs suggested it, I decided not to use [`esutil.NewBulkIndexer`](https://pkg.go.dev/github.com/elastic/go-elasticsearch/v8@v8.17.0/esutil#BulkIndexer) because it has a it's own internal concurrency and flushing logic that we don't really need since the benthos engine provides that for us.
- I didn't support `create` action because it is not idempotent which I don't think would play well with the assumptions that Connect makes of it's plugins.


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 - Field `arguments` added to the `amqp_0_9` input and output. (@calini)
+- New `elasticsearch_v8` output which supersedes the existing `elasticsearch` output that uses a deprecated Elasticsearch library. (@ooesili)
```
rockwotj, line 11: Does it also supersede the `opensearch` output?

internal/impl/elasticsearchv8/output.go
```diff
@@ -0,0 +1,373 @@
...
+func esConfigFromParsed(pConf *service.ParsedConfig) (*esConfig, error) {
+	conf := &esConfig{}
+
+	if os.Getenv("ELASTICSEARCH_DEBUG") != "" {
+		conf.clientOpts.Logger = &elastictransport.CurlLogger{
+			Output:             os.Stdout,
+			EnableRequestBody:  true,
+			EnableResponseBody: true,
+		}
+	}
```
rockwotj, line 69: Can we wrap this logger with our service logger at trace level instead?

internal/impl/elasticsearchv8/output.go
```diff
@@ -0,0 +1,373 @@
...
+				Optional(),
+			service.NewBatchPolicyField(esFieldBatching),
+		).
+		Example("Updating Documents", "When updating documents, the request body should contain a combination of a `doc`, `upsert`, and/or `script` fields at the top level, this should be done via mapping processors. `doc` updates using a partial document, `script` performs an update using a scripting language such as the built in Painless language, and `upsert` updates an existing document or inserts a new one if it doesn’t exist. For more information on the structures and behaviors of these fields, please see the https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html[Elasticsearch Update API^]", `
+output:
+  processors:
+    - mapping: |
+        meta id = this.id
+        root.doc = this
+  opensearch:
```
rockwotj, line 180: wrong component?

internal/impl/elasticsearchv8/output.go
```diff
@@ -0,0 +1,373 @@
...
+					Default(""),
+				service.NewStringField(esFieldAuthPassword).
+					Description("A password to authenticate with.").
+					Default("").Secret(),
+			).Description("Allows you to specify basic authentication.").
+				Advanced().
+				Optional(),
+			service.NewBatchPolicyField(esFieldBatching),
+		).
+		Example("Updating Documents", "When updating documents, the request body should contain a combination of a `doc`, `upsert`, and/or `script` fields at the top level, this should be done via mapping processors. `doc` updates using a partial document, `script` performs an update using a scripting language such as the built in Painless language, and `upsert` updates an existing document or inserts a new one if it doesn’t exist. For more information on the structures and behaviors of these fields, please see the https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html[Elasticsearch Update API^]", `
```
rockwotj, line 174: Can we get a couple of solid examples? Maybe Redpanda -> elasticsearch_v8 and s3 -> elasticsearch_v8?

internal/impl/elasticsearchv8/output.go
```diff
@@ -0,0 +1,373 @@
...
+					batchErr.Failed(i, err)
+				}
+			}
+		}
+		return batchErr
+	}
+
+	// time.Duration is an int64 counting nanonseconds, result.Took is an int64
+	// counting milliseconds
+	tookDuration := time.Duration(result.Took * 1e6)
```
rockwotj, line 269: this is serverside duration right?

also can we please do `result.Took * time.Millisecond`?

internal/impl/elasticsearchv8/output.go
rockwotj: Also can we add the `update` docs to either the `action` field or the overall summary? That's pretty important.

rockwotj (commented):
Will review the tests a bit later

CHANGELOG.md
ooesili: No it does not, my understanding is elasticsearch and opensearch have diverged a bit so it's useful for us to support both

internal/impl/elasticsearchv8/output.go
ooesili: Copy paste error lol

internal/impl/elasticsearchv8/output.go
ooesili: Yes it's server side. That doesn't work, you can't multiply an `int64` and a `time.Duration`. Stuff like `time.Millisecond * 300` only works because Go's shapeshifting integer constant magic

rockwotj: Yeah you could do this tho

```
time.Duration(result.Took)*time.Millsecond
```

internal/impl/elasticsearchv8/output.go
ooesili: There's not an easy way to do that, `elastictransport.CurlLogger` is a decorator around a custom roundtripper type interface. I could write my own struct for that but it doesn't really seem worth it for a debugger that I intended to really only be used by us during development.

internal/impl/elasticsearchv8/output.go
ooesili: Done, lmk what you think.

mihaitodor (approved):
That looks great, nice one! Feel free to 🐑 🚀

rockwotj (approved):

mihaitodor:
PS: Before merging, would you mind adding a note to the Changelog about `retry_on_conflict` being added to the `elasticsearch` output? I merged #3147 so we have that as well.

defgenx:
> PS: Before merging, would you mind adding a note to the Changelog about `retry_on_conflict` being added to the `elasticsearch` output? I merged #3147 so we have that as well.

It means the v8 will not implement the `retry_on_conflict` feature (preventing people needing this to upgrade to this new output) ? Can you maybe add it in this PR ?

PR #3177, opened 2025-02-10
snowflake: force singleshot object uploads

rockwotj:
- **snowflake: don't use multipart uploads in AWS**
- **snowflake: don't use multipart uploads in GCP**
- Azure is already good 👍

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ Changelog
 
 All notable changes to this project will be documented in this file.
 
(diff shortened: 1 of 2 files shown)


mihaitodor (approved):
LGTM 🐑  🚀 

PS: Do we need a changelog entry?

rockwotj:
Yeah good catch I just pushed it, also going to cut a patch release. Thanks @mihaitodor

PR #3178, opened 2025-02-11
snowflake: add lint rule that private_key is in PEM format

rockwotj:


internal/impl/snowflake/output_snowflake_streaming.go
```diff
@@ -101,7 +101,7 @@ You can monitor the output batch size using the `+"`snowflake_compressed_output_
 			service.NewStringField(ssoFieldDB).Description("The Snowflake database to ingest data into.").Example("MY_DATABASE"),
 			service.NewStringField(ssoFieldSchema).Description("The Snowflake schema to ingest data into.").Example("PUBLIC"),
 			service.NewInterpolatedStringField(ssoFieldTable).Description("The Snowflake table to ingest data into.").Example("MY_TABLE"),
-			service.NewStringField(ssoFieldKey).Description("The PEM encoded private RSA key to use for authenticating with Snowflake. Either this or `private_key_file` must be specified.").Optional().Secret(),
+			service.NewStringField(ssoFieldKey).Description("The PEM encoded private RSA key to use for authenticating with Snowflake. Either this or `private_key_file` must be specified.").Optional().Secret().LintRule(`root = if !this.re_match("(?s)^-----BEGIN [A-Z ]+-----\\n[0-9A-Za-z+/=\\n]+-----END [A-Z ]+-----\\n?$") && !this.re_match("[0-9A-Za-z+/=]") { ["field private_key must be in PEM format"] }`),
```
t-eckert, line 104: This is not my area of expertise so I could be way off, but should this be `||` instead of `&&`? 

If I logic this out.

```
if 
// It doesn't match the BEGIN/END
!this.re_match(<REGEX-FOR-BEGIN/END>) 
// AND
&& 
// it is not base64 encoded, 
!this.re_match("[0-9A-Za-z+/=]") 
// then we error.
{ ["field private_key must be in PEM format"] }
```

Should it error if the input is _either_ not BEGIN/END-ed correctly _or_ not base64?

rockwotj: It's right, probably should read as `if !(this.IsPEMFormat() || this.IsBase64()) throw error` which according to https://en.wikipedia.org/wiki/De_Morgan%27s_laws is the same as `if !this.IsPEMFormat() && !this.IsBase64() throw error`

mihaitodor (approved):
LGTM 🐑 🚀

PR #3179, opened 2025-02-12
fix(elasticsearch): output was not using retry_on_conflict option

defgenx:
## Description
This PR addresses an issue with the `retry_on_conflict` parameter in the Elasticsearch output implementation. The fix ensures that the `retry_on_conflict` parameter is correctly applied during update and upsert operations.


mihaitodor (approved):
Thank you for fixing this! I should've spotted it when merging #3147... Also, if possible, please switch to `elasticsearch_v8`. That's the one which we'll be maintaining going forward since this one relies on the deprecated `github.com/olivere/elastic` library.

defgenx:
@mihaitodor I migrated just after I created this PR. ;)

PR #3182, opened 2025-02-12
Fix Redpanda Migrator consumer group offset migration logic

mihaitodor:
- Field `is_high_watermark` added to the `redpanda_migrator_offsets` output.
- Metadata field `kafka_is_high_watermark` added to the `redpanda_migrator_offsets` input.
- Fixed a bug with the `redpanda_migrator_offsets` input and output where the consumer group update migration logic based on timestamp lookup should no longer skip ahead in the destination cluster. This should enforce at-least-once delivery guarantees.
- The `redpanda_migrator_bundle` output no longer drops messages if either the `redpanda_migrator` or the `redpanda_migrator_offsets` child output throws an error. Connect will keep retrying to write the messages and apply backpressure to the input.

TODO:
- [x] Push extra integration tests

### DIFF
--- a/internal/impl/kafka/enterprise/integration_test.go
+++ b/internal/impl/kafka/enterprise/integration_test.go
@@ -985,31 +985,27 @@ func TestRedpandaMigratorOffsetsIntegration(t *testing.T) {
 			dummyConsumerGroup := "test_cg"
 			messageCount := 5
 
-			// Produce messages in the source cluster
-			// The message timestamps are produced in ascending order, starting from 1 all the way to messageCount
+			// Produce messages in the source cluster.
+			// The message timestamps are produced in ascending order, starting from 1 all the way to messageCount.
 			produceMessages(t, source, dummyTopic, dummyMessage, 0, messageCount, false)
 
-			// Produce the exact same messages in the destination cluster
+			// Produce the exact same messages in the destination cluster.
 			produceMessages(t, destination, dummyTopic, dummyMessage, 0, messageCount, false)
 
-			// Read the messages from the source cluster using a consumer group
+			// Read the messages from the source cluster using a consumer group.
 			readMessagesWithCG(t, source, dummyTopic, dummyConsumerGroup, dummyMessage, 5, false)
 
-			if !test.cgAtEndOffset {
+			if test.extraCGUpdate || !test.cgAtEndOffset {
+				// Make sure both source and destination have extra messages after the current consumer group offset.
 				// The next messages need to have more recent timestamps than the existing messages, so we use
-				// `messageCount` as an offset for their timestamps
-				produceMessages(t, source, dummyTopic, dummyMessage, messageCount, messageCount, false)
-			}
-
-			if test.extraCGUpdate {
-				// Make sure both source and destination have extra messages after the current consumer group offset
+				// `messageCount` as an offset for their timestamps.
 				produceMessages(t, source, dummyTopic, dummyMessage, messageCount, messageCount, false)
 				produceMessages(t, destination, dummyTopic, dummyMessage, messageCount, messageCount, false)
 			}
 
 			t.Log("Finished setting up messages in the source and destination clusters")
 
-			// Migrate the consumer group offset
+			// Migrate the consumer group offsets.
 			streamBuilder := service.NewStreamBuilder()
 			require.NoError(t, streamBuilder.SetYAML(fmt.Sprintf(`
 input:
@@ -1030,7 +1026,7 @@ output:
 				return nil
 			}))
 
-			// Ensure the callback function is called after the output wrote the message
+			// Ensure the callback function is called after the output wrote the message.
 			streamBuilder.SetOutputBrokerPattern(service.OutputBrokerPatternFanOutSequential)
 
 			stream, err := streamBuilder.Build()
@@ -1038,7 +1034,7 @@ output:
 
 			license.InjectTestService(stream.Resources())
 
-			// Run stream in the background
+			// Run stream in the background.
 			migratorCloseChan := make(chan struct{})
 			go func() {
 				err = stream.Run(context.Background())
@@ -1058,7 +1054,7 @@ output:
 			migratorUpdateWG.Wait()
 
 			if test.extraCGUpdate {
-				// Trigger another consumer group update to get it to point to the end of the topic
+				// Trigger another consumer group update to get it to point to the end of the topic.
 				migratorUpdateWG.Add(1)
 				readMessagesWithCG(t, source, dummyTopic, dummyConsumerGroup, dummyMessage, messageCount, false)
 				migratorUpdateWG.Wait()
@@ -1066,7 +1062,9 @@ output:
 
 			client, err := kgo.NewClient(kgo.SeedBrokers([]string{destination.brokerAddr}...))
 			require.NoError(t, err)
-			defer client.Close()
+			t.Cleanup(func() {
+				client.Close()
+			})
 
 			adm := kadm.NewClient(client)
 			offsets, err := adm.FetchOffsets(context.Background(), dummyConsumerGroup)
--- a/internal/impl/kafka/enterprise/integration_test.go
+++ b/internal/impl/kafka/enterprise/integration_test.go
@@ -17,6 +17,7 @@ import (
 	"log/slog"
 	"net/http"
 	"strconv"
+	"sync"
 	"testing"
 	"time"
 
@@ -25,6 +26,7 @@ import (
 	"github.com/ory/dockertest/v3/docker"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"github.com/twmb/franz-go/pkg/kadm"
 	"github.com/twmb/franz-go/pkg/kerr"
 	"github.com/twmb/franz-go/pkg/kgo"
 	"github.com/twmb/franz-go/pkg/kmsg"
@@ -654,20 +656,30 @@ func startRedpanda(t *testing.T, pool *dockertest.Pool, exposeBroker bool, autoc
 	}, nil
 }
 
-func produceMessage(t *testing.T, rpe redpandaEndpoints, topic, message string) {
+// produceMessages produces `count` messages to the given `topic` with the given `message` content. The
+// `timestampOffset` indicates an offset which gets added to the `counter()` Bloblang function which is used to generate
+// the message timestamps sequentially, the first one being `1 + timestampOffset`.
+func produceMessages(t *testing.T, rpe redpandaEndpoints, topic, message string, timestampOffset, count int, encode bool) {
 	streamBuilder := service.NewStreamBuilder()
-	require.NoError(t, streamBuilder.SetYAML(fmt.Sprintf(`
+	config := ""
+	if encode {
+		config = fmt.Sprintf(`
 pipeline:
   processors:
     - schema_registry_encode:
         url: %s
         subject: %s
         avro_raw_json: true
+`, rpe.schemaRegistryURL, topic)
+	}
+	config += fmt.Sprintf(`
 output:
   kafka_franz:
     seed_brokers: [ %s ]
     topic: %s
-`, rpe.schemaRegistryURL, topic, rpe.brokerAddr, topic)))
+    timestamp_ms: ${! counter() + %d}
+`, rpe.brokerAddr, topic, timestampOffset)
+	require.NoError(t, streamBuilder.SetYAML(config))
 	require.NoError(t, streamBuilder.SetLoggerYAML(`level: OFF`))
 
 	inFunc, err := streamBuilder.AddProducerFunc()
@@ -682,7 +694,9 @@ output:
 	t.Cleanup(done)
 
 	go func() {
-		require.NoError(t, inFunc(ctx, service.NewMessage([]byte(message))))
+		for range count {
+			require.NoError(t, inFunc(ctx, service.NewMessage([]byte(message))))
+		}
 
 		require.NoError(t, stream.StopWithin(3*time.Second))
 	}()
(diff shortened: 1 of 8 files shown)


internal/impl/kafka/enterprise/redpanda_migrator_bundle_output.tmpl.yaml
```diff
@@ -87,29 +87,15 @@ mapping: |
...
             output:
-              fallback:
-                - label: %s_redpanda_migrator_offsets_output
-                  redpanda_migrator_offsets: %s
-                # TODO: Use a DLQ
-                - drop: {}
-                  processors:
-                    - log:
-                        message: |
-                          Dropping message: ${! content() } / ${! metadata() }
```
mihaitodor, line 112: We really shouldn't be dropping messages here... I left this as a TODO initially thinking that we might want to configure DLQs, but it's best to just block if we can't write data. If it's a transitive issue, it will resume when the broker becomes responsive again. If not, an operator will have to look into it. Restarting Migrator should be a safe operation (even if it potentially introduces duplicate records).

internal/impl/kafka/enterprise/redpanda_migrator_bundle_output.tmpl.yaml
```diff
@@ -134,29 +119,15 @@ mapping: |
...
```
mihaitodor, line 159: Same here (it's the same logic, just repeated for the case when we don't use schemas).

internal/impl/kafka/enterprise/redpanda_migrator_bundle_output.tmpl.yaml
```diff
@@ -119,8 +105,7 @@ mapping: |
                     cases:
                       - check: '@fallback_error == "request returned status: 422"'
                         output:
-                          # TODO: Use a DLQ
-                          drop: {}
+                          drop: {} # TODO: Use a DLQ?
```
mihaitodor, line 108: We want to drop these messages (and potentially forward them to a DLQ?), because they indicate that the provided schema already exists in the destination.

internal/impl/kafka/enterprise/redpanda_migrator_offsets_input.go
```diff
@@ -175,53 +171,134 @@ func (rmoi *redpandaMigratorOffsetsInput) matchesTopic(topic string) bool {
...
+
+func (rmoi *redpandaMigratorOffsetsInput) getCommitTimestamp(ctx context.Context, topic string, partition int32, offset int64) (int64, bool, error) {
+	client, err := kgo.NewClient(rmoi.clientOpts...)
+	if err != nil {
+		return 0, false, fmt.Errorf("failed to create Kafka client: %s", err)
+	}
+	defer client.Close()
+
+	// The default kadm client timeout is 15s. Do we need to make this configurable?
+	offsets, err := kadm.NewClient(client).ListCommittedOffsets(ctx, topic)
```
mihaitodor, line 218: There are 2 API methods that we can use here:
- [`ListCommittedOffsets`](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kadm#Client.ListCommittedOffsets)
- [`ListEndOffsets`](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kadm#Client.ListEndOffsets)

From the `ListCommittedOffsets` docs:

> A committed offset may be slightly less than the
> latest offset. In Kafka terms, committed means the last stable offset, and
> newest means the high watermark. Record offsets in active, uncommitted
> transactions will not be returned.

Also, under the hood, `ListCommittedOffsets` uses the `READ_COMMITTED` isolation level which is documented [here](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kmsg#ListOffsetsRequest):

> IsolationLevel configures which record offsets are visible in the
> response. READ_UNCOMMITTED (0) makes all records visible. READ_COMMITTED
> (1) makes non-transactional and committed transactional records visible.
> READ_COMMITTED means all offsets smaller than the last stable offset and
> includes aborted transactions (allowing consumers to discard aborted
> records).

With my simple tests, both `ListCommittedOffsets` and `ListEndOffsets` return the same values. However, it's not clear to me under which scenarios they might produce different results. What matters is that the returned offset matches the offset we get from the `__consumer_offsets` topic when a client reads the last record of a topic.

Does it make sense to stick with `ListCommittedOffsets`?

mihaitodor: @twmb suggested to use `ListEndOffsets` instead so I switched to that.

internal/impl/kafka/enterprise/redpanda_migrator_offsets_output.go
```diff
@@ -265,6 +317,36 @@ func (w *redpandaMigratorOffsetsWriter) Write(ctx context.Context, msg *service.
 					delete(partitionOffsets, partition)
 				}
 				offset.Metadata = offsetMetadata
+
+				// As an optimisation to try and avoid unnecessary duplicates in the common case, we check if the
```
rockwotj, line 321: were you able to write tests for these cases or is that too complicated?

rockwotj (commented):
Generally looks great! Is it possible to write any tests for this case? I will take a deeper look at the PR on Monday. Thanks Mihai

internal/impl/kafka/enterprise/redpanda_migrator_offsets_output.go
mihaitodor: I have to clean them up, but yeah, I have a few that are reproducible. I'll try to push them in the next few days.

mihaitodor:
Thank you for having a look Tyler! By the way, I think we won't have to decrement the timestamps in any way, since `ListOffsetsAfterMilli()` is inclusive as you mentioned in https://github.com/twmb/franz-go/pull/897. Will think more through it as I clean up the tests...

internal/impl/kafka/enterprise/redpanda_migrator_offsets_output.go
mihaitodor: OK, I was able to push some tests to cover this logic. They cover the happy flows, but I can come up with some error cases as well if needed. They're a bit more annoying to assert though.

internal/impl/kafka/enterprise/redpanda_migrator_offsets_output.go
```diff
@@ -265,6 +317,36 @@ func (w *redpandaMigratorOffsetsWriter) Write(ctx context.Context, msg *service.
...
+						if err != nil {
+							return fmt.Errorf("failed to read the end offset for topic %q and partition %q: %s", topic, partition, err)
+						}
+
+						committedOffset, ok := offsets.Lookup(topic, partition)
+						if !ok {
+							return fmt.Errorf("failed to find the end offset for topic %q and partition %q: %s", topic, partition, err)
+						}
+
+						offset.At = committedOffset.Offset
```
rockwotj, line 346: This is almost the same as doing `offset.At++` but is racy right? If new records come in during `getLastRecordOffset` and `ListEndOffsets` then we can skip records because ListEndOffsets will pick up the new records?

I am also trying to think through if comparing the offsetCommitTimestamp to the record timestamp from `getLastRecord` would also be benifical to ensure we're not skipping data... Let me noodle on an algorithm here.

mihaitodor: > This is almost the same as doing offset.At++ but is racy right?

Yep! I should've called that `committedOffset` variable `endOffset`. Fixing now...

It's racy indeed either way... I just realised now that I could get new messages written to the topic between `getLastRecordOffset` and `ListEndOffsets` which will lead to a fast forward... I was thinking initially that I'd have to call `getLastRecordOffset` again after `ListEndOffsets` to check if `lastOffset` changed in the meantime. Not sure why I changed my mind 🤦‍♂️ 

> Let me noodle on an algorithm here.

OK, curious if we can get away with simpler logic. In practice `isEndOffset` will be `true` frequently only when topics don't get a lot of traffic and consumers catch up immediately, so maybe the performance isn't a concern.

internal/impl/kafka/enterprise/redpanda_migrator_offsets_output.go
```diff
@@ -194,11 +202,43 @@ func (w *redpandaMigratorOffsetsWriter) Connect(ctx context.Context) error {
...
+	if err := fetches.Err(); err != nil {
+		return 0, fmt.Errorf("failed to read last record for topic %q partition %q: %s", topic, partition, err)
+	}
+
+	it := fetches.RecordIter()
+	if it.Done() {
+		return 0, fmt.Errorf("couldn't find the last record for topic %q partition %q: %s", topic, partition, err)
+	}
+
+	return it.Next().Offset, nil
```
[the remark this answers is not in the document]
mihaitodor: > I believe we can replace this function with a ListEndOffsets call right? 

That's what I wanted to do... However, `ListEndOffsets` returns the offset after the offset of the last message (which I denoted as "last offset" - not sure what's a better name for it) in the topic. Here's an experiment:

```
> docker run --rm -it --name=source -p 8081:8081 -p 9092:9092 -p 9644:9644 redpandadata/redpanda redpanda start --node-id 0 --mode dev-container --set "rpk.additional_start_flags=[--reactor-backend=epoll]" --kafka-addr 0.0.0.0:9092 --advertise-kafka-addr host.docker.internal:9092 --schema-registry-addr 0.0.0.0:8081
```

```
$ rpk topic create test -p1 -X brokers=localhost:9092
TOPIC  STATUS
test   OK
$ echo foobar | rpk topic produce test -X brokers=localhost:9092
Produced to partition 0 at offset 0 with timestamp 1739834099974.
```

`getLastRecordOffset()`:
```
Offset:  0
```

`ListEndOffsets()` /  `ListCommittedOffsets()`:
```
Offset_end:  1
Offset_comm:  1
```

<details>
<summary>Click this to reveal the messy code if you wish to play with it:</summary>

`getLastRecordOffset()`:
```go
func main() {
	seeds := []string{"localhost:9092"}
	consumeStart := map[string]map[int32]kgo.Offset{
		"test": {
			// The default offset begins at the end.
			0: kgo.NewOffset().Relative(-1),
		},
	}

	for {
		client, err := kgo.NewClient([]kgo.Opt{kgo.SeedBrokers(seeds...), kgo.ConsumePartitions(consumeStart)}...)
		if err != nil {
			log.Fatalf("failed to create Kafka client: %s", err)
		}

		ctx, _ := context.WithTimeout(context.Background(), 10*time.Second)
		fetches := client.PollFetches(ctx)
		if fetches.IsClientClosed() {
			log.Fatalf("failed to read message: client closed")
		}

		if err := fetches.Err(); err != nil {
			log.Fatalf("failed to read message: %s", err)
		}

		it := fetches.RecordIter()
		if it.Done() {
			log.Fatalf("couldn't find message: %s", err)
		}

		rec := it.Next()

		fmt.Println("Offset: ", rec.Offset)
		fmt.Println("Timestamp: ", rec.Timestamp)

		client.Close()

		time.Sleep(200 * time.Millisecond)
	}
}
```

`ListEndOffsets()` /  `ListCommittedOffsets()`:
```go
func main() {
	seeds := []string{"localhost:9092"}

	client, err := kgo.NewClient([]kgo.Opt{kgo.SeedBrokers(seeds...)}...)
	if err != nil {
		log.Fatalf("failed to create Kafka client: %s", err)
	}
	defer client.Close()
	adm := kadm.NewClient(client)
	for {

		offsets, err := adm.ListEndOffsets(context.Background(), "test")
		if err != nil {
			log.Fatalf("failed to read offsets: %s", err)
		}

		o, ok := offsets.Lookup("test", 0)
		if !ok {
			log.Fatalf("failed to find offset for test-0")
		}

		fmt.Println("Offset_end: ", o.Offset)
		fmt.Println("Timestamp_end: ", o.Timestamp)

		offsets, err = adm.ListCommittedOffsets(context.Background(), "test")
		if err != nil {
			log.Fatalf("failed to read offsets: %s", err)
		}

		o, ok = offsets.Lookup("test", 0)
		if !ok {
			log.Fatalf("failed to find offset for test-0")
		}

		fmt.Println("Offset_comm: ", o.Offset)
		fmt.Println("Timestamp_comm: ", o.Timestamp)

		time.Sleep(200 * time.Millisecond)
	}
}
```
</details>

rockwotj (commented):
Just one question about retries and then I think it LGTM

internal/impl/kafka/enterprise/redpanda_migrator_offsets_output.go
```diff
@@ -257,21 +280,50 @@ func (w *redpandaMigratorOffsetsWriter) Write(ctx context.Context, msg *service.
...
+		offset, ok := listedOffsets.Lookup(topic, partition)
+		if !ok {
+			// This should never happen, but we check just in case.
+			return fmt.Errorf("no offsets returned for topic %q", topic)
+		}
+
+		if !isEndOffset && offset.Timestamp == -1 {
+			// This can happen if we received an offset update, but the record which was read from the source cluster to
+			// trigger it has not been replicated to the destination cluster yet. In this case, we raise an error so the
+			// operation is retried.
```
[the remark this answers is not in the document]
mihaitodor: That would result in some duplicates rather than data loss, which I think should be acceptable (we can't do exactly-once-delivery anyway). Happy to make it less prone to such scenarios if you have any ideas on how to handle this better...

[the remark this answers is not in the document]
rockwotj: I have some ideas but let's do it in a followup.

rockwotj (approved):

internal/impl/kafka/enterprise/integration_test.go
```diff
@@ -691,34 +705,46 @@ output:
 	require.NoError(t, err)
 }
 
-func readMessageWithCG(t *testing.T, rpe redpandaEndpoints, topic, consumerGroup, message string) {
+// readMessagesWithCG reads `count` messages from the given `topic` with the given `consumerGroup`.
+// TODO: Since `read_until` can't guarantee that we will read `count` messages, `topic` needs to have exactly `count`
+// messages in it. We should add some mechanism to the Kafka inputs to allow us to read a range of offsets if possible
+// (or up to a certain offset).
```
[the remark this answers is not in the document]
mihaitodor: > So do we ack the messages after `check` returns true?

Yep, if I use `read_until` there, instead of reading 5 messages out of 10, it reads about 6-7... `read_until` should tell the child input to shut down at that point, but it's not immediate, so I think it gets a bit of time to ack before `kafka_franz` tears down the `franz-go` client. I'm not entirely sure how it's wired under the hood, but I recall bumping into issues with `read_until` before. For example, this generates 10 messages:
```yaml
input:
  read_until:
    check: counter() == 2
    input:
      generate:
        count: 20
        batch_size: 5
        interval: 0s
        mapping: root = counter()
```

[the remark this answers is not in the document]
rockwotj: Yeah I was reading the code for `read_until` and was surprised that `check` only runs on the first message of every batch and the whole batch is still passed through..

mihaitodor:
I did the cleanup & I'll merge it later tomorrow in case you think of other stuff I should cover in there. Thank you again!

PR #3188, opened 2025-02-18
snowflake: add more logging around errors

rockwotj:


internal/impl/snowflake/streaming/rest.go
```diff
@@ -36,10 +36,12 @@ import (
...
-	responseSuccess         = 0
-	responseTableNotExist   = 4
-	responseErrQueueFull    = 7
-	responseErrRetryRequest = 10
+	responseSuccess                   = 0
+	responseTableNotExist             = 4
+	responseErrQueueFull              = 7
+	responseErrRetryRequest           = 10
+	responseErrInvalidClientSequencer = 20
+	responseErrReopenChannelNeeded    = 35 // this can happen if the schema changes for example
```
mihaitodor, line 44: Are these used?

mihaitodor (approved):
LGTM :shipit:

internal/impl/snowflake/streaming/rest.go
rockwotj: No but I need to track them somewhere and this is a good place for now 🙃

PR #3192, opened 2025-02-18
telemetry updates

rockwotj:
Similar to what we do in the broker: https://github.com/redpanda-data/redpanda/blob/bea4cc4d3661d73f9c8c70e81e9b26f7338382b3/src/v/security/license.cc#L178


mihaitodor (approved):
🐑 🚀

PR #3195, opened 2025-02-19
snowflake: simplify bytes converter

rockwotj:


internal/impl/snowflake/streaming/userdata_converter.go
```diff
@@ -252,26 +251,9 @@ func (c binaryConverter) ValidateAndConvert(stats *statsBuffer, val any, buf typ
...
-			v = []byte{}
-		}
-	case []byte:
-		v = t
-	default:
-		b, err := bloblang.ValueAsBytes(val)
-		if err != nil {
-			return err
-		}
-		v = b
```
mihaitodor, line 274: Nice cleanup! Was this added initially due to some perf issues in the bloblang code?

mihaitodor (approved):
LGTM! 🐑 🚀

internal/impl/snowflake/streaming/userdata_converter.go
rockwotj: I was over optimizing the string conversion. I don't think it's worth the risk now

Issue #3198, opened 2025-02-20
postgres_cdc: prevent low frequency updates from retaining WAL files


rockwotj:
From Debezium docs:
> Heartbeat messages are needed when there are many updates in a database that is being tracked but only a tiny number of updates are related to the table(s) and schema(s) for which the connector is capturing changes. In this situation, the connector reads from the database transaction log as usual but rarely emits change records to Kafka. This means that no offset updates are committed to Kafka and the connector does not have an opportunity to send the latest retrieved LSN to the database. The database retains WAL files that contain events that have already been processed by the connector. Sending heartbeat messages enables the connector to send the latest retrieved LSN to the database, which allows the database to reclaim disk space being used by no longer needed WAL files.

We need to do something similar, except I think we can do this with logical messages that is simpler and more transparent.

PR #3199, opened 2025-02-21
pgcdc: add heartbeat mechanism

rockwotj:
See the commits for more details

- **pgcdc: don't prefix the replication slot name**
- **pgcdc: add heartbeat mechanism using logical wal messages**

Fixes: https://github.com/redpanda-data/connect/issues/3198

### DIFF
--- a/internal/impl/postgresql/pglogicalstream/heartbeat.go
+++ b/internal/impl/postgresql/pglogicalstream/heartbeat.go
@@ -1,0 +1,52 @@
+// Copyright 2025 Redpanda Data, Inc.
+//
+// Licensed as a Redpanda Enterprise file under the Redpanda Community
+// License (the "License"); you may not use this file except in compliance with
+// the License. You may obtain a copy of the License at
+//
+// https://github.com/redpanda-data/connect/v4/blob/main/licenses/rcl.md
+
+package pglogicalstream
+
+import (
+	"context"
+	"database/sql"
+	"time"
+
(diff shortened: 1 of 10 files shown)


internal/impl/postgresql/pglogicalstream/heartbeat.go
```diff
@@ -0,0 +1,52 @@
...
+	logger        *service.Logger
+	prefix, value string
+}
+
+func newHeartbeat(dsn string, interval time.Duration, logger *service.Logger, prefix, value string) (*heartbeat, error) {
+	dbConn, err := openPgConnectionFromConfig(dsn)
+	if err != nil {
+		return nil, err
+	}
+	h := &heartbeat{dbConn, nil, logger, prefix, value}
```
c4milo, line 33: I personally prefer decoupling the composite literal from the struct definition by explicitly specifying field names but since this is a private type and very close to the instantiation, it doesn't seem as important, unless you all care about consistent style. 

https://google.github.io/styleguide/go/decisions.html#literal-formatting

c4milo (approved):
LGTM!

internal/impl/postgresql/pglogicalstream/heartbeat.go
rockwotj: Yes agreed. I was just being lazy and going quickly, will cleanup. Thanks for the review!

PR #3201, opened 2025-02-21
snowflake: automatically retry transient errors in snowflake

rockwotj:
These errors can be expected in some cases. If we fail to commit,
we should re-open the channel, and in these cases where the schema
changes then we know it's safe to retry the write, so do this
automatically to prevent log noise or concerns.


internal/impl/snowflake/integration_test.go
```diff
@@ -303,20 +353,80 @@ snowflake_streaming:
           }
 `)
 	RunStreamInBackground(t, stream)
-	require.NoError(t, produce(context.Background(), Batch([]map[string]any{
+	require.NoError(t, produce([]map[string]any{
```
rockwotj, line 357: I tried to remove a bunch of the boilerplate in these tests

internal/impl/snowflake/integration_test.go
```diff
@@ -36,6 +43,51 @@ func EnvOrDefault(name, fallback string) string {
...
+		"$DB", config.db,
+		"$ROLE", config.role,
+		"$SCHEMA", config.schema,
+		"$PRIVATE_KEY_FILE", config.privateKeyFile,
+		"$PRIVATE_KEY", config.privateKey,
+		"$DSN", config.dsn,
+	).Replace(s)
+}
+
+var setup = sync.OnceFunc(func() {
```
mihaitodor, line 70: I wonder if having a `func TestMain(m *testing.M)` here would help. Then again, I'm not really sure why this func must run only once (would be handy to have a comment).

mihaitodor (approved):
LGTM :shipit: I just left one small nit. Feel free to ignore it

internal/impl/snowflake/integration_test.go
rockwotj: Yeah actually it's probably fine to have it just run everytime you call setup. I'll make that change 👍 Thanks @mihaitodor !

PR #3203, opened 2025-02-22
mongo/cdc: add operation time metadata

rockwotj:
Also update the config to be more generic so in a followup we can support a mode that sends out partial updates if you don't want to pay for the full lookup of documents for updates.


mihaitodor (approved):
LGTM 🐑 🚀

PR #3207, opened 2025-02-25
Add topic properties to Migrator

mihaitodor:
Also make sure ACLs are always migrated even if the destination topic already exists.

### DIFF
--- a/internal/impl/kafka/enterprise/redpanda_migrator_offsets_output.go
+++ b/internal/impl/kafka/enterprise/redpanda_migrator_offsets_output.go
@@ -273,24 +273,26 @@ func (w *redpandaMigratorOffsetsWriter) Write(ctx context.Context, msg *service.
 		// timestamps of all the records in the topic. It also sets the timestamp of the returned offset to -1 in this case.
 		listedOffsets, err := w.client.ListOffsetsAfterMilli(ctx, offsetCommitTimestamp, topic)
 		if err != nil {
-			return fmt.Errorf("failed to translate consumer offsets: %s", err)
+			return fmt.Errorf("failed to list offsets for topic %q and timestamp %d: %s", topic, offsetCommitTimestamp, err)
 		}
 
+		w.mgr.Logger().Tracef("Listed offsets for topic %q and timestamp %d: %+v", topic, offsetCommitTimestamp, listedOffsets)
+
 		if err := listedOffsets.Error(); err != nil {
-			return fmt.Errorf("listed offsets error: %s", err)
+			return fmt.Errorf("failed to read offsets for topic %q and timestamp %d: %s", topic, offsetCommitTimestamp, err)
 		}
 
 		offset, ok := listedOffsets.Lookup(topic, partition)
 		if !ok {
 			// This should never happen, but we check just in case.
-			return fmt.Errorf("committed offset not yet replicated to the destination %q topic: lookup failed", topic)
+			return fmt.Errorf("record for timestamp %d not yet replicated to the destination topic %q partition %d: lookup failed", offsetCommitTimestamp, topic, partition)
 		}
 
 		if !isHighWatermark && offset.Timestamp == -1 {
 			// This can happen if we received an offset update, but the record which was read from the source cluster to
 			// trigger it has not been replicated to the destination cluster yet. In this case, we raise an error so the
 			// operation is retried.
-			return fmt.Errorf("committed offset not yet replicated to the destination %q topic", topic)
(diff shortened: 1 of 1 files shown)


internal/impl/kafka/enterprise/topic_crud.go
```diff
@@ -56,7 +57,38 @@ func createTopic(ctx context.Context, topic string, replicationFactorOverride bo
...
+				c.Key == "retention.ms" ||
+				c.Key == "segment.ms" ||
+				c.Key == "segment.bytes" ||
+				c.Key == "compression.type" ||
+				c.Key == "message.timestamp.type" ||
+				c.Key == "max.message.bytes" ||
+				c.Key == "replication.factor" ||
+				c.Key == "write.caching" ||
+				c.Key == "redpanda.iceberg.mode" ||
+				strings.HasPrefix(c.Key, "redpanda.remote.") {
```
mihaitodor, line 83: Is this correct? Do we have to first query the destination server settings to ensure that we're not overwriting a custom setting?

internal/impl/kafka/enterprise/integration_test.go
```diff
@@ -776,8 +838,6 @@ input:
       topics: [ %s ]
       consumer_group: migrator_cg
       start_from_oldest: true
-      replication_factor_override: true
-      replication_factor: -1
```
mihaitodor, line 780: The template linter is a bit annoying, since it doesn't check if there are any unrecognised fields... Something to improve in Benthos.

internal/impl/kafka/enterprise/redpanda_migrator_output.go
```diff
@@ -226,12 +226,10 @@ func init() {
 											// message from it.
 											mgr.Logger().Errorf("Failed to create topic %q and ACLs: %s", topic, err)
 										}
-
-										continue
+									} else {
+										mgr.Logger().Infof("Created topic %q", topic)
```
mihaitodor, line 230: There's a debug message which gets printed if the topic already exists...

internal/impl/kafka/enterprise/topic_crud.go
```diff
@@ -56,7 +57,40 @@ func createTopic(ctx context.Context, topic string, replicationFactorOverride bo
...
+			c.Key == "retention.ms" ||
+			c.Key == "segment.ms" ||
+			c.Key == "segment.bytes" ||
+			c.Key == "compression.type" ||
+			c.Key == "message.timestamp.type" ||
+			c.Key == "max.message.bytes" ||
+			c.Key == "replication.factor" ||
+			c.Key == "write.caching" ||
+			c.Key == "redpanda.iceberg.mode" ||
+			strings.HasPrefix(c.Key, "redpanda.remote.") {
```
rockwotj, line 87: I'm not sure we want to mirror `redpanda.remote.*` but just want to take defaults in the destination cluster. The thing is in Redpanda Cloud we never want this to be off, and the default is always on. Maybe we should just not mirror it...

rockwotj (approved):
LGTM, the only thing is that I think we don't want to mirror redpanda.remote.* properties

internal/impl/kafka/enterprise/topic_crud.go
mihaitodor: Sounds good, I removed `redpanda.remote.` from that list.

internal/impl/kafka/enterprise/topic_crud.go
```diff
@@ -56,7 +56,39 @@ func createTopic(ctx context.Context, topic string, replicationFactorOverride bo
...
+			c.Key == "retention.bytes" ||
+			c.Key == "retention.ms" ||
+			c.Key == "segment.ms" ||
+			c.Key == "segment.bytes" ||
+			c.Key == "compression.type" ||
+			c.Key == "message.timestamp.type" ||
+			c.Key == "max.message.bytes" ||
+			c.Key == "replication.factor" ||
+			c.Key == "write.caching" ||
+			c.Key == "redpanda.iceberg.mode" {
```
rockwotj, line 85: Maybe simpler as a switch or a slices.Contain?

rockwotj (approved):

internal/impl/kafka/enterprise/topic_crud.go
mihaitodor: I went for a map since that's what Copilot suggested when I asked it to make the code look nicer 😅

rockwotj (approved):
lgtm

PR #3208, opened 2025-02-25
snowflake: add an option to override the URL

rockwotj:
Needed for private link and gov cloud type deployments


mihaitodor (approved):
LGTM 🐑 🚀

rockwotj:
Docs here:

https://docs.snowflake.com/en/user-guide/admin-security-privatelink#specify-hostname-for-snowflake-clients

PR #3211, opened 2025-02-26
snowflake: add option to ignore null values in schema evolution

rockwotj:
Mor options! This allows folks to include nulls when managing the
schema. If you are managing the schema explicitly (ie. using a schema in
schema registry) this can make a lot of sense, by default we expect
people to just throw data at the connector and we should just drop nulls
because we don't know what type the column will eventually be.

### DIFF
--- a/internal/impl/snowflake/output_snowflake_streaming.go
+++ b/internal/impl/snowflake/output_snowflake_streaming.go
@@ -51,6 +51,7 @@ const (
 	ssoFieldBuildChunkSize                      = "chunk_size"
 	ssoFieldSchemaEvolution                     = "schema_evolution"
 	ssoFieldSchemaEvolutionEnabled              = "enabled"
+	ssoFieldSchemaEvolutionIgnoreNulls          = "ignore_nulls"
 	ssoFieldSchemaEvolutionNewColumnTypeMapping = "new_column_type_mapping"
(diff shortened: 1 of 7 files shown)


internal/impl/snowflake/output_snowflake_streaming.go
```diff
@@ -121,6 +122,7 @@ ALTER TABLE t1 ADD COLUMN a2 NUMBER;
 `),
 			service.NewObjectField(ssoFieldSchemaEvolution,
 				service.NewBoolField(ssoFieldSchemaEvolutionEnabled).Description("Whether schema evolution is enabled."),
+				service.NewBoolField(ssoFieldSchemaEvolutionIgnoreNulls).Description("If `true`, then new columns that are `null` are ignored and schema evolution is not triggered. If `false` then null columns trigger schema migrations in Snowflake. NOTE: unless you already know what type this column will be in advanced, it's highly encouraged to ignore null values.").Default(true).Advanced(),
```
mihaitodor, line 125: Typo: `advanced` -> advance

mihaitodor (approved):
LGTM, just one tiny typo :) :shipit:

internal/impl/snowflake/output_snowflake_streaming.go
rockwotj: Nice catch :)

PR #3217, opened 2025-02-27
Fix Go versions in CI

mihaitodor:
This is a small cleanup to switch to the [`stable`/`oldstable` aliases](https://github.com/actions/setup-go?tab=readme-ov-file#using-stableoldstable-aliases) and I'm not 100% sure if it works well in practice. We have the `toolchain` specified in the `go.mod` [for Benthos](https://github.com/redpanda-data/benthos/blob/4d6fce617665d27d5599fc4a6b44769193d09cb9/go.mod#L78) and I'm not sure exactly why it doesn't get added for Connect as well (might have something to do with the [`replace` directive](https://github.com/redpanda-data/connect/blob/4196ff3294a2af12ef6df79cbfba5ae0b0030272/go.mod#L5) we have in `go.mod`), so I think Go will download the appropriate version anyway...

WDYT?

### DIFF
--- a/.github/workflows/cross_build.yml
+++ b/.github/workflows/cross_build.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.21.x, 1.22.x]
+        go-version: [oldstable, stable]
         os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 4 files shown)


mihaitodor:
OK, I'll merge it and we'll see how it goes 😬

PR #3218, opened 2025-02-27
mongo/cdc: support partial updates

rockwotj:


internal/impl/mongodb/cdc/input.go
```diff
@@ -276,6 +298,14 @@ type mongoBatch struct {
 	ackFn     service.AckFunc
 }
 
+type documentMode int
+
+const (
+	documentModePreAndPostImage = iota
```
mihaitodor, line 304: Shouldn't this be `documentModePreAndPostImage documentMode = iota`?

mihaitodor (approved):
LGTM, just one tiny nit. Feel free to 🐑 🚀

internal/impl/mongodb/cdc/input.go
rockwotj: Ah yes thanks. Good catch

mihaitodor: Really wish Go didn't have these implicit conversions...

PR #3220, opened 2025-02-28
snowflake: fix timestamps at extreme ranges

rockwotj:
Fix a bug where timestamps at the extreme ends of the spectrum could be encoded incorrectly due to a bug in `int128.Div`. Also fix Time types in avro to be something more usable.


internal/impl/snowflake/streaming/int128/int128.go
```diff
@@ -145,7 +145,7 @@ func fls128(n Num) int {
 	if n.hi != 0 {
 		return 127 - bits.LeadingZeros64(uint64(n.hi))
 	}
-	return 64 - bits.LeadingZeros64(n.lo)
+	return 63 - bits.LeadingZeros64(n.lo)
```
rockwotj, line 148: This was the bug fix, the rest is mostly tests or refactoring/clean up :)

internal/impl/snowflake/streaming/int128/division.go
```diff
@@ -58,6 +58,28 @@ func Div(dividend, divisor Num) Num {
...
+// CompareUnsigned returns -1 if |a| < |b|, 0 if a == b, and 1 if |a| > |b|.
+func CompareUnsigned(a, b Num) int {
+	if uint64(a.hi) < uint64(b.hi) || (a.hi == b.hi && a.lo < b.lo) {
+		return -1
+	}
+	if a.hi == b.hi && a.lo == b.lo {
+		return 0
+	}
+	return 1
+}
```
mihaitodor, line 81: I wonder if we can leverage https://pkg.go.dev/cmp#Compare for these. Not sure if that's possible

internal/impl/snowflake/streaming/int128/int128.go
mihaitodor: Heh, that reminded me of https://pkg.go.dev/math/rand#Int63 :)

mihaitodor (approved):
LGTM, just a small nit. Feel free to 🐑 🚀

internal/impl/snowflake/streaming/int128/division.go
rockwotj: Yes we can make this much more readable :) Thanks for this!

PR #3221, opened 2025-03-01
Fix aws_s3 input missing SQS keys

mihaitodor:
We currently keep retrying SQS messages which correspond to notifications about S3 files which have been deleted before we got a chance to process them. It makes more sense to just log a warning and drop the notification.


rockwotj (approved):
LGTM, thanks for fixing this one Mihai!

PR #3225, opened 2025-03-03
snowflake: don't mutate stats values

rockwotj:
- **snowflake: don't mutate stats values**
- **Revert "snowflake: simplify bytes converter"**


rockwotj:
This bug is incredibly subtle. We have to have an error after an upload and see the original data is mutated and recreate the file with the new corrupted data.

PR #3232, opened 2025-03-04
ci: free more bytes in release runs

rockwotj:
Looks like Android alone is something like ~12GB of space we can save.


c4milo (approved):

mihaitodor (approved):
:shipit:

PR #3239, opened 2025-03-05
feat(parquet): add support for more logical types

ooesili:


internal/impl/parquet/schema_coersion.go
```diff
@@ -0,0 +1,148 @@
...
+
+type decodingCoersionVisitor struct{}
+
+func (decodingCoersionVisitor) visitLeaf(value any, schemaNode parquet.Node) (any, error) {
+	logicalType := schemaNode.Type().LogicalType()
+	if logicalType == nil {
+		return value, nil
+	}
+	if logicalType.Timestamp != nil {
+		tsNum, ok := value.(int64)
```
rockwotj, line 120: https://pkg.go.dev/github.com/redpanda-data/benthos/v4@v4.44.1/public/bloblang#ValueAsTimestamp

ooesili: The fact that parquet lets the timestamp be stored as nanos, millis, or micros makes that function less useful

rockwotj: What do you mean? Just because it's stored as some form doesn't mean we should interpret it differently from the bloblang semantics.

ooesili: `ValueAsTimestamp` assumes that the given `int64` is the number of seconds since the epoch. Parquet allows the timestamp to be stored as nano, milli, or microseconds, so we have to cover each case here and can't just call that function.

internal/impl/parquet/processor_decode.go
```diff
@@ -72,13 +87,27 @@ func init() {
 //------------------------------------------------------------------------------
 
 func newParquetDecodeProcessorFromConfig(conf *service.ParsedConfig, logger *service.Logger) (*parquetDecodeProcessor, error) {
+	handleLogicalTypes, err := conf.FieldString(pFieldHandleLogicalTypes)
+	if err != nil {
+		return nil, err
+	}
+
+	logicalTypesVersion, err := strconv.Atoi(handleLogicalTypes[1:])
```
mihaitodor, line 95: I know it's extra plumbing, but can we have a const for each version and use a `switch` statement? If not, users can pass arbitrary data here if running via the `--chilled` flag.

internal/impl/parquet/processor_decode.go
```diff
@@ -72,13 +87,27 @@ func init() {
...
+	if err != nil {
+		return nil, fmt.Errorf("invalid field format: %s", handleLogicalTypes)
+	}
+
 	return &parquetDecodeProcessor{
 		logger: logger,
+		visitor: decodingCoersionVisitor{
```
mihaitodor, line 102: Did you mean `coercion`?

internal/impl/parquet/schema_coersion.go
mihaitodor, line 1: Same `coercion` in the filename

internal/impl/parquet/schema_coersion.go
```diff
@@ -0,0 +1,172 @@
...
+			unit := logicalType.Timestamp.Unit
+			if unit.Millis != nil {
+				return ts.UnixMilli(), nil
+			} else if unit.Micros != nil {
+				return ts.UnixMicro(), nil
+			} else if unit.Nanos != nil {
+				return ts.UnixNano(), nil
+			} else {
+				panic("unreachable")
+			}
```
mihaitodor, line 94: Nit: This might look better as a `switch` statement and, also, it would be best to return an error instead of having a `panic` even if it's never going to happen so we stick with the same error handling pattern.

internal/impl/parquet/schema_coersion.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+type encodingCoersionVisitor struct{}
+
+func (encodingCoersionVisitor) visitLeaf(value any, schemaNode parquet.Node) (any, error) {
+	logicalType := schemaNode.Type().LogicalType()
+	if logicalType == nil {
+		return value, nil
+	}
+	if logicalType.Timestamp != nil {
```
mihaitodor, line 78: Nit: I'd consider turning this into a `switch`

internal/impl/parquet/schema_coersion.go
```diff
@@ -0,0 +1,172 @@
...
+				return ts.UnixNano(), nil
+			} else {
+				panic("unreachable")
+			}
+		default:
+			return nil, errors.New("TIMESTAMP values must be RFC3339-formatted strings")
+		}
+	} else if logicalType.Json != nil {
+		switch value.(type) {
+		case map[string]any, []any:
```
mihaitodor, line 100: Why are we only calling `json.Marshal` for these 2 cases? I think the marshaler handles primitive types as well. The reason I'm asking is because I think there are some string escaping rules and number formatting which is done by `json.Marshal` and maybe that's the intent when the logical type is JSON.

internal/impl/parquet/schema_coersion.go
```diff
@@ -0,0 +1,172 @@
...
+			schemaSpec := logicalType.Timestamp
+			var ts time.Time
+			if schemaSpec.Unit.Millis != nil {
+				ts = time.UnixMilli(tsNum)
+			} else if schemaSpec.Unit.Micros != nil {
+				ts = time.UnixMicro(tsNum)
+			} else if schemaSpec.Unit.Nanos != nil {
+				ts = time.Unix(tsNum/1e9, tsNum%1e9)
+			} else {
+				panic("unreachable")
```
mihaitodor, line 151: Same about panics and `switch`

mihaitodor (approved):
LGTM, just left a bunch of really small nits. Feel free to ignore them if they're too nitpicky :) :shipit:

internal/impl/parquet/processor_decode.go
ooesili: Makes sense, turned it into a switch statement and made sure that invalid values throw an error.

internal/impl/parquet/processor_decode.go
ooesili: Lol, fixed.

internal/impl/parquet/schema_coersion.go
ooesili: Turned into a switch statement

internal/impl/parquet/schema_coersion.go
ooesili: Done.

internal/impl/parquet/schema_coersion.go
ooesili: I'm not sure why I wrote it that way. Updated the code to use `json.Marshal` in all cases and it can now handle primitive values like strings and numbers. Good catch.

mihaitodor (approved):
LGTM, thank you for all the changes! 🐑 🚀

PR #3241, opened 2025-03-05
feat(sftp): add private_key field

ooesili:
This allows users to directly include their private key into the config if they don't or can't write it to a file.


rockwotj (approved):
LGTM

mihaitodor (approved):
LGTM 🐑 🚀

PR #3244, opened 2025-03-06
Add new flags for overriding redpanda logs and status updates

Jeffail:
This PR adds a suite of hidden cli flags to the `run` subcommand. The intention here is to allow the cloud product to override the `redpanda` broker connection and log/status fields of a config without actually mutating the config itself. When this is done the logs and status updates will be routed via the cli flags regardless of how the config is structured, and if the config specifies its own series of `redpanda` fields then they will be used for any `redpanda_common` components and nothing else.

We could potentially go one step further and make the logging and status updates respect both mechanisms, so that the cloud logs work and also customers can route logs themselves somewhere else. However, I'll leave that as a future task.


internal/impl/kafka/enterprise/global_redpanda_writer.go
```diff
@@ -0,0 +1,402 @@
...
+	}
+
+	f := franzTopicLoggerWriter{
+		log:         log,
+		connDetails: connDetails,
+	}
+
+	f.clientOpts = f.connDetails.FranzOpts()
+
+	// All other options (producer, etc) is currently set to the defaults.
```
mihaitodor, line 299: Nit: are

mihaitodor (approved):
LGTM 🐑 🚀

PR #3246, opened 2025-03-07
Add permission check to gcp pubsub input to prevent StreamingPull errors

volkerhess2:
This PR adds a permission check to the gcp pubsub input to prevent running into this gcp error: `Quota exceeded for quota metric 'Regional StreamingPull number of open connections' and limit 'Regional StreamingPull number of open connections per minute per region'`
Benthos tries to reconnect very fast to a subscription if there is a wrong subscription name in input config or change of  subscription permissions resulting in this error and blocking all other services that use the StreamingPull api.

### DIFF
--- a/internal/impl/gcp/input_pubsub.go
+++ b/internal/impl/gcp/input_pubsub.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"strings"
 	"sync"
+	"time"
 
 	"cloud.google.com/go/pubsub"
 	"google.golang.org/api/option"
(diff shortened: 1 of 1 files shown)


internal/impl/gcp/input_pubsub.go
```diff
@@ -234,6 +235,14 @@ func (c *gcpPubSubReader) Connect(ignored context.Context) error {
 	sub.ReceiveSettings.MaxOutstandingBytes = c.conf.MaxOutstandingBytes
 	sub.ReceiveSettings.Synchronous = c.conf.Sync
 
+	p, err := sub.IAM().TestPermissions(context.Background(), []string{"pubsub.subscriptions.consume"})
+	if err != nil {
+		return service.NewErrBackOff(err, 5*time.Second)
+	}
+	if len(p) == 0 {
+		return service.NewErrBackOff(errors.New("missing subscription permissions"), 5*time.Second)
+	}
```
rockwotj, line 244: Do these get printed to logs?

```
The testIamPermissions() method is intended for third-party graphical user interfaces (GUIs) that need to display Google Cloud resources based on what the authenticated user has permissions to see. For example, the Google Cloud console internally uses the testIamPermissions() method to determine what resources and functionality are visible to you after authenticating. Different users are typically granted different permissions, and the Google Cloud console hides or exposes items accordingly.
```

Should we really be using this if it's made for GUIs and not programmatic usage?

volkerhess2: The IAM API package itself doesn't have a hint that it is intended for GUIs. Afaik it's not uncommon to use this programmatically, for example kubernetes and core-go do it:

[https://github.com/kubernetes/release/blob/81bb652f045df84f569896d85a7980e9985ed44d/pkg/cve/impl.go#L77](https://github.com/kubernetes/release/blob/81bb652f045df84f569896d85a7980e9985ed44d/pkg/cve/impl.go#L77)
[https://github.com/core-go/log/blob/159ec1677dc4d967c17572e873178c87a894ac24/pubsub/health_checker.go#L49](https://github.com/core-go/log/blob/159ec1677dc4d967c17572e873178c87a894ac24/pubsub/health_checker.go#L49)

rockwotj (approved):

PR #3254, opened 2025-03-11
snowflake: support NaN + Infinity values in floating point stats

rockwotj:
Also fixes a bug with stats for `-0.0`

### DIFF
--- a/internal/impl/snowflake/integration_test.go
+++ b/internal/impl/snowflake/integration_test.go
@@ -16,6 +16,9 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"iter"
+	"math"
+	"math/bits"
 	"os"
 	"strings"
 	"sync"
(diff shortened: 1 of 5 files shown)


internal/impl/snowflake/streaming/stats.go
```diff
@@ -42,7 +44,12 @@ func (s *statsBuffer) UpdateFloat64Stats(v float64) {
 		s.maxRealVal = v
 		s.hasData = true
 	} else {
-		s.minRealVal = min(s.minRealVal, v)
+		if !math.IsNaN(v) {
+			// NaN is bigger than any other number in Snowflake,
+			// so it's only possible to be the minimum value if
+			// every value is NaN
+			s.minRealVal = min(s.minRealVal, v)
```
rockwotj, line 51: ah this is not enough, if the first value is nan we need to overwrite

internal/impl/snowflake/integration_test.go
```diff
@@ -531,3 +535,72 @@ snowflake_streaming:
...
+		t,
+		stream,
+		`SELECT min(a), max(a) FROM integration_test_floats`,
+	)
+	require.Equal(t, [][]string{
+		{"-inf", "NaN"},
+	}, rows)
+}
+
+func powerSet[T any](items []T, minCount int) iter.Seq[[]T] {
```
mihaitodor, line 583: Neat! 😎

mihaitodor (approved):
LGTM :shipit:

PR #3258, opened 2025-03-11
oai: support tool calling

rockwotj:


internal/impl/openai/chat_processor.go
```diff
@@ -51,8 +51,24 @@ const (
 	ocpFieldSchemaRegistryNamePrefix      = "name_prefix"
 	ocpFieldSchemaRegistryURL             = "url"
 	ocpFieldSchemaRegistryTLS             = "tls"
+	// Tool options
+	ocpFieldTool                     = "tools"
```
mihaitodor, line 55: I wonder if this should be `ocpFieldTools`

internal/impl/openai/chat_processor.go
```diff
@@ -495,14 +627,80 @@ func (p *chatProcessor) Process(ctx context.Context, msg *service.Message) (serv
 			}},
 		})
 	}
-	resp, err := p.client.CreateChatCompletion(ctx, body)
-	if err != nil {
-		return nil, err
+	if len(p.tools) > 0 {
+		// TODO: Support parallel tool calls
```
mihaitodor, line 631: Not sure if this helps, but the `parallel` processor could help with that if there's more than one message. Otherwise, the `workflow` processor performs parallel processing when the defined topology supports it.

mihaitodor (approved):
LGTM, feel free to 🐑 🚀

internal/impl/openai/chat_processor.go
rockwotj: Oh it just means that the model returns multiple tools, it's not really "parallel". otherwise. Setting this to true means that the check below for only a single tool call returned would have to be lifted

internal/impl/openai/chat_processor.go
rockwotj: done

PR #3265, opened 2025-03-17
goreleaser: add connect-fips

clee:
This adds a new build target and a new step in the GitHub actions that should install the required microsoft-go compiler so that we can actually build said new target successfully.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,6 +55,16 @@ jobs:
       with:
         go-version: stable
 
+    - name: Install Microsoft Go
+      run: |
+        GO_VERSION=$(go mod edit -json | jq -r .Go)
+        curl -sSLf -o "$RUNNER_TEMP/msgo.tgz" https://aka.ms/golang/release/latest/go${GO_VERSION}-1.linux-amd64.tar.gz
+        [[ -d "$RUNNER_TEMP/bin" ]] || install -d -m 0755 "$RUNNER_TEMP/bin"
+        [[ -d "$RUNNER_TEMP/microsoft" ]] || install -d -m 0755 "$RUNNER_TEMP/microsoft"
+        tar -C "$RUNNER_TEMP/microsoft" -xf "$RUNNER_TEMP/msgo.tgz"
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -19,10 +19,20 @@ jobs:
...
+        [[ -d "$RUNNER_TEMP/microsoft" ]] || install -d -m 0755 "$RUNNER_TEMP/microsoft"
+        tar -C "$RUNNER_TEMP/microsoft" -xf "$RUNNER_TEMP/msgo.tgz"
+        [[ -x "$RUNNER_TEMP/microsoft/go/bin/go" ]] && ln -s "$RUNNER_TEMP/microsoft/go/bin/go" "$RUNNER_TEMP/bin/microsoft-go"
+        echo "$RUNNER_TEMP/bin" >> "$GITHUB_PATH"
+
     - name: Install Go
       uses: actions/setup-go@v5
       with:
-        go-version: oldstable
+        go-version: stable
```
mihaitodor, line 35: Good catch! We should definitely run the tests against the newer version and only test the old version in `cross_build.yml`

.github/workflows/release.yml
```diff
@@ -50,6 +50,16 @@ jobs:
         sudo rm -rf "/usr/local/share/boost"
         sudo rm -rf "$AGENT_TOOLSDIRECTORY"
 
+    - name: Install Microsoft Go
+      run: |
+        GO_VERSION=$(sed -n 3p ./go.mod | awk '{ print $2; }')
```
mihaitodor, line 55: Just a nitpick, but here's a safer alternative if you can run the "Install Go" job first:
```
go mod edit -json | jq -r .Go
```

mihaitodor (approved):
LGTM, thank you! 🐑 🚀

.github/workflows/test.yml
clee: Realized I missed this in the release.yml so changed oldstable->stable there, too.

.github/workflows/release.yml
clee: Awesome, updated to use that

Jeffail:
Hey @clee, looks like this introduced conflicting archive names which breaks goreleaser: https://github.com/redpanda-data/connect/actions/runs/14035011229/job/39290956041, we'll need this to be fixed before the next release is tagged

clee:
Got it, working on a fix, will send out a PR today for review.

PR #3271, opened 2025-03-20
Add a consistency test for Redpanda Migrator

mihaitodor:
This new test ensures that we don't skip records when switching a consumer from source to destination.

### DIFF
--- a/internal/impl/kafka/enterprise/integration_test.go
+++ b/internal/impl/kafka/enterprise/integration_test.go
@@ -856,11 +856,8 @@ output:
 	stream, err := streamBuilder.Build()
 	require.NoError(t, err)
 
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-
 	go func() {
-		err = stream.Run(ctx)
+		err = stream.Run(context.Background())
 		require.NoError(t, err)
 	}()
 
@@ -953,6 +950,7 @@ input:
         - check: '@input_label == "redpanda_migrator_offsets_input"'
           processors:
             - log:
+                level: INFO
                 message: Migrating Kafka offset
                 fields:
                   kafka_offset_topic:            ${! @kafka_offset_topic }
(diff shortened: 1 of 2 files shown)


internal/impl/kafka/enterprise/integration_test.go
```diff
@@ -1343,11 +1351,177 @@ output:
...
+	time.Sleep(1 * time.Second)
+
+	// Fetch the first record key from the source to create the consumer group
+	dummyConsumerGroup := "foobar_cg"
+	keys := fetchRecordKeys(t, source.brokerAddr, dummyTopic, dummyConsumerGroup, 5)
+	require.Len(t, keys, 5)
+	require.Equal(t, 1, keys[0])
+
+	// Wait for the topic and consumer group to be replicated in the destination first
+	for {
```
rockwotj, line 1462: Can we add a condition (even if long) so we don't loop forever? Is something like 1 minute a good enough bound here?

internal/impl/kafka/enterprise/integration_test.go
```diff
@@ -1343,11 +1351,177 @@ output:
...
+			time.Sleep(1 * time.Second)
+			continue
+		}
+		client.Close()
+
+		break
+	}
+
+	var prevSrcKeys []int
+	for {
```
rockwotj, line 1489: Same here let's add an upper bound timeout

rockwotj (approved):
Thanks for this!

internal/impl/kafka/enterprise/integration_test.go
mihaitodor: That got me to experiment with `require.Eventually` and it's a lot cleaner that way, thank you! :)

internal/impl/kafka/enterprise/integration_test.go
```diff
@@ -978,21 +982,27 @@ output:
   redpanda_migrator_bundle:
     redpanda_migrator:
       seed_brokers: [ %s ]
-      topic_prefix: %s
+      topic_prefix: "%s"
       replication_factor_override: true
       replication_factor: -1
+      # TODO: Remove this compression setting once https://github.com/redpanda-data/redpanda/issues/25769 is fixed
+      compression: none
```
mihaitodor, line 989: This was a "fun" rabbit hole...

PR #3275, opened 2025-03-22
Bugfix: amqp_0_9 output: only enforce dots in routing keys for topic exchanges

brknstrngz:
- we currently replace all slashes in routing keys and message types with dots; this is however not mandated for any other exchange type than 'topic'. This change adds a check that skips the replacement for non-topic exchanges, such as 'direct'
- add support for the 'headers' exchange type


mihaitodor (approved):
Thank you for the contribution @brknstrngz! LGTM

PR #3284, opened 2025-03-24
snowflake: fix rare data race

rockwotj:
Technically it's possible to have a data race here where `append` can
realloc and the work function could write to `rowGroups[j]` while the
realloc is happening. In general that should be very rare since I expect
that loop to finish first, but still we should fix by moving the allocs
outside the loop :)

Also create a new writer for every parquet file incase that's an issue...


mihaitodor (approved):
LGTM 🐑 🚀

PR #3285, opened 2025-03-24
git: initial commit for new input

weeco:
A new input connector to clone and poll files from a git repository.


internal/impl/git/input.go
```diff
@@ -0,0 +1,565 @@
...
+	f, err := os.Open(filePath)
+	if err != nil {
+		in.log.Debugf("failed to open file %q for MIME detection: %v. Using fallback text/plain.", filePath, err)
+		return "unknown/error", false
+	}
+	defer f.Close()
+
+	buffer := make([]byte, 512)
+	n, err := f.Read(buffer)
+	if err != nil {
```
rockwotj, line 419: Can `err` be EOF here?

internal/impl/git/input.go
```diff
@@ -0,0 +1,565 @@
...
+		in.log.Debugf("failed to read file %q for MIME detection: %v. Using fallback text/plain.", filePath, err)
+		return "unknown/error", false
+	}
+
+	// Detect content type and check if binary
+	contentType := http.DetectContentType(buffer[:n])
+
+	// For known file extensions, override the content type
+	if mimeType, exists := extensionToMIME[contentType]; exists {
```
rockwotj, line 428: I'm confused, why are we using a contentType as a key in a map that is key'd by file extension.

internal/impl/git/input.go
```diff
@@ -0,0 +1,550 @@
...
+	if err != nil {
+		panic(err)
+	}
+}
+
+// newInput creates a new Git input instance from a parsed configuration.
+func newInput(cfg inputCfg, mgr *service.Resources) (*input, error) {
+	i := &input{
+		cfg:       cfg,
+		filesChan: make(chan fileEvent, 100),
```
rockwotj, line 93: Why 100? I would suggest just making an unbuffered channel for now unless we see performance issues

internal/impl/git/input.go
```diff
@@ -0,0 +1,565 @@
...
+				// For deleted files, create a message with empty content and metadata
+				msg := service.NewMessage(nil)
+				relPath, err := filepath.Rel(in.tempDir, event.path)
+				if err != nil {
+					return nil, nil, fmt.Errorf("failed to get relative path for %s: %w", event.path, err)
+				}
+				msg.MetaSet("git_file_path", relPath)
+				msg.MetaSet("git_commit", in.lastCommit.String())
+				msg.MetaSet("git_deleted", "true")
+				return msg, func(ctx context.Context, err error) error { return nil }, nil
```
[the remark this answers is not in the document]
rockwotj: What do you want to do here @weeco? Punt or are you planning to add the cache?

internal/impl/git/input.go
```diff
@@ -0,0 +1,550 @@
...
+		panic(err)
+	}
+}
+
+// newInput creates a new Git input instance from a parsed configuration.
+func newInput(cfg inputCfg, mgr *service.Resources) (*input, error) {
+	i := &input{
+		cfg:       cfg,
+		filesChan: make(chan fileEvent, 100),
+		closeCh:   make(chan struct{}),
```
rockwotj, line 94: Consider using a `*shutdown.Signaller`

internal/impl/git/input.go
```diff
@@ -0,0 +1,550 @@
...
+		log:       mgr.Logger(),
+	}
+
+	return i, nil
+}
+
+// Connect implements service.Input. It initializes the Git repository by creating
+// a temporary directory, cloning the repository, and starting the polling routine.
+func (in *input) Connect(context.Context) error {
```
rockwotj, line 103: If an input is opened and closed multiple times do we need to reset `closeCh`?

internal/impl/git/input_config.go
```diff
@@ -0,0 +1,342 @@
...
+				Default([]any{}).
+				Optional(),
+			service.NewStringListField("exclude_patterns").
+				Description("A list of file patterns to exclude (e.g., '.git/**', '**/*.png'). These patterns take precedence over include_patterns. "+
+					"Supports glob patterns: *, /**/, ?, and character ranges [a-z]. Any character with a special meaning can be escaped with a backslash.").
+				Default([]any{}).
+				Optional(),
+			service.NewBoolField("recursive").
+				Description("Whether to recursively read files in subdirectories.").
+				Default(true),
```
rockwotj, line 78: Personally I think we should remove this option and never emit directories but instead emit full files, this recursive behavior can be changed with the `include_patterns`.

internal/impl/git/integration_test.go
```diff
@@ -0,0 +1,128 @@
...
+	git:
+		repository_url: "test-dir"
+		branch: "master"
+		poll_interval: 1s
+
+pipeline: {}
+
+output:
+	stdout:
+		codec: lines
```
rockwotj, line 109: You can just omit these

internal/impl/git/integration_test.go
```diff
@@ -0,0 +1,128 @@
...
+`)
+
+	integration.StreamTests(
+		integration.StreamTestOpenClose(),
+		integration.StreamTestStreamSequential(20),
+		integration.StreamTestSendBatchCount(10),
+		integration.StreamTestStreamParallelLossyThroughReconnect(20),
```
rockwotj, line 116: I also find the service.StreamBuilder method of writing tests to be good.

internal/impl/git/input.go
```diff
@@ -0,0 +1,550 @@
...
+}
+
+// detectMimeType determines the MIME type of a file by examining its contents or by looking
+// at the file name's extension.
+func (in *input) detectMimeType(filePath string) (string, bool) {
+	// Read the first 512 bytes of the file for MIME detection
+	f, err := os.Open(filePath)
+	if err != nil {
+		in.log.Debugf("failed to open file %q for MIME detection: %v. Using fallback text/plain.", filePath, err)
+		return "unknown/error", false
```
rockwotj, line 412: Log disagrees. Should we be falling back to `application/octet-stream` and binary?

rockwotj (commented):
Looking good, I just have a question if we're doing the cache, the git notes checkpointing or punting on checkpointing

internal/impl/git/input.go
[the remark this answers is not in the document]
weeco: I'll add it support for a cache today. Expect a PR in a bit

internal/impl/git/input.go
```diff
@@ -0,0 +1,636 @@
...
+	return nil
+}
+
+// Read implements service.Input. It returns the next available file content as a message,
+// or returns an error if the context is cancelled or shutdown is signaled.
+func (in *input) Read(ctx context.Context) (*service.Message, service.AckFunc, error) {
+	for {
+		select {
+		case <-ctx.Done():
+			in.shutSig.TriggerHasStopped()
```
weeco, line 155: This is annoying, but Close() is not called while Read is still blocking.

rockwotj: This feels wrong. `TriggerHasStopped` should be called when `pollChanges` goroutine finishes meaning that everything has stopped and shutdown before Close exits.

internal/impl/git/input.go
```diff
@@ -0,0 +1,636 @@
...
+	if err != nil {
+		return fmt.Errorf("git clone failed: %w", err)
+	}
+
+	ref, err := in.repository.Head()
+	if err != nil {
+		return fmt.Errorf("failed to get HEAD reference: %w", err)
+	}
+
+	in.setLastCommit(ctx, ref.Hash())
```
rockwotj, line 233: This should only happen once we `ack` stuff otherwise it's not at least once

rockwotj:
@weeco please take a look at my changes and see what you think

rockwotj:
Force push to fix conflicts

internal/impl/git/input.go
```diff
@@ -0,0 +1,709 @@
...
+	in.lastCommitMu.RLock()
+	defer in.lastCommitMu.RUnlock()
+	return in.lastCommit
+}
+
+// pollChanges runs in a separate goroutine and periodically checks for updates
+// in the Git repository according to the configured poll interval.
+func (in *input) pollChanges(ctx context.Context, cachedCommit plumbing.Hash) {
+	hasCheckpoint := cachedCommit != plumbing.ZeroHash
+	var initialScanWg *sync.WaitGroup
```
weeco, line 292: Nicely done with the waitgroup.
I ignored the need for acking the messages before I can mark them committed 🤦

internal/impl/git/input.go
```diff
@@ -0,0 +1,709 @@
...
+	in.lastCommit = newCommit
+	in.lastCommitMu.Unlock()
+
+	if in.cfg.checkpointCache == "" {
+		return
+	}
+	if err := in.updateCheckpointCache(ctx, newCommit); err != nil {
+		in.log.Errorf("failed to update checkpoint cache: %v", err)
+	}
+}
```
weeco, line 279: NIT: We could add a param here that decides whether we want to save the commit as a checkpoint so that we don't need to lock/unlock repeatedly. May be opinionated, so please feel free to ignore.

```suggestion
// setLastCommit sets the in.lastCommit field and updates the checkpoint cache (if configured).
func (in *input) setLastCommit(ctx context.Context, newCommit plumbing.Hash, saveCheckpoint bool) {
	in.lastCommitMu.Lock()
	in.lastCommit = newCommit
	in.lastCommitMu.Unlock()
	
	if !saveCheckpoint {
		return
	}

	if in.cfg.checkpointCache == "" {
		return
	}
	if err := in.updateCheckpointCache(ctx, newCommit); err != nil {
		in.log.Errorf("failed to update checkpoint cache: %v", err)
	}
}
```

rockwotj (approved):
LGTM - I did some local testing on a gist and seems to work nicely.

PR #3292, opened 2025-03-25
text: add text chunking processors

rockwotj:
- **text: import textsplitters from langchain-go**
- **text: add embeddings document chunker**


internal/impl/text/chunker/options.go
```diff
@@ -0,0 +1,181 @@
+// The MIT License
```
emaxerrno, line 1: apache2?

emaxerrno: ah this is an import nvm.

internal/impl/text/text_chunker_processor.go
```diff
@@ -0,0 +1,293 @@
+// Copyright 2025 Redpanda Data, Inc.
```
emaxerrno, line 1: very cool. functionally lgtm, will let others comment on the details.

internal/impl/text/chunker/options.go
rockwotj: Yeah it's a fork from langchain go.

Using that library as a dependency would be an option but it's got a massive number of transitive dependencies and this package is all we need right now (and it has very little in terms of deps)

rockwotj: I will add additional comments to make this more clear

Jeffail (approved):
Looks great!

mihaitodor (approved):
LGTM, just left a few nits. Feel free to :shipit:

internal/impl/text/text_chunker_processor.go
```diff
@@ -0,0 +1,294 @@
...
+	splitter  chunker.TextSplitter
+	secondary *processorSplitter
+}
+
+// Process implements service.Processor.
+func (t *textChunker) Process(ctx context.Context, msg *service.Message) (service.MessageBatch, error) {
+	if t.secondary != nil {
+		t.secondary.context = ctx
+		defer func() { t.secondary.context = nil }()
+	}
```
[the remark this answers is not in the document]
rockwotj: Oof good catch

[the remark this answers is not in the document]
rockwotj: Just removed this for now

PR #3294, opened 2025-03-26
Fix azure_blob_storage input missing blobs

### LINKED ISSUE #3189: Files are not all removed for azure_blob_storage with max_in_flight value greater than 1
Hi, we're observing that, if we set `max_in_flight` greater than 1, all files are not removed. With log level at `DEBUG` we can see that all files are handled, so it's seems to works.

This is our input resource configuration

```
---
input_resources:
  - label: main
    azure_blob_storage:
      storage_connection_string: ${AZURE_STORAGE_CONNECTION_STRING}
      storage_sas_token: ${AZURE_STORAGE_SAS_TOKEN:""}
      container: ${AZURE_CONTAINER}
      delete_objects: ${AZURE_STORAGE_INPUT_DELETE_OBJECT:true}
      prefix: "replay/"
      scanner:
        to_the_end: {}
      targets_input:
        azure_queue_storage:
          storage_connection_string: ${AZURE_STORAGE_CONNECTION_STRING}
          queue_name: ${AZURE_QUEUE_IN}
          dequeue_visibility_timeout: ${AZURE_DEQUEUE_TIMEOUT:30s}
          max_in_flight: ${AZURE_QUEUE_MAX_IN_FLIGHT:1}
          track_properties: false
        processors:
          - mapping: |-
              let json = content().decode("base64").parse_json()
              if $json.eventType == "Microsoft.Storage.BlobCreated" {
                root.name = $json.data.url.parse_url().path.trim_prefix("/" + env("AZURE_CONTAINER") + "/")
              } else {
                root = deleted()
              }
```

mihaitodor:
Fixes #3189, or so I hope...

This also updates the Azure dependencies.

Annoyingly, the original CosmosDB emulator doesn't have an ARM-based container and the new one ([`mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:vnext-preview`](https://learn.microsoft.com/en-us/azure/cosmos-db/emulator-linux)) has some gaps which break the existing tests such as: https://github.com/Azure/azure-cosmos-db-emulator-docker/issues/182.


rockwotj (approved):

mihaitodor:
Thanks @rockwotj! I sneaked in there a small fix for the `aws_s3` input as well. I think the warning message should be more consistent now if users somehow end up processing multiple deleted objects in the same batch. I somehow overlooked that in #3221.

PR #3298, opened 2025-03-27
slack: add input

rockwotj:
Adds a slack input using SocketMode.

There isn't really a mock server anywhere for socketmode, but I did manual testing that it works.


internal/impl/slack/input.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+}
+
+const (
+	iFieldAppToken = "app_token"
+	iFieldBotToken = "bot_token"
+)
+
+func inputSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
```
mihaitodor, line 37: Missing docs about metadata

mihaitodor (approved):
Oh, cool! I'd like to play with this one 😁 Just a small nit & feel free to 🐑 🚀 

PS: It's missing a changelog entry

internal/impl/slack/input.go
rockwotj: It's in there

mihaitodor: Oops... My bad. I was looking for the typical pattern. Never mind 🙈

rockwotj: yeah was certainly lazy and am relying on Angela to make it right

PR #3303, opened 2025-03-28
ai: add history to ollama and openai chat processors

rockwotj:


emaxerrno (approved):
ah neat!

internal/impl/openai/chat_processor.go
```diff
@@ -587,6 +602,21 @@ func (p *chatProcessor) Process(ctx context.Context, msg *service.Message) (serv
 			Content: s,
 		})
 	}
+	if p.history != nil {
+		msg, err := msg.BloblangQuery(p.history)
```
emaxerrno, line 606: only thing would be if history should be a storage query to postgres or whatever memory the user configured.

malinskibeniamin (approved):
What will be the final schema for me to use for the chat, right now I just append up to 30 messages to the only prompt.

internal/impl/openai/chat_processor.go
rockwotj: Right now that will have to be done manually, we can add that on in another PR

rockwotj:
@malinskibeniamin I added an example to the PR of how to use memory for history (using a cache)

```
What color is the sky in one word?
Blue.
What did I just ask you?
You asked me, "What color is the sky in one word?"
How many times have you responded to me?
I have responded to you twice.
You're cool
Thank you! If you have any more questions or need assistance, feel free to ask.
```

internal/impl/openai/chat_processor.go
rockwotj: Provided an example of manually hooking up storage

PR #3306, opened 2025-03-28
fips: add wrapper script, build hook

clee:
Also generates separate redpanda-connect-fips packages, but no tar.gz archives.


mihaitodor (approved):
LGTM 🐑 🚀

PR #3316, opened 2025-04-02
git: clone commit history for catching up

weeco:
When catching up with a given commit we need to
also clone the commit history. Otherwise, the commit we try to continue with, will be unknown.


mihaitodor (approved):
LGTM 🐑 🚀 🐏 🚀 🐐 🚀

rockwotj (approved):

PR #3322, opened 2025-04-04
fix call fmt.Errorf with wrong error

alingse:


mihaitodor (approved):
LGTM, nice catch @alingse!

PR #3340, opened 2025-04-14
Upload connect-fips plugin

clee:
Distinct from the standard connect plugin, should generate a manifest under the /connect-fips/ prefix in the destination.


r-vasquez (approved):
LGTM, thanks!

mihaitodor (approved):
:shipit:

PR #3341, opened 2025-04-15
Add topic prefix support for Redpanda Migrator

mihaitodor:
Also clean up logic a bit to only iterate the batch records once in the `redpanda_migrator` output. The existing code goes over them once when translating schema IDs and another time when creating topics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - Field `start_from_oldest` for the `kafka_franz`, `ockam_kafka`, `redpanda`, `redpanda_common`, and `redpanda_migrator` inputs is now deprecated in favour of `start_offset`. (@mihaitodor)
+- Field `topic_prefix` added to the `redpanda_migrator` output. (@mihaitodor)
+- Field `offset_topic_prefix` added to the `redpanda_migrator_offsets` output. (@mihaitodor)
 
 ## 4.52.0 - 2025-04-03
 
--- a/internal/impl/kafka/enterprise/redpanda_migrator_bundle_output.tmpl.yaml
+++ b/internal/impl/kafka/enterprise/redpanda_migrator_bundle_output.tmpl.yaml
@@ -51,8 +51,8 @@ mapping: |
     root = throw("redpanda_migrator_bundle.translate_schema_ids must be used instead of schema_registry.translate_ids and redpanda_migrator.translate_schema_ids")
   }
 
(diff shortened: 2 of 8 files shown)


internal/impl/kafka/enterprise/redpanda_migrator_bundle_output.tmpl.yaml
```diff
@@ -62,7 +62,6 @@ mapping: |
       "partition": "${! metadata(\"kafka_partition\").or(throw(\"missing kafka_partition metadata\")) }",
       "partitioner": "manual",
       "timestamp_ms": "${! metadata(\"kafka_timestamp_ms\").or(timestamp_unix_milli()) }",
-      "max_in_flight": this.redpanda_migrator.max_in_flight.or(1),
```
mihaitodor, line 65: I forgot to update this when I switched to the new `FranzReaderOrdered` / `FranzWriter` APIs. The default value (256) should be fine and it should not break message ordering.

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ All notable changes to this project will be documented in this file.
 - Fields `region`, `endpoint` and `credentials` added to the `dynamodb` configuration section of the `aws_kinesis` input. (@jreyeshdez, @mihaitodor)
 - Field `isolation_level` added to `kafka`, `kafka_franz`, `ockam_kafka`, `redpanda`, `redpanda_common`, and `redpanda_migrator` inputs. (@rockwotj)
 - New `cohere_rerank` processor to rerank documents in RAG pipelines using Cohere. (@rockwotj)
+- Field `topic_prefix` added to the `redpanda_migrator` output. (@mihaitodor)
+- Field `offset_topic_prefix` added to the `redpanda_migrator_offsets` output. (@mihaitodor)
```
mihaitodor, line 18: I decided to go with the easy route and only support prefixes. It's still a bit awkward, since the `topic` field on the `redpanda_migrator` output allows users to configure any interpolated value, but if they do that, it's no longer possible to know what the input topic was, so we can't fetch its config. Might need to revise this in the future (or maybe hardcode the `topic` field to `${! @kafka_topic }` and prevent users from changing it...

rockwotj: +1 to hard coding if that breaks things.

rockwotj (approved):

CHANGELOG.md
mihaitodor: I thought about it and the `redpanda_migrator_bundle` output template will reject configs which attempt to set the `topic` field for `redpanda_migrator`, since that's set by the template to the `@kafka_topic` metadata. In general, I don't expect people to use the `redpanda_migrator` output directly, so would rather not make the `FranzWriter` code overly-complex for now.

PR #3345, opened 2025-04-16
Add record order test for the `redpanda` components

mihaitodor:
This test checks for out-of-order records being transferred between two Redpanda containers using the `redpanda` input and output with default settings. It used to fail occasionally before this fix was put in place: https://github.com/redpanda-data/connect/pull/3386.

### DIFF
--- a/internal/impl/kafka/enterprise/integration_test.go
+++ b/internal/impl/kafka/enterprise/integration_test.go
@@ -1542,3 +1542,127 @@ func TestRedpandaMigratorConsumerGroupConsistencyIntegration(t *testing.T) {
 		return false
 	}, 30*time.Second, 1*time.Nanosecond)
 }
+
+func TestRedpandaRecordOrder(t *testing.T) {
+	// This test checks for out-of-order records being transferred between two Redpanda containers using the `redpanda`
+	// input and output with default settings. It used to fail occasionally before this fix was put in place:
+	// https://github.com/redpanda-data/connect/pull/3386.
+	//
+	// Normally, you'll want to let it run multiple times in a loop over night:
+	// ```shell
+	// $ nohup go test -timeout 0 -v -count 10000 -run ^TestRedpandaRecordOrder$ ./internal/impl/kafka/enterprise > test.log 2>&1 &`
+	// ```
+	integration.CheckSkip(t)
+
+	pool, err := dockertest.NewPool("")
+	require.NoError(t, err)
+	pool.MaxWait = time.Minute
+
+	source, err := startRedpanda(t, pool, true, false)
+	require.NoError(t, err)
+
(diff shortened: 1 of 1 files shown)


internal/impl/kafka/enterprise/integration_test.go
```diff
@@ -1542,3 +1542,115 @@ func TestRedpandaMigratorConsumerGroupConsistencyIntegration(t *testing.T) {
...
+    seed_brokers: [ %s ]
+    topics: [ %s ]
+    consumer_group: migrator_cg
+    start_from_oldest: true
+
+output:
+  redpanda:
+    seed_brokers: [ %s ]
+    topic: ${! @kafka_topic }
+    key: ${! @kafka_key }
```
rockwotj, line 1587: The original produced data does mot have a key set, so it goes to a random partition. We will ensure order only because there is a single partition then right? I know we auto create topics in this test, I am assuming those are created with a single partition otherwise I am confused how this works

rockwotj (approved):

internal/impl/kafka/enterprise/integration_test.go
mihaitodor: I just pushed a change to force a single partition. Also, I do set the key to `${! counter() }` [here](https://github.com/redpanda-data/connect/blob/35ed595740a20b29bd3e66f2ef73d415fbc6de62/internal/impl/kafka/enterprise/integration_test.go#L847) so we should have deterministic keys.

internal/impl/kafka/enterprise/integration_test.go
```diff
@@ -1542,3 +1542,127 @@ func TestRedpandaMigratorConsumerGroupConsistencyIntegration(t *testing.T) {
 		return false
 	}, 30*time.Second, 1*time.Nanosecond)
 }
+
+func TestRedpandaRecordOrder(t *testing.T) {
+	// This test checks for out-of-order records being transferred between two Redpanda containers using the `redpanda`
+	// input and output with default settings. It used to fail occasionally before this fix was put in place:
+	// https://github.com/redpanda-data/connect/pull/3386.
+	//
+	// Normally, you'll want to let it run multiple times in a loop over night:
```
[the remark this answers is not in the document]
mihaitodor: Neat! I didn't know about this tool, thank you! In this particular case, it wouldn't really offer much insight unfortunately. We can only detect the out-of-order messages via this test after the data migration and offset translation has occurred, so one should bump up logging as needed if it starts failing again (which is also tricky since it didn't break as often with trace logging 😬). Too many moving pieces unfortunately...

It might be more interesting to use it with tests like the one Ash added [here](https://github.com/redpanda-data/connect/pull/3386).

PR #3346, opened 2025-04-17
feat(deltalake): initial delta lake output

ooesili:


internal/impl/deltalake/output.go
```diff
@@ -0,0 +1,202 @@
...
+		Beta().
+		Categories("Services").
+		Version("4.53.0").
+		Summary("Ingests data into Delta Lake tables such as those managed by Databricks").
+		Description(`
+    Ingests parquet data into Delta Lake tables, supporting both Databricks-managed and open-source Spark deployments.
+`).
+		Fields(
+			service.NewStringField(dlFieldTableURL).
+				Description("The URL to table to ingest data into. Infers the storage backend to use from the scheme in the given table URL."),
```
rockwotj, line 38: Examples would be good

internal/impl/deltalake/output.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/rivian/delta-go"
+	"github.com/rivian/delta-go/lock/nillock"
+	"github.com/rivian/delta-go/state"
+	"github.com/rivian/delta-go/storage"
+	"github.com/rivian/delta-go/storage/s3store"
+)
+
+const (
+	dlFieldTableURL = "table_url"
+	dlFieldBatching = "batching"
```
rockwotj, line 24: We should probably make this a batching output

internal/impl/deltalake/output.go
```diff
@@ -0,0 +1,202 @@
...
+	d.objectStore = objectStore
+	return nil
+}
+
+func (d *deltaLakeWriter) Write(ctx context.Context, msg *service.Message) error {
+	id := "2b0046bb-f9d1-46b1-9476-8eeaa6c27ce5"
+	// id := uuid.New().String()
+	fileName := fmt.Sprintf("part-%s.snappy.parquet", id)
+
+	msgBytes, err := msg.AsBytes()
```
rockwotj, line 115: Is the API assumption that there is an upstream `parquet` processor first? I think that will hurt usability a bit, as it will be more difficult to add schema evolution, etc.

go.mod
```diff
@@ -28,6 +28,7 @@ require (
 	github.com/Masterminds/squirrel v1.5.4
 	github.com/PaesslerAG/gval v1.2.2
 	github.com/PaesslerAG/jsonpath v0.1.1
+	github.com/apache/arrow/go/parquet v0.0.0-20211112161151-bc219186db40
```
mmatczuk, line 31: Please send go.mod file changes in separate commits. It improves readability and it's easier to rebase commits lated.

internal/impl/deltalake/output.go
```diff
@@ -0,0 +1,202 @@
...
+		func(conf *service.ParsedConfig, mgr *service.Resources) (out service.Output, maxInFlight int, err error) {
+			if maxInFlight, err = conf.FieldMaxInFlight(); err != nil {
+				return
+			}
+			out, err = newDeltaLakeWriterFromParsed(conf, mgr)
+			return
+		})
+	if err != nil {
+		panic(err)
+	}
```
mmatczuk, line 54: Would be nice to have `service.MustRegisterOutput` function

internal/impl/deltalake/output.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func newDeltaLakeWriterFromParsed(conf *service.ParsedConfig, mgr *service.Resources) (*deltaLakeWriter, error) {
+	d := &deltaLakeWriter{logger: mgr.Logger()}
```
mmatczuk, line 58: It would be easier to read if deltaLakeWriter was instantiated by the end of the function.
We'd also avoid `var err error`.

internal/impl/deltalake/output.go
mmatczuk: It's not used, do we need it?

internal/impl/deltalake/output.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	d.table = table
+	d.objectStore = objectStore
+	return nil
+}
+
+func (d *deltaLakeWriter) Write(ctx context.Context, msg *service.Message) error {
+	id := "2b0046bb-f9d1-46b1-9476-8eeaa6c27ce5"
+	// id := uuid.New().String()
```
mmatczuk, line 112: What's happening here?

internal/impl/deltalake/output.go
```diff
@@ -0,0 +1,202 @@
...
+	transaction.AddAction(add)
+	operation := delta.Write{Mode: delta.Append}
+	// TODO: metadata
+	transaction.SetOperation(operation)
+	version, err := transaction.Commit()
+	if err != nil {
+		return fmt.Errorf("committing transaction: %w", err)
+	}
+
+	fmt.Printf("version %d\n", version)
```
mmatczuk, line 140: Maybe we'd want to have debug log with more details?

internal/impl/deltalake/output.go
ooesili: @rockwotj Ok, I wasn't sure about this because with parquet the messages are already sort of in a batch. Do you think it's still useful to use a batching output in that case?

internal/impl/deltalake/output.go
ooesili: Whoops, debugging code that I left in

internal/impl/deltalake/output.go
ooesili: There is a follow up ticket to add metrics, I was going to flesh out logging in at that time

PR #3354, opened 2025-04-18
configs: rag pipelines

rockwotj:
In this PR 
* Fix a small bug in the cohere embeddings processor
* Add some example pipelines for RAG using various LLMs (more vector DBs coming soon)
- * I don't think these really need much look over
* Add a qdrant processor for querying from qdrant


mmatczuk:
It would be very cool if we could run the Benthos tests in CI.

rockwotj:
Can you clarify? We do this in CI: https://github.com/redpanda-data/connect/blob/main/Makefile#L80

mmatczuk:
Do you think we could be running the Benthos tests added in this PR in CI?
Correct me if I'm wrong but it looks like we run tests in `./config/test/...` but not in config/rag.

config/rag/docker-compose.yml
```diff
@@ -0,0 +1,102 @@
...
+      - the_network
+
+volumes:
+  pgdata: null
+  redpandadata: null
+  esdata: null
+  qdrant_storage: null
+
+networks:
+  the_network:
```
mihaitodor, line 100: Nit: Might be worth giving this a more descriptive name

internal/impl/qdrant/processor.go
```diff
@@ -0,0 +1,283 @@
...
+	for _, result := range results {
+		b, err := protojson.Marshal(result)
+		if err != nil {
+			return nil, err
+		}
+		points = append(points, json.RawMessage(b))
+	}
+	b, err := json.Marshal(points)
+	if err != nil {
+		return nil, err
```
mihaitodor, line 273: Nit: Might be worth annotating these errors

mihaitodor (approved):
LGTM, just a few small nits. eel free to 🐑 🚀

config/rag/docker-compose.yml
rockwotj: done

PR #3358, opened 2025-04-22
feat(sftp): added modtime metadata to sftp input

anthonyvitale:
Related: #2955, #2970 

Adds modTime to sftp input metadata


mihaitodor:
Thank you for your contribution! I picked up this change in #3498 which should get merged soon. I added you as a Co-author.

PR #3359, opened 2025-04-22
gh: add vulnerability check action

mmatczuk:


mihaitodor (approved):
LGTM, just one small question since we're using `stable` [for releases](https://github.com/redpanda-data/connect/blob/056932209e14e24f94673f0eec743b45c5f391b1/.github/workflows/release.yml#L56) and [for CI](https://github.com/redpanda-data/connect/blob/056932209e14e24f94673f0eec743b45c5f391b1/.github/workflows/test.yml#L25), I'm wondering if it might be best to use `stable` here as well. Feel free to 🐑 🚀

mmatczuk:
I confirm it works with `stable` https://github.com/redpanda-data/connect/actions/runs/14637163125/job/41070827040

PR #3361, opened 2025-04-23
Mcp followup

Jeffail:
This PR adds a tags filter for reducing the scope of resources registered as tools `mcp-server` subcommand along with open telemetry support.


mmatczuk:
It's timeout, not sure why it takes so long now. I'd revert it to 30m in CI.

mmatczuk:
@Jeffail https://github.com/redpanda-data/connect/pull/3362 should fix it

internal/cli/mcp_server.go
```diff
@@ -14,6 +14,7 @@ import (
 	"log/slog"
 	"os"
 
+	"github.com/go-faker/faker/v4/pkg/slice"
```
mmatczuk: How about using https://pkg.go.dev/slices#Contains

PR #3366, opened 2025-04-24
ai: add licensed connector support to agent/mcp server commands

rockwotj:


Jeffail (approved):
Approved but on the condition that https://github.com/redpanda-data/benthos/pull/218 is merged and pulled in first.

PR #3367, opened 2025-04-25
mk: add run target to streamline running dev configs

mmatczuk:
Calling make run would compile redpanda-connect CLI and call it with --config ./config/dev.yaml.


mihaitodor (approved):
LGTM :shipit: You might want to mirror it in Benthos as well

PR #3374, opened 2025-04-29
Fix flag init to use config spec defaults

Jeffail:


nicolaferraro (approved):
I see FranzConnectionFields using a default of 10s

PR #3378, opened 2025-04-29
schema_registry_decode: Fix Schema Decoding Reference Order for Correct Schema Resolution Behavior

### LINKED ISSUE #3363: Error in Schema Decoding with References in Redpanda Connect Processor(schema_registry_decode)
I encountered an issue when trying to decode a avro message using a schema that references another schema. The error message is:

```
{"@service":"redpanda-connect","label":"","level":"debug","msg":"Processor failed: unable to parse schema avro: unknown type: SchemaA","path":"root.pipeline.processors.0"}
```
Steps to Reproduce:
1- Define SchemaA:
```
{"schema":"{\"type\":\"record\",\"name\":\"SchemaA\",\"namespace\":\"namespaceA\",\"fields\":[{\"name\":\"field8\",\"type\":\"double\"},{\"name\":\"field9\",\"type\":\"double\"}]}"}


```

2- Define SchemaB:
```
{"schema":"{\"type\":\"record\",\"name\":\"SchemaB\",\"namespace\":\"namespaceA\",\"fields\":[{\"name\":\"field7\",\"type\":\"SchemaA\"}]}","references":[{"name":"namespaceA.SchemaA","subject":"namespaceA.SchemaA","version":1}]}

```
3- Define SchemaC:
```
{"schema":"{\"type\":\"record\",\"name\":\"SchemaC\",\"namespace\":\"namespaceA\",\"fields\":[{\"name\":\"field1\",\"type\":\"SchemaB\"},{\"name\":\"field2\",\"type\":\"string\"}]}","references":[{"name":"namespaceA.SchemaB","subject":"namespaceA.SchemaB","version":1}]}
```

4- Define SchemaD:
```
{"schema":"{\"type\":\"record\",\"name\":\"SchemaD\",\"namespace\":\"namespaceA\",\"fields\":[{\"name\":\"field3\",\"type\":\"namespaceA.SchemaA\"}]}","references":[{"name":"namespaceA.SchemaA","subject":"namespaceA.SchemaA","version":1}]}
```

5- Define SchemaE:
```
{"schema":"{\"type\":\"record\",\"name\":\"SchemaE\",\"namespace\":\"namespaceB\",\"fields\":[{\"name\":\"field4\",\"type\":\"string\"},{\"name\":\"field5\",\"type\":\"namespaceA.SchemaD\"},{\"name\":\"field6\",\"type\":\"namespaceA.SchemaC\"}]}","references":[{"name":"namespaceA.SchemaD","subject":"namespaceA.SchemaD","version":1},{"name":"namespaceA.SchemaC","subject":"namespaceA.SchemaC","version":1}]}
```

6- Use the following processor configuration:
```
pipeline:
  processors:
    - schema_registry_decode:
        url: http://redpanda:8081
        avro:
          preserve_logical_types: true
```

Expected Behavior:
The message should decode successfully.

Actual Behavior:
The error message indicates a failure to parse the schema.

The message published in Avro format that triggers the error:
```
{
  "field5": {
    "field3": {
      "field8": 1.1,
      "field9": 1.2
    }
  },
  "field6": {
    "field1": {
      "field7": {
        "field8": 1.5,
        "field9": 1.6
      }
    },
    "field2": "value2"
  },
  "field4": "value5"
}
```
I investigated the issue and it seems to be related to referencing SchemaA in both SchemaB and SchemaD. It looks like SchemaB cannot resolve the reference to SchemaA.
Interestingly, when I create a new schema identical to SchemaA and reference that new schema in SchemaB instead, everything works correctly.  It seems that we might not be allowed to reference the same schema (SchemaA) from multiple other schemas that are ultimately used within a single base schema (SchemaE)

geniegeist:
This PR attempts to fix #3363. First reported here: https://redpandacommunity.slack.com/archives/C075M18MAEA/p1745906748439179?thread_ts=1745853266.775599&cid=C075M18MAEA

## Description

Given these Avro schemas
```
// foo
{
  "schema":"{
    \"type\":\"record\",
    \"name\":\"foo\",
    \"namespace\":\"example\",
    \"fields\":[{\"name\":\"A\",\"type\":\"string\"}]
  }"
}

// bar
{
  "schema":"{
    \"type\":\"record\",
    \"name\":\"bar\",
    \"namespace\":\"example\",
    \"fields\":[{\"name\":\"A\",\"type\":\"foo\"}]
  }"
}

// baz
{
  "schema":"{
    \"type\":\"record\",
    \"name\":\"baz\",
    \"namespace\":\"example\",
    \"fields\":[{\"name\":\"A\",\"type\":\"foo\"}]
  }"
}
```
the schema decoder fails with `Processor failed: unable to parse schema avro: unknown type: foo`.

## Issue

The walk references function does not invoke the callback function when a schema has already been visited causing issues upstream.

Let's say
```
Schema B references Schema A
Schema C references Schema A
```

Let's assume we invoke the walkReferences function in this order
```
walkReferences(A)
walkReferences(B)
walkReferences(C)
```
Since walkReferences is recursive, we have the following call stack.
```
walkReferences(A)
walkReferences(B)
    walkReferences(A)
walkReferences(C)
    walkReferences(A)
```

Note that the second and third time `walkReferences(A)` is called, the provided callback won't be invoked since schema A has already been visited.

So, this [supplier of the callbacks](https://github.com/redpanda-data/connect/blob/21f21aaa09a79f66f258edadbad2a6a1c6c27330/internal/impl/confluent/serde_hamba_avro.go#L34) processes the resolved schemas in the following (reversed) order
```
C
B
A
```

The correct order would be
```
A
C
B
``` 

Note that C depends on A; hence A should be processed first.

## Fix

~~This PR calls the provided callback even for already visited schemas.~~
This PR implements a proper topological sort.

### DIFF
--- a/internal/impl/confluent/sr/client_test.go
+++ b/internal/impl/confluent/sr/client_test.go
@@ -31,7 +31,7 @@ func mustJBytes(t testing.TB, obj any) []byte {
 }
 
 func TestWalkReferences(t *testing.T) {
-	tCtx, done := context.WithTimeout(context.Background(), time.Second*10)
+	tCtx, done := context.WithTimeout(t.Context(), time.Second*10)
 	defer done()
 
 	rootSchema := `[
@@ -117,61 +117,61 @@ func TestWalkReferences(t *testing.T) {
 				"schema": barSchema,
 			}), nil
 		case "/subjects/baz/versions/1", "/schemas/ids/4":
-			return mustJBytes(t, map[string]any {
-				"id": 4, 
-				"version": 1,
-				"schema": bazSchema,
+			return mustJBytes(t, map[string]any{
+				"id":         4,
+				"version":    1,
+				"schema":     bazSchema,
 				"schemaType": "AVRO",
 				"references": []any{
 					map[string]any{"name": "benthos.namespace.com.foo", "subject": "foo", "version": 1},
 				},
 			}), nil
-		}		
+		}
 		return nil, nil
 	})
 
 	tests := []struct {
-		name				string 
-		schemaId    int
-		output     	[]string
+		name     string
+		schemaId int
+		output   []string
 	}{
 		{
-			name: "root",
+			name:     "root",
 			schemaId: 1,
 			output: []string{
 				"benthos.namespace.com.foo",
 				"benthos.namespace.com.bar",
-  			"benthos.namespace.com.baz",
+				"benthos.namespace.com.baz",
 			},
 		},
 		{
-			name: "foo",
+			name:     "foo",
 			schemaId: 2,
-			output: []string{},
+			output:   []string{},
 		},
 		{
-			name: "baz",
+			name:     "baz",
 			schemaId: 4,
 			output: []string{
 				"benthos.namespace.com.foo",
 			},
 		},
 		{
-			name: "root2",
+			name:     "root2",
 			schemaId: 5,
 			output: []string{
 				"benthos.namespace.com.foo",
 				"benthos.namespace.com.baz",
-  			"benthos.namespace.com.bar",
+				"benthos.namespace.com.bar",
 			},
 		},
 		{
-			name: "root3",
+			name:     "root3",
 			schemaId: 6,
 			output: []string{
 				"benthos.namespace.com.bar",
 				"benthos.namespace.com.foo",
-  			"benthos.namespace.com.baz",
+				"benthos.namespace.com.baz",
 			},
 		},
 	}
@@ -188,7 +188,7 @@ func TestWalkReferences(t *testing.T) {
 			walkErr := client.WalkReferences(tCtx, schema.References, func(ctx context.Context, name string, schema franz_sr.Schema) error {
 				schemas = append(schemas, name)
 				return nil
-			});
+			})
 			require.NoError(t, walkErr)
 			require.Len(t, schemas, len(test.output))
 			for i, name := range schemas {
@@ -197,7 +197,6 @@ func TestWalkReferences(t *testing.T) {
 		})
 	}
 }
-
 
 func runSchemaRegistryServer(t testing.TB, fn func(path string) ([]byte, error)) string {
 	t.Helper()
@@ -222,4 +221,3 @@ func runSchemaRegistryServer(t testing.TB, fn func(path string) ([]byte, error))
 
 	return ts.URL
 }
-
--- a/internal/impl/confluent/sr/client_test.go
+++ b/internal/impl/confluent/sr/client_test.go
@@ -1,0 +1,225 @@
+package sr
+
+import (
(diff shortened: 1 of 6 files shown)


rockwotj:
This is good - we now do a full depth first search, which is required for topological sort.

rockwotj:
This only works with `preserve_logical_types: true` as the goavro schema walking doesn't preserve the order from the walk. Thats an easy fix tho

rockwotj:
This might break the json decoder tho, as now it can get duplicates

geniegeist:
Thanks for the comments. I'll try to address them tomorrow. If you or anyone from the Redpanda team wants to jump in and take over, feel free to go ahead :)

geniegeist:
> This might break the json decoder tho, as now it can get duplicates

@rockwotj The `WalkReferences()` function now traverses the references in proper topological order, hence no more duplicates.

geniegeist:
> This is good - we now do a full depth first search, which is required for topological sort.

Right, topological sort is exactly what we needed. Thanks for pointing that out. I’ve now implemented a proper topological sort with DFS.

geniegeist:
>This only works with preserve_logical_types: true as the goavro schema walking doesn't preserve the order from the walk. Thats an easy fix tho

I'm trying to fix this issue and other resolver issues with the goavro encoder in #3388.

internal/impl/confluent/sr/client_test.go
```diff
@@ -0,0 +1,223 @@
...
+	"encoding/json"
+	"io/fs"
+	"net/http"
+	"net/http/httptest"
+	"sync"
+	"testing"
+	"time"
+
+	franz_sr "github.com/twmb/franz-go/pkg/sr"
+	//"github.com/stretchr/testify/assert"
```
rockwotj, line 14: remove

rockwotj (commented):
One small comment to cleanup an import, otherwise LGTM. Thanks!

It would be great to have checks for this in protobuf and json. My quick look at protobuf this doesn't look to be an issue because the APIs are better. We can add follow up tests for other formats later.

As for when this can be released. I can't speak to when we will release next at the moment, but can ping you here when we have that set. Generally we don't go more than a couple weeks without a release.

internal/impl/confluent/sr/client_test.go
rockwotj: We can merge after we fix this up. If you could also add a changelog entry, that would be much appreciated (not required, I can also do this)

geniegeist: Done!

rockwotj (approved):

rockwotj:
@geniegeist the linter seems to be upset, once that is fixed I will merge

rockwotj:
thank you for fixing this!

PR #3381, opened 2025-04-30
kafka: add consumer lag metric to kafka_franz

rockwotj:


internal/impl/kafka/lag.go
```diff
@@ -0,0 +1,86 @@
...
+	adminClient := kadm.NewClient(client)
+	topicLagCache := &sync.Map{}
+	lagUpdater := asyncroutine.NewPeriodicWithContext(topicLagRefreshPeriod, func(ctx context.Context) {
+		ctx, done := context.WithTimeout(ctx, topicLagRefreshPeriod)
+		defer done()
+		lags, err := adminClient.Lag(ctx, consumerGroup)
+		if err != nil {
+			logger.Debugf("Failed to fetch group lags: %s", err)
+		}
+		lags.Each(func(gl kadm.DescribedGroupLag) {
```
mmatczuk, line 53: Consider calling in else branch..

internal/impl/kafka/lag.go
```diff
@@ -0,0 +1,86 @@
...
+// NewConsumerLag creates a new ConsumerLag instance.
+func NewConsumerLag(
+	client *kgo.Client,
+	consumerGroup string,
+	logger *service.Logger,
+	topicLagGauge *service.MetricGauge,
+	topicLagRefreshPeriod time.Duration,
+) *ConsumerLag {
+	adminClient := kadm.NewClient(client)
+	topicLagCache := &sync.Map{}
```
mmatczuk, line 45: How about new(sync.Map)

mihaitodor (approved):
LGTM :sheep: 🚀

internal/impl/kafka/lag.go
rockwotj: I tend to use this syntax - I'm not sure if there is a difference either way, but happy to switch

internal/impl/kafka/lag.go
rockwotj: used an early refer in the logger case

PR #3388, opened 2025-05-02
Draft: Goavro resolves references

geniegeist:
> [!NOTE]
> This is still in draft but the resolver algorithm is done. Before I proceed, I would like to hear feedback if this approach is feasible.

## Problem

The Goavro encoder is not able to resolve schemas with references such as

```
// foo
{
    "id":         1,
    "version":    1,
    "schema":    {
	"namespace": "benthos.namespace.com",
	"type": "record",
	"name": "foo",
	"fields": [
		{ "name": "Woof", "type": "bar"}
	]
    }, 
    "schemaType": "AVRO",
    "references": [{"name": "benthos.namespace.com.bar", "subject": "bar", "version": 1}]
}

// bar
{
     "id": 2, "version": 1, "schemaType": "AVRO",
     "schema": {
	"namespace": "benthos.namespace.com",
	"type": "record",
	"name": "bar",
	"fields": [
		{ "name": "Moo", "type": "string"}
	]
     },
}
```

It usually fails with
```
failed to parse root schema as enum: json: cannot unmarshal object into Go value of type
```

This PR attempts to fix this.

## Solution
Assume we want to encode payload `P` with schema `Q`.

### Solution 1: Inline all schemas
The obvious solution is to resolve all schemas and inline the required schemas in `Q`.

### Solution 2: `union` type solution
I propose an alternative (hacky) solution. Let me know your thoughts.

1. The Goavro encoder resolves all required schemas of `P` and collects them in an array `A`. 
2. We build a new Avro schema `S` of type `union` to hold all the required schemas in `A`.
3. We encode `P` using schema `S` obtaining encoding `B`.
4. Finally, we remove the first bytes of `B` that hold information about the union type. We need to remove these bytes because `P` is was encoded using a `union` type `S`. After removing these bytes, we obtain the encoding of `P` with schema `Q`.

## Limitations

There is still one limitation when we want to encode some object with a schema that is a union type such as `benthos.namespace.com.root`
```
// root
{
    "id":         1,
    "version":    1,
    "schema":    {
	"namespace": "benthos.namespace.com",
	"type": "record",
	"name": "root",
	"fields": ["benthos.namespace.com.bar"]
    }, 
    "schemaType": "AVRO",
    "references": [{"name": "benthos.namespace.com.bar", "subject": "bar", "version": 1}]
}
```
and where `benthos.namespace.com.root` references another schema `benthos.namespace.com.bar` which references another schema `benthos.namespace.com.foo`, e.g.
```
// bar
{
    "id":         2,
    "version":    1,
    "schema":    {
	"namespace": "benthos.namespace.com",
	"type": "record",
	"name": "bar",
	"fields": [{ "name": "Moo", "type": "foo"}]
    }, 
    "schemaType": "AVRO",
    "references": [{"name": "benthos.namespace.com.foo", "subject": "foo", "version": 1}]
}
```
I'm working on this by improving the [switch statement](https://github.com/redpanda-data/connect/blob/ada07c01539dad6ef10c1ef821522d9b05b42f75/internal/impl/confluent/serde_goavro.go#L85).

## State of this PR

Almost everything is done. I just need to fix the limitation above, and maybe write more tests.

### DIFF
--- a/internal/impl/confluent/processor_schema_registry_encode_test.go
+++ b/internal/impl/confluent/processor_schema_registry_encode_test.go
@@ -320,7 +320,7 @@ func TestSchemaRegistryEncodeAvroLogicalTypes(t *testing.T) {
 		{
 			name:        "message doesnt match schema codec",
 			input:       `{"int_time_millis":35245000,"long_time_micros":20192000000000,"long_timestamp_micros":null,"pos_0_33333333":"!"}`,
-			errContains: "cannot decode textual union: expected:",
+			errContains: "cannot decode textual union:",
 		},
 		{
 			name:        "message doesnt match schema",
--- a/internal/impl/confluent/serde_goavro.go
+++ b/internal/impl/confluent/serde_goavro.go
@@ -16,6 +16,7 @@ package confluent
 
 import (
 	"context"
+	"encoding/binary"
 	"encoding/json"
 	"fmt"
 
@@ -28,7 +29,7 @@ import (
 	"github.com/redpanda-data/connect/v4/internal/impl/confluent/sr"
 )
 
-func resolveGoAvroReferences(ctx context.Context, client *sr.Client, mapping *bloblang.Executor, schema franz_sr.Schema) (string, error) {
+func resolveGoAvroReferences(ctx context.Context, client *sr.Client, mapping *bloblang.Executor, schema franz_sr.Schema) ([]json.RawMessage, error) {
 	mapSchema := func(s franz_sr.Schema) (string, error) {
 		if mapping == nil {
 			return s.Schema, nil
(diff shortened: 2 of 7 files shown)


rockwotj:
To be perfectly honest, I think I would like to remove goavro. AFAIK the only difference now is the logical type behavior, and I can easily add post processing to make hamba act like goavro. I am inclined to do that other than continue to try hacking around issues like this in goavro.

geniegeist:
> To be perfectly honest, I think I would like to remove goavro. AFAIK the only difference now is the logical type behavior, and I can easily add post processing to make hamba act like goavro. I am inclined to do that other than continue to try hacking around issues like this in goavro.

Sounds good! I'll close this PR for now as we also don't need the encoder right now.

geniegeist:
Btw, I’m super interested in contributing here. If you’re not planning to implement an encoder with Hamba right away, I’d be happy to give it a shot myself.

rockwotj:
Would happily accept a hamba encoder PR. No current plans to work on this ourselves

PR #3389, opened 2025-05-02
Feature: add support to extras field on sentry_capture processor

peczenyj:
I may need to add extra/ unstructured information on `sentry_capture`

also, I add a mock transport constructor that performs the registration of  `AssertExpectations` on test cleanup. mockery creates a constructor with this same pattern, and it simplifies the code.

### DIFF
--- a/internal/impl/sentry/processor_capture.go
+++ b/internal/impl/sentry/processor_capture.go
@@ -51,6 +51,12 @@ func newCaptureProcessorConfig() *service.ConfigSpec {
 				Example(`root = {"order": {"product_id": "P93174", "quantity": 5}}`).
 				Example(`root = deleted()`),
 
+			service.NewBloblangField("extras").
+				Description("A mapping that must evaluate to an object. If this mapping produces a value, then it is set on a sentry event as extras.").
+				Optional().
+				Example(`root.foo = "bar"`).
+				Example(`root = this.without("password")`),
+
 			service.NewInterpolatedStringMapField("tags").
 				Optional().
 				Description("Sets key/value string tags on an event. Unlike context, these are indexed and searchable on Sentry but have length limitations."),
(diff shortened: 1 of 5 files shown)


internal/impl/sentry/processor_capture.go
```diff
@@ -51,6 +51,13 @@ func newCaptureProcessorConfig() *service.ConfigSpec {
 				Example(`root = {"order": {"product_id": "P93174", "quantity": 5}}`).
 				Example(`root = deleted()`),
 
+			service.NewBloblangField("extras").
+				Version("TBD").
+				Description("A mapping that must evaluate to an object-of-objects. If this mapping produces a value, then it is set on a sentry event as extras.").
+				Optional().
+				Example(`root.foo = "bar"`).
```
rockwotj, line 58: Is this misleading? The description states it must be an "object of objects" which this does not seem to be that

rockwotj: Seems like it just need to be an object from looking at the code (and probably copy/pasted from above). Can we update the description?

internal/impl/sentry/processor_capture.go
```diff
@@ -51,6 +51,13 @@ func newCaptureProcessorConfig() *service.ConfigSpec {
 				Example(`root = {"order": {"product_id": "P93174", "quantity": 5}}`).
 				Example(`root = deleted()`),
 
+			service.NewBloblangField("extras").
+				Version("TBD").
```
rockwotj, line 55: You can drop this, we've been not using this anymore and have different plans to keep this updated.

internal/impl/sentry/processor_capture.go
peczenyj: Fixed

internal/impl/sentry/processor_capture.go
peczenyj: Fixed

peczenyj:
I think I made a mistake and I cant regenerate the docs right now. I will ask @mihaitodor to help

mihaitodor (approved):
LGTM, thank you!

PR #3395, opened 2025-05-03
improve source code formatting with gofumpt and goimports

peczenyj:
using gofumpt and goimports to improve source code formatting

next step could be include gci


mmatczuk:
Hey @peczenyj, thanks a lot for your contribution! 🙌
Would you mind squashing the formatting changes into a single commit and adding a short explanation of why you think this new formatting is better than the current one?
Also, it’d be great if you could keep the application of the new formatting in a separate commit.

Thanks again!

peczenyj:
hey @mmatczuk 

I think the diff `internal/agent/agent_plugin.go` is a good example

https://github.com/redpanda-data/connect/pull/3395/files#diff-f89aca45bd015bff1f3d4e231532d8cdf017af7e1bbcc0ef19a967a38df7a374

about goimports

we have all imports that belongs to `github.com/redpanda-data/` group by project

it means not only `github.com/redpanda-data/connect` but also `github.com/redpanda-data/benthos` imports are close to each other and not mixed with other thirdparty imports

also, gofumpt do things like

```go
var x foo
var y bar
```
became

```go
var (
   x foo
   y bar
)
```

so gofumpt can merge variable and constant declarations, but also other things like fix octal notation, etc.

In my opinion, gofumpt helps to enhance what gofmt does.

peczenyj:
I don't know how can I split the existing commits except if I revert some commits and try again

mmatczuk:
Thanks for your contribution, @peczenyj! I’ve opened #3421 to implement the changes you suggested. Much appreciated! I'll be sending a follow up patch to the benthos repo.

PR #3401, opened 2025-05-07
mcp: use parameters as the input

rockwotj:
Instead of adding extra parameters as metadata, add the whole object as
a structured value if specified. If not specified, then do the value
string trick.


Jeffail (commented):
LGTM, can you make the same adjustment for outputs?

rockwotj:
Done

Jeffail (approved):
Awesome, thanks @rockwotj!

PR #3405, opened 2025-05-09
kafka: add redpanda cache

rockwotj:
This is useful for connects that need a cache (CDC for example), but we don't want to force people to add a new system. Those use cases the cache being a bit slow on reads is fine because it's only used for checkpointing.


mihaitodor (approved):
Nice one! :shipit:

PR #3407, opened 2025-05-12
schema_registry_decode: support debezium temporal type decoding

rockwotj:


mmatczuk:
Looks good to me

internal/impl/confluent/processor_schema_registry_decode.go
```diff
@@ -87,6 +87,57 @@ When raw_unions is set to true then the above union schema is decoded as the fol
...
+|timestamp
+|Timestamp with microsecond precision
+
+|io.debezium.time.NanoTimestamp
+|timestamp
+|Timestamp with nanosecond precision
+
+|io.debezium.time.ZonedTime
+|timestamp
+|Time with timezone (ISO-8601 format)
```
mihaitodor, line 116: Should there be a case for this in `translateKafkaConnectValue`? Also, it might be worth saying `ISO-8601 format with nanosecond precision` (same for `ZonedTimestamp` and also in the error message: `expected valid ISO formatted`)

mihaitodor (approved):
Looks good, just one small nit and then :sheep: 🚀

internal/impl/confluent/processor_schema_registry_decode.go
rockwotj: Good catch! This type doesn't exist :)

PR #3408, opened 2025-05-12
Change Redpanda Migrator license to Apache

mihaitodor:


internal/impl/kafka/enterprise/integration_test.go
```diff
@@ -200,11 +191,19 @@ max_message_bytes: 1MB
 	license.InjectTestService(conf.Resources())
 
 	gmgr := enterprise.NewGlobalRedpandaManager("foo")
-	require.NoError(t, gmgr.InitWithCustomDetails("meowcustom", logsTopicOverride, statusTopicOverride, &kafka.FranzConnectionDetails{
-		SeedBrokers: []string{brokerAddr},
-		MetaMaxAge:  time.Minute * 5,
-		ClientID:    "foobar",
-	}))
```
mihaitodor, line 207: This was an interesting footgun... We added some more fields which cannot be set to their default Go value.

internal/impl/kafka/enterprise/integration_test.go
```diff
@@ -40,14 +32,13 @@ import (
...
 	"github.com/redpanda-data/connect/v4/internal/impl/kafka"
 	"github.com/redpanda-data/connect/v4/internal/impl/kafka/enterprise"
+	"github.com/redpanda-data/connect/v4/internal/impl/kafka/redpandatest"
 	"github.com/redpanda-data/connect/v4/internal/license"
 	"github.com/redpanda-data/connect/v4/internal/protoconnect"
 	_ "github.com/redpanda-data/connect/v4/public/components/confluent"
 )
 
-func createKafkaTopic(ctx context.Context, address, id string, partitions int32) error {
-	topicName := fmt.Sprintf("topic-%v", id)
```
mihaitodor, line 49: I'm not sure why we had it like this, but it seems that the topic names are injected directly

rockwotj:
LGTM as long as the tests pass and this is just moving code around. There doesn't seem to be functional changes outside of the license check removals

rockwotj (approved):

PR #3411, opened 2025-05-13
Add `max_yield_batch_size` field to redpanda inputs

Jeffail:
This now sets a more sensible cap to the batches yielded by the redpanda input, so it's no longer directly tied to the poll limit. I believe this fixes the performance issues being seen in the wild as users were trying to flush outrageously large batches.


mmatczuk:
How about `max_batch_size_bytes`?

mihaitodor (approved):
Nice one! ❤️ 🐑 🚀

Jeffail:
> How about `max_batch_size_bytes`?

Good shout, I've renamed it `max_batch_yield_bytes`, so that it's more consistent with other bytes based fields, but also keeps `yield` in there, I feel it's helpful even if a bit vague because it distinguishes this field from the other batch fields.

PR #3420, opened 2025-05-16
Fix RP input reconnection logic

Jeffail:
Previously we were stalling after closing the client during broker outages. In reality we can back off our polling and continue using the client as it will recover on its own.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
-- Fixed an issue where the `kafka_franz`, `redpanda` and `redpanda_common` inputs could stall for an unreasonable length of time after losing connection to a broker. (@Jeffail)
+- Fixed an issue where the `kafka_franz`, `redpanda`, `redpanda_common`, `redpanda_migrator`, `redpanda_migrator_offsets` and `ockam_kafka` inputs could stall for an unreasonable length of time after losing connection to a broker. (@Jeffail)
 
 ## 4.55.0 - 2025-05-15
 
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ Changelog
 
 All notable changes to this project will be documented in this file.
 
+## 4.55.1 - 2025-05-16
+
+### Fixed
+
+- Fixed an issue where the `kafka_franz`, `redpanda` and `redpanda_common` inputs could stall for an unreasonable length of time after losing connection to a broker. (@Jeffail)
```
mihaitodor, line 10: Also `redpanda_migrator`, `redpanda_migrator_offsets` and `ockam_kafka`

internal/impl/kafka/franz_reader_unordered.go
```diff
@@ -560,9 +566,12 @@ func (f *FranzReaderUnordered) Connect(ctx context.Context) error {
 					}
 				}
 
-				if nonTemporalErr {
-					cl.Close()
-					return
+				if nonTemporalErr && fetches.Empty() {
+					select {
+					case <-time.After(connErrBackOff.NextBackOff()):
```
rockwotj, line 571: don't we need to reset the backoff at some point?

mihaitodor (approved):
Nice, thank you for figuring this out! 🐑 🚀

internal/impl/kafka/franz_reader_unordered.go
Jeffail: Whoops, you have good eye-meat, thanks

PR #3427, opened 2025-05-21
Use UTC timestamps when decoding Debezium temporal types

mihaitodor:
I'm not 100% sure if this is the right thing to do here, but the tests fail on my laptop without this change (cc @rockwotj)

### DIFF
--- a/internal/impl/confluent/serde_hamba_avro.go
+++ b/internal/impl/confluent/serde_hamba_avro.go
@@ -241,34 +241,34 @@ func (w *avroSchemaWalker) translateKafkaConnectValue(value any, schema avro.Pro
 		if err != nil {
 			return nil, fmt.Errorf("expected number for io.debezium.time.Date got: %T", value)
 		}
-		return time.UnixMilli(0).AddDate(0, 0, int(v)), nil
+		return time.UnixMilli(0).UTC().AddDate(0, 0, int(v)), nil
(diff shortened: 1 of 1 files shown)


internal/impl/confluent/serde_hamba_avro.go
```diff
@@ -241,38 +241,38 @@ func (w *avroSchemaWalker) translateKafkaConnectValue(value any, schema avro.Pro
...
-		return time.UnixMilli(0).Add(time.Duration(v) * time.Nanosecond), nil
+		return time.UnixMilli(0).UTC().Add(time.Duration(v) * time.Nanosecond), nil
 	case "io.debezium.time.ZonedTimestamp":
 		v := bloblang.ValueToString(value)
 		t, err := time.Parse(time.RFC3339Nano, v)
 		if err != nil {
 			return nil, fmt.Errorf("expected valid ISO formatted timestamp for io.debezium.time.ZonedTimestamp got: %q", v)
 		}
-		return t, nil
+		return t.UTC(), nil
```
rockwotj, line 275: This is the one I'm not quite sure about... Maybe it's fine to convert the zoned timestamp to UTC? I could see arguments for customers wanting to keep the configured timezone from the string

rockwotj: Maybe we should use `ParseInLocation` with the UTC timezone

Issue #3425, opened 2025-05-20
mongodb_cdc connector snapshot restarting if database has no changes


jordancastrodev:
The mongodb_cdc connector has an odd behaviour. If the database is not in use (no inserts, updates or deletes), after consuming all snapshot records the rpk process will wait 30 seconds, send an error message like `ERRO[2025-05-20T19:20:47-03:00] Failed to read message: error watching MongoDB change stream: timed out while checking out a connection from connection pool: context deadline exceeded; total connections: 1, maxPoolSize: 100, idle connections: 0, wait duration: 250ns @service=redpanda-connect label="" path=root.input` and restart consuming the snapshot from scratch. Also, no checkpoints are ever recorded to the checkpoint_cache in the snapshot phase as described in the [docs](https://docs.redpanda.com/redpanda-connect/components/inputs/mongodb_cdc/#snapshot-mode), only if a change operation actually happens. I'd expect the pipeline to just sit idle, not start from scratch.

Environment dump:

```yaml
# docker-compose.yml
services:
  mongo:
    container_name: mongo
    image: mongo:8.0.8
    command: ["--replSet", "test-platform-shard-0", "--bind_ip_all"]
    volumes:
      - "${MONGO_DATA_FOLDER:-./data/mongo}:/data/db"
    ports:
      - 27017:27017
    networks:
      - test-network
    healthcheck:
      test: >
        mongosh --eval "
          try {
            rs.status();
          } catch (e) {
            if (e.message.includes('no replset config')) {
              rs.initiate({
                _id: 'test-platform-shard-0',
                members: [{ _id: 0, host: 'localhost:27017' }]
              });
            } else {
              throw e;
            }
          }
        "
      interval: 10s
      timeout: 5s
      retries: 5
      start_period: 10s

  redpanda-broker:
    container_name: redpanda-broker
    image: docker.redpanda.com/redpandadata/redpanda:v25.1.4
    command:
      - redpanda
      - start
      - --kafka-addr internal://0.0.0.0:9092,external://0.0.0.0:19092
      - --advertise-kafka-addr internal://redpanda-broker:9092,external://localhost:19092
      - --pandaproxy-addr internal://0.0.0.0:8082,external://0.0.0.0:18082
      - --advertise-pandaproxy-addr internal://redpanda-broker:8082,external://localhost:18082
      - --schema-registry-addr internal://0.0.0.0:8081,external://0.0.0.0:18081
      - --rpc-addr redpanda-broker:33145
[...]

PR #3430, opened 2025-05-22
mongodb/cdc: make progress on low activity collections

rockwotj:
When having a database that recieves no writes, we need to snapshot before streaming so that we don't resnapshot on restart.

I can't reproduce the timeouts when there are no changes yet.

Fixes: #3425

Commits:
- **chore: bump mongodb version**
- **mongodb/cdc: checkpoint when completing snapshot**


rockwotj:
@Jeffail if you merge https://github.com/Jeffail/checkpoint/pull/4 then I can make this PR trivial

Issue #3454, opened 2025-06-01
High CVE-2025-1975 in ollama


moshiko-s:
https://github.com/advisories/GHSA-wrh5-cmwx-q2qr

rockwotj:
We're not effected FYI, we don't embed the ollama code

PR #3457, opened 2025-06-02
ollama: upgrade to latest version

rockwotj:
Fixes: https://github.com/redpanda-data/connect/issues/3454


internal/impl/ollama/base_processor.go
```diff
@@ -69,10 +68,6 @@ func commonFields() []*service.ConfigField {
 				Optional().
 				Advanced().
 				Description("Map the model into memory. This is only support on unix systems and allows loading only the necessary parts of the model as needed."),
-			service.NewBoolField(bopFieldUseMLock).
```
mihaitodor, line 72: Might be worth leaving this in place as `deprecated` to avoid breaking configs where people just set it blindly.

mihaitodor (approved):
LGTM, just one small nit. Feel free to 🐑 🚀

internal/impl/ollama/base_processor.go
rockwotj: If this processor was used more I would probably have done this...

PR #3458, opened 2025-06-03
Adjust `redpanda_migrator_offsets` input

mihaitodor:
This change switches the `redpanda_migrator_offsets` input implementation to a poll-based mechanism which queries the source broker consumer groups periodically and emits updates when changes are detected.

Also, a new test is added to assert behaviour for non-monotonically increasing timestamps. As expected, `ListOffsetsAfterMilli` will return the offset of the first record which has the specified timestamp (or newer), so, if record timestamps are not always increasing, consumer groups migration will lead to duplicate records for consumers. Assuming that timestamps do increase eventually, the number of duplicates is limited.

### DIFF
--- a/internal/impl/kafka/redpanda_migrator_offsets_input.go
+++ b/internal/impl/kafka/redpanda_migrator_offsets_input.go
@@ -20,12 +20,13 @@ import (
 	"fmt"
 	"regexp"
 	"slices"
+	"time"
 
 	"github.com/twmb/franz-go/pkg/kadm"
 	"github.com/twmb/franz-go/pkg/kgo"
-	"github.com/twmb/franz-go/pkg/kmsg"
 
 	"github.com/redpanda-data/benthos/v4/public/service"
+	"github.com/redpanda-data/connect/v4/internal/asyncroutine"
 )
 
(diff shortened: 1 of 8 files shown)


internal/impl/kafka/redpanda_migrator_offsets_input.go
```diff
@@ -270,45 +241,116 @@ func (rmoi *redpandaMigratorOffsetsInput) getTimestampForCommittedOffset(ctx con
...
-		// Delete the records that we skipped
-		batch = slices.Delete(batch, i, len(batch))
+	return nil
+}
 
-		if len(batch) == 0 {
-			_ = ack(ctx, nil) // TODO: Log this error?
-			continue
+func (rmoi *redpandaMigratorOffsetsInput) Read(ctx context.Context) (*service.Message, service.AckFunc, error) {
+	for {
```
Jeffail: looks like this`for` loop isn't necessary as all paths lead to `return`,

mihaitodor: Oops, good catch! Thank you!

PR #3466, opened 2025-06-06
feat: add extra metadata for pubsub input

### LINKED ISSUE #3465: input: gcp_pubsub doesnt have messageID in metadata
When reading from the gcp_pubsub input the plugin is not forwarding the messageID or orderingKey when/if available.

It would be good to add these in as headers for downstream systems to handle if they wish to

anicoll:
Add in message metadata for pubsub input where missing.
Updated docs.
Added unit tests to validate changes

resolves #3465


rockwotj (approved):
Thank you! Looks awesome 🎉

PR #3469, opened 2025-06-08
Schema registry decode forward options for protobuf

ZijunHui:
Adds 4 options for the processor schema_registry_decode that are propagated to the protojson library. This adds flexibility for the deserialisation of protobuf using schema registries.


internal/impl/confluent/processor_schema_registry_decode.go
```diff
@@ -237,6 +262,30 @@ func newSchemaRegistryDecoderFromConfig(conf *service.ParsedConfig, mgr *service
 			return nil, err
 		}
 	}
+	if conf.Contains("protobuf", "use_proto_names") {
```
rockwotj: If there are defaults you don't need contains checks

ZijunHui: Thanks for the hints. Removed the redundant checks

rockwotj (approved):

PR #3476, opened 2025-06-10
Ensure linter reports all issues

mihaitodor:


mmatczuk:
I think we should also remove `only-new-issues: true`

mihaitodor:
Good catch, I forgot that one...

rockwotj (approved):

PR #3480, opened 2025-06-11
add side_effect processor

rockwotj:


mmatczuk:
How about `tee` processor?

internal/impl/sideeffect/processor.go
```diff
@@ -0,0 +1,62 @@
...
+
+type sideEffectProcessor struct {
+	output *service.OwnedOutput
+}
+
+var _ service.BatchProcessor = (*sideEffectProcessor)(nil)
+
+// ProcessBatch implements service.BatchProcessor.
+func (s *sideEffectProcessor) ProcessBatch(ctx context.Context, batch service.MessageBatch) ([]service.MessageBatch, error) {
+	err := s.output.WriteBatch(ctx, batch)
```
mihaitodor, line 55: I wonder if this might interfere with the dispatch mechanism from the `redpanda` input & output.

mihaitodor (approved):
Neat! Thanks for prototyping this Tyler! Calling it `side_effect` might be a bit cryptic, since some processors have side-effects.

Guess `tee` would work as Michal suggested, but maybe `tap` might suit better. Here's an example: https://angular.love/tap-rxjs-reference

Also, I'd consider pushing this to Benthos instead.

internal/impl/sideeffect/processor.go
rockwotj: Good point let me remove the dispatch stuff from the messages

rockwotj:
Good points. I will move to benthos and maybe call it just output, tee is very unix-y and not what I would search for

Issue #3383, opened 2025-04-30
sftp input creating new connection per file


shivchelwa:
We are seeing an issue with sftp input. A new connection is created for each scanned file and the connection are not closed until the process is stopped. The sshd file descriptors are going up steadily on sftp server. 

Tried to reproduce this issue locally by setting up sftp server using jmcombs/sftp:alpine with docker-compose.
```
  sftp:
    image: jmcombs/sftp:alpine
    container_name: sftp
    ports:
      - "2222:22"
    command: [ "foo:pass:1001" ]
    volumes:
      - /Users/username/test/uploads:/home/foo/uploads 
```
Here are my test artifacts:

_redis.yaml_
```
cache_resources:
  - label: redis_sftp
    redis:
      url: redis://${REDIS_HOST:localhost}:${REDIS_PORT:6379}
      prefix: "benthos:sftp:file:"
```
_sftp_test.yaml_
```
input:
  sftp:
    address: ${SFTP_ADDR:"localhost:2222"}
    credentials:
      username: ${SFTP_USER:foo}
      password: ${SFTP_PASS:pass}
    paths: [ ${SFTP_UPLOAD_DIR:/uploads/files/*.txt} ]
    scanner:
      to_the_end: {}
    watcher:
      enabled: true
      minimum_age: ${FILE_MINIMUM_AGE:1m}
      poll_interval: ${FILE_POLL_INTERVAL:15s}
      cache: redis_sftp
pipeline:
  processors:
  - try:
    - log:
        message: processing file ${! @sftp_path }
    - mapping: root = deleted()
```

The logs from pipelines and sftp server indicates a connection being opened per file.

> INFO Running without a main config file            @service=benthos benthos_version=v4.48.0
> INFO Listening for HTTP requests at: http://0.0.0.0:4195/  @service=benthos
> INFO Launching Benthos in streams mode, use CTRL+C to close  @service=benthos
> INFO Output type stdout is now active              @service=benthos label="" path=root.output stream=sftp_conn_test
> INFO Input type sftp is now active                 @service=benthos label="" path=root.input stream=sftp_conn_test
> INFO processing file /uploads/files/file1.txt      @service=benthos custom_source=true label="" path=root.pipeline.processors.0.try.0 stream=sftp_conn_test
> INFO processing file /uploads/files/file2.txt      @service=benthos custom_source=true label="" path=root.pipeline.processors.0.try.0 stream=sftp_conn_test
[...]

PR #3498, opened 2025-06-17
Fix `sftp` input connection leak

mihaitodor:
Fixes https://github.com/redpanda-data/connect/issues/3383.

TODO:
- [ ] Fix integration tests: The updated input should not close the SSH connection before all the file reads have completed or were cancelled.


internal/impl/sftp/resources/docker-compose.yaml
mihaitodor, line 1: This doesn't work on ARM, so I decided to switch to `drakkan/sftpgo` for the integration tests.

internal/pool/pool.go
```diff
@@ -29,7 +29,7 @@ type (
 		// Acquire gets an object T out of the pool if available, otherwise will create a new item.
 		// The context can be used to abort waiting for an item from the queue, otherwise an error
 		// is only ever returned if creating the object in the pool fails.
-		Acquire(context.Context) (T, error)
+		Acquire(ctx context.Context) (T, error)
```
mmatczuk: This looks unrelated, I'm not sure if we need it.

internal/pool/pool.go
```diff
@@ -127,3 +130,19 @@ func (p *cappedImpl[T]) Reset() {
 		}
 	}
 }
+
+func (p *cappedImpl[T]) Range(f func(item T) bool) {
```
mmatczuk: * Would be nice if this took context
* Typical name is ForEach
* Could we have a test
* Could we move this change to separate commit

internal/impl/sftp/config.go
```diff
@@ -0,0 +1,171 @@
...
+	scFieldCredentialsPassword          = "password"
+	scFieldCredentialsHostPublicKey     = "host_public_key"
+	scFieldCredentialsHostPublicKeyFile = "host_public_key_file"
+	scFieldCredentialsPrivateKey        = "private_key"
+	scFieldCredentialsPrivateKeyFile    = "private_key_file"
+	scFieldCredentialsPrivateKeyPass    = "private_key_pass"
+)
+
+func credentialsFields() []*service.ConfigField {
+	return []*service.ConfigField{
```
mmatczuk, line 44: I wonder what is our approach to marking fields as optional?

internal/impl/sftp/config.go
```diff
@@ -0,0 +1,171 @@
...
+
+	var privateKey []byte
+	if privateKeyStr != "" {
+		privateKey = []byte(privateKeyStr)
+	} else if privateKeyFile != "" {
+		privateKey, err = service.ReadFile(mgr.FS(), privateKeyFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read private key file: %s", err)
+		}
+	}
```
mmatczuk: Consider extracting to method, and adding test.

internal/impl/sftp/config.go
```diff
@@ -0,0 +1,171 @@
...
+
+	var hostPubKey []byte
+	if hostPublicKeyStr != "" {
+		hostPubKey = []byte(hostPublicKeyStr)
+	} else if hostPublicKeyFile != "" {
+		hostPubKey, err = service.ReadFile(mgr.FS(), hostPublicKeyFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read host public key file: %s", err)
+		}
+	}
```
mmatczuk: Consider extracting to method, and adding test.

internal/impl/sftp/config.go
```diff
@@ -0,0 +1,171 @@
...
+		var u *user.User
+		if u, err = user.Current(); err == nil {
+			keyCallback, err = knownhosts.New(filepath.Join(u.HomeDir, ".ssh", "known_hosts"))
+		} else {
+			keyCallback, err = knownhosts.New("/etc/ssh/known_hosts")
+		}
+		if err != nil {
+			return nil, fmt.Errorf("failed to read known_hosts file: %s", err)
+		}
+	}
```
mmatczuk: It feels like this could be move to hostPubKey handling

internal/impl/sftp/input.go
```diff
@@ -134,14 +153,23 @@ func newSFTPReaderFromParsed(conf *service.ParsedConfig, mgr *service.Resources)
...
 	}
+	if conf.Contains(siFieldConnectionTimeout) {
+		if s.sshConfig.Timeout, err = conf.FieldDuration(siFieldConnectionTimeout); err != nil {
+			return
+		}
+	}
+	var maxSFTPSessions int
+	if maxSFTPSessions, err = conf.FieldInt(siFieldMaxSFTPSessions); err != nil {
+		return nil, err
+	}
```
mmatczuk: How about moving closer to `s.sftpClientPool = pool`

internal/impl/sftp/input.go
```diff
@@ -167,31 +195,48 @@ func newSFTPReaderFromParsed(conf *service.ParsedConfig, mgr *service.Resources)
...
+		}
+
+		return client, nil
+	})
+
 	return
 }
 
-func (s *sftpReader) Connect(context.Context) error {
+func (s *sftpReader) Connect(ctx context.Context) error {
```
mmatczuk, line 204: Is this caused by some linter?

internal/impl/sftp/input.go
```diff
@@ -167,31 +195,48 @@ func newSFTPReaderFromParsed(conf *service.ParsedConfig, mgr *service.Resources)
...
 	s.stateLock.Lock()
 	defer s.stateLock.Unlock()
 
-	client, err := newClientPool(func() (*sftp.Client, error) {
-		return s.creds.GetClient(s.mgr.FS(), s.address)
-	})
+	if s.sshClient != nil {
+		return nil
```
mmatczuk, line 210: Worth logging that I think

internal/impl/sftp/input.go
```diff
@@ -167,31 +195,48 @@ func newSFTPReaderFromParsed(conf *service.ParsedConfig, mgr *service.Resources)
...
+	}
+
+	var err error
+	s.sshClient, err = ssh.Dial("tcp", s.address, s.sshConfig)
 	if err != nil {
-		if errors.Is(err, sftp.ErrSSHFxConnectionLost) {
-			err = service.ErrNotConnected
-		}
-		return err
+		return fmt.Errorf("failed to connect to SFTP server: %s", err)
```
mmatczuk: %w

internal/impl/sftp/input.go
```diff
@@ -167,31 +195,48 @@ func newSFTPReaderFromParsed(conf *service.ParsedConfig, mgr *service.Resources)
...
 	}
 
-	s.client = client
 	if s.pathProvider == nil {
```
mmatczuk: `s.pathProvider` is set only once in L230, this means that, probably, we'd never reconnect.
Am I missing something @mihaitodor ?

internal/pool/pool.go
```diff
@@ -42,6 +42,9 @@ type (
 		Cap() int
 		// Reset deletes all items currently in the pool and resets the allocated count.
 		Reset()
+		// Range calls f sequentially for each key and value present in the map.
+		// If f returns false, range stops the iteration.
+		Range(f func(item T) bool)
```
mmatczuk: Perhaps it would be more natural to add a function that get's called when item is removed.

internal/impl/sftp/output.go
```diff
@@ -57,7 +70,13 @@ func sftpOutputSpec() *service.ConfigSpec {
...
 				Default("all-bytes"),
 			service.NewObjectField(soFieldCredentials, credentialsFields()...).
-				Description("The credentials to use to log into the target server."),
+				Description("The credentials to use to log into the target server.").
+				LintRule(`
+root = match {
+  this.exists("host_public_key") && this.exists("host_public_key_file") => "both host_public_key and host_public_key_file can't be set simultaneously"
+  this.exists("private_key") && this.exists("private_key_file") => "both private_key and private_key_file can't be set simultaneously"
+}`,
+				),
```
mmatczuk: Should be also applied to input and tested.

internal/pool/pool.go
mihaitodor: It's just for consistency with `TryAcquireExisting() (item T, ok bool)`. I'll move it in a separate commit (or remove it if that's easier).

internal/pool/pool.go
mihaitodor: I followed https://pkg.go.dev/sync#Map.Range. Would you mind sharing how that would work? I'm looking for a mechanism to clear all the remaining items in the pool when shutting down so we don't end up with dangling SFTP sessions when tearing down the SSH connection.

internal/pool/pool.go
mihaitodor: Indeed, I followed https://pkg.go.dev/sync#Map.Range. I like ForEach more and adding a context sounds good, thank you for the suggestion!

internal/impl/sftp/input.go
mihaitodor: It's because now I'm doing `s.sftpClientPool.Acquire(ctx)` in this method.

internal/impl/sftp/input.go
mihaitodor: The errors are a bit messy in this code. Normally I default to `%s` unless I want to bubble something specific higher up (it's less common). In Connect, we just need to return some error, so no wrapping needed.

internal/impl/sftp/input.go
mihaitodor: Ouch, nice catch, that deserves a bit more attention... It's fine in watcher mode, since it uses `s.sftpClientPool` to discover new files, but it's not fine for the `staticPathProvider`.

I removed that `getFilePathProvider` utility and moved the logic in Connect. Please have another look.

internal/impl/sftp/config.go
mihaitodor: `.Optional()` instead of `.Default("")`. They kind of work in a similar way, but if you use `.Optional()` and the field is omitted from the config completely, we can detect that via `conf.Contains` or `.exists(...)` in Bloblang.

internal/pool/pool.go
mihaitodor: Ugh, the code I pushed before was very silly / broken and Range isn't actually the right name for this method since it also acquires the items when iterating on the channel...

Since some items from the pool might be acquired by some other goroutine, my naive first attempt was to only iterate on the available ones, which meant potentially not closing all of them... The updated implementation implies waiting until we iterate through `pool.Size()` items, so basically waiting until all the goroutines which acquired items release them. However, this leads to deadlocks since there's no signalling mechanism to tell the other goroutines to release their SFTP sessions. I can maybe use a `shutdown.Signaller` for this purpose, but I worry the code will end up too tangled...

For now, I added a `context.WithTimeout` for calling `ForEach` to get the tests to pass, but that means we could get some dangling sessions released back into the pool after `Close` completes. Dunno if this is a big issue...

Or maybe I should just not clear the pool at all and let the SFTP sessions dangle after closing the SSH connection. Need to check the sftp code to see if that leaks memory... That way, I don't need to add this `ForEach` utility at all. Can Connect be called again after Close finishes?

@Jeffail WDYT?

internal/impl/sftp/output.go
mihaitodor: Sorted

internal/pool/pool.go
Jeffail: Connect should not be called ever if a Close has been called, if that ever came up then I would fix the engine rather than trying to accommodate that behaviour within sftp. As long as we don't have memory leaks then I'm happy to let them dangle, LEAVE THE SESSIONS BE.

mihaitodor: OK, I did some digging and I think it's safe to let the GC tear down sessions when reading data, since the SFTP client Close method just calls https://github.com/pkg/sftp/blob/36ce9cfd5c257416ed074f58de17e3e6c2149857/conn.go#L36-L40. I'm a bit unsure about writes, though. That code does `c.WriteCloser.Close()`, so I wonder if it might not flush all data unless the session is closed explicitly. If that's the case, we'll need to add some extra logic for the output... WDYT?

mihaitodor:
Looking at the output code in more detail, it caches `handle` and `handlePath` on the `sftpWriter` when using append mode. However, the `path` field supports interpolation. We'd need a cache of handles instead, no? (cc @Jeffail)

Jeffail:
> Looking at the output code in more detail, it caches `handle` and `handlePath` on the `sftpWriter` when using append mode. However, the `path` field supports interpolation. We'd need a cache of handles instead, no? (cc @Jeffail)

So this is a bit of an awkward one, ideally yes we'd have a cache of filenames to handles that we'd keep open and you'd be able to have multiple files being appended to at any given time. However, this would put us in a situation where it's not clear when a file has been "finished", so we'd need to have some sort of timeout whereby we close the file. Instead, we currently assume the path will be consistent throughout the entirety of a file being written so, this suits use cases such as timed files and it gives us an explicit signal to close the current handle (the path changes).

mihaitodor:
@Jeffail Then I think the output doesn't even need a sftp session pool since it'll only write to one file at a time (no matter if it's in append mode or not). I was thinking we should allow parallel writes somehow... I pushed the updated output code.

internal/impl/sftp/input.go
mmatczuk: Maybe, but this also prevents us from building common helper functions that to help with errors processing and may be misleading if you wan to check some os errors.

internal/impl/sftp/output.go
```diff
@@ -182,52 +197,68 @@ func (s *sftpWriter) Write(_ context.Context, msg *service.Message) error {
 		flag |= os.O_TRUNC
 	}
 
-	if err := s.client.MkdirAll(filepath.Dir(path)); err != nil {
-		if errors.Is(err, sftp.ErrSshFxConnectionLost) {
-			return service.ErrNotConnected
-		}
-		return err
+	s.sftpClient, err = sftp.NewClient(s.sshClient)
```
mmatczuk, line 206: If we create SFTP client for every write maybe it should be a local variable not a field.

internal/impl/sftp/output.go
```diff
@@ -150,11 +156,18 @@ func (s *sftpWriter) writeTo(wtr io.Writer, p *service.Message) error {
 	return nil
 }
 
-func (s *sftpWriter) Write(_ context.Context, msg *service.Message) error {
+func (s *sftpWriter) Write(_ context.Context, msg *service.Message) (err error) {
```
mmatczuk: How about ferr or werr to avoid conflicts with internal error handling

internal/impl/sftp/output.go
mihaitodor: We create a SFTP session (client) for each opened file. Otherwise it would be very wasteful if you have a static `path` or if the `path` doesn't change for every message (which would be common for this output). See lines 179-191.

internal/impl/sftp/input.go
mihaitodor: OK, I don't mind changing it (see updated code), but I'm not sure I agree with optimising error handling for future refactors, since one should go through the code in detail when doing so. Otherwise, we should add tests to cover the expected behaviour. Or maybe some custom linter?

internal/impl/sftp/output.go
mmatczuk: I see. It would be worth a comment on how it works.

mihaitodor: Sorted!

mmatczuk: Thanks.

mmatczuk (approved):

PR #3499, opened 2025-06-18
fix cohere rerank processor output format

birdayz:
Fix ordering for cohere reranker, it just appends and does not honor the ordering, mixing up results basically.
Change output from array of strings + metadata to array of objects with document and relevance_score fields.

### DIFF
--- a/internal/impl/cohere/rerank_processor_test.go
+++ b/internal/impl/cohere/rerank_processor_test.go
@@ -1,0 +1,336 @@
+// Copyright 2024 Redpanda Data, Inc.
+//
+// Licensed as a Redpanda Enterprise file under the Redpanda Community
+// License (the "License"); you may not use this file except in compliance with
+// the License. You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


internal/impl/cohere/rerank_processor_test.go
```diff
@@ -0,0 +1,344 @@
...
+			require.NoError(t, msgs[0].GetError())
+		})
+	}
+}
+
+func TestCohereRerankProcessorIntegration(t *testing.T) {
+	integration.CheckSkip(t)
+
+	// Read API key from /tmp/cohere
+	apiKeyBytes, err := os.ReadFile("/tmp/cohere")
```
rockwotj: We generally use environment variables instead.

rockwotj (commented):
Plz fix linter and add a changelog entry about the output change.

internal/impl/cohere/rerank_processor_test.go
birdayz: updated, this is ofc nonsense

rockwotj (approved):

PR #3506, opened 2025-06-26
Fix `redpanda_migrator_offsets` input

mihaitodor:
Fetch record timestamps in parallel and discard consumer groups which point to truncated records.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file.
 
 - The `sftp` input no longer creates new SSH connections for each file it reads. (@mihaitodor, @TColl)
 - Fixed a bug with the `redpanda_migrator_offsets` output where it was attempting to rewind consumer groups if it got restarted after consumers were migrated to the destination cluster. (@mihaitodor)
+- The `redpanda_migrator_offsets` input now fetches record timestamps in parallel and discards consumer groups which point to truncated records. (@mihaitodor)
 
 ## 4.58.2 - 2025-06-17
 
--- a/internal/impl/kafka/input_redpanda_migrator_offsets.go
+++ b/internal/impl/kafka/input_redpanda_migrator_offsets.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"regexp"
 	"slices"
 	"time"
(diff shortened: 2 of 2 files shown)


internal/impl/kafka/integration_test.go
```diff
@@ -1292,8 +1292,8 @@ func TestRedpandaMigratorIntegration(t *testing.T) {
 				}))
 
 				validate(string(b), meta)
-			case <-time.After(5 * time.Second):
-				t.Error("timed out waiting for migrator transfer")
+			case <-time.After(20 * time.Second):
```
mihaitodor, line 1295: We're a bit slower to propagate consumer group updates now...

go.mod
```diff
@@ -137,9 +137,10 @@ require (
 	github.com/timeplus-io/proton-go-driver/v2 v2.0.17
 	github.com/tmc/langchaingo v0.1.13
 	github.com/trinodb/trino-go-client v0.315.0
-	github.com/twmb/franz-go v1.18.0
-	github.com/twmb/franz-go/pkg/kadm v1.13.0
-	github.com/twmb/franz-go/pkg/kmsg v1.9.0
+	github.com/twmb/franz-go v1.19.1
+	github.com/twmb/franz-go/pkg/kadm v1.15.0
+	github.com/twmb/franz-go/pkg/kfake v0.0.0-20250625174842-669b18eeee83
```
mihaitodor, line 142: This could use a tag...

internal/impl/kafka/input_redpanda_migrator_offsets_test.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	partition, ok := msg.MetaGet("kafka_offset_partition")
+	assert.True(t, ok)
+	assert.Equal(t, "0", partition)
+	assertCGUpdate(t, msg, dummyTopic, dummyGroup, dummyMetadata, 2, false)
+}
+
+/*
+// TODO: Fix this test...
```
mmatczuk: What about this test?

mihaitodor: Sorted! Turns out I had to `kgo.PurgeTopicsFromClient("test_topic")` to prevent the underlying client from reading stale data after the call to `kadm.DeleteRecords()`. It's a bit odd, since the `streamBuilder` creates a stream from scratch which initialises a new franz-go client, so I wasn't expecting it to have access to cached data... Maybe something to fix in franz-go.

mmatczuk:
In MM2 we have

```
                // sync offset to the target cluster only if the state of current consumer group is:
                // (1) idle: because the consumer at target is not actively consuming the mirrored topic
                // (2) dead: the new consumer that is recently created at source and never existed at target
                //           This case will be reported as a GroupIdNotFoundException
```

https://github.com/apache/kafka/blob/56a3c6dde929763aaf74a801bd043fdd474a8ed2/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorCheckpointTask.java#L301C17-L304C86

This code keeps updating CG all the time unless there are readers in the destination.
Don't you think that not migrating CG would be more secure and efficient?

mihaitodor:
> This code keeps updating CG all the time unless there are readers in the destination.
> Don't you think that not migrating CG would be more secure and efficient?

I have a cache to prevent emitting updates which have been sent already: https://github.com/redpanda-data/connect/blob/b64827e750c68237415c71a36d6b732df9d6abe8/internal/impl/kafka/input_redpanda_migrator_offsets.go#L425-L428. This check in the output should prevent writing identical updates if we restart Migrator and the cache is empty: https://github.com/redpanda-data/connect/blob/b64827e750c68237415c71a36d6b732df9d6abe8/internal/impl/kafka/output_redpanda_migrator_offsets.go#L332-L338

Not sure what you mean about the secure part.

mmatczuk:
> Not sure what you mean about the secure part.

If we explicitly only migrate the CG that are eligible it's less likely to mess things up :)

mihaitodor:
@mmatczuk We don’t have any means to query only a subset of CGs, so I used the list of input topics to filter out CGs which aren’t associated with those topics. We could also ask the users to provide an explicit list of CGs, but that could be a very long list…

mmatczuk:
This is fine, actually MM2 [requires users to provide CGs](https://github.com/apache/kafka/blob/27383970b629775bf6368bf10e58122b3b709522/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorCheckpointTaskConfig.java#L35).

It then asks for CG details and filters based on that.

mihaitodor:
I think that's optional if I read that code right. Happy to expose it as an extra config param similar to `topics` and, if left empty, we return all the groups that match the topics. WDYT?

PR #3511, opened 2025-06-27
fix(mysql_cdc): fix snapshotting AWS RDS managed MySQL

mmatczuk:


internal/impl/mysql/snapshot.go
```diff
@@ -68,78 +72,90 @@ func (s *Snapshot) prepareSnapshot(ctx context.Context) (*position, error) {
...
-		return nil, fmt.Errorf("failed to get binlog position: %v", err)
+		return nil, errors.Join(
+			fmt.Errorf("get binlog position: %w", err),
+			unlockTables(),
+			s.tx.Rollback())
 	}
 
-	// Release the global read lock immediately after getting the binlog position
+	// Release the table locks immediately after getting the binlog position
 	if _, err := s.lockConn.ExecContext(ctx, "UNLOCK TABLES"); err != nil {
```
rockwotj, line 125: We don't have to specify the table names here is that right?

mmatczuk: Correct, the syntax is `UNLOCK {TABLE | TABLES}`. See https://dev.mysql.com/doc/refman/8.4/en/lock-tables.html and https://dev.mysql.com/doc/refman/8.4/en/flush.html#flush-tables-with-read-lock-with-list

PR #3519, opened 2025-07-01
GCP Cloud Storage output now interpolates the collision mode

Jeffail:
This adds interpolation to the collision mode so that dynamic modes can work without needing multiple instances of an output.


mihaitodor (approved):
LGTM 🐑 🚀

PR #3520, opened 2025-07-01
cohere_rerank: add index field

birdayz:
this makes it possible to do a mapping afterwards, and combine fields from this processor's output with the original input (e.g. other fields found in the vector DB).


Jeffail (approved):
LGTM, thanks @birdayz!

PR #3522, opened 2025-07-02
cohere rerank: make it possible to interpolate top_n

birdayz:
This allows us to take a value from http (gateway) input, and use it as value for cohere_rerank.top_n.


mmatczuk (approved):
Quite cool.

PR #3529, opened 2025-07-08
Back off unrecoverable conn errors

Jeffail:
Currently we silently retry kafka broker connections indefinitely even when the error is something unrecoverable, such as a SASL credentials issue.

Under these circumstances the connections remained open and our attempts were being made so frequently that it was causing issues at the broker level.

As a fix this commit attempts to detect unrecoverable connection errors and, when detected, forces a client closure as well as a full minute of back off before another attempt is made.

### DIFF
--- a/internal/impl/kafka/franz_client.go
+++ b/internal/impl/kafka/franz_client.go
@@ -17,13 +17,9 @@ package kafka
 import (
 	"context"
 	"crypto/tls"
-	"errors"
-	"net"
 	"strings"
-	"sync"
 	"time"
 
-	"github.com/twmb/franz-go/pkg/kerr"
 	"github.com/twmb/franz-go/pkg/kgo"
 	"github.com/twmb/franz-go/pkg/sasl"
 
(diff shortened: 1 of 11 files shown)


internal/impl/kafka/franz_client.go
```diff
@@ -151,3 +156,92 @@ func FranzConnectionOptsFromConfig(conf *service.ParsedConfig, log *service.Logg
...
+		return
+	}
+
+	var ke *kerr.Error
+	if !errors.As(err, &ke) {
+		return
+	}
+
+	if !ke.Retriable {
+		k.unrecoverableErr = err
```
twmb: You may want to use `kgo.IsRetryableBrokerErr`, which returns true for some other common problems (notably, temporary connection errors) and false for things that are truly not retryable.

internal/impl/kafka/franz_client.go
```diff
@@ -151,3 +156,92 @@ func FranzConnectionOptsFromConfig(conf *service.ParsedConfig, log *service.Logg
...
+		if unrecovErr != nil {
+			return unrecovErr
+		}
+
+		select {
+		case <-time.After(time.Millisecond * 10):
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+	}
```
twmb: You could switch to using a sync.Cond in `kgoConnectionChecker`, and broadcast is after `unrecoverableBrokerErr` is set in OnBrokerConnect.

internal/impl/kafka/franz_client.go
```diff
@@ -151,3 +156,92 @@ func FranzConnectionOptsFromConfig(conf *service.ParsedConfig, log *service.Logg
...
+
+// PingAndWait waits until either a successful connection has been established
+// or an unrecoverable error has been encountered.
+func (k *kgoConnectionChecker) PingAndWait(ctx context.Context, client *kgo.Client) error {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	var pingErr error
+	go func() {
+		err := client.Ping(ctx)
```
twmb: `Ping` itself issues a Metadata request, which triggers the full SASL flow. Since this is detecting valid SASL only at client initialization, you can probably remove the whole OnBrokerConnect chunk of code and rely entirely on Ping.

OnBrokerConnect would be needed if you want to handle SASL changes _after_ a client is initialized, to kill an initialized and in-use client. Since that's not happening here, well, probably no need for the complication.

twmb: (only in franz-go 1.19+)

Jeffail: Ahhhhh nice, that makes it way simpler, will update and retest.

mihaitodor (approved):
Nice one, LGTM! 🐑 🚀

PR #3538, opened 2025-07-13
Adding support for consumer audience for serverless

chapmanc:
This change enables gcp serverless to use workload identity federation to authenticate against another GCP workload federation identity using oidc. https://github.com/redpanda-data/common-go/pull/69


chapmanc:
Only to be merged after this is merged https://github.com/redpanda-data/common-go/pull/69 and we have updated from a branch reference to a fully qualified release.

PR #3539, opened 2025-07-14
Trigger docs update on tags

JakeSCahill:
Our workflow wasn't getting triggered on releases. I checked the other workflows and they rely on tags.


mmatczuk (approved):
LGTM

PR #3540, opened 2025-07-14
Auto schemas from avro to parquet

Jeffail:
This PR uses the new common schema package from benthos in order to create a reusable method for creating parquet files with dynamic schemas extracted during avro decoding.

### DIFF
--- a/docs/modules/components/pages/processors/parquet_encode.adoc
+++ b/docs/modules/components/pages/processors/parquet_encode.adoc
@@ -38,7 +38,8 @@ Common::
 # Common config fields, showing default values
 label: ""
 parquet_encode:
-  schema: [] # No default (required)
+  schema: [] # No default (optional)
+  schema_metadata: ""
   default_compression: uncompressed
 ```
 
@@ -51,7 +52,8 @@ Advanced::
 # All config fields, showing default values
 label: ""
 parquet_encode:
-  schema: [] # No default (required)
+  schema: [] # No default (optional)
+  schema_metadata: ""
   default_compression: uncompressed
   default_encoding: DELTA_LENGTH_BYTE_ARRAY
 ```
@@ -171,6 +173,15 @@ fields:
   - name: bar
     type: BYTE_ARRAY
 ```
+
+=== `schema_metadata`
+
+Optionally specify a metadata field containing a schema definition to use for encoding instead of a statically defined schema. For batches of messages the first messages schema will be applied for all subsequent messages of the batch.
+
+
+*Type*: `string`
+
+*Default*: `""`
 
 === `default_compression`
 
--- a/docs/modules/components/pages/processors/parquet_encode.adoc
+++ b/docs/modules/components/pages/processors/parquet_encode.adoc
@@ -176,7 +176,7 @@ fields:
 
 === `schema_metadata`
 
-Optionally specify a metadata field containing a schema definition to use for encoding instead of a statically defined schema. For batches of messages the first messages schema will be applied for all subsequent messages of the batch.
+Optionally specify a metadata field containing a schema definition to use for encoding instead of a statically defined schema. For batches of messages, the first message's schema will be applied to all subsequent messages of the batch.
 
 
 *Type*: `string`
--- a/internal/impl/confluent/serde_goavro.go
+++ b/internal/impl/confluent/serde_goavro.go
@@ -17,14 +17,12 @@ package confluent
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 
 	"github.com/linkedin/goavro/v2"
 	franz_sr "github.com/twmb/franz-go/pkg/sr"
 
 	"github.com/redpanda-data/benthos/v4/public/bloblang"
-	"github.com/redpanda-data/benthos/v4/public/schema"
 	"github.com/redpanda-data/benthos/v4/public/service"
 
 	"github.com/redpanda-data/connect/v4/internal/impl/confluent/sr"
@@ -123,20 +121,6 @@ func (s *schemaRegistryEncoder) getAvroEncoder(ctx context.Context, schema franz
 	}, nil
 }
 
-func tryExtractCommonSchemaFromAvroBytes(rawUnion bool, specBytes []byte) (any, error) {
-	var as any
-	if err := json.Unmarshal(specBytes, &as); err != nil {
-		return nil, err
-	}
-
-	commonSchema, err := avroAnyToCommonSchema(rawUnion, as)
-	if err != nil {
-		return nil, err
-	}
-
-	return commonSchema.ToAny(), nil
-}
-
 func (s *schemaRegistryDecoder) getGoAvroDecoder(ctx context.Context, aschema franz_sr.Schema) (schemaDecoder, error) {
 	schemaSpec, err := resolveGoAvroReferences(ctx, s.client, s.cfg.avro.mapping, aschema)
 	if err != nil {
@@ -153,9 +137,11 @@ func (s *schemaRegistryDecoder) getGoAvroDecoder(ctx context.Context, aschema fr
 		return nil, err
 	}
 
-	var commonSchemaAny any
+	var commonSchema any
 	if s.cfg.avro.storeSchemaMeta != "" {
-		if commonSchemaAny, err = tryExtractCommonSchemaFromAvroBytes(s.cfg.avro.rawUnions, []byte(schemaSpec)); err != nil {
+		if commonSchema, err = ecsAvroFromBytes(ecsAvroConfig{
+			rawUnion: s.cfg.avro.rawUnions,
+		}, []byte(schemaSpec)); err != nil {
 			s.logger.With("error", err).Error("Failed to extract common schema for meta storage")
 		}
 	}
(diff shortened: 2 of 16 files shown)


internal/impl/confluent/serde_goavro.go
```diff
@@ -154,8 +177,175 @@ func (s *schemaRegistryDecoder) getGoAvroDecoder(ctx context.Context, schema fra
...
+					}
+					c.Children = append(c.Children, tmpC)
+				}
+			}
+		}
+	case string:
+		c.Type = avroTypeToCommonType(t)
+	case map[string]any:
+		// This is so ridiculous, I can't believe they've allowed the type field
+		// to be a union of three different types SMDH.
```
rockwotj: I agree it's a mess. BTW hamba abstracts over this quite nicely if you wanted to just traverse their parsed representation. We do this in the hamba processors to normalize some fields.

internal/impl/confluent/serde_goavro.go
```diff
@@ -154,8 +177,175 @@ func (s *schemaRegistryDecoder) getGoAvroDecoder(ctx context.Context, schema fra
...
+				case string:
+					c.Children = append(c.Children, &schema.Common{
+						Type: avroTypeToCommonType(ut),
+					})
+				case map[string]any:
+					tmpC, err := avroAnyMapToCommonSchema(ut)
+					if err != nil {
+						return nil, fmt.Errorf("union `%v` child '%v': %w", c.Name, i, err)
+					}
+					c.Children = append(c.Children, tmpC)
```
rockwotj: Unions are wierd in avro, we probably need to respect if the flags are saving the names or not. IE there is a flag `raw_unions` or something that toggles unions from:

```
{"my_union":{"string":"foo"}}
```

vs

```
{"my_union":"foo"}
```

For the schema where `my_union` is `["string", "int", "null"]`

Jeffail: Arghhhh, good point. Got an implementation that should cover it, there's probably some edge cases in there that we might need to follow up on.

internal/impl/parquet/processor_encode.go
```diff
@@ -30,19 +32,24 @@ func parquetEncodeProcessorConfig() *service.ConfigSpec {
 		// Stable(). TODO
 		Categories("Parsing").
 		Summary("Encodes https://parquet.apache.org/docs/[Parquet files^] from a batch of structured messages.").
-		Field(parquetSchemaConfig()).
-		Field(service.NewStringEnumField("default_compression",
-			"uncompressed", "snappy", "gzip", "brotli", "zstd", "lz4raw",
+		Fields(
+			parquetSchemaConfig().Optional(),
+			service.NewStringField("schema_metadata").
+				Description("Optionally specify a metadata field containing a schema definition to use for encoding instead of a statically defined schema. For batches of messages the first messages schema will be applied for all subsequent messages of the batch.").
```
josephwoodward:
```suggestion
				Description("Optionally specify a metadata field containing a schema definition to use for encoding instead of a statically defined schema. For batches of messages, the first message's schema will be applied to all subsequent messages of the batch.").
```

internal/impl/parquet/processor_encode.go
```diff
@@ -312,3 +347,92 @@ func (s *parquetEncodeProcessor) ProcessBatch(_ context.Context, batch service.M
...
+	case schema.Int64:
+		n = parquet.Int(64)
+	case schema.Float32:
+		n = parquet.Leaf(parquet.FloatType)
+	case schema.Float64:
+		n = parquet.Leaf(parquet.DoubleType)
+	case schema.String:
+		n = parquet.String()
+	case schema.Timestamp:
+		// TODO: add field to specify timestamp unit
```
josephwoodward: I presume this todo is for a later date (as opposed to something that was missed)? Would it be worth creating an issue for it?

docs/modules/components/pages/processors/parquet_encode.adoc
```diff
@@ -172,6 +174,15 @@ fields:
     type: BYTE_ARRAY
 ```
 
+=== `schema_metadata`
+
+Optionally specify a metadata field containing a schema definition to use for encoding instead of a statically defined schema. For batches of messages, the first message's schema will be applied to all subsequent messages of the batch.
```
mmatczuk, line 179: Could we perhaps have metadata on batch messages in benthos?
What about rebatching in the pipeline?

internal/impl/confluent/serde_goavro.go
```diff
@@ -121,8 +123,22 @@ func (s *schemaRegistryEncoder) getAvroEncoder(ctx context.Context, schema franz
...
 
-func (s *schemaRegistryDecoder) getGoAvroDecoder(ctx context.Context, schema franz_sr.Schema) (schemaDecoder, error) {
-	schemaSpec, err := resolveGoAvroReferences(ctx, s.client, s.cfg.avro.mapping, schema)
+func tryExtractCommonSchemaFromAvroBytes(rawUnion bool, specBytes []byte) (any, error) {
+	var as any
+	if err := json.Unmarshal(specBytes, &as); err != nil {
+		return nil, err
+	}
+
+	commonSchema, err := avroAnyToCommonSchema(rawUnion, as)
```
mmatczuk: Could we avoid passing the bool parameter from function to function? Maybe adding a type would help.

mmatczuk: It's not clear to me what is the benefit of this API that accepts any. It's coupled with what json package produces - consider changing the top level APIs to work on json as byte slice.

internal/impl/confluent/serde_goavro.go
```diff
@@ -121,8 +123,22 @@ func (s *schemaRegistryEncoder) getAvroEncoder(ctx context.Context, schema franz
...
+	if err != nil {
+		return nil, err
+	}
+
+	return commonSchema.ToAny(), nil
+}
+
+func (s *schemaRegistryDecoder) getGoAvroDecoder(ctx context.Context, aschema franz_sr.Schema) (schemaDecoder, error) {
```
mmatczuk, line 124: The `get` is not idiomatic Go - how about `GoAvroDecoderForSchema`?

internal/impl/confluent/serde_goavro.go
```diff
@@ -137,6 +153,13 @@ func (s *schemaRegistryDecoder) getGoAvroDecoder(ctx context.Context, schema fra
 		return nil, err
 	}
 
+	var commonSchemaAny any
```
mmatczuk: Why add type Any suffix to the name?

internal/impl/confluent/serde_goavro.go
```diff
@@ -154,8 +177,224 @@ func (s *schemaRegistryDecoder) getGoAvroDecoder(ctx context.Context, schema fra
...
+		c.Children = append(c.Children, schema.Common{
+			Type:     schema.Object,
+			Children: []schema.Common{childT},
+		})
+	}
+
+	return nil
+}
+
+func avroAnyMapToCommonSchema(rawUnion bool, as map[string]any) (schema.Common, error) {
```
mmatczuk: It feels like all the Avro helper functions that call themselves recursively could have a dedicated namespace / type and file.

internal/impl/confluent/serde_goavro.go
```diff
@@ -154,8 +177,224 @@ func (s *schemaRegistryDecoder) getGoAvroDecoder(ctx context.Context, schema fra
...
+	var c schema.Common
+	c.Name, _ = as["name"].(string)
+
+	switch t := as["type"].(type) {
```
mmatczuk: Consider extracting to function / method.

internal/impl/confluent/serde_goavro.go
```diff
@@ -154,8 +177,224 @@ func (s *schemaRegistryDecoder) getGoAvroDecoder(ctx context.Context, schema fra
...
+
+			root.Children = append(root.Children, cObj)
+		}
+		return root, nil
+	}
+	return schema.Common{}, fmt.Errorf("expected either an array or object at root of schema, got %T", aRoot)
+}
+
+func avroHydrateRawUnion(c *schema.Common, types []any) error {
+	if c.Type, c.Optional = isUnionJustOptional(types); !c.Optional {
```
mmatczuk: Return early if condition is not met.

internal/impl/confluent/serde_goavro.go
```diff
@@ -121,8 +123,22 @@ func (s *schemaRegistryEncoder) getAvroEncoder(ctx context.Context, schema franz
 	}, nil
 }
 
-func (s *schemaRegistryDecoder) getGoAvroDecoder(ctx context.Context, schema franz_sr.Schema) (schemaDecoder, error) {
-	schemaSpec, err := resolveGoAvroReferences(ctx, s.client, s.cfg.avro.mapping, schema)
+func tryExtractCommonSchemaFromAvroBytes(rawUnion bool, specBytes []byte) (any, error) {
```
josephwoodward: (subjective): 'tryDoSomething` feels off here. Most functions attempt to do something and return an error if they can't, so perhaps try feels kind of redundant?

internal/impl/parquet/processor_encode.go
Jeffail: Yeah exactly, done https://github.com/redpanda-data/connect/issues/3570

docs/modules/components/pages/processors/parquet_encode.adoc
Jeffail: We could potentially expand the functionality to group messages within the processor by their schema hash but I figured to start with we enforce that all batched messages must match in schema. If users are mixing schemas in a pipeline they'll be expected to use a `group_by` processor beforehand, which is how this processor currently works (with the static schemas).

internal/impl/confluent/serde_goavro.go
Jeffail: its the same for all the other schema formats, we can change it later if we need to refactor the package.

internal/impl/parquet/processor_encode.go
josephwoodward: If the TODO is staying, would it be helpful to reference the issue number/link in the TODO?

mmatczuk (approved):

internal/impl/parquet/processor_encode.go
Jeffail: Good catch, done 👍

PR #3543, opened 2025-07-15
gh: restore build task binary due to arduino rate limits

mmatczuk:
Fix

```
Run arduino/setup-task@v2
Error: API rate limit exceeded for 104.209.5.144. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```


mihaitodor (approved):
:shipit:

PR #3544, opened 2025-07-15
gh: skip unstable integration tests in CI

mmatczuk:


taskfiles/test.yml
```diff
@@ -21,13 +21,51 @@ tasks:
...
+        - ./internal/impl/gcp/enterprise/changestreams/metadata
+        - ./internal/impl/hdfs
+        - ./internal/impl/kafka
+        - ./internal/impl/mongodb
+        - ./internal/impl/mqtt
+        - ./internal/impl/mysql
+        - ./internal/impl/nanomsg
+        - ./internal/impl/nats
+        - ./internal/impl/opensearch
+        - ./internal/impl/postgresql
```
rockwotj, line 40: Why is this skipped? It should work fine in CI

mmatczuk: But it didn't. Maybe we should run each in separate runner to make it parallel and not dependent on other test failures.

Issue #3348, opened 2025-04-17
cassandra: add support for RackAwareRoundRobinPolicy and DCAwareRoundRobinPolicy


mmatczuk:
Add possibility to specify DC and Rack for ScyllaDB/Cassandra (using gocql).

Modify cassandra [inputs](https://docs.redpanda.com/redpanda-connect/components/inputs/cassandra/) and [outputs](https://docs.redpanda.com/redpanda-connect/components/outputs/cassandra/) by adding `dc` and `rack` fields.

* If `dc` and `rack` is specified use RackAwareRoundRobinPolicy
* If only `dc` is specified use DCAwareRoundRobinPolicy
* If only `rack` is specified report error

PR #3545, opened 2025-07-15
Cassandra: Add support for DCAware, RackAware and TokenAware Selection Policies

jonny7:
Fixes https://github.com/redpanda-data/connect/issues/3348

This adds support for specifying the various Cassandra `HostSelectionPolicies`. It's implemented in a way that is non-breaking. E.g. not using the new `host_selection_policy` maintains the existing driver default, which is `gocql.RoundRobinHostPolicy`. Though in practice, this likely shouldn't be the default, as the `TokenAwareHostPolicy` allows requests to be routed to the node where the data resides, whereas the round-robin will often require a secondary hop via the coordinator node. As per the driver [docs ](https://github.com/apache/cassandra-gocql-driver) you can see that they also recommend this default.

> Some tips for getting more performance from the driver:
> Use the TokenAware policy

This is my first time working on the project, so let me know if there are any idioms or things I didn't follow correctly and i'll adjust

### DIFF
--- a/internal/impl/cassandra/shared.go
+++ b/internal/impl/cassandra/shared.go
@@ -197,7 +197,7 @@ func clientConfFromParsed(conf *service.ParsedConfig) (c clientConf, err error) 
 func newHostSelectionPolicy(localDC, localRack string) (gocql.HostSelectionPolicy, error) {
 	if localRack != "" {
 		if localDC == "" {
-			return nil, fmt.Errorf("localDC cannot be empty when localRack is set")
+			return nil, errors.New("localDC cannot be empty when localRack is set")
 		}
 		return gocql.RackAwareRoundRobinPolicy(localDC, localRack), nil
 	}


mmatczuk:
Thanks for contributing @jonny7. I noted your ScyllaDB certification, is this for Cassandra or ScyllaDB?

jonny7:
> Thanks for contributing @jonny7. I noted your ScyllaDB certification, is this for Cassandra or ScyllaDB?

Hi @mmatczuk 

The change is for Cassandra, if you see the linked driver docs and in particular it pertains to this https://github.com/apache/cassandra-gocql-driver/blob/trunk/policies.go

mmatczuk:
Thank you for the clarification, @jonny7 - much appreciated.

Personally, I think it would be preferable to expose only the `local_dc` and `local_rack` fields to users, and internally select the most suitable host selection policy based on that information.

Ideally, we should use `gocql.TokenAwareHostPolicy` as the primary policy. As a fallback, we might consider one of the following, in order of preference:
	•	gocql.RackAwareRoundRobinPolicy
	•	gocql.DCAwareRoundRobinPolicy
	•	gocql.RoundRobinPolicy
assuming the configuration is viable in each case.

Also worth noting: `TokenAwareHostPolicy` supports options like `gocql.ShuffleReplicas` and `gocql.NonLocalReplicasFallback`. In my opinion, both should be enabled by default.

jonny7:
OK, If you're happy to change the default behaviour to `TokenAwareHostPolicy` I can do that. It actually makes the code easier.  I'll also set those other properties by default

internal/impl/cassandra/shared.go
```diff
@@ -84,6 +87,16 @@ func clientFields() []*service.ConfigField {
...
+		service.NewObjectField(cFieldHostSelectionPolicy,
+			service.NewStringField(cFieldHostSelectionPolicyLocalDC).
+				Description("The local DC to use, this is only applicable for the DC Aware & Rack Aware policies").
+				Optional(),
+			service.NewStringField(cFieldHostSelectionPolicyLocalRack).
+				Description("The local Rack to use, this is only applicable for the Rack Aware Policy").
+				Optional(),
+		).
+			Description("Optional host selection policy configurations. Defaults to TokenAwareHostPolicy with fallback of RoundRobinHostPolicy").
+			Advanced(),
```
mmatczuk, line 101: You can add a lint rule like that 

```
LintRule(`root = if this.local_rack != "" && this.local_dc == "" { "local_dc must be set if local_rack is set" }`).
```

internal/impl/cassandra/shared_test.go
```diff
@@ -0,0 +1,52 @@
...
+			localDC:            "",
+			localRack:          "",
+			expectedPolicyType: gocql.RoundRobinHostPolicy(),
+		},
+		{
+			name:               "Round Robin - Only Rack provided",
+			localDC:            "",
+			localRack:          "rack2",
+			expectedPolicyType: gocql.RoundRobinHostPolicy(),
+		},
```
mmatczuk: It seems like this should result in an error.

mmatczuk:
It looks like the docs could be updated to the new code, otherwise LGTM.

internal/impl/cassandra/shared.go
jonny7: Can I confirm this rule, I applied it, but it only triggers a lint failure if you actually supply `local_dc` in the config and leave it empty. If you skip the property all together, it doesn't trigger a linter fail.  I presume this is because the `local_dc` is optional and just not present. Is that the intended behavior you wanted?

mmatczuk: Good catch, I think there is also `this.exists("<name>")` function.

jonny7: OK thanks, I did find this in the bloblang docs, but wanted to check I understood correctly. I'll update

jonny7: Updated

internal/impl/cassandra/shared_test.go
jonny7: Thanks, made it so the `newHostPolicy` can return an error now that matches the linter rule and updated tests

jonny7:
> It looks like the docs could be updated to the new code, otherwise LGTM.

I re-ran the docs_gen tool also

internal/impl/cassandra/shared.go
```diff
@@ -164,5 +182,27 @@ func clientConfFromParsed(conf *service.ParsedConfig) (c clientConf, err error)
...
+			return
+		}
+	}
 	return
 }
+
+func newHostSelectionPolicy(localDC, localRack string) (gocql.HostSelectionPolicy, error) {
+	if localRack != "" {
+		if localDC == "" {
+			return nil, fmt.Errorf("localDC cannot be empty when localRack is set")
```
mmatczuk:
```suggestion
			return nil, errors.New("localDC cannot be empty when localRack is set")
```

mmatczuk (approved):

internal/impl/cassandra/shared.go
jonny7: Thanks

PR #3546, opened 2025-07-15
kafka(migrator): extract migrator integration tests to separate file and refactor for maintainability

mmatczuk:


internal/impl/kafka/integration_migrator_test.go
```diff
@@ -0,0 +1,929 @@
...
+
+		<-closeChan
+	})
+}
+
+// migratorPipeline is a helper type that encapsulates a
+// redpanda_migrator_offsets pipeline and provides methods for interacting with
+// it during tests.
+type migratorPipeline struct {
+	t          *testing.T
```
mihaitodor: I find it a bit odd to cache `t` in a struct. Would it not make more sense to pass it as a parameter to `Cleanup()`?

mihaitodor (approved):
That looks great, thank you for the refactor @mmatczuk! ❤️ Just one small comment, feel free to 🐑 🚀

internal/impl/kafka/integration_migrator_test.go
mmatczuk: Fixed.

PR #3549, opened 2025-07-16
gh: use matrix test for running integration tests

mmatczuk:
Remove task test:integration and leave test:integration-package. Move alias it to integration-package.


rockwotj:
> LGTM, but maybe a bit annoying to maintain since we'd have to update this list when adding new packages... Not sure if there's any way around it, though. Feel free to 🐑 🚀

We could have a CI task that ensures all the packages are listed here.

PR #3553, opened 2025-07-16
rpcplugin/python: implement clean shutdown

rockwotj:


public/plugin/python/src/redpanda_connect/_grpc.py
```diff
@@ -345,13 +354,19 @@ async def _serve_component(register: Callable[[grpc.aio.Server], None]):
...
     async def stop(sig: int):
-        _logger.info("stopping server")
-        await server.stop(grace=None)
+        if sig == signal.SIGTERM:
+            _logger.warning("Recieved SIGTERM stopping server immediately")
+            await server.stop(grace=None)
+        else:
+            _logger.info(f"Recieved {signal.strsignal(sig)} waiting for server close")
+            await closed_event.wait()
+            await server.stop(grace=30)
```
mmatczuk, line 369: Maybe I'm missing something but shouldn't the order here be reversed?

rockwotj: No we want to get the close RPC first, then shut the server down. If we shut the server down we immediately stop serving RPCs

mmatczuk: Gotcha

PR #3554, opened 2025-07-17
Add extra fields to `schema_registry` output

mihaitodor:
- `normalize` - Normalize schemas.
- `remove_metadata` - Remove metadata from schemas.
- `remove_rule_set` - Remove rule set from schemas.


internal/impl/kafka/output_schema_registry.go
```diff
@@ -373,22 +394,47 @@ func (o *schemaRegistryOutput) createSchemaDeps(ctx context.Context, ss franz_sr
 	return nil
 }
 
+// TODO: Remove this utility after https://github.com/redpanda-data/redpanda/issues/26331 is resolved.
+func areSchemasEqual(lhs, rhs franz_sr.SubjectSchema) bool {
+	// Remove newlines and leading/trailing spaces from the schemas before comparison.
```
mihaitodor: Unfortunately, Redpanda applies some normalisation even when `?normalize=true` isn't set.

internal/impl/kafka/output_schema_registry.go
```diff
@@ -23,10 +23,10 @@ import (
 	"net/http"
 	"net/url"
 	"slices"
+	"strings"
 	"sync"
 	"sync/atomic"
 
-	"cuelang.org/go/pkg/strings"
```
mihaitodor, line 29: My bad, I didn't notice when this import got added automatically in a previous PR...

internal/impl/kafka/output_schema_registry.go
```diff
@@ -83,6 +86,9 @@ func schemaRegistryOutputConfigFields() []*service.ConfigField {
 		service.NewInterpolatedStringField(sroFieldSubject).Description("Subject."),
 		service.NewBoolField(sroFieldBackfillDependencies).Description("Backfill schema references and previous versions.").Default(true).Advanced(),
 		service.NewBoolField(sroFieldTranslateIDs).Description("Translate schema IDs.").Default(false).Advanced(),
+		service.NewBoolField(sroFieldNormalize).Description("Normalize schemas.").Default(true).Advanced(),
```
mihaitodor, line 88: Do we want this to be `true` by default? Schema Registry dedupes protobuf schemas with reordered fields even when `?normalize=true` isn't set. See here: https://github.com/redpanda-data/redpanda/issues/26858.

internal/impl/kafka/output_schema_registry.go
```diff
@@ -373,22 +394,47 @@ func (o *schemaRegistryOutput) createSchemaDeps(ctx context.Context, ss franz_sr
...
+
+	if lhsSchema != rhsSchema {
+		return false
+	}
+
+	// Compare the rest of the fields.
+	lhs.Schema.Schema = ""
+	rhs.Schema.Schema = ""
+
+	return cmp.Equal(lhs, rhs)
```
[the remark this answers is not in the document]
mihaitodor: Nah, it's fine.

internal/impl/kafka/integration_test.go
```diff
@@ -821,6 +822,188 @@ output:
...
+			normalize:  true,
+		},
+		{
+			name:       "removes metadata",
+			iterations: 1,
+			metadata: `{
+  "properties": {
+    "confluent:version": "1"
+  }
+}`,
```
mmatczuk: It feels like those should be subtests, not table driven tests.

internal/impl/kafka/integration_test.go
```diff
@@ -821,6 +822,188 @@ output:
...
+      "type": "CEL",
+      "mode": "WRITE",
+      "expr": "size(message.ssn) == 9"
+    }
+  ]
+}`,
+		},
+	}
+
+	writeSchema := func(t *testing.T, schema []byte, normalize, removeMetadata, removeRuleSet bool) {
```
mmatczuk: Could it be extracted from as a normal function?

internal/impl/kafka/integration_test.go
```diff
@@ -821,6 +822,188 @@ output:
...
+		streamBuilder := service.NewStreamBuilder()
+
+		// Set up a dummy `schema_registry` input which the output can connect to even though it won't need to fetch any
+		// schemas from it.
+		err = streamBuilder.AddInputYAML(fmt.Sprintf(`
+schema_registry:
+  url: %s
+  subject_filter: does_not_exist
+`, sr.SchemaRegistryURL))
+		require.NoError(t, err)
```
mmatczuk: Could we extract yaml creation to variable and inline err here.

internal/impl/kafka/integration_test.go
```diff
@@ -821,6 +822,188 @@ output:
...
+		err = streamBuilder.AddOutputYAML(fmt.Sprintf(`
+schema_registry:
+  url: %s
+  subject: ${! json("subject") }
+  backfill_dependencies: true
+  normalize: %t
+  remove_metadata: %t
+  remove_rule_set: %t
+`, sr.SchemaRegistryURL, normalize, removeMetadata, removeRuleSet))
+		require.NoError(t, err)
```
mmatczuk: Same as above.

internal/impl/kafka/integration_test.go
```diff
@@ -821,6 +822,188 @@ output:
...
+			dummySubject := "foo"
+
+			defer func() {
+				// Clean up the subject at the end of each subtest.
+				deleteSubject(t, sr.SchemaRegistryURL, dummySubject, false)
+				deleteSubject(t, sr.SchemaRegistryURL, dummySubject, true)
+			}()
+
+			for range test.iterations {
+				dummyProtoSchema := `syntax = "proto3";
```
mmatczuk: How about a const.

internal/impl/kafka/integration_test.go
```diff
@@ -821,6 +822,188 @@ output:
...
+package com.mycorp.mynamespace;
+
+message SampleRecord {
+  int32 my_field1 = 1;
+  double my_field2 = 2;
+  string my_field3 = 3;
+}`
+
+				dummyDenormalizedProtoSchema := `syntax = "proto3";
```
mmatczuk: How about a const. Also, please add a comment it took me a while to spot the difference.

internal/impl/kafka/integration_test.go
```diff
@@ -821,6 +822,188 @@ output:
...
+}`
+
+				dummySchema := dummyProtoSchema
+				if test.normalize {
+					dummySchema = dummyDenormalizedProtoSchema
+				}
+
+				var metadata *franz_sr.SchemaMetadata
+				if test.metadata != "" {
+					require.NoError(t, json.Unmarshal([]byte(test.metadata), &metadata))
```
mmatczuk: Maybe it would be worth adding more assertions on the value?

internal/impl/kafka/output_schema_registry.go
mmatczuk: What is the value of disabling it? If none we probably can remove this field and always normalize - code would become cleaner, and configuration would be simpler.

internal/impl/kafka/output_schema_registry.go
```diff
@@ -83,6 +85,9 @@ func schemaRegistryOutputConfigFields() []*service.ConfigField {
 		service.NewInterpolatedStringField(sroFieldSubject).Description("Subject."),
 		service.NewBoolField(sroFieldBackfillDependencies).Description("Backfill schema references and previous versions.").Default(true).Advanced(),
 		service.NewBoolField(sroFieldTranslateIDs).Description("Translate schema IDs.").Default(false).Advanced(),
+		service.NewBoolField(sroFieldNormalize).Description("Normalize schemas.").Default(true).Advanced(),
+		service.NewBoolField(sroFieldRemoveMetadata).Description("Remove metadata from schemas.").Default(true).Advanced(),
+		service.NewBoolField(sroFieldRemoveRuleSet).Description("Remove rule set from schemas.").Default(true).Advanced(),
```
mmatczuk, line 90: Why do we want to trim schemas by default?

internal/impl/kafka/schema_registry.go
```diff
@@ -14,5 +14,31 @@
...
 // srResourceKey is a type that represents a key for registering a `schema_registry` resource.
 type srResourceKey string
+
+// AreSchemasEqual compares two schema objects for equality, ignoring newlines and leading/trailing spaces in the schema string.
+func AreSchemasEqual(lhs, rhs franz_sr.SubjectSchema) bool {
+	// TODO: Remove this utility after https://github.com/redpanda-data/redpanda/issues/26331 is resolved.
+
+	// Remove newlines and leading/trailing spaces from the schemas before comparison.
+	lhsSchema := strings.TrimSpace(strings.ReplaceAll(lhs.Schema.Schema, "\n", ""))
+	rhsSchema := strings.TrimSpace(strings.ReplaceAll(rhs.Schema.Schema, "\n", ""))
```
mmatczuk, line 34: Can's we replace all whitespace with regex `/"\s"/""/g`?

internal/impl/kafka/schema_registry.go
```diff
@@ -14,5 +14,31 @@
...
+
+	if lhsSchema != rhsSchema {
+		return false
```
mmatczuk, line 37: Maybe https://pkg.go.dev/strings#Compare?

internal/impl/kafka/schema_registry.go
```diff
@@ -14,5 +14,31 @@
...
+	}
+
+	// Compare the rest of the fields.
+	lhs.Schema.Schema = ""
+	rhs.Schema.Schema = ""
+
+	return cmp.Equal(lhs, rhs)
```
mmatczuk: We could do https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#IgnoreFields instead of destroying the passed objects.

internal/impl/kafka/output_schema_registry.go
mihaitodor: If users don't want to do schema ID translation and the normalisation leads to more aggressive deduplication than what Schema Registry normally does, then they could get fewer schemas in the destination which also means some schemas will have different IDs... However, https://github.com/redpanda-data/redpanda/issues/26858 would have to be addressed before they can try to run with `normalize: false`.

internal/impl/kafka/integration_test.go
mihaitodor: But they are subtests. See below... How should I refactor them?

internal/impl/kafka/integration_test.go
mihaitodor: Do you mean `const dummyProtoSchema = `...``? I haven't really optimised for const correctness so far in Go unless it makes sense to extract the consts outside of funcs. It can help avoid issues, but it also adds overhead and the rest of the code is no longer consistent.

internal/impl/kafka/integration_test.go
mihaitodor: This is just static data that I set up in the test table. Don't think it's worth doing that here (maybe in the `sr` package in `franz-go`).

internal/impl/kafka/output_schema_registry.go
mihaitodor: Because Redpanda doesn't support metadata and rule sets.

internal/impl/kafka/schema_registry.go
mihaitodor: Sure, but I was trying to be conservative. It should fail if somehow the schema contains some sort of value which has spaces in it, because that can have semantic meaning. Newlines, hopefully not...

internal/impl/kafka/schema_registry.go
mihaitodor: Why? We only need to know if they differ.

> It is usually clearer and always faster to use the built-in string comparison operators ==, <, >, and so on.

internal/impl/kafka/schema_registry.go
mihaitodor: Nice, thank you! ❤️

internal/impl/kafka/schema_registry.go
mmatczuk: Unlikely but ok

internal/impl/kafka/output_schema_registry.go
mmatczuk: How can we automatically detect Redpanda?

internal/impl/kafka/integration_test.go
mmatczuk: It's a static text what's the problem, you could as well put it to testdata and embed it.

internal/impl/kafka/integration_test.go
mmatczuk: The point is that they are different enough that the table doe not add value i.e. table could be removed.

internal/impl/kafka/schema_registry.go
```diff
@@ -14,5 +14,31 @@
...
+
+	"github.com/google/go-cmp/cmp"
+	franz_sr "github.com/twmb/franz-go/pkg/sr"
+)
+
 // srResourceKey is a type that represents a key for registering a `schema_registry` resource.
 type srResourceKey string
+
+// AreSchemasEqual compares two schema objects for equality, ignoring newlines and leading/trailing spaces in the schema string.
+func AreSchemasEqual(lhs, rhs franz_sr.SubjectSchema) bool {
```
mmatczuk: Could we rename it ti just SchemasEqual?

internal/impl/kafka/output_schema_registry.go
mihaitodor: I don't think we can through standard API calls... Maybe something to ask @BenPope. Also, this should only connect to Schema Registry, so there would need to be some non-standard REST API endpoint that Redpanda maintains. For now, I'd just leave it as a config option.

internal/impl/kafka/integration_test.go
mihaitodor: I'd struggle to remember to add `const` whenever I add static strings, numbers etc given that the rest of the code does not enforce this. Like, should we have `const i = 3` instead of `i = 3` in a function with a few lines of code? I could get used to that, but it should be uniform.

mihaitodor: We can have testdata too, but then you have to look into another file and that's a bit annoying... I'd do that only if the data would be dozens of lines and fills my screen.

internal/impl/kafka/integration_test.go
mihaitodor: Yeah, I'm on the fence with that. I'd need to add a wrapper function which the tests call. Let me see how it ends up looking.

internal/impl/kafka/output_schema_registry.go
BenPope:
```
curl http://127.0.0.5:18081/v1/  | grep -E "Pandaproxy|Redpanda"
```

mmatczuk: Sweet, let's autodetect it then.

nicolaferraro: Does not work in serverless (unless we implement it)

mihaitodor: Thanks for chiming in Nicola! Then I think it's best to leave these fields as user-configurable for now.

internal/impl/kafka/integration_test.go
mihaitodor: Done. Guess it looks a bit less convoluted than the table-driven version I had before. WDYT?

mmatczuk (approved):

PR #3563, opened 2025-07-21
 kafka: schema registry migrate the subject compatibility level

mmatczuk:


mihaitodor (commented):
Hey @mmatczuk, thank you so much for jumping on this one! Unfortunately it fell off the rails last week... The code looks great!

I think there's one more edge case which I think needs to be addressed: After the existing schemas have been copied over and the `redpanda_migrator` output starts writing message, it's possible that new schemas get added in the source cluster. In that case, the `redpanda_migrator` output gets a message containing a schema ID which it doesn't have in the `schemaIDCache`, so then it calls `srOutput.GetDestinationSchemaID()`. However, this method uses a backchannel to the `schema_registry` input to fetch info about this new schema ID and then it calls `getOrCreateSchemaID()`, bypassing `maybeUpdateCompatibilityLevel()`.

mmatczuk:
This is quite weird, it's a different implementation, that does not take into account the interpolation.

mihaitodor:
I know... But this is the best I could think of without having the output tell the `schema_registry` input to generate more messages and then somehow wait for those to be written.

One more thing to check: Right now, `getOrCreateSchemaID` can lead to extra subjects being created if the schema has references which are associated with other subjects. In that case, I believe `maybeUpdateCompatibilityLevel` isn't called for those subjects.

internal/impl/confluent/sr/client.go
```diff
@@ -223,3 +224,65 @@ func (c *Client) walkReferencesTracked(ctx context.Context, seen map[string]int,
...
 }
+
+// CompatibilityLevelUnknown is used when the compatibility level of a subject
+// could not be determined.
+const CompatibilityLevelUnknown = sr.CompatibilityLevel(0)
+
+// GetCompatibilityLevel returns the compatibility level of the given subjects.
+//
+// If the client could not query the compatibility level for a subject (i.e. due
+// to a network error) the subject is associated with the
```
josephwoodward:
```suggestion
// to a network error), the subject is associated with the
```

internal/impl/confluent/sr/client.go
```diff
@@ -223,3 +224,65 @@ func (c *Client) walkReferencesTracked(ctx context.Context, seen map[string]int,
 	}
 	return nil
 }
+
+// CompatibilityLevelUnknown is used when the compatibility level of a subject
+// could not be determined.
+const CompatibilityLevelUnknown = sr.CompatibilityLevel(0)
+
+// GetCompatibilityLevel returns the compatibility level of the given subjects.
```
josephwoodward, line 232:
```suggestion
// GetCompatibilityLevel returns the compatibility level of the given subject(s).
```

internal/impl/confluent/sr/client.go
```diff
@@ -223,3 +224,65 @@ func (c *Client) walkReferencesTracked(ctx context.Context, seen map[string]int,
...
+		} else {
+			levels[i] = res.Level
+		}
+	}
+
+	return levels
+}
+
+// UpdateCompatibilityLevel updates the compatibility level of a subject if it
+// differs from the given `level`. If the `level` is `CompatibilityLevelUnknown`
```
josephwoodward:
```suggestion
// differs from the given `level`. If the `level` is `CompatibilityLevelUnknown`,
```

internal/impl/kafka/output_schema_registry.go
```diff
@@ -288,10 +343,18 @@ func (o *schemaRegistryOutput) GetDestinationSchemaID(ctx context.Context, id in
 		return -1, fmt.Errorf("no subjects found for schema ID %d", id)
 	}
 
-	// Register the schema with all the subjects it's associated with in the source Schema Registry. Each call should
-	// return the same destination schema ID.
+	// Register the schema with all the subjects it's associated with in the
+	// source Schema Registry. Each call should return the same destination schema ID.
 	var destinationID int
 	for _, subject := range schemaSubjects {
+		// Update compatibility level for the subject before creating the schema.		// the schema
```
josephwoodward:
```suggestion
		// Update compatibility level for the subject before creating the schema.
```

josephwoodward (approved):

Jeffail (approved):

PR #3564, opened 2025-07-23
Redpanda Migrator V2

mmatczuk:
This PR introduces new `redpanda_migrator2` input and output components for comprehensive data migration between Apache Kafka and Redpanda clusters. These components replace and improve upon the existing redpanda_migrator components.

Key Features:
- Topics: Automatic creation with mirrored partition counts, configurable replication factor, optional ACL replication
- Schema Registry: One-shot or periodic syncing with subject selection, ID translation, and schema normalisation
- Consumer Groups: Periodic syncing with timestamp-based offset translation and safety guarantees

### DIFF
--- a/internal/impl/redpanda/migrator/migrator.go
+++ b/internal/impl/redpanda/migrator/migrator.go
@@ -179,7 +179,6 @@ output:
 		// Kafka fields
 		Fields(kafka.FranzConnectionFields()...).
 		Fields(kafka.FranzProducerFields()...).
-		Fields(kafka.FranzWriterConfigFields()...).
 		// Schema registry fields
 		Field(schemaRegistryField(schemaRegistryMigratorFields()...).Optional()).
 		// Consumer groups fields
--- a/docs/modules/components/pages/outputs/redpanda_migrator.adoc
+++ b/docs/modules/components/pages/outputs/redpanda_migrator.adoc
@@ -40,12 +40,6 @@ output:
   label: ""
   redpanda_migrator:
     seed_brokers: [] # No default (required)
-    topic: ${! metadata('kafka_topic') } # No default (optional)
-    key: "" # No default (optional)
-    partition: ${! meta("partition") } # No default (optional)
-    metadata:
-      include_prefixes: []
-      include_patterns: []
     schema_registry:
       url: http://localhost:8081 # No default (required)
       enabled: true
@@ -64,6 +58,7 @@ output:
       include: [] # No default (optional)
       exclude: [] # No default (optional)
       only_empty: false
+    topic: ${! metadata('kafka_topic') } # No default (optional)
     topic_replication_factor: "3" # No default (optional)
     sync_topic_acls: false
 ```
@@ -98,13 +93,6 @@ output:
     timeout: 10s
     max_message_bytes: 1MiB
     broker_write_max_bytes: 100MiB
-    topic: ${! metadata('kafka_topic') } # No default (optional)
-    key: "" # No default (optional)
-    partition: ${! meta("partition") } # No default (optional)
-    metadata:
-      include_prefixes: []
-      include_patterns: []
-    timestamp_ms: ${! timestamp_unix_milli() } # No default (optional)
     schema_registry:
       url: http://localhost:8081 # No default (required)
       tls:
@@ -146,6 +134,7 @@ output:
       include: [] # No default (optional)
       exclude: [] # No default (optional)
       only_empty: false
+    topic: ${! metadata('kafka_topic') } # No default (optional)
     topic_replication_factor: "3" # No default (optional)
     sync_topic_acls: false
     serverless: false
@@ -768,114 +757,6 @@ broker_write_max_bytes: 128MB
 broker_write_max_bytes: 50mib
 ```
 
-=== `topic`
-
-The topic to write messages to. Use interpolation to derive destination topic names from source topics. The source topic name is available as 'kafka_topic' metadata.
-This field supports xref:configuration:interpolation.adoc#bloblang-queries[interpolation functions].
-
-
-*Type*: `string`
-
-
-```yml
-# Examples
-
-topic: ${! metadata('kafka_topic') }
-
-topic: prod_${! metadata('kafka_topic') }
-```
-
-=== `key`
-
-An optional key to populate for each message.
-This field supports xref:configuration:interpolation.adoc#bloblang-queries[interpolation functions].
-
-
-*Type*: `string`
-
-
-=== `partition`
-
-An optional explicit partition to set for each message. This field is only relevant when the `partitioner` is set to `manual`. The provided interpolation string must be a valid integer.
-This field supports xref:configuration:interpolation.adoc#bloblang-queries[interpolation functions].
-
-
-*Type*: `string`
-
-
-```yml
-# Examples
-
-partition: ${! meta("partition") }
-```
-
-=== `metadata`
-
-Determine which (if any) metadata values should be added to messages as headers.
-
-
-*Type*: `object`
-
-
-=== `metadata.include_prefixes`
-
-Provide a list of explicit metadata key prefixes to match against.
-
-
-*Type*: `array`
-
-*Default*: `[]`
-
-```yml
-# Examples
-
-include_prefixes:
-  - foo_
-  - bar_
-
-include_prefixes:
-  - kafka_
-
-include_prefixes:
-  - content-
-```
-
-=== `metadata.include_patterns`
-
-Provide a list of explicit metadata key regular expression (re2) patterns to match against.
-
-
-*Type*: `array`
-
-*Default*: `[]`
-
-```yml
-# Examples
-
-include_patterns:
-  - .*
-
-include_patterns:
-  - _timestamp_unix$
-```
-
-=== `timestamp_ms`
-
-An optional timestamp to set for each message expressed in milliseconds. When left empty, the current timestamp is used.
-This field supports xref:configuration:interpolation.adoc#bloblang-queries[interpolation functions].
-
-
-*Type*: `string`
-
-
-```yml
-# Examples
-
-timestamp_ms: ${! timestamp_unix_milli() }
-
-timestamp_ms: ${! metadata("kafka_timestamp_ms") }
-```
-
 === `schema_registry`
 
 Configuration for schema registry integration. Enables migration of schema subjects, versions, and compatibility settings between clusters.
@@ -1427,6 +1308,23 @@ Whether to only migrate Empty consumer groups. When false (default), all statuse
 *Type*: `bool`
 
 *Default*: `false`
+
+=== `topic`
+
+The topic to write messages to. Use interpolation to derive destination topic names from source topics. The source topic name is available as 'kafka_topic' metadata.
+This field supports xref:configuration:interpolation.adoc#bloblang-queries[interpolation functions].
+
+
+*Type*: `string`
+
+
+```yml
+# Examples
+
+topic: ${! metadata('kafka_topic') }
+
+topic: prod_${! metadata('kafka_topic') }
+```
 
 === `topic_replication_factor`
 


internal/impl/redpanda/migrator/migrator_groups.go
```diff
@@ -0,0 +1,363 @@
...
+
+// readRecordTimestamp sends a fetch request to the Redpanda cluster to read the
+// timestamp of the record at the given topic, partition, and offset.
+//
+// The function returns the timestamp of the record at the given offset, or an
+// error if the request fails or if the record is not found.
+//
+// The function does not retry the request if it fails, as the caller should
+// handle retries according to their needs.
+func readRecordTimestamp(
```
mmatczuk, line 716: @twmb could you please review this function to see if I'm not missing anything?

internal/impl/redpanda/migrator/migrator_groups.go
```diff
@@ -0,0 +1,363 @@
...
+	ctx context.Context,
+	client *kgo.Client,
+	topic string,
+	partition int32,
+	offset int64,
+) (time.Time, error) {
+	// Build fetch request
+	req := kmsg.NewPtrFetchRequest()
+	req.MaxWaitMillis = 1000 // If data is not available we wait at most 1 second
```
twmb: I'm not positive how this works in the context of tiered storage -- I don't know if the broker replies AT one second EVEN IF data is being fetched from remote storage, or if the broker waits.

internal/impl/redpanda/migrator/migrator_groups.go
```diff
@@ -0,0 +1,363 @@
...
+	req.MinBytes = 1
+	req.MaxBytes = 1 // The response can exceed MaxBytes if the first record is larger than MaxBytes
+	req.SessionID = -1
```
twmb: My doc line is wrong, SessionID 0 does *not* opt in -- this can be deleted.
SessionEpoch -1 is explicitly what opts out (and -1 is the default, so it does not need to be set)

internal/impl/redpanda/migrator/migrator_groups.go
```diff
@@ -0,0 +1,363 @@
...
+
+	topicReq := kmsg.NewFetchRequestTopic()
+	topicReq.Topic = topic
```
twmb, line 732: This is going to need a topic ID soon -- fetch request v13 relies on topic IDs only, and Redpanda just started adding support for topic IDs in the 25.2 release.

internal/impl/redpanda/migrator/migrator_groups.go
```diff
@@ -0,0 +1,363 @@
...
+
+	partitionReq := kmsg.NewFetchRequestTopicPartition()
+	partitionReq.Partition = partition
+	partitionReq.FetchOffset = offset
```
twmb, line 737: I think you need a partitionReq.PartitionMaxBytes as well? Not positive here to be honest (it's been a while since I've seen a use case of trying to fetch only one record max)

mmatczuk:
```
        // PartitionMaxBytes is the maximum bytes to return for this partition.
	// This can be used to limit how many bytes an individual partition in
	// a request is allotted so that it does not dominate all of MaxBytes.
```

I had a look at the usage, I think I'd skip it.

internal/impl/redpanda/migrator/migrator_groups.go
mmatczuk: Based on my reading of code it seems that it calls `SetReadDeadline` on a connection.

mmatczuk:
cc @mihaitodor

emaxerrno:
why redpanda_migrator2 vs just replacing the current impl? I don't see value in keeping both. Just one makes sense. we can manage the internal migration w/ customers

internal/impl/redpanda/migrator/franz.go
```diff
@@ -0,0 +1,146 @@
...
+	"sync/atomic"
+
+	"github.com/twmb/franz-go/pkg/kgo"
+
+	"github.com/redpanda-data/benthos/v4/public/service"
+	"github.com/redpanda-data/connect/v4/internal/impl/kafka"
+)
+
+func newFranzReaderOrdered(pConf *service.ParsedConfig, mgr *service.Resources) (*kafka.FranzReaderOrdered, error) {
+	var opst []kgo.Opt
```
josephwoodward: nit: I presume this is meant to be `opts`? Would it also be worth moving it down closer to where it's first used?

internal/impl/redpanda/migrator/migrator.go
```diff
@@ -0,0 +1,507 @@
...
+			m.srcAdm = kadm.NewClient(fr.Client)
+
+			return service.AutoRetryNacksBatchedToggled(pConf, migratorBatchInput{fr, m})
+		})
+
+	service.MustRegisterBatchOutput("redpanda_migrator2", migratorOutputConfig(),
+		func(pConf *service.ParsedConfig, mgr *service.Resources) (out service.BatchOutput, batchPolicy service.BatchPolicy, maxInFlight int, err error) {
+			m := newMigratorFrom(mgr)
+
+			err = m.initOutputFromParsed(pConf, mgr)
```
josephwoodward, line 291: nit: Seems a prime candidate for inlining, like you did in the similar `initInputFromParsed` above.

internal/impl/redpanda/migrator/migrator.go
```diff
@@ -0,0 +1,507 @@
...
+
+			fw, err := newFranzWriter(pConf, mgr)
+			if err != nil {
+				return
+			}
+			fw.OnWrite = m.onWrite
+			out = migratorBatchOutput{fw, m}
+
+			maxInFlight, err = pConf.FieldInt(rmoFieldMaxInFlight)
+			if err != nil {
```
josephwoodward: This err check is redundant, is it there for intentionality purposes?

internal/impl/redpanda/migrator/franz.go
```diff
@@ -0,0 +1,146 @@
...
+		return nil, err
+	}
+
+	return fr, nil
+}
+
+// lazyFranzSharedClientInfo defers client creation until Connect due to
+// API restrictions.
+//
+// TODO(mmt): initialize the Franz writer the same way as reader
```
josephwoodward: Is this TODO still required? No problem it being there if it's something you're coming back to, just wanted to check. Would it be worth creating an issue for it if it's something you won't be able to get around to until a late date?

internal/impl/redpanda/migrator/migrator_schema_registry.go
```diff
@@ -0,0 +1,796 @@
...
+			return nil, nil
+		}
+	}
+
+	srURL, err := pConf.FieldURL(srObjectField, srFieldURL)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO(mmt): this does not work with objects
```
josephwoodward: Is this TODO still relevant?

internal/impl/redpanda/migrator/migrator.go
```diff
@@ -0,0 +1,507 @@
...
+	}
+	if m.plumbing&outputInitialized == 0 {
+		return errors.New("output not initialized")
+	}
+	// If schema registry migration is disabled, allow client mismatch.
+	if !m.sr.conf.Enabled {
+		return nil
+	}
+	if m.sr.src != nil && m.sr.dst == nil || m.sr.dst != nil && m.sr.src == nil {
+		return errors.New("schema registry mismatch both input and output must be set")
```
josephwoodward: Would this read better?

```suggestion
		return errors.New("schema registry mismatch: both input and output must be set")
```

mihaitodor:
Hey @mmatczuk, thanks for revamping this! I had a quick look through the code and I think it looks good and it's easier to maintain if there's no need to expose the subcomponents and use templates.

Might be worth having something similar to https://github.com/redpanda-data/connect/blob/main/internal/impl/kafka/redpanda_migrator.md for this implementation.

internal/impl/redpanda/migrator/migrator.go
mmatczuk: maxInFlight has been removed.

internal/impl/redpanda/migrator/migrator_schema_registry.go
mmatczuk: Yes will address that.

internal/impl/redpanda/migrator/migrator.go
mmatczuk: Actually this was for a reason, it plays a role of variable initialization

internal/impl/redpanda/migrator/migrator_schema_registry.go
mmatczuk: Fixed using `Namespace()`

internal/impl/redpanda/migrator/migrator_groups.go
rockwotj: No we need more time to read from tiered storage, we've seen this cause issues in customer clusters having this set to long (there are no keep-alives in the kafka protocol AFAIK) so this needs to be higher and user configurable.

internal/impl/redpanda/migrator/migrator.go
```diff
@@ -0,0 +1,507 @@
...
+}
+
+func parseSchemaID(b []byte) (int, error) {
+	if b == nil {
+		return 0, nil
+	}
+
+	var ch sr.ConfluentHeader
+	schemaID, _, err := ch.DecodeID(b)
+	if err != nil && !errors.Is(err, sr.ErrBadHeader) {
```
rockwotj, line 525: I suppose this is defensive? As `ErrBadHeader` is the only error `DecodeID` returns

internal/impl/redpanda/migrator/migrator.go
```diff
@@ -0,0 +1,507 @@
...
+func (m *Migrator) OnOutputConnected(_ context.Context, fw franzWriter) error { //nolint:revive
+	if err := m.validateInitialized(); err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		defer cancel()
+		<-m.stopSig.SoftStopChan()
+	}()
```
rockwotj: Can this just be `m.stopSig.SoftStopCtx(context.Background())`?

internal/impl/redpanda/migrator/migrator_schema_registry.go
```diff
@@ -0,0 +1,797 @@
...
+
+// SyncLoop runs the schema registry sync in a loop at the configured interval
+// until ctx is done. If interval is <= 0, the loop is not started.
+func (m *schemaRegistryMigrator) SyncLoop(ctx context.Context) {
+	if !m.enabled() {
+		m.log.Info("Schema migration: schema registry sync disabled")
+		return
+	}
+	if m.conf.Interval <= 0 {
+		return
```
rockwotj, line 400: Should this log too?

internal/impl/redpanda/migrator/migrator_schema_registry.go
```diff
@@ -0,0 +1,797 @@
...
+		if !cmp.Equal(as, bs) {
+			return false
+		}
+	case sr.TypeProtobuf:
+		// Remove newlines and leading/trailing spaces from the schemas
+		as := strings.TrimSpace(strings.ReplaceAll(a, "\n", ""))
+		bs := strings.TrimSpace(strings.ReplaceAll(b, "\n", ""))
+		if as != bs {
+			return false
+		}
```
rockwotj, line 628: I guess this is mostly OK but but proto files have comments, etc so it can be incorrect

The proper way is probably too complicated to be worth it.

internal/impl/redpanda/migrator/migrator_groups.go
```diff
@@ -0,0 +1,803 @@
...
+
+// groupsMigratorFields returns the config fields for consumer groups migrator.
+func groupsMigratorFields() []*service.ConfigField {
+	return []*service.ConfigField{
+		service.NewBoolField(srFieldEnabled).
+			Description("Whether consumer group offset migration is enabled. When disabled, no consumer group operations are performed.").
+			Default(true),
+		service.NewDurationField(srFieldInterval).
+			Description("How often to synchronise consumer group offsets. Regular syncing helps maintain offset accuracy during ongoing migration.").
+			Example("0s     # Disabled").
```
rockwotj, line 69: What's the difference from interval 0s and the enabled flag?

internal/impl/redpanda/migrator/migrator_groups.go
```diff
@@ -0,0 +1,803 @@
...
+
+	// Filter out active groups, possible values are:
+	// * Dead – the group has no members and no active metadata; effectively removed.
+	// * Empty – no active members, but group metadata (like offsets) still exists.
+	// * PreparingRebalance – group is in the process of rebalancing, waiting for members to rejoin.
+	// * CompletingRebalance – all members have joined, and assignments are being finalized.
+	// * Stable – group has members, assignments are completed, and it is operating normally.
+	// See: https://kafka.apache.org/40/javadoc/org/apache/kafka/common/GroupState.html
+	groups = slices.DeleteFunc(groups, func(g string) bool {
+		del := cg[g].State != "Empty"
```
rockwotj: Why are we deleting non empty groups? I would think we would replicate active groups as well, that was the behavior for the previous migrator.

internal/impl/redpanda/migrator/migrator_groups.go
```diff
@@ -0,0 +1,803 @@
...
+	if len(respTopic.Partitions) == 0 {
+		return time.Time{}, errors.New("no partitions in response")
+	}
+	respPartition := &respTopic.Partitions[0]
+	if respPartition.ErrorCode != 0 {
+		return time.Time{}, fmt.Errorf("partition error: %d", respPartition.ErrorCode)
+	}
+
+	// Extract record timestamp
+	fp, _ := kgo.ProcessFetchPartition(kgo.ProcessFetchPartitionOpts{
```
rockwotj, line 764: ah OK - this also trims the batch to give back the correct offsets too, nice.

internal/impl/redpanda/migrator/migrator_groups.go
```diff
@@ -0,0 +1,803 @@
...
+	switch tpo.Timestamp {
+	case -1:
+		m.log.Debugf("Consumer group migration: no offsets found for topic '%s' partition %d after timestamp %s",
+			dstTopic, partition, ts)
+		return unknownOffset, nil
+	case ts.UnixMilli():
+		// Actually ListOffsetsAfterMilli returns the offset that is greater or
+		// equal to the requested timestamp, so if the timestamps match, we return.
+		o1 += 1
+	}
```
rockwotj: Why does adding one suffice here? The "we return" part of the comment makes me think this is out of date.

If we're trying to skip this record (since we don't do this if the cluster has a timestamp that is greater), there is no guarantee that the incremented offset doesn't point to another record with the same timestamp.

mmatczuk:
- Rebased on https://github.com/redpanda-data/connect/pull/3611
- Added E2E integration test processing 10k records
- Removed previous version and replaced it

internal/impl/redpanda/migrator/migrator.go
mmatczuk: Future compatibility, @rockwotj do you suggest handling it differently?

internal/impl/redpanda/migrator/migrator.go
mmatczuk: Lovely, thanks.

internal/impl/redpanda/migrator/migrator.go
rockwotj: Nope, this is good, just wanted to make sure I understood

internal/impl/redpanda/migrator/migrator_schema_registry.go
mmatczuk: Added in both SR and CG

internal/impl/redpanda/migrator/migrator_schema_registry.go
mmatczuk: Yes.

internal/impl/redpanda/migrator/migrator_groups.go
mmatczuk: There is no difference it's here for compatibility with schema registry.

internal/impl/redpanda/migrator/migrator_groups.go
mmatczuk: This behavior is inspired by MM2 

```
                // sync offset to the target cluster only if the state of current consumer group is:
                // (1) idle: because the consumer at target is not actively consuming the mirrored topic
                // (2) dead: the new consumer that is recently created at source and never existed at target
                //           This case will be reported as a GroupIdNotFoundException
```

Why would you like to migrate Stable groups? I think there is no point in migrating groups in other states.

rockwotj: Because during a migration you will need to switch over to the destination and if they don't have offsets replicated (because they were a stable group during the migration) they may (depending on configuration) reprocess all existing data in the cluster.

Only migrating empty groups assumes that all consumers are stopped during the migration period, which we don't require people to do.

internal/impl/redpanda/migrator/migrator_groups.go
mmatczuk: It is configurable as `fetch_timeout` I bumped the t/o to 10s

internal/impl/redpanda/migrator/migrator_groups.go
mmatczuk: OK, changed that, and added `ignore_stable` field to only migrate cg when stopped.

internal/impl/redpanda/migrator/franz.go
mmatczuk: Not really, but the fact remains. I removed it.

internal/impl/redpanda/migrator/migrator_groups.go
mmatczuk: @rockwotj I added more context in this comment https://github.com/redpanda-data/connect/pull/3564/files#diff-e02e97d9b4b3c97141816b51252b37b47989619845aae4606108dc21728d99ccR681-R693

PTAL

internal/impl/redpanda/migrator/migrator_groups.go
```diff
@@ -0,0 +1,833 @@
...
+	// Filter out active groups, possible values are:
+	// * Dead – the group has no members and no active metadata; effectively removed.
+	// * Empty – no active members, but group metadata (like offsets) still exists.
+	// * PreparingRebalance – group is in the process of rebalancing, waiting for members to rejoin.
+	// * CompletingRebalance – all members have joined, and assignments are being finalized.
+	// * Stable – group has members, assignments are completed, and it is operating normally.
+	// See: https://kafka.apache.org/40/javadoc/org/apache/kafka/common/GroupState.html
+	groups = slices.DeleteFunc(groups, func(g string) bool {
+		st := cg[g].State
+		allowed := st == "Empty" || (!m.conf.IgnoreStable && st == "Stable")
```
rockwotj: This logic should be:

```
allowed = state != Dead
if ignore stable {
  allowed = state == Empty
}
```

Just because a group decides to start rebalancing doesn't mean it should be excluded from a live migration

internal/impl/redpanda/migrator/migrator_groups.go
```diff
@@ -0,0 +1,833 @@
...
+	// Since we queried for the timestamp of offset-1, we need to adjust the result:
+	//
+	// Case 1: Found timestamp > requested timestamp
+	//   - The exact record wasn't found (may be deleted or destination has newer data)
+	//   - Return the found offset as best approximation
+	//
+	// Case 2: Found timestamp == requested timestamp
+	//   - We found a record with the same timestamp as the record at offset-1
+	//   - Since ListOffsetsAfterMilli returns the FIRST offset with that timestamp,
+	//     we need to add 1 to get the correct translated offset
```
rockwotj, line 700: I see now, I missed we queried for offset-1. Why do we query for offset-1? Is this to handle the case where we've committed the high watermark?

internal/impl/redpanda/migrator/migrator_groups.go
mmatczuk: Sure I'll fix that.

internal/impl/redpanda/migrator/migrator_groups.go
mmatczuk: This way we can elegantly cover cases where we are at the end offset.
The tradeoff is potentially replaying 1ms worth of data.
I did design this before I realizing the resolution limitation.

Let's try to improve it.

mmatczuk: I implemented two code paths and conducted some experiments within `monotonic sub millisecond timestamp` test case.

Given 4 records/ms `t_0, t_1/4, t_2/4, t_3/4, t_1, t_1+1/4, t_1+2/4, t_1+2/4...`

Current version gives `// 1-4 -> 1, 5-8 -> 5, 9-12 -> 9`
Fixed version gives `// 0-3 -> 0, 4-7 -> 4, 8-9 -> 8 (floor)` and `10->9`

Improvement is only visible in cases where you have more than one record with the same timestamp (ms) AND group points to the FIRST record with a given timestamp.

I think it's better to just keep it the way it is.

internal/impl/redpanda/migrator/migrator.go
```diff
@@ -0,0 +1,535 @@
...
+    schema_registry:
+      url: "http://dest-registry:8081"
+      translate_ids: true
+    consumer_groups:
+      interval: 1m
+`).
+		// Kafka fields
+		Fields(kafka.FranzConnectionFields()...).
+		Fields(kafka.FranzProducerFields()...).
+		Fields(kafka.FranzWriterConfigFields()...).
```
rockwotj: One other thing I noticed is that by default we should be preserving the key, partition, and topic. We should not require everyone setup manual partitioning and all the other things that they expect from a migration. We need to tweak the defaults for the FranzWriter here for this. I'm not sure how we want to do it, but in your above example they should preserve the key, the partition and topic by default, and be able to override the topic for example if they want to remap things.

rockwotj (approved):
The code LGTM, but is not quite ready to replace the old migrator.

We need to preserve keys at the very least and the other options we chatted about over Slack.

internal/impl/redpanda/migrator/migrator.go
mmatczuk: @rockwotj please note [0f080d4](https://github.com/redpanda-data/connect/pull/3564/commits/0f080d426ce822faec4fd4b47deb812e9eb109cd)

mmatczuk:
I changed base to a feature branch and I'm merging this. I'll send followup items to the feature branch.

PR #3566, opened 2025-07-29
Use absolute tools dir

Jeffail:
Updates the tools install tasks to use TOOLS_BIN_DIR, it needed to be made absolute in order to satisfy GOBIN.

### DIFF
--- a/taskfiles/tools.yml
+++ b/taskfiles/tools.yml
@@ -10,17 +10,20 @@ tasks:
     desc: Install golangci-lint
     required:
       vars:
+        - TOOLS_BIN_DIR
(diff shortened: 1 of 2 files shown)


taskfiles/tools.yml
```diff
@@ -10,17 +10,20 @@ tasks:
     desc: Install golangci-lint
     required:
       vars:
+        - TOOLS_BIN_DIR
```
mmatczuk, line 13: I think we could use `env` in this file to avoid both GOBIN and ad-hoc TOOLS_BIN_DIR var requirements

PR #3581, opened 2025-08-08
input_pubsub: slow down connections when topic does not exit

donotnoot:
Without it it seems like it'll try to connect as fast as it can forever. A fun typo will cause the usage in GCP to quickly go through the roof (guess how I know).

### DIFF
--- a/internal/impl/gcp/input_pubsub.go
+++ b/internal/impl/gcp/input_pubsub.go
@@ -17,6 +17,7 @@ package gcp
 import (
 	"context"
 	"errors"
+	"fmt"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


internal/impl/gcp/input_pubsub.go
```diff
@@ -233,6 +234,20 @@ func (c *gcpPubSubReader) Connect(context.Context) error {
 	}
 
 	sub := c.client.Subscription(c.conf.SubscriptionID)
+
+	subExists, err := sub.Exists(context.Background())
```
mmatczuk, line 238: Please use the context provided to Connect.

internal/impl/gcp/input_pubsub.go
```diff
@@ -233,6 +234,20 @@ func (c *gcpPubSubReader) Connect(context.Context) error {
...
+	if err != nil {
+		return service.NewErrBackOff(
+			fmt.Errorf("could not check whether subscription %q exists: %w",
```
mmatczuk, line 241: nit: it would be nicer to avoid "could not" and alike

internal/impl/gcp/input_pubsub.go
```diff
@@ -233,6 +234,20 @@ func (c *gcpPubSubReader) Connect(context.Context) error {
...
+				c.conf.SubscriptionID, err),
+			5*time.Second)
+	}
+	if !subExists {
+		return service.NewErrBackOff(
+			fmt.Errorf("subscription %q does not exist", c.conf.SubscriptionID),
+			5*time.Second)
```
mmatczuk, line 248: Consider extracting to const.

internal/impl/gcp/input_pubsub.go
donotnoot: https://github.com/redpanda-data/connect/blob/96c6efeda3da2d8f7bf3baf4936f23f5de7ab4e0/internal/impl/gcp/input_pubsub.go#L227 FYI

donotnoot:
Closing cause no time and we're finally (!) phasing this library out. Maintainers free to reopen and fix/address comments though!

PR #3582, opened 2025-08-08
logging: ensure topic logger respects configured log level

josephwoodward:


internal/cli/enterprise.go
```diff
@@ -102,6 +103,28 @@ func InitEnterpriseCLI(binaryName, version, dateBuilt string, schema *service.Co
 			// Kick off license service, it's important we do this before chroot and telemetry
 			license.RegisterService(pConf.Resources(), licenseConfig)
 
+			// Now we've parsed config, ensure preconfigured topic logger level matches logger.level
+			cfg := pConf.Namespace("logger")
+			logsLevelStr, err := cfg.FieldString("level")
+			if err != nil {
+				return err
```
Jeffail: I'd maybe suggest we only do an error log on the fallback logger in this case, given it's only a minor observability failure. We wouldn't ever expect to hit this error unless something very funky is going on with custom builds, but I think it's better to be a bit defensive and not reject the entire pipeline.

josephwoodward: Good suggestion. I've also applied this to parsing the log level (see https://github.com/redpanda-data/connect/pull/3582/commits/96bdb9f169da2e254efee88ce47ee1a52c4bf246)

Jeffail (approved):
LGTM

internal/cli/enterprise.go
```diff
@@ -102,6 +103,29 @@ func InitEnterpriseCLI(binaryName, version, dateBuilt string, schema *service.Co
...
+				logsLevel = slog.LevelInfo
+			case "warn":
+				logsLevel = slog.LevelWarn
+			case "error":
+				logsLevel = slog.LevelError
+			default:
+				fbLogger.Errorf("Log level '%s' not recognized, using to default level %s", logsLevelStr, logsLevel)
+			}
+
+			rpMgr.SetTopicLoggerLevel(&logsLevel)
```
mmatczuk, line 127: Consider extracting to function.

PR #3590, opened 2025-08-23
mcp: apply CORS as in gateway input

birdayz:
Currently, CORS is not handled at all in MCP mode.
This PR adds CORS handling, as in gateway input. This is in line with how other features like JWT issuer are handled.

### DIFF
--- a/internal/gateway/cors.go
+++ b/internal/gateway/cors.go
@@ -47,7 +47,8 @@ func (conf CORSConfig) WrapHandler(handler http.Handler) http.Handler {
 	}
 	return handlers.CORS(
 		handlers.AllowedOrigins(conf.allowedOrigins),
-		handlers.AllowedHeaders([]string{"Content-Type", "Authorization"}),
+		handlers.AllowedHeaders([]string{"Content-Type", "Authorization", "Mcp-Session-Id"}),
+		handlers.ExposedHeaders([]string{"Mcp-Session-Id"}),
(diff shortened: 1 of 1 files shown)


internal/mcp/mcp.go
```diff
@@ -158,7 +181,16 @@ func NewServer(
...
 
-	return &Server{s, mux, rpJWT}, nil
+	var cors corsConfig
+	if v := os.Getenv(rpEnvCorsOrigins); v != "" {
+		cors.enabled = true
+		cors.allowedOrigins = strings.Split(v, ",")
+		for i, o := range cors.allowedOrigins {
+			cors.allowedOrigins[i] = strings.TrimSpace(o)
+		}
+	}
```
mmatczuk: This could be extracted to function.

mmatczuk:
Consider extracting cors to separate file.

birdayz:
so it's a shared function across gateway and mcp you mean? or only within mcp package?

mmatczuk:
If it can be shared it would be good.

birdayz:
@mmatczuk i moved it to shared pkg, what do you think?

rockwotj (commented):
LGTM outside the linter failures

birdayz:
@rockwotj i added another header (expose-headers, because response headers are also required, when testing from a browser). could you please take another look? thanks

rockwotj (approved):

PR #3597, opened 2025-08-27
Execute plugin commands rel to config dir

Jeffail:
Currently dynamic plugin commands are executed relative to the directory that RPCN is running. However, this directory will differ depending on the nature of each deployment environment. This PR updates plugins so that these commands are executed relative to the directory that the config itself was found in, which means we can use relative paths in our config and it'll be more portable.


internal/rpcplugin/testdata/catshout/main.go
```diff
@@ -0,0 +1,47 @@
...
+
+// ProcessBatch implements service.BatchProcessor.
+func (p *myProcessor) ProcessBatch(_ context.Context, batch service.MessageBatch) ([]service.MessageBatch, error) {
+	for _, m := range batch {
+		mBytes, err := m.AsBytes()
+		if err != nil {
+			return nil, err
+		}
+		m.SetBytes(slices.Concat(
+			[]byte("MEOW! "),
```
rockwotj, line 36: You'll appreciate this thread: https://redpandadata.slack.com/archives/CPKPE6RU5/p1756306515640069?thread_ts=1756305973.111209&cid=CPKPE6RU5

internal/rpcplugin/config.go
```diff
@@ -222,6 +223,17 @@ func (c *Config) Validate() error {
 	return nil
 }
 
+func (c *Config) setDefaultCWD(cpath string) {
+	configDir := filepath.Dir(cpath)
+	if c.Cwd != "" {
+		if tmpPath, err := filepath.Rel(configDir, cpath); err == nil {
```
rockwotj: shouldn't this be `configDir, c.Cwd`?

Jeffail: Ah good spot, needs more tests 😅

PR #3600, opened 2025-08-27
protobuf: Ignore hidden files to fix duplicate descriptor errors.

### LINKED ISSUE #3138: Feature request for protobuf ignore files that starts with dot
The feature is alrady trying to fix the issue I will describe later, So if there is other way to resolve the problem please do.

When using kafka connect on kubernetes if we add proto files in a configMap and we mount that for example in /proto folder

Kubernetes create symlinks so if there are:
person.proto in /proto folder 
kubernetes created that file as a symlink to /proto/..timestamp_person.proto

so if we use in protobuf processor the import path /proto it fails becasue it complains that person defined in /proto/person.proto is already defined in /proto/..timestamp_person.proto

So I think we could add a filter to ignore files starting with dot
or we can pass specifig list of  files instead of a a dir

Other option is add flag that ignoreSymlinks so if we use /proto import _path by ignoring symlinks it could find the defintions on the ../..timestamp_... protos

do you think in other ways?

Currently I noticed that if I use mounting path as a specifc file and use subpath there is no symlinks but that is not an ideal fix for me.

dubyte:
This pull request resolves an issue where Connect fails to load protocol buffer descriptors when running in a Kubernetes environment using ConfigMaps. It does this by ignoring hidden files (those starting with a .) during the file walk, preventing duplicate descriptor errors.

```go
func loadDescriptors(f fs.FS, importPaths []string) (*protoregistry.Files, *protoregistry.Types, error) {
...
			if filepath.Ext(info.Name()) == ".proto" && !strings.HasPrefix(info.Name(), ".") {
...
}
```

This PR fix #3138

### DIFF
--- a/config/test/protobuf/schema/.envelope_.proto
+++ b/config/test/protobuf/schema/.envelope_.proto
@@ -1,0 +1,1 @@
+envelope.proto
--- a/config/test/protobuf/schema/.house_.proto
+++ b/config/test/protobuf/schema/.house_.proto
@@ -1,0 +1,1 @@
+house.proto
--- a/config/test/protobuf/schema/.person_.proto
+++ b/config/test/protobuf/schema/.person_.proto
@@ -1,0 +1,1 @@
+person.proto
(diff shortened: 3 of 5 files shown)


dubyte:
cc: @mmatczuk

mmatczuk (approved):

PR #3602, opened 2025-08-28
SNS Output: Topic ARN string interpolation

josephwoodward:
This change updates the `topic_arn` field for the SNS output to support string interpolation.

### DIFF
--- a/internal/impl/aws/output_sns.go
+++ b/internal/impl/aws/output_sns.go
@@ -41,7 +41,7 @@ const (
 )
 
 type snsoConfig struct {
-	TopicArn               string
+	TopicArn               *service.InterpolatedString
 	MessageGroupID         *service.InterpolatedString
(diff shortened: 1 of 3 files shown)


internal/impl/aws/output_sns.go
```diff
@@ -218,12 +220,21 @@ func (a *snsWriter) Write(wctx context.Context, msg *service.Message) error {
...
 
+	var topicArn *string
+	if a.conf.TopicArn != nil {
+		topicArnStr, err := a.conf.TopicArn.TryString(msg)
+		if err != nil {
+			return fmt.Errorf("%s interpolation error: %s", snsoFieldTopicARN, err)
+		}
+		topicArn = aws.String(topicArnStr)
+	}
+
```
mmatczuk: Consider refactoring to method.

PR #3605, opened 2025-09-01
(Microsoft SQL Server): Add CDC Support

josephwoodward:
Context helpful for reviewers:

- SQL Server works by capturing changes in table specific change tables (ie `dbo.products` has a corresponding `ctc.dbo_products_CT`). This is different to some other SQL based databases that track all changes in a single changelog table.
- LSNs are unique across the database, multiple changes in a single transaction will result in changes sharing the same LSN at which point the sequence number and command ID can be used to order changes.
- When in streaming mode the SQL Server CDC connector merges each change from all change tables into a single stream ordered by LSN, Sequence Value and Command ID

---

### Features:

- Snapshotting of existing tables (can be enabled/disabled)
- Streaming of change table updates
- Checkpointing of streaming changes only (not snapshotting) 
**Note:** Once a checkpoint is set, snapshotting will be skipped similar to other CDC components, wishing to snapshot new tables will require clearing of cache and snapshotting all tables again)
- Tables can be configured by "include" and "exclude" regex during component start up.
**Note:** This means newly added tables (or tables modified that match the include/exclude pattern) won't be matched until service restart. If the user has performed a snapshot then they'll need to clear their cache to force a new snapshot of all tables.

---

Metrics over an hour with 6 spikes of 10,000,000 records across two tables (also includes initial snapshot of 10,000,000 rows).

image

image

---

Before and after profiling of insertion of 10,000,000 records across two tables (5,000,000 each)

image


internal/impl/mssqlserver/input_mssql_server_cdc.go
```diff
@@ -0,0 +1,385 @@
...
+			Table:         c.table,
+			Data:          c.columns,
+			Operation:     OpType(c.operation),
+			SequenceValue: c.seqVal,
+			CommandID:     c.commandID,
+			StartLSN:      c.startLSN,
+		}
+		return c.startLSN, nil
+	})
+	return fmt.Errorf("streaming from change tables: %w", err)
```
Jeffail: Does this need protecting with`err != nil`?

internal/impl/mssqlserver/input_mssql_server_cdc.go
```diff
@@ -0,0 +1,385 @@
...
+				if err != nil {
+					return fmt.Errorf("flush batch error: %w", err)
+				}
+				if err := r.flushBatch(ctx, r.checkpoint, flushedBatch); err != nil {
+					return fmt.Errorf("failed to flush batch: %w", err)
+				}
+			} else {
+				d, ok := r.batchPolicy.UntilNext()
+				if ok {
+					nextTimedBatchChan = time.After(d)
```
Jeffail: Pretty sure we need an `if nextTimedBatchChan == nil` guard on this to prevent spawning multiple `time.After`'s.

internal/impl/mssqlserver/input_mssql_server_cdc.go
```diff
@@ -0,0 +1,385 @@
...
+				}
+			}
+		}
+	}
+}
+
+func (r *msSqlServerCDCReader) readMessages(ctx context.Context, stream *changeTableStream) error {
+	err := stream.readChangeTables(ctx, r.db, func(c *change) (LSN, error) {
+		// fmt.Printf("LSN=%x, CommandID=%d, SeqVal=%x, op=%d table=%s cols=%d\n", c.startLSN, c.commandID, c.seqVal, c.operation, c.table, len(c.columns))
+		r.rawMessageEvents <- MessageEvent{
```
Jeffail: It's probably worth wrapping this in a select that detects shutdown and allows escape.

Jeffail (commented):
Looking good, just a couple small bits.

internal/impl/mssqlserver/input_mssql_server_cdc.go
josephwoodward: Strictly speaking, no it doesn't or at least not yet (the function _currently_ doesn't return a nil error). Adding a nil check causes the linter to shout at me and require a number of no lint static check rules to please it. I'll most likely add the lint exclusions and wrap it in a nil check but just waiting to see what the final read messages function looks like.

internal/impl/mssqlserver/input_mssql_server_cdc.go
```diff
@@ -0,0 +1,388 @@
...
+	fieldSnapshotMaxBatchSize = "snapshot_max_batch_size"
+	fieldCheckpointLimit      = "checkpoint_limit"
+	fieldTables               = "tables"
+	fieldCheckpointCache      = "checkpoint_cache"
+	fieldCheckpointKey        = "checkpoint_key"
+	fieldBatching             = "batching"
+)
+
+func init() {
+	service.MustRegisterBatchInput("mssql_server_cdc", mssqlStreamConfigSpec, newMssqlCDCReader)
```
mmatczuk: Could we maybe unify naming a bit atm we have 
-  mssqlStream
- MssqlCDC
- msSqlServerCDC

docs/modules/components/pages/inputs/mssql_server_cdc.adoc
```diff
@@ -0,0 +1,272 @@
...
+
+=== `stream_snapshot`
+
+If set to true, the connector will query all the existing data as a part of snapshot process. Otherwise, it will start from the current Log Sequence Number position.
+
+
+*Type*: `bool`
+
+
+=== `snapshot_max_batch_size`
```
mmatczuk, line 126: Could we align on the batch size? Say we'd take `max(batch size, 1k)` to avoid reading single row.

internal/impl/mssqlserver/input_mssql_server_cdc.go
```diff
@@ -0,0 +1,388 @@
...
+		Default("mssql_cdc_position"),
+	).
+	Field(service.NewIntField(fieldCheckpointLimit).
+		Description("The maximum number of messages that can be processed at a given time. Increasing this limit enables parallel processing and batching at the output level. Any given Log Sequence Number (LSN) will not be acknowledged unless all messages under that offset are delivered in order to preserve at least once delivery guarantees.").
+		Default(1024),
+	).
+	Field(service.NewAutoRetryNacksToggleField()).
+	Field(service.NewBatchPolicyField(fieldBatching))
+
+type asyncMessage struct {
```
mmatczuk: @Jeffail this is defined in almost every package, maybe we could promote it to benthos service pkg WDYT?

internal/impl/mssqlserver/input_mssql_server_cdc.go
```diff
@@ -0,0 +1,388 @@
...
+
+var mssqlStreamConfigSpec = service.NewConfigSpec().
+	Beta().
+	Categories("Services").
+	Version("4.45.0").
+	Summary("Creates an input that consumes from a Microsoft SQL Server's change log.").
+	Description(``).
+	Field(service.NewStringField(fieldConnectionString).
+		Description("The connection string of the Microsoft SQL Server database to connect to.").
+		Example("sqlserver://username:password@host/instance?param1=value&param2=value"),
```
mmatczuk: Super smart to provide example here.

internal/impl/mssqlserver/input_mssql_server_cdc.go
```diff
@@ -0,0 +1,388 @@
...
+type msSqlServerCDCReader struct {
+	connectionString     string
+	streamSnapshot       bool
+	snapshotMaxBatchSize int
+	tables               []string
+	lsnCache             string
+	lsnCacheKey          string
+	batching             service.BatchPolicy
+	batchPolicy          *service.Batcher
+	checkpoint           *checkpoint.Capped[LSN]
```
mmatczuk: Consider extracting that to dedicated config type. It's not needed during execution.

internal/impl/mssqlserver/input_mssql_server_cdc.go
```diff
@@ -0,0 +1,388 @@
...
+
+	logger *service.Logger
+	res    *service.Resources
+
+	dbMu             sync.Mutex
```
mmatczuk: db is not thread safe?

mmatczuk: This is only used sync db field when writing in Connect and reading in Close. I doubt it serves any purpose the way it's used now. 

We can either ignore it and assume the framework will take care of synchronization of Connect, ReadBatch, Close or use atomic.Pointer.

internal/impl/mssqlserver/input_mssql_server_cdc.go
```diff
@@ -0,0 +1,388 @@
...
+	streamSnapshot       bool
+	snapshotMaxBatchSize int
+	tables               []string
+	lsnCache             string
+	lsnCacheKey          string
+	batching             service.BatchPolicy
+	batchPolicy          *service.Batcher
+	checkpoint           *checkpoint.Capped[LSN]
+
+	logger *service.Logger
```
mmatczuk: nit: IMO log is more ergonomic name when using it

internal/impl/mssqlserver/input_mssql_server_cdc.go
```diff
@@ -0,0 +1,388 @@
...
+	if reader.snapshotMaxBatchSize, err = conf.FieldInt(fieldSnapshotMaxBatchSize); err != nil {
+		return nil, err
+	}
+	var checkpointLimit int
+	if checkpointLimit, err = conf.FieldInt(fieldCheckpointLimit); err != nil {
+		return nil, err
+	}
+	reader.checkpoint = checkpoint.NewCapped[LSN](int64(checkpointLimit))
+
+	// TODO: support regular expression on tablenames
```
mmatczuk: I added regex helper in https://github.com/redpanda-data/connect/pull/3564/commits/7fd2bde9d15d355bbc2b7e89a02d64782dc24686
maybe it makes sense to promote it to shared package.

josephwoodward: Great, thanks for making me aware of this 👍

internal/impl/mssqlserver/input_mssql_server_cdc.go
mmatczuk: Also, it would be good to unify the name in the logs/error messages

mmatczuk:
Could we rename package to mssql to match Xsql used for other databases.

internal/impl/mssqlserver/input_mssql_server_cdc.go
```diff
@@ -0,0 +1,388 @@
...
+	batchPolicy          *service.Batcher
+	checkpoint           *checkpoint.Capped[LSN]
+
+	logger *service.Logger
+	res    *service.Resources
+
+	dbMu             sync.Mutex
+	db               *sql.DB
+	msgChan          chan asyncMessage
+	rawMessageEvents chan MessageEvent
```
mmatczuk: Could we remove this channel and move batch creation to a dedicated type / factory function that you would feed to readChangeTables.

internal/impl/mssqlserver/input_mssql_server_cdc.go
```diff
@@ -0,0 +1,388 @@
...
+	stopSig          *shutdown.Signaller
+}
+
+func newMssqlCDCReader(conf *service.ParsedConfig, res *service.Resources) (s service.BatchInput, err error) {
+	if err := license.CheckRunningEnterprise(res); err != nil {
+		return nil, err
+	}
+
+	reader := msSqlServerCDCReader{
```
mmatczuk: Idiomatically would `r` :)

internal/impl/mssqlserver/input_mssql_server_cdc.go
```diff
@@ -0,0 +1,388 @@
...
+
+	if err := ctStream.verifyChangeTables(ctx, r.db, r.tables); err != nil {
+		return fmt.Errorf("verifying MS MSQL Server change tables: %s", err)
+	}
+
+	go func() {
+		ctx, _ = r.stopSig.SoftStopCtx(context.Background())
+		wg, ctx := errgroup.WithContext(ctx)
+		wg.Go(func() error { return r.readMessages(ctx, ctStream) })
+		wg.Go(func() error { return r.batchMessages(ctx) })
```
mmatczuk: See my comment in L102

internal/impl/mssqlserver/input_mssql_server_cdc.go
```diff
@@ -0,0 +1,388 @@
...
+			CommandID:     c.commandID,
+			StartLSN:      c.startLSN,
+		}:
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		}
+		return c.startLSN, nil
+	})
+
+	// if err != nil { //nolint:staticcheck
```
mmatczuk: Dead code.

internal/impl/mssqlserver/input_mssql_server_cdc.go
josephwoodward: I agree, though I noticed the other CDC inputs (MySQL and PostgreSQL at least) use `logger` so I decided to follow suit for consistency.

internal/impl/mssqlserver/input_mssql_server_cdc.go
```diff
@@ -0,0 +1,388 @@
...
+
+func (r *msSqlServerCDCReader) ReadBatch(ctx context.Context) (service.MessageBatch, service.AckFunc, error) {
+	select {
+	case m := <-r.msgChan:
+		return m.msg, m.ackFn, nil
+	case <-r.stopSig.HasStoppedChan():
+		return nil, nil, service.ErrNotConnected
+	case <-ctx.Done():
+	}
+	return nil, nil, ctx.Err()
```
mmatczuk: For clarity you could return context error early.

internal/impl/mssqlserver/integration_test.go
```diff
@@ -0,0 +1,492 @@
...
+		IF NOT EXISTS (SELECT 1 FROM sys.tables WHERE name = '%s' AND schema_id = SCHEMA_ID('dbo'))
+		BEGIN
+			%s
+			%s
+		END;`, tableName, query, cdc)
+	db.exec(q)
+}
+
+func setupTestWithMSSQLServerVersion(t *testing.T, version string) (string, *testDB) {
+	t.Parallel()
```
mmatczuk: This does not belong here.

internal/impl/mssqlserver/input_mssql_server_cdc.go
josephwoodward: Yeah, this is an artefact of when I first started writing the input and synchronising access to the connection for CDC purposes and snapshotting was still unknown. I've left it there for the time being but agree that it's not going to be needed.

internal/impl/mssqlserver/input_mssql_server_cdc.go
josephwoodward: This is more a reminder for [this](https://github.com/redpanda-data/connect/pull/3605#discussion_r2336157649) conversation earlier.

internal/impl/mssqlserver/integration_test.go
```diff
@@ -0,0 +1,492 @@
...
+			IF NOT EXISTS (SELECT name FROM sys.databases WHERE name = N'testdb')
+			BEGIN
+				CREATE DATABASE testdb;
+			END;`)
+		if err != nil {
+			return err
+		}
+		db.Close()
+
+		connectionString = fmt.Sprintf("sqlserver://sa:YourStrong!Passw0rd@localhost:%s?database=%s&encrypt=disable", port, "testdb")
```
mmatczuk, line 813: Would be nice to provide explanations what happens here and why

internal/impl/mssqlserver/integration_test.go
```diff
@@ -0,0 +1,492 @@
+// Copyright 2025 Redpanda Data, Inc.
+//
+// Licensed as a Redpanda Enterprise file under the Redpanda Community
+// License (the "License"); you may not use this file except in compliance with
+// the License. You may obtain a copy of the License at
+//
+// https://github.com/redpanda-data/connect/blob/main/licenses/rcl.md
+
+package mssqlserver
```
mmatczuk: Typically ITs run in the _test package.

internal/impl/mssqlserver/integration_test.go
```diff
@@ -0,0 +1,492 @@
...
+	// 		FROM sys.all_objects a
+	// 		CROSS JOIN sys.all_objects b
+	// 	)
+	// 	INSERT INTO products (id, name)
+	// 	SELECT n, CONCAT('product-', n) FROM Numbers;
+	// `)
+	// require.NoError(t, err)
+}
+
+func TestIntegration_MSSQLServerCDC(t *testing.T) {
```
mmatczuk: Remember to add 	integration.CheckSkip(t)

internal/impl/mssqlserver/integration_test.go
josephwoodward: Yeah, though I don't see that convention followed in the Connect repo. Not sure if this is something we want to start encourgaging?

internal/impl/mssqlserver/stream.go
```diff
@@ -0,0 +1,312 @@
...
+		table:   ct.table,
+		columns: make(map[string]any, len(ct.cols)),
+	}
+	for i, c := range ct.cols {
+		v := *(vals[i].(*any))
+		switch c {
+		case "__$start_lsn":
+			if b, ok := v.([]byte); ok {
+				ch.startLSN = b
+			}
```
mmatczuk, line 185: Log ERRORs when it's not ok.

internal/impl/mssqlserver/stream.go
```diff
@@ -0,0 +1,312 @@
...
+	seqVal     []byte
+	commandID  int
+	columns    map[string]any
+}
+
+type changeTableRowIter struct {
+	table   string
+	rows    *sql.Rows
+	cols    []string
+	current *change
```
mmatczuk, line 89: You could not use pointer here and make `valsToChange` accept out param *change where you would fill the same memory for every row.

internal/impl/mssqlserver/stream.go
```diff
@@ -0,0 +1,312 @@
...
+	startLSN   LSN // varbinary(10)
+	endLSN     LSN // varbinary(10)
+	operation  int // 1=delete, 2=insert, 3=update (before), 4=update (after), 5=merge
+	updateMask []byte
+	seqVal     []byte
+	commandID  int
+	columns    map[string]any
+}
+
+type changeTableRowIter struct {
```
mmatczuk, line 85: Comment would be nice.

internal/impl/mssqlserver/stream.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	if len(r.trackedTables) != len(configTables) {
+		return errors.New("could not find all change tables")
+	}
+
+	return nil
+}
+
+func (r *changeTableStream) readChangeTables(ctx context.Context, db *sql.DB, handle func(c *change) (LSN, error)) error {
```
mmatczuk: I'd change that to func(change) to be more defensive. Why you want to return LSN?

internal/impl/mssqlserver/input_mssql_server_cdc.go
josephwoodward: Yeah, would be really good to get rid of one of these. I was going to look at revisiting this once the snapshotting is complete and I have a clearer picture of what propagating data between CDC and Snapshotting would look like.

internal/impl/mssqlserver/input_mssql_server_cdc.go
josephwoodward: Renamed type to input and went with `i` to follow consistency of others (MySQLCDC and PostgreSQL) https://github.com/redpanda-data/connect/pull/3605/commits/a770b91c0fd6b7b5006e28676d7b6db7b73dd782

internal/impl/mssqlserver/integration_test.go
josephwoodward: Done in https://github.com/redpanda-data/connect/pull/3605/commits/f8e2ad0c058764108f6a1649a716c0fd3edb9bd5

internal/impl/mssqlserver/input_mssql_server_cdc.go
josephwoodward: Great catch, done in https://github.com/redpanda-data/connect/pull/3605/commits/0d1defedb21c791dc623e060dd308fc2a5f6fc34

internal/impl/mssqlserver/integration_test.go
josephwoodward: Removed in https://github.com/redpanda-data/connect/pull/3605/commits/f8e2ad0c058764108f6a1649a716c0fd3edb9bd5

internal/impl/mssqlserver/integration_test.go
josephwoodward: Done

internal/impl/mssqlserver/input_mssql_server_cdc.go
josephwoodward: Done.

internal/impl/mssqlserver/integration_test.go
josephwoodward: Done

internal/impl/mssqlserver/stream.go
josephwoodward: The function has since been removed so marking this as resolved.

internal/impl/mssqlserver/input_mssql_server_cdc.go
josephwoodward: This has now been removed.

internal/impl/mssqlserver/input_mssql_server_cdc.go
josephwoodward: This has since been improved in the latest version.

internal/impl/mssqlserver/input_mssql_server_cdc.go
josephwoodward: This has since been reworked so marking as resolved.

docs/modules/components/pages/inputs/mssql_server_cdc.adoc
josephwoodward: The default is now set to 1000.

internal/impl/mssqlserver/input_mssql_server_cdc.go
josephwoodward: I've changed it to log as it does read much nicer when logging (ie "log warn" instead of "logging warn".

internal/impl/mssqlserver/input_mssql_server_cdc.go
josephwoodward: This has been done.

internal/impl/mssqlserver/replication/snapshot.go
```diff
@@ -0,0 +1,341 @@
...
+	tx *sql.Tx
+
+	snapshotConn *sql.Conn
+
+	Tables    []UserTable
+	publisher ChangePublisher
+	log       *service.Logger
+}
+
+// NewSnapshot creates a new instance of Snapshot capable of snapshotting provided tables.
```
mmatczuk, line 37: We generally try to wrap text at 80 it's easier to read

internal/impl/mssqlserver/replication/snapshot.go
```diff
@@ -0,0 +1,341 @@
...
+)
+
+// Snapshot is responsible for creating snapshots of existing tables based on the Tables configuration value.
+type Snapshot struct {
+	db *sql.DB
+	tx *sql.Tx
+
+	snapshotConn *sql.Conn
+
+	Tables    []UserTable
```
mmatczuk: Why Tables are exported while other fields are not?

internal/impl/mssqlserver/replication/snapshot.go
```diff
@@ -0,0 +1,341 @@
...
+	if s.snapshotConn, err = s.db.Conn(ctx); err != nil {
+		return nil, fmt.Errorf("creating snapshot connection: %v", err)
+	}
+
+	// Use context.Background() because we want the Tx to be long lived, we explicitly close it in the close method
+	if s.tx, err = s.snapshotConn.BeginTx(context.Background(), &sql.TxOptions{Isolation: sql.LevelSnapshot}); err != nil {
+		return nil, fmt.Errorf("starting snapshot transaction: %v", err)
+	}
+
+	var toLSN LSN
```
mmatczuk: could this be maxLSN

internal/impl/mssqlserver/replication/snapshot.go
```diff
@@ -0,0 +1,341 @@
...
+		var numRowsProcessed int
+
+		s.log.Infof("Beginning snapshot process for table '%s'", table.FullName())
+		for {
+			var batchRows *sql.Rows
+			if numRowsProcessed == 0 {
+				batchRows, err = s.querySnapshotTable(ctx, table, tablePks, nil, maxBatchSize)
+			} else {
+				batchRows, err = s.querySnapshotTable(ctx, table, tablePks, &lastSeenPksValues, maxBatchSize)
+			}
```
mmatczuk, line 118: Why we need distinct calls here?

internal/impl/mssqlserver/replication/snapshot.go
```diff
@@ -0,0 +1,341 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to execute snapshot table query: %s", err)
+			}
+
+			types, err := batchRows.ColumnTypes()
+			if err != nil {
+				return fmt.Errorf("failed to fetch column types: %w", err)
+			}
+
+			values, mappers := prepSnapshotScannerAndMappers(types)
```
mmatczuk, line 128: Is it really scanner that is returned?

mmatczuk: Seems like it's more a values template

internal/impl/mssqlserver/replication/snapshot.go
```diff
@@ -0,0 +1,341 @@
...
+		return nil, fmt.Errorf("discovering primary keys for table '%s': %w", table.FullName(), err)
+	}
+
+	if len(pks) == 0 {
+		return nil, fmt.Errorf("unable to find primary key for table '%s' - does the table exist and does it have a primary key set?", table.FullName())
+	}
+	return pks, nil
+}
+
+func (s *Snapshot) querySnapshotTable(ctx context.Context, table UserTable, pk []string, lastSeenPkVal *map[string]any, limit int) (*sql.Rows, error) {
```
mmatczuk: Try declaring parameter per line it there is too many

mmatczuk: Sometimes it's more handy to declare a helper type to avoid passing map pointers

internal/impl/mssqlserver/replication/snapshot.go
```diff
@@ -0,0 +1,341 @@
...
+	snapshotQueryParts := []string{
```
mmatczuk, line 224: Does this driver support named parameters, this is prone to SQL injection.

internal/impl/mssqlserver/replication/snapshot.go
```diff
@@ -0,0 +1,341 @@
...
+				break
+			}
+		}
+		s.log.Infof("Completed snapshot process for table '%s'", table.FullName())
+	}
+	return nil
+}
+
+func (s *Snapshot) getTablePrimaryKeys(ctx context.Context, table UserTable) ([]string, error) {
+	pkSql := `
```
mmatczuk: I believe it should be pkSQL like ID not Id

internal/impl/mssqlserver/replication/stream.go
```diff
@@ -0,0 +1,411 @@
...
+	item := old[n-1]
+	*h = old[:n-1]
+	return item
+}
+
+// change represents a logical change row from the change table.
+type change struct {
+	startLSN   LSN // varbinary(10)
+	endLSN     LSN // varbinary(10)
+	operation  int // 1=delete, 2=insert, 3=update (before), 4=update (after)
```
mmatczuk: Maybe a type would be helpful.

internal/impl/mssqlserver/replication/stream.go
```diff
@@ -0,0 +1,411 @@
...
+	rows    *sql.Rows
+	cols    []string
+	current *change
+	log     *service.Logger
+
+	vals []any
+}
+
+// newChangeTableRowIter returns an custom row iterator for the given changeTable.
+func newChangeTableRowIter(ctx context.Context, db *sql.DB, changeTable UserTable, fromLSN, toLSN LSN, logger *service.Logger) (*changeTableRowIter, error) {
```
mmatczuk: Another candidate for parameter per line.

internal/impl/mssqlserver/replication/stream.go
```diff
@@ -0,0 +1,411 @@
...
+	// Note: LSN is varbinary type so can sort correctly for LSNs
+	// Inspired by Debezium https://github.com/debezium/debezium/blob/main/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java?plain=1#L177
```
mmatczuk, line 104: Maybe use permalink, this points to empty line.

internal/impl/mssqlserver/replication/snapshot.go
```diff
@@ -0,0 +1,341 @@
...
+
+	if lastSeenPkVal == nil {
+		snapshotQueryParts = append(snapshotQueryParts, buildOrderByClause(pk))
+
+		q := strings.Join(snapshotQueryParts, " ")
+		return s.tx.QueryContext(ctx, q)
+	}
+
+	var lastSeenPkVals []any
+	var placeholders []string
```
mmatczuk: nit: maybe a var group

internal/impl/mssqlserver/replication/stream.go
mmatczuk: I'd suggest to add relevant chunks of docs here, I think this comment shall be extended for clarity.

internal/impl/mssqlserver/replication/stream.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+
+	// pre-allocate slice of pointers for sql.Scan operations
+	vals := make([]any, len(cols))
+	for i := range vals {
+		var v any
+		vals[i] = &v
+	}
+
+	itor := &changeTableRowIter{
```
mmatczuk: Is that a typo or intentional?

internal/impl/mssqlserver/replication/stream.go
```diff
@@ -0,0 +1,411 @@
...
+		table: changeTable,
+		rows:  rows,
+		cols:  cols,
+		vals:  vals,
+		log:   logger,
+	}
+	// Prime the iterator by loading the first row
+	if err := itor.next(); err != nil {
+		// Already exhausted iterator
```
mmatczuk, line 136: It can be a read error can't it?

internal/impl/mssqlserver/replication/stream.go
```diff
@@ -0,0 +1,411 @@
...
+		}
+		return sql.ErrNoRows
+	}
+
+	// read row into ct.vals, reusing pre-allocated slice of pointer
+	if err := ct.rows.Scan(ct.vals...); err != nil {
+		return err
+	}
+
+	if ct.current == nil {
```
mmatczuk, line 158: Could this be allocated in new? Also it can be a value not pointer it seems.

internal/impl/mssqlserver/replication/stream.go
```diff
@@ -0,0 +1,411 @@
...
+
+	return nil
+}
+
+func (ct *changeTableRowIter) Close() error {
+	return ct.rows.Close()
+}
+
+// mapValsToChange maps the values from vals to the dst out parameter.
+func (ct *changeTableRowIter) mapValsToChange(vals []any, dst *change) error {
```
mmatczuk, line 176: Could this be defined on change? ex. initFromVals

internal/impl/mssqlserver/replication/stream.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+	return s
+}
+
+// ReadChangeTables streams the change events from the configured SQL Server change tables.
+func (r *ChangeTableStream) ReadChangeTables(ctx context.Context, db *sql.DB, startPos LSN) error {
+	r.log.Infof("Starting streaming %d change table(s)", len(r.tables))
+	var (
+		fromLSN LSN // load last checkpoint; nil means start from beginning in tables
+		toLSN   LSN // often set to fn_cdc_get_max_lsn(); nil means no upper bound
```
mmatczuk: We have some naming discrepancy, could we use startLSN and endLSN here too?

internal/impl/mssqlserver/replication/stream.go
```diff
@@ -0,0 +1,411 @@
...
+		lastLSN LSN
+	)
+
+	if len(startPos) != 0 {
+		fromLSN = startPos
+		lastLSN = startPos
+		r.log.Debugf("Resuming from recorded LSN position '%s'", startPos)
```
mmatczuk: Maybe Info

internal/impl/mssqlserver/replication/stream.go
```diff
@@ -0,0 +1,411 @@
...
+					_ = it.Close()
+				}
+				return err
+			} else {
+				// next page
+				lastLSN = cur.startLSN
+			}
+
+			// Advance the iterator and push back on heap to be sorted
+			if err := item.iter.next(); err != nil {
```
mmatczuk, line 335: Maybe next() should return bool if it's only used to check end of data?

internal/impl/mssqlserver/replication/stream.go
```diff
@@ -0,0 +1,411 @@
...
+				heap.Push(h, item)
+			}
+		}
+
+		if len(lastLSN) != 0 {
+			if !bytes.Equal(fromLSN, lastLSN) {
+				fromLSN = lastLSN
+			} else {
+				r.log.Debug("No more changes across all change tables, backing off...")
+				time.Sleep(backoffDuration)
```
mmatczuk: How about making it a config field.

internal/impl/mssqlserver/replication/snapshot.go
josephwoodward: `Tables` was exported as the SQL Server input was using it to log the total number of tables to be snapshotted, however I've now moved that log output _into_ the snapshot where it makes more sense and also allowed me to unexport the field.

Addressed in https://github.com/redpanda-data/connect/pull/3605/commits/7fe74ce7f01f860bb4200f5da5c478676de6e6fc

internal/impl/mssqlserver/replication/snapshot.go
josephwoodward: Updated in latest change.

internal/impl/mssqlserver/replication/snapshot.go
josephwoodward: Done in latest commit (6743111f2)

internal/impl/mssqlserver/replication/stream.go
josephwoodward: Updated in latest commit (6743111f2)

internal/impl/mssqlserver/replication/stream.go
josephwoodward: Good idea 👍

internal/impl/mssqlserver/replication/stream.go
josephwoodward: Updated in f0fb4b859

internal/impl/mssqlserver/replication/stream.go
josephwoodward: Updated in f0fb4b859

internal/impl/mssqlserver/replication/snapshot.go
josephwoodward: Updated in latest set of changes.

internal/impl/mssqlserver/replication/snapshot.go
josephwoodward: Parameters have been split per line. As for the map, it doesn't need to be a pointer type so  I've removed that.

internal/impl/mssqlserver/replication/stream.go
josephwoodward: Intentional (as an abbreviation of iterator). Changed it to `iter` in recent change though.

internal/impl/mssqlserver/replication/snapshot.go
josephwoodward: The items in the `values` slice all implement the `sql.Scan` interface which is the reason for the `Scanner` part of the function name.

internal/impl/mssqlserver/replication/stream.go
josephwoodward: Updated to use the `OpType` type which makes things clearer: https://github.com/redpanda-data/connect/pull/3605/commits/387bec9ead5b20ce509241f1e9a4357603ec85ae

internal/impl/mssqlserver/replication/stream.go
josephwoodward: Was on the fence about this but probably not a bad idea for tables that are quieter. I've given the config a default value. f8891acc

internal/impl/mssqlserver/replication/snapshot.go
josephwoodward: It does, however unfortunately you can't use named parameters for table names. This data has been sanitised during table validation [here](https://github.com/redpanda-data/connect/pull/3605/files#diff-08a7edae20de1ff8226a3cde8a610c581447ea2fb938f01efcab8e94f5cab516R390-R398) and is driven by the underlying table name as opposed to user input.

internal/impl/mssqlserver/replication/snapshot.go
josephwoodward: Perhaps we should look at the lll linter to force this?

Updated in this commit https://github.com/redpanda-data/connect/pull/3605/commits/387948a894c17763c0ddec4649c3f01a5f34ecf2

Jeffail (approved):

internal/impl/mssqlserver/replication/stream.go
josephwoodward: I considered making it a value type but given the data it contains its actually cheaper to remain a pointer than passing by value.

mmatczuk: Fair.

mmatczuk (approved):

PR #3606, opened 2025-09-01
RPC connector for TigerBeetle

batiati:
Implements an input connector for streaming [TigerBeetle](https://tigerbeetle.com/) events via CDC.

The CDC process is stateless, so this connector requires a [`cache_resource`](https://docs.redpanda.com/redpanda-connect/components/caches/about) to persist progress and resume from the last acknowledged event when restarted.

- Includes documentation and integration tests.  
- Licensing information in the file header is pending.

For more information about TigerBeetle CDC, see https://docs.tigerbeetle.com/operating/cdc/

### DIFF
--- a/internal/impl/tigerbeetle/input_tigerbeetle.go
+++ b/internal/impl/tigerbeetle/input_tigerbeetle.go
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build cgo
+
 package tigerbeetle
 
 import (
--- a/internal/impl/tigerbeetle/config_test.go
+++ b/internal/impl/tigerbeetle/config_test.go
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build cgo
+
 package tigerbeetle
 
 import (
--- a/internal/impl/tigerbeetle/integration_test.go
+++ b/internal/impl/tigerbeetle/integration_test.go
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build cgo
+
 package tigerbeetle
 
 import (
--- a/public/components/tigerbeetle/cgo.go
+++ b/public/components/tigerbeetle/cgo.go
@@ -12,4 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build cgo
+
 package tigerbeetle
+
+import (
+	// Bring in the internal plugin definitions.
+	_ "github.com/redpanda-data/connect/v4/internal/impl/tigerbeetle"
+)


internal/impl/tigerbeetle/input_tigerbeetle.go
```diff
@@ -0,0 +1,629 @@
...
+			Flags:          0,
+			Timestamp:      "1745328270103401031",
+		},
+	}, "", "  ")
+	assertTrue(err == nil, "invalid JSON object")
+
+	return service.NewConfigSpec().
+		Beta().
+		Categories("Services").
+		Version("0.0.1").
```
batiati, line 103: I suppose this version should be defined only once we know the release, right?

public/components/cloud/package.go
batiati, line 1: TODO: Decide whether it will be included in the `cloud` bundle.

internal/impl/tigerbeetle/input_tigerbeetle.go
```diff
@@ -0,0 +1,629 @@
...
+	case batchedMessage := <-input.consumerChan:
+		return batchedMessage.batch, batchedMessage.ackFunc, nil
+	case <-input.stopSignaller.HasStoppedChan():
+		return nil, nil, service.ErrNotConnected
+	case <-ctx.Done():
+	}
+	return nil, nil, ctx.Err()
+}
+
+func newTigerbeetleInput(config *service.ParsedConfig, resources *service.Resources) (s service.BatchInput, err error) {
```
batiati, line 284: TODO: Decide whether to check the enterprise license.

```suggestion
func newTigerbeetleInput(config *service.ParsedConfig, resources *service.Resources) (s service.BatchInput, err error) {
	if err := license.CheckRunningEnterprise(resources); err != nil {
		return nil, err
	}

```

internal/impl/tigerbeetle/input_tigerbeetle.go
Jeffail: Yeah, we're actually deprecating this method as we update the docs manually these days.

internal/impl/tigerbeetle/input_tigerbeetle.go
Jeffail: Thanks for setting this up. I believe we're happy to go ahead with a FOSS license so no need for license checks 👍

internal/impl/tigerbeetle/input_tigerbeetle.go
```diff
@@ -0,0 +1,629 @@
...
+	}
+	if len(addresses) == 0 {
+		return nil, fmt.Errorf("invalid config: %s is empty", fieldAddresses)
+	}
+
+	if progressCache, err = config.FieldString(fieldProgressCache); err != nil {
+		return nil, err
+	}
+	if !config.Resources().HasCache(progressCache) {
+		return nil, fmt.Errorf("cache resource '%s' not defined", progressCache)
```
Jeffail:
```suggestion
		return nil, fmt.Errorf("cache resource '%s' not found", progressCache)
```

internal/impl/tigerbeetle/input_tigerbeetle.go
```diff
@@ -0,0 +1,629 @@
...
+	}
+
+	if eventCountMax, err = config.FieldInt(fieldEventCountMax); err != nil {
+		return nil, err
+	} else if eventCountMax <= 0 {
+		return nil, fmt.Errorf("property '%s' must be greater than zero", fieldEventCountMax)
```
Jeffail, line 330: Checks like this can be added to the field definition as a `LintRule`. You'd still need to have a check here just in case the linter is disabled, but if you're willing to add a lint rule it helps with other tools that guard against invalid configs before deployment.

internal/impl/tigerbeetle/input_tigerbeetle.go
```diff
@@ -0,0 +1,629 @@
...
+	}
+
+	if idleInterval, err = config.FieldInt(fieldIdleInterval); err != nil {
+		return nil, err
+	} else if idleInterval <= 0 {
+		return nil, fmt.Errorf("property '%s' must be greater than zero", fieldIdleInterval)
```
Jeffail, line 336: Same as above (and below)

internal/impl/tigerbeetle/input_tigerbeetle.go
```diff
@@ -0,0 +1,629 @@
...
+	}); err != nil {
+		return fmt.Errorf("unable to access cache for writing: %w", err)
+	}
+	if cErr != nil {
+		return fmt.Errorf("unable to persist the last timestamp to cache:: %w", cErr)
+	}
+	return nil
+}
+
+func assertTrue(ok bool, message string) {
```
Jeffail: I think asserts during init such as the json marshaller for the config spec is fine, but ideally for functions executing at normal runtime we should be avoiding panics where possible.

batiati:
Thanks for the review @Jeffail!

The `LintRule`s are really cool! I've implemented the validations and added unit tests for checking valid and invalid configurations.

Also, I replaced assertions that would `panic` with errors that will cause the connector to shut down.

Please let me know if I should add the license disclaimer to the source files. I didn't want to just copy from other files, as most say `Copyright 2024 ...`.

internal/impl/tigerbeetle/input_tigerbeetle.go
batiati: The same for `Beta()`, when should we remove it?

internal/impl/tigerbeetle/input_tigerbeetle.go
```diff
@@ -0,0 +1,646 @@
...
+		consumerChan: make(chan batchedMesssage, 1),
+		logger:       logger,
+		resources:    resources,
+	}
+
+	return service.AutoRetryNacksBatchedToggled(config, input)
+}
+
+// Extracts events from TigerBeetle.
+func (input *tigerbeetleInput) produce(ctx context.Context, client tb.Client, timestampLast uint64) error {
```
mmatczuk, line 368: nit: how about using short variable name for self?

batiati: I guess it would usually be named just `i` in Go!

I ended up using the [TigerStyle](https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/TIGER_STYLE.md#naming-things) coding style, favoring short but meaningful names and avoiding single-letter variables and abbreviations (except for trivial counters/indexes and well-known abbreviations like `ctx`, `config`, etc).  

Although TigerStyle was originally designed for another programming language (Zig), I think we can apply most of the rules as long as they don't conflict with Go's idiomatic conventions. I don't want this code to feel alien 👽 compared with the rest of the code base!

It's your call, I'd be glad to follow your coding style otherwise.

internal/impl/tigerbeetle/input_tigerbeetle.go
```diff
@@ -0,0 +1,646 @@
...
+	resources     *service.Resources
+}
+
+type batchedMesssage struct {
+	batch   []*service.Message
+	ackFunc service.AckFunc
+}
+
+func init() {
+	service.MustRegisterBatchInput("tigerbeetle", configSpec(), newTigerbeetleInput)
```
josephwoodward: Great work!

Just one small observation. We seem to have an emerging convention of having `cdc` in the suffix of CDC  specific components to help discoverability. See the below screenshot from [here](https://docs.redpanda.com/redpanda-connect/components/about/?search=cdc&type=input%2Coutput%2Cprocessor%2Cscanner%2Cmetric%2Ccache%2Ctracer%2Crate_limit%2Cbuffer&support=certified%2Ccommunity&cloud=yes%2Cno&enterprise=yes%2Cno), so perhaps `tigerbeetle_cdc` to follow that convention?

Image

david-yu:
Hi @batiati Is this PR ready for another re-review?

batiati:
> Hi @batiati Is this PR ready for another re-review? 

Yes, it is ready for review. I've addressed all comments so far.

batiati:
CI is failing due to `CGO_ENABLED: 0`.

The TigerBeetle client requires CGO to consume the static library `lib_tbclient` (the underlying implementation of the VSR client). It includes pre-built libraries in-repo for all supported platforms.

But it seems there's a policy on builds that require external dependencies and CGO, like the ZeroMQ connector:

> By default Redpanda Connect does not build with components that require linking to external libraries. If you wish to build Redpanda Connect locally with this component then set the build tag x_benthos_extra:

internal/impl/tigerbeetle/input_tigerbeetle.go
```diff
@@ -0,0 +1,701 @@
...
+			} else if len(results) > int(input.config.eventCountMax) {
+				return errors.New("assertion failed: too many results")
+			} else if len(batch) != 0 {
+				return errors.New("assertion failed: pending messages to flush")
+			}
+
+			for _, result := range results {
+				bytes, err := jsonSerialize(result)
+				if err != nil {
+					input.logger.Errorf("Unable to serialize as JSON: %w", err)
```
josephwoodward: This should be %s as we're logging. That said, given the code is returning an error below which is eventually logged as part of the errorgroup, perhaps returning a wrapped error with the same error message would suffice to reduce double logging?

internal/impl/tigerbeetle/input_tigerbeetle.go
```diff
@@ -0,0 +1,701 @@
...
+		case <-input.stopSignaller.HardStopChan(): // Hard stop.
+		}
+		client.Close()
+	}()
+
+	idleTimer := time.NewTimer(0)
+	_ = idleTimer.Stop()
+
+	for {
+		err := input.checkRateLimit(ctx)
```
josephwoodward: nitpick (non-blocking): perhaps a candidate for an inline error check.

internal/impl/tigerbeetle/input_tigerbeetle.go
```diff
@@ -0,0 +1,701 @@
...
+
+		// No events returned from the query,
+		// waiting for the timeout to resume the producer.
+		if len(results) == 0 {
+			// NB: We could go idle if `len(results) < eventCountMax`, since the client
+			// likely won’t return new results if queried again immediately.
+			// However, we wait for the *consumer* to begin flushing the current results
+			// before issuing a new query, avoiding unnecessary idle time for workloads
+			// with high frequency but low volume per batch.
+			rescheduled := idleTimer.Reset(input.config.idleInterval)
```
josephwoodward: nitpick (non-blocking): another a candidate for an inline check.

internal/impl/tigerbeetle/input_tigerbeetle.go
```diff
@@ -0,0 +1,701 @@
...
+
+			if duration > 0 {
+				input.logger.Debugf("rate_limit: waiting for %d ms", duration.Milliseconds())
+				_ = <-time.After(duration)
+			} else {
+				break
+			}
+		}
+		if attempt == max_tries {
+			return fmt.Errorf("failed to access the rate limit after %d attemps", max_tries)
```
josephwoodward:
```suggestion
			return fmt.Errorf("failed to access the rate limit after %d attempts", max_tries)
```

docs/modules/components/pages/inputs/tigerbeetle_cdc.adoc
```diff
@@ -0,0 +1,200 @@
...
+     https://github.com/redpanda-data/connect/tree/main/cmd/tools/docs_gen/templates/plugin.adoc.tmpl
+////
+
+// © 2024 Redpanda Data Inc.
+
+
+component_type_dropdown::[]
+
+
+Enables TigerBeetle CDC streaming for RedPanda Connect.
```
josephwoodward:
```suggestion
Enables TigerBeetle CDC streaming for Redpanda Connect.
```

internal/impl/tigerbeetle/input_tigerbeetle.go
```diff
@@ -0,0 +1,701 @@
...
+			UserData32:     result.CreditAccountUserData32,
+			Code:           result.CreditAccountCode,
+			Flags:          result.CreditAccountFlags,
+			Timestamp:      strconv.FormatUint(result.CreditAccountTimestamp, 10),
+		},
+	})
+}
+
+// StringToUint128 parses a base 10 string and returns the corresponding value as a Uint128.
+func StringToUint128(str string) (tb_types.Uint128, bool) {
```
josephwoodward: It doesn't appear that this function needs to be exported, nor `Uint128ToString` below.

internal/impl/tigerbeetle/input_tigerbeetle.go
```diff
@@ -0,0 +1,701 @@
...
+		addresses           []string
+		progressCache       string
+		rateLimit           string
+		eventCountMax       int
+		idleInterval        int
+		timestampInitialStr string
+		timestampInitial    uint64 = 0
+	)
+
+	logger := resources.Logger()
```
josephwoodward: nitpick (non-blocking): perhaps move this closer to where it's used? Being declared so high up made me question whether it was required in the configuration parsing below it.

josephwoodward:
@batiati A couple more of minor comments but otherwise it's looking great!

docs/modules/components/pages/inputs/tigerbeetle_cdc.adoc
batiati: I used the mysql connector as a template, which appears to have the same typo.
I'll open another PR to fix it there as well.

https://github.com/redpanda-data/connect/blob/96c6efeda3da2d8f7bf3baf4936f23f5de7ab4e0/internal/impl/mysql/input_mysql_stream.go#L54

batiati:
> @batiati A couple more of minor comments but otherwise it's looking great!

Thanks @josephwoodward! It's ready for another round.

mmatczuk:
Could we add file headers to the Go files?

internal/impl/tigerbeetle/input_tigerbeetle.go
mmatczuk: I had a look at various inputs it feels like the "idiomatic" way in connect would be `t`.

Jeffail:
> CI is failing due to `CGO_ENABLED: 0`.
> 
> The TigerBeetle client requires CGO to consume the static library `lib_tbclient` (the underlying implementation of the VSR client). It includes pre-built libraries in-repo for all supported platforms.
> 
> But it seems there's a policy on builds that require external dependencies and CGO, like the ZeroMQ connector:
> 
> > By default Redpanda Connect does not build with components that require linking to external libraries. If you wish to build Redpanda Connect locally with this component then set the build tag x_benthos_extra:

Ideally we'd want to find a way of getting pure Go builds so that this isn't hidden away from the majority of users. Would it be possible to use something like https://github.com/ebitengine/purego here?

mmatczuk:
I think we lack batch policy configuration `fieldBatchPolicy = "batching"`

batiati:
> lack batch policy configuration `fieldBatchPolicy = `

@mmatczuk, initially I went with the batch policy, but some of the properties didn’t make sense for how TigerBeetle streams out CDC events (such as batching by size in bytes). Instead, we use the `event_count_max` config to limit the maximum number of events produced per batch.

Perhaps the batch policy could still be useful for pipeline composability. Please let me know.

batiati:
> Ideally we'd want to find a way of getting pure Go builds so that this isn't hidden away from the majority of users. Would it be possible to use something like https://github.com/ebitengine/purego here?

Unfortunately, Go’s standard library does not expose a general-purpose FFI or dlopen-style interface! That’s how some of the other clients work. 

I know CGO can be a huge burden of complexity, but for the TigerBeetle client, adding a third-party dependency would not be worth the tradeoff either.

mmatczuk:
@batiati would you be open to exploring an approach similar to [opendal’s Go bindings](https://opendal.apache.org/bindings/go/) without CGO.

batiati:
> @batiati would you be open to exploring an approach similar to [opendal’s Go bindings](https://opendal.apache.org/bindings/go/) without CGO.

They also use `purego`.

https://github.com/apache/opendal/blob/a1b810910c6d288be8bd698521f843c622673c16/bindings/go/go.mod

We are looking for alternatives in the meantime, but for the short term we can't get rid of CGO.

batiati:
> but for the short term we can't get rid of CGO.

Hi @Jeffail and @mmatczuk! I did some research to find alternatives, and we really can't remove the CGO dependency anytime soon. However, as we ship the client with pre-built static libraries for multiple targets, we don't depend too heavily on the system's libraries (as is the case with installing libzmq3-dev when the `x_benthos_extra` flag is present).

Please let me know what you suggest doing to teach the CI to build the TigerBeetle connector conditionally, and which pieces of documentation warnings we should include about it.

internal/impl/tigerbeetle/input_tigerbeetle.go
Jeffail: Yeah you can get rid of that, but it's also harmless to leave in.

Jeffail:
Hey @batiati, we might try and enable CGO for some of our standard distributions of RPCN, but we almost definitely need to exclude this from our fips builds.

In order to proceed I think we should guard all of these files with a CGO build constraint so that non-cgo builds aren't broken. If you could update your PR to do this then we can merge and handle the rest in a follow-up PR, where we'll probably create fips builds with a cherry-picked list of connectors.

PR #3613, opened 2025-09-16
Initial Jira connector commit

zoltancsontosness:
# Redpanda Jira Connector

| Connector | Path                                            |
| ----------- |-------------------------------------------------|
| Jira | connect/public/components/jira |

## Prerequisites for development

1. Install **go**
2. Install **go-task**
    
    `go install github.com/go-task/task/v3/cmd/task@latest`

## Running & testing the connector

1. Set **GOBIN** variable
    
    `go env -w GOBIN=$(pwd)/bin`

2. Lint and format the code

    ```
    task lint
    task fmt
    ```

3. Build the application
    
    `go build`

4. Set variables specific to **connector** that you are running based on **config.yaml** file (see [redpanda secrets](https://docs.redpanda.com/redpanda-connect/configuration/secrets/)):
    
    `export JIRA_API_TOKEN=yourToken`

5. Run the connector 
    
    `./connect --config config.yaml`

## Config & usage examples

- Example `processor` configuration:
```
pipeline:
  threads: 1
  processors:
    - jira:
        base_url: "https://{your_org}.atlassian.net"
        username: "jiraUsername"
        api_token: "${JIRA_API_TOKEN}"
        max_results_per_page: 200
        max_retries: 50
```

- Example `input` configuration that produces one message into the pipeline:
```
input:
  generate:
    interval: "1m"
    count: 1 # 0 for infinite @ specified interval
    mapping: |
      root = {
        "project": "RPC",
        "fields": ["summary", "Story Points", "assignee.displayName", "Sprint.name"],
        "updated": ">= -4w",
        "created": ">= -4w"
      }
```

- Example message using JQL query - this retrieves the **summary**, **name of sprint** and **story points** for an issue from an user that is having status **In progress**
```
{
    "fields": ["summary", "Sprint.name", "Story Points"],
    "jql": "assignee=\"Andrei Arcan\" and status=\"In Progress\" order by created DESC"
}
```

- Example message using Project - this gets the **summary** and **status** for **all issues** that are **created** in the **last 4 weeks** and **updated** in the **last 2 days** for **RPC project**
```
{
    "project": "RPC",
    "fields": ["summary", "status.name"],
    "created": ">= -4w",
    "updated": ">= -2d"
}
```

- Example message using Project - this gets **all fields** for **all issues**
```
{
    "project": "RPC"
}
```

### DIFF
--- a/internal/impl/jira/README.md
+++ b/internal/impl/jira/README.md
@@ -1,0 +1,1 @@
+package jira
--- a/internal/impl/jira/config.go
+++ b/internal/impl/jira/config.go
@@ -1,0 +1,128 @@
+package jira
+
+import (
+	"errors"
+	"net/http"
+	"net/url"
+
+	"github.com/redpanda-data/connect/v4/internal/impl/jira/helpers/http_helper"
+	"github.com/redpanda-data/connect/v4/internal/impl/jira/helpers/http_metrics"
+
+	"github.com/redpanda-data/benthos/v4/public/service"
+)
+
+type jiraProc struct {
+	baseURL    string
+	username   string
+	apiToken   string
+	maxResults int
+	retryOpts  http_helper.RetryOptions
+	httpClient *http.Client
+	log        *service.Logger
+}
+
+// Configuration specification for the Jira processor
+var jiraProcessorConfigSpec = service.NewConfigSpec().
+	Summary("Fetches data from Jira based on input messages").
+	Description(`This processor takes input messages containing Jira queries and returns Jira data.
+
+Supports the following Jira resources:
+- issues
+- issue transitions
+- users
+- roles
+- project versions
+- project categories
+- project types
+- projects
+
+Configuration examples:
+
+` + "```yaml" + `
+# Minimal configuration
+pipeline:
+  processors:
+    - jira:
+        base_url: "https://your-domain.atlassian.net"
+        username: "${JIRA_USERNAME}"
+        api_token: "${JIRA_API_TOKEN}"
+
+# Full configuration
+pipeline:
+  processors:
+    - jira:
+        base_url: "https://your-domain.atlassian.net"
+        username: "${JIRA_USERNAME}"
+        api_token: "${JIRA_API_TOKEN}"
+        max_results_per_page: 200
+        request_timeout: "30s"
+        max_retries: 50
+` + "```").
+	Field(service.NewStringField("base_url").
+		Description("Jira instance base URL (e.g., https://your-domain.atlassian.net)")).
+	Field(service.NewStringField("username").
+		Description("Jira instance account username/email")).
+	Field(service.NewStringField("api_token").
+		Description("Jira API token for the specified account").
+		Secret()).
+	Field(service.NewIntField("max_results_per_page").
+		Description("Maximum number of results to return per page when calling JIRA API").
+		Default(50)).
+	Field(service.NewDurationField("request_timeout").
+		Description("HTTP request timeout").
+		Default("30s")).
+	Field(service.NewIntField("max_retries").
+		Description("Maximum number of retries in case of 429 HTTP Status Code").
+		Default(10))
+
+func newJiraProcessor(conf *service.ParsedConfig, mgr *service.Resources) (*jiraProc, error) {
+	baseURL, err := conf.FieldString("base_url")
+	if err != nil {
+		return nil, err
+	}
+
+	if _, err := url.ParseRequestURI(baseURL); err != nil {
+		return nil, errors.New("base_url is not a valid URL")
+	}
+
+	username, err := conf.FieldString("username")
+	if err != nil {
+		return nil, err
+	}
+
+	apiToken, err := conf.FieldString("api_token")
+	if err != nil {
+		return nil, err
+	}
+
+	timeout, err := conf.FieldDuration("request_timeout")
+	if err != nil {
+		return nil, err
+	}
+
+	maxResults, err := conf.FieldInt("max_results_per_page")
+	if err != nil {
+		return nil, err
+	}
+
+	maxRetries, err := conf.FieldInt("max_retries")
+	if err != nil {
+		return nil, err
+	}
+
+	httpClient := &http.Client{Timeout: timeout}
+
+	return &jiraProc{
+		baseURL:    baseURL,
+		username:   username,
+		apiToken:   apiToken,
+		maxResults: maxResults,
+		retryOpts: http_helper.RetryOptions{
+			MaxRetries: maxRetries,
+		},
+		httpClient: http_metrics.NewInstrumentedClient(
+			mgr.Metrics(), "jira_http",
+			httpClient),
+		log: mgr.Logger(),
+	}, nil
+}
--- a/internal/impl/jira/constants.go
+++ b/internal/impl/jira/constants.go
@@ -1,0 +1,4 @@
+package jira
+
+// JiraAPIBasePath is the base path for Jira Rest API
+const JiraAPIBasePath = "/rest/api/3"
--- a/internal/impl/jira/filter.go
+++ b/internal/impl/jira/filter.go
@@ -1,0 +1,174 @@
+package jira
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+/*
+selectorTree is used to build a tree from the elements present in Fields input message
+The tree is then used for filtering output messages and including only what is present in the Fields
+*/
+type selectorTree map[string]selectorTree
+
+/*
+Function to build selectorTree from the Fields []string object in the input message used for the attribute filtering
+
+Example: '"fields": ["summary", "assignee.displayName", "status.name", "parent.key", "parent.fields.status.name"]'
+Will result in returning a tree of the form:
+
+	{
+		"assignee": {
+			"displayName": {}
+		},
+		"parent": {
+			"fields": {
+				"status": {
+					"name": {}
+				}
+			},
+			"key": {}
+		},
+		"status": {
+			"name": {}
+		},
+		"summary": {}
+	}
+
+If custom fields are present, they will also be included in the selectorTree
+Example: '"fields": ["summary", "Sprint.name", "assignee.displayName", "Story Points"]'
+Will result in returning a tree of the form:
+
+	{
+	"assignee": {
+		"displayName": {}
+	},
+	"custom_field_10022": {
+		"name": {}
+	},
+	"custom_field_10100": {},
+	"summary": {}
+	}
+*/
+func (j *jiraProc) buildSelectorTree(fields []string, custom map[string]string) (selectorTree, error) {
+	j.debug("building selector tree based on filters: %v", fields)
+	tree := make(selectorTree)
+	for _, field := range fields {
+		if strings.TrimSpace(field) == "" {
+			return nil, errors.New("invalid field: empty string")
+		}
+		parts := strings.Split(field, ".")
+		cur := tree
+		for _, part := range parts {
+			if strings.TrimSpace(part) == "" {
+				return nil, fmt.Errorf("invalid field path: %q", field)
+			}
+			if _, ok := cur[part]; !ok {
+				cur[part] = make(selectorTree)
+			}
+			cur = cur[part]
+		}
+	}
+	for _, value := range custom {
+		if strings.TrimSpace(value) == "" {
+			return nil, errors.New("invalid field: empty string")
+		}
+		if _, ok := tree[value]; !ok {
+			tree[value] = make(selectorTree)
+		}
+	}
+	return tree, nil
+}
+
+/*
+The filter function takes the data JSON and selectorTree and returns only what is
+found in the selectorTree by comparing keys from data and keys from selectorTree.
+If customFields are present in the data, they will also be replaced with their real name;
+example: custom_field_10100 will be replaced with "Story Points"
+*/
+func (j *jiraProc) filter(data interface{}, selectors selectorTree, custom map[string]string) (interface{}, error) {
+	switch val := data.(type) {
+	case map[string]interface{}:
+		res := make(map[string]interface{})
+		for key, sub := range selectors {
+			if subData, ok := val[key]; ok {
+				if len(sub) > 0 {
+					filtered, err := j.filter(subData, sub, custom)
+					if err != nil {
+						return nil, err
+					}
+					if value, exists := custom[key]; exists {
+						res[value] = filtered
+					} else {
+						res[key] = filtered
+					}
+				} else {
+					if value, exists := custom[key]; exists {
+						res[value] = subData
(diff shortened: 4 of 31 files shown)


internal/impl/jira/logging.go
```diff
@@ -0,0 +1,46 @@
+package jira
+
+/*
+Package jira implements JIRA API integration and processing functionality
+debug logs a debug level message if the logger is initialized
+format: the format string for the message
+args: optional arguments for format string interpolation
+*/
+func (j *jiraProc) debug(format string, args ...interface{}) {
+	if j != nil && j.log != nil {
```
Jeffail: It's actually safe to call methods on a nil `*service.Logger` as it already guards against nil references: https://github.com/redpanda-data/benthos/blob/main/public/service/logger.go#L49, so we don't really need this wrapper and are probably better off without it.

internal/impl/jira/constants.go
```diff
@@ -0,0 +1,4 @@
+package jira
+
+// JiraAPIBasePath is the base path for Jira Rest API
+const JiraAPIBasePath = "/rest/api/3"
```
Jeffail: This probably doesn't need to be its own file, lets move this into client.go.

Jeffail (commented):
Hey @zoltancsontosness! Just a couple of comments, code looks really neat on a first pass through.

mmatczuk:
Would you mind converting block comments to line comments especially for docstrings?

internal/impl/jira/README.md
```diff
@@ -0,0 +1 @@
+package jira
```
mmatczuk: This is probably not needed.

mmatczuk:
Also, would be good to add file headers.

internal/impl/jira/config.go
```diff
@@ -0,0 +1,128 @@
...
+	"net/http"
+	"net/url"
+
+	"github.com/redpanda-data/connect/v4/internal/impl/jira/helpers/http_helper"
+	"github.com/redpanda-data/connect/v4/internal/impl/jira/helpers/http_metrics"
+
+	"github.com/redpanda-data/benthos/v4/public/service"
+)
+
+type jiraProc struct {
```
mmatczuk: I like the split to different resources files, I'm confused why this one is called config?

Probably jiraProc  belongs to resources, the rest to processor.

mmatczuk:
I’d like to suggest we reduce the number of exported types in the package to keep the API simple and easy to use. It would also be clearer if all Jira-specific code lived in a single package, for example `jirahttp`. In that case, I suggest we migrate all of the helpers content (and its subpackages) into jirahttp, so everything Jira-related is grouped in one place. Our internal Jira library can then live under the jira package.

mmatczuk:
Please add enterprise license check 

Look for

```
	if err := license.CheckRunningEnterprise(mgr); err != nil {
		return nil, err
	}
```

internal/impl/jira/filter.go
```diff
@@ -0,0 +1,174 @@
...
+	return tree, nil
+}
+
+/*
+The filter function takes the data JSON and selectorTree and returns only what is
+found in the selectorTree by comparing keys from data and keys from selectorTree.
+If customFields are present in the data, they will also be replaced with their real name;
+example: custom_field_10100 will be replaced with "Story Points"
+*/
+func (j *jiraProc) filter(data interface{}, selectors selectorTree, custom map[string]string) (interface{}, error) {
```
mmatczuk: Use `any` instead of `interface{}`

mmatczuk: Could this be a function not method? Or method of selectorTree?

internal/impl/jira/filter.go
```diff
@@ -0,0 +1,174 @@
...
+		"displayName": {}
+	},
+	"custom_field_10022": {
+		"name": {}
+	},
+	"custom_field_10100": {},
+	"summary": {}
+	}
+*/
+func (j *jiraProc) buildSelectorTree(fields []string, custom map[string]string) (selectorTree, error) {
```
mmatczuk: How about `selectorTreeFrom()` - a constructor not a method of jiraProc

internal/impl/jira/helpers/http_helper/http_helper.go
```diff
@@ -0,0 +1,344 @@
+/*
+Package http_helper usage
+
+Overview
+- Provides a thin, robust wrapper around net/http to:
```
mmatczuk: Improve rendering in go doc.

internal/impl/jira/helpers/jira_helper/jira_helper.go
```diff
@@ -0,0 +1,268 @@
...
+    // 200 with X-Seraph-LoginReason or other 4xx/5xx
+    }
+    }
+    return
+    }
+    // proceed to parse resp.Body
+
+ 5. Example helper wrapping a Jira call
+
+    func callJira(ctx context.Context, client *http.Client, req *http.Request) ([]byte, *jira_helper.JiraError) {
```
mmatczuk: Improve code formatting - maybe use examples https://go.dev/blog/examples

internal/impl/jira/logging.go
atudose-ness: done :heavy_check_mark:

internal/impl/jira/constants.go
atudose-ness: done :heavy_check_mark:

internal/impl/jira/README.md
atudose-ness: done :heavy_check_mark:

internal/impl/jira/config.go
atudose-ness: done :heavy_check_mark:

internal/impl/jira/filter.go
atudose-ness: done :heavy_check_mark:

internal/impl/jira/filter.go
atudose-ness: done :heavy_check_mark:

internal/impl/jira/helpers/http_helper/http_helper.go
atudose-ness: done :heavy_check_mark:

internal/impl/jira/processor.go
```diff
@@ -0,0 +1,161 @@
...
+		ApiToken:   apiToken,
+		MaxResults: maxResults,
+		RetryOpts: http_helper.RetryOptions{
+			MaxRetries: maxRetries,
+		},
+		HttpClient: http_metrics.NewInstrumentedClient(
+			mgr.Metrics(), "jira_http",
+			httpClient),
+		Log: mgr.Logger(),
+	}, nil
```
mmatczuk: We'd like to have all the connect integration code here - in the main jira package, and all the rest, http related, in jirahttp. Also, the helpers belong to jirahttp, as they contain logic specific to JIRA http needs and semantics.

internal/impl/jira/processor.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/redpanda-data/connect/v4/internal/impl/jira/helpers/http_helper"
+	"github.com/redpanda-data/connect/v4/internal/impl/jira/helpers/http_metrics"
+	"github.com/redpanda-data/connect/v4/internal/impl/jira/jirahttp"
+	"github.com/redpanda-data/connect/v4/internal/license"
+
+	"github.com/redpanda-data/benthos/v4/public/service"
+)
+
+// Configuration specification for the Jira processor
+var jiraProcessorConfigSpec = service.NewConfigSpec().
```
mmatczuk: Use function instead of a global variable.

internal/impl/jira/processor.go
```diff
@@ -0,0 +1,161 @@
+// Copyright 2024 Redpanda Data, Inc.
```
mmatczuk: Rename this file to `processor_jira.go`

internal/impl/jira/processor.go
```diff
@@ -0,0 +1,161 @@
...
+		Description("Maximum number of results to return per page when calling JIRA API").
+		Default(50)).
+	Field(service.NewDurationField("request_timeout").
+		Description("HTTP request timeout").
+		Default("30s")).
+	Field(service.NewIntField("max_retries").
+		Description("Maximum number of retries in case of 429 HTTP Status Code").
+		Default(10))
+
+func newJiraProcessor(conf *service.ParsedConfig, mgr *service.Resources) (*jirahttp.JiraProc, error) {
```
mmatczuk: I'm confused, we have a constructor for a type defined in a different package.

internal/impl/jira/processor.go
```diff
@@ -0,0 +1,161 @@
...
+		Username:   username,
+		ApiToken:   apiToken,
+		MaxResults: maxResults,
+		RetryOpts: http_helper.RetryOptions{
+			MaxRetries: maxRetries,
+		},
+		HttpClient: http_metrics.NewInstrumentedClient(
+			mgr.Metrics(), "jira_http",
+			httpClient),
+		Log: mgr.Logger(),
```
mmatczuk: The type should have a real constructor and those fields should not be exported.

internal/impl/jira/processor_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestJiraProcessorConfigValidation(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name       string
+		yaml       string
```
mmatczuk: It's safer to avoid names that conflict with common package names

mmatczuk:
The only type exported from `jirahttp` is the `JiraProc` which should be split and possibly renamed. The `Process` and `searchResource` belong to the main package. We need `jirahttp` to be more flexible, it shall export all the functionalities separately for ease of extension and maintenance.

internal/impl/jira/processor_test.go
atudose-ness: done ✔️

internal/impl/jira/processor.go
atudose-ness: done ✔️

internal/impl/jira/processor.go
atudose-ness: done ✔️

internal/impl/jira/processor.go
atudose-ness: done ✔️

internal/impl/jira/processor.go
atudose-ness: done ✔️

internal/impl/jira/processor.go
atudose-ness: done ✔️

mmatczuk:
I wonder why we have decided to implement custom retries for http instead of using something established like https://github.com/hashicorp/go-retryablehttp. It would be faster, easier and more secure. To make this retry scheme work in general case you need a more graceful handling of request body, for example note that (from Do)

```
// The request Body, if non-nil, will be closed by the underlying
// Transport, even on errors. The Body may be closed asynchronously after
// Do returns.
``` 

Given that at this point I suggest to move http_helper.go to jirahttp.

mmatczuk:
Also, the jira_helper hosting only 

```
// CheckJiraAuth inspects an HTTP response from Jira and
// returns a *JiraError if an authentication or rate-limiting
// problem is detected.
//
// The following cases are handled:
//   - 401 Unauthorized: Invalid credentials (e.g., bad API token).
//   - 403 Forbidden: Valid credentials but insufficient permissions.
//   - 429 Too Many Requests: Jira API throttling; caller should retry
//     after the delay indicated in the Retry-After header.
//   - 200 OK with X-Seraph-LoginReason header: Jira sometimes returns
//     200 but indicates authentication issues in this header.
//
// For all other 4xx/5xx statuses, a JiraError is returned with the status text as the reason.
// On success (no detected problem), CheckJiraAuth returns nil and the caller should proceed to read from the response body.
func CheckJiraAuth(resp *http.Response) ([]byte, error)
```

seems to have a lot in common with the retries and most likely be integrated there - maybe we can reuse the error now each package has it's own.

mmatczuk:
I'd like to suggest we add `internal/httpclient` package for all the general http concerns and configurations. We already have one in benthos. We could either move it from internal to public there or use it as inspiration and copy relevant parts to connect. I think the latter makes more sense given that we'd avoid work coordinate across repos. 

Given that we'd like to add more HTTP based processors it makes sense to put pure HTTP concerns is a dedicated package. Then http_metrics should go there, as well as logging, PROXY setting and autodetection that I think will be important.

cc @Jeffail

mmatczuk:
nit: could we rename `jirahttp.JiraHttp` to `jirahttp.Client`, and merge client.go and jira_http.go?

mmatczuk:
We'd take care of `internal/httpclient` later.

internal/impl/jira/processor_jira.go
```diff
@@ -0,0 +1,201 @@
...
+	"github.com/redpanda-data/connect/v4/internal/license"
+
+	"github.com/redpanda-data/benthos/v4/public/service"
+)
+
+// jiraProcessor is the Benthos jiraProcessor implementation for Jira queries.
+// It holds the client state and orchestrates calls into the jirahttp package.
+type jiraProcessor struct {
+	log      *service.Logger
+	jiraHttp *jirahttp.Client
```
mmatczuk: For clarity let's rename that to client.

mmatczuk:
@zoltancsontosness I think that in general we are good to go. 

Last items:

* please move `jira_helper` contents to `jirahttp` directly as it belongs there
* please move `http_metrics` package under `jirahttp` package i.e. `jirahttp/http_metrics`, and remove the `helpers` now empty dir
* please add row in `info.csv` with enterprise license
* please run `task docs` to generate the docs

mmatczuk:
Please make sure lining and testing works.

atudose-ness:
@mmatczuk can you please run the pipeline again to see if the linting passes? locally it runs successfully

atudose-ness:
tried generating the docs with `task docs`, but nothing is generated. this is the output of the task:
```
task docs
task: [target-dir] mkdir -p 'target'
task: [build:redpanda-connect] go build  -tags "" -ldflags "-w -s -X main.Version=v -X main.DateBuilt=2025-10-15T17:09:16Z" -o target/redpanda-connect ./cmd/redpanda-connect
task: [docs] go run -tags "" ./cmd/tools/docs_gen
task: [docs] go run -tags "" ./cmd/tools/plugins_csv_fmt
task: [docs] target/redpanda-connect lint --deprecated "./config/examples/*.yaml" "./docs/modules/**/*.md"
task: [docs] target/redpanda-connect template lint "./config/template_examples/*.yaml"
```

mmatczuk:
Work moved to #3699 closing this.

PR #3636, opened 2025-09-23
 Add CyborgDB Output Connector

ahellegit:
### Summary

This PR adds support for https://github.com/cyborginc/cyborgdb, a privacy-focused embedded vector database with client-side
encryption, as an output connector for Redpanda Connect.

### Motivation

CyborgDB provides a unique value proposition in the vector database ecosystem:
- Client-side encryption: All data is encrypted locally before transmission
- Privacy-first design: Ideal for sensitive data pipelines in regulated industries (healthcare, finance, legal)
- Simple embedded deployment: No complex infrastructure required, reducing operational overhead

This connector enables Redpanda Connect users to build secure, privacy-preserving data pipelines for vector search and AI
applications.

### Implementation Details

The connector provides:
- Batch operations: Efficient upsert and delete operations with configurable batching
- Automatic index management: Creates indexes if missing
- Metadata support: Store additional context alongside vectors


internal/plugins/info.csv
```diff
@@ -61,6 +61,7 @@ couchbase                 ,processor ,Couchbase                 ,4.11.0  ,commun
 crash                     ,processor ,crash                     ,4.47.0  ,certified  ,n          ,n     ,n
 csv                       ,input     ,csv                       ,0.0.0   ,certified  ,n          ,n     ,n
 csv                       ,scanner   ,csv                       ,0.0.0   ,certified  ,n          ,y     ,y
+cyborgdb                  ,output    ,cyborgdb                  ,4.40.0  ,community  ,n          ,y     ,y
```
rockwotj: Thanks can you set the version here to 4.66.0?

internal/impl/cyborgdb/output.go
```diff
@@ -0,0 +1,614 @@
...
+	poFieldMetadataMapping = "metadata_mapping"
+	poFieldCreateIfMissing = "create_if_missing"
+	
+	// KeySize is the required size for CyborgDB encryption keys (32 bytes for AES-256)
+	KeySize = 32
+)
+
+func outputSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Version("4.40.0").
```
rockwotj: Please remove this, we track it in the `info.csv` file now

internal/impl/cyborgdb/output.go
```diff
@@ -0,0 +1,614 @@
...
+}
+
+// resolveIndexKey resolves encryption key with priority: ENV > existing file > generate new
+func resolveIndexKey(indexName string, logger *service.Logger) ([]byte, error) {
+	// Check environment variable
+	if envKey := os.Getenv("CYBORGDB_INDEX_KEY"); envKey != "" {
+		logger.Infof("Using encryption key from CYBORGDB_INDEX_KEY environment variable")
+		return decodeBase64Key(envKey)
+	}
+
```
rockwotj: The proper way to do this in Redpanda Connect is to make this a configuration property and use https://docs.redpanda.com/redpanda-connect/configuration/secrets/ to inject the value. This way you can load from secrets managers as well. Additionally some deployments we don't actually want to load these variables as well as they may be locked down.

internal/impl/cyborgdb/output.go
```diff
@@ -0,0 +1,614 @@
...
+	keyFile := filepath.Join(".cyborgdb", fmt.Sprintf("%s.key", indexName))
+	if keyData, err := os.ReadFile(keyFile); err == nil {
+		logger.Debugf("Reusing existing encryption key from: %s", keyFile)
+		return decodeKeyFromFile(string(keyData))
+	}
+	
+	// Generate and store new key using CyborgDB SDK (DEV/TEST ONLY)
+	logger.Infof("No existing key found. Generating new encryption key for index: %s", indexName)
+	logger.Warnf("Auto-generating encryption keys is for DEVELOPMENT and TESTING only!")
+	logger.Warnf("For production, set CYBORGDB_INDEX_KEY environment variable!")
```
rockwotj: All this behavior ideally would be left out, otherwise if we must keep it needs to be put behind an configuration flag. Some environments (k8s for example) run with read only disks so the `.cyborgdb` will likely not be writable (and certainly won't be in Redpanda Cloud).

rockwotj: If you want a configuration reference, for our snowflake connector we support an inline key or a private key from the filesystem - the filesystem version is just disabled in these more strict environments: https://docs.redpanda.com/redpanda-connect/components/outputs/snowflake_streaming/#private_key

internal/impl/cyborgdb/output.go
```diff
@@ -0,0 +1,614 @@
...
+	
+	return nil
+}
+
+func (w *outputWriter) WriteBatch(ctx context.Context, batch service.MessageBatch) error {
+	if !w.init {
+		if err := w.Connect(ctx); err != nil {
+			return err
+		}
+	}
```
rockwotj: You can remove this - the connect framework calls connect for you, and if it didn't this would need a lock or be prone to races

internal/impl/cyborgdb/output.go
```diff
@@ -0,0 +1,614 @@
...
+		}
+		
+		// Create new index with hardcoded ivfflat type
+		// CyborgDB will auto-detect dimension and auto-train
+		w.logger.Infof("Creating new CyborgDB index %s with IVFFlat (auto-dimension, auto-train)", w.indexName)
+		
+		index, err = w.client.CreateIndex(ctx, w.indexName, w.indexKey)
+		if err != nil {
+			return fmt.Errorf("failed to create index %s: %w", w.indexName, err)
+		}
```
rockwotj, line 304: Do we need to handle races with multiple connectors creating the index at once? Or is this idempotent?

internal/impl/cyborgdb/output.go
```diff
@@ -0,0 +1,614 @@
...
+					vector[i] = float32(val)
+				case json.Number:
+					f, err := val.Float64()
+					if err != nil {
+						return fmt.Errorf("vector element %d cannot be converted to float: %w", i, err)
+					}
+					vector[i] = float32(f)
+				default:
+					return fmt.Errorf("vector element %d is not a number: %T", i, val)
+				}
```
rockwotj: You should instead call [bloblang.ValueAsFloat32](https://pkg.go.dev/github.com/redpanda-data/benthos/v4@v4.57.0/public/bloblang#ValueAsFloat32), this gives similar semantics elsewhere in Redpanda Connect

internal/impl/cyborgdb/output.go
ahellegit: You're absolutely right, I've updated the implementation to use a required index_key configuration field instead of environment variable reading.

internal/impl/cyborgdb/output.go
ahellegit: I've removed all the automatic key generation and file storage behavior as the connector now requires an explicit index_key configuration field, which works with Redpanda Connect's secrets system.

internal/impl/cyborgdb/output.go
ahellegit: CreateIndex operation is idempotent, multiple connectors can safely call it simultaneously.

rockwotj:
@ahellegit you're going to need to run `task docs` as well to auto generate the documentation, as well as `task deps` to fixup dependencies. After that I think we'll be all good to go!

rockwotj:
Looks like tests are failing still @ahellegit 

```
-test.shuffle 1759156596169815973
--- FAIL: TestComponentExamples (0.80s)
    --- FAIL: TestComponentExamples/processor:openai_embeddings (0.00s)
        schema_test.go:44: 
            	Error Trace:	/home/runner/work/connect/connect/cmd/tools/docs_gen/schema_test.go:44
            	            	yaml: line 16: found character that cannot start any token
            	Test:       	TestComponentExamples/processor:openai_embeddings
FAIL
```

ahellegit:
> Looks like tests are failing still @ahellegit
> 
> ```
> -test.shuffle 1759156596169815973
> --- FAIL: TestComponentExamples (0.80s)
>     --- FAIL: TestComponentExamples/processor:openai_embeddings (0.00s)
>         schema_test.go:44: 
>             	Error Trace:	/home/runner/work/connect/connect/cmd/tools/docs_gen/schema_test.go:44
>             	Error:      	Received unexpected error:
>             	            	yaml: line 16: found character that cannot start any token
>             	Test:       	TestComponentExamples/processor:openai_embeddings
> FAIL
> ```

rockwotj:
@ahellegit Just a `task docs` is needed and we're good here

ahellegit:
Thank you for your help! Just updated the failing doc

PR #3642, opened 2025-09-26
lang: add with_confluent_header function

mmatczuk:


internal/impl/lang/bloblang.go
```diff
@@ -403,3 +409,73 @@ func registerULID() error {
...
+		Example(
+			"Add header with schema ID 123",
+			`root = with_confluent_header(123, content())`,
+		).
+		Example(
+			"Add header with schema ID from metadata",
+			`root = with_confluent_header(meta("schema_id").number(), content())`,
+		)
+
+	return bloblang.RegisterFunctionV2("with_confluent_header", spec, func(args *bloblang.ParsedParams) (bloblang.Function, error) {
```
rockwotj: two things: 

* can we call this `with_schema_registery_id_prefix` or remove the confluent name?
* can we put this in the `confluent` package, it's not a generic thing, so I think from an imports prospective it only makes sense if you include the schema registry stuff.

mmatczuk:
@rockwotj updated

rockwotj (approved):

PR #3644, opened 2025-09-26
Redpanda Migrator V2

mmatczuk:
Check the details in [redpanda-migrator-upgrade-guide](https://github.com/redpanda-data/connect/blob/feature/migrator_v2/docs/modules/guides/pages/redpanda-migrator-upgrade-guide.adoc).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 
 ## 4.67.0 - TBD
 
+### Changed
+
+- Unified migrator: Introduced a single `redpanda_migrator` input/output pair replacing legacy `redpanda_migrator_bundle`, `redpanda_migrator_offsets`, and the standalone `schema_registry` output; pair components by matching `label`; all migration logic is centralised in the output. (@mmatczuk)
+
 ### Fixed
 
 - (MongoDB CDC) Fixed an issue with connecting to sharded databases. (@rockwotj)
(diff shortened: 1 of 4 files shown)


david-yu:
Apologies for adding more requirements here but based on feedback we should do the following

1. Document scale testing that has gone into migrator testing. Has this been tested for 500-1000 MB ingress traffic? 
2. Perform soak testing for 100 or so migrations: https://redpandadata.slack.com/archives/C076AV2M5MH/p1740331194937279?thread_ts=1740138300.451189&cid=C076AV2M5MH
3. Build in automation on the Cloud Connect Operator to automatically migrate config over to v2 config prior to upgrading the connector, so as to mitigate any outages on the customer side.

david-yu:
Thank you for the changes based on recent review. Adding the response from slack:

> 1. Testing is documented in https://github.com/redpanda-data/connect/pull/3644/files#diff-494214a5d26d0a8ded3cc30a51f145b491a4e2fc406f8e321329b7cd169e9c98, I'm working in perf testing this week - it should not be a blocker to merging what we already have
> 2. I ported and ran the soak test, it passes and it's stable
> 3. The legacy config redpanda_migrator_bundle is supported and falls back to the legacy implementation

Approving, but we should have someone else from eng approve before merging.

mmatczuk:
Rebased on main

rockwotj (approved):

mmatczuk:
Rebased on main again resolving conflicts on CL.

PR #3648, opened 2025-09-30
ci: automate bundle creation

rockwotj:


rockwotj:
We'll have to do a release to test.

Jeffail:
When we tag the bundle releases would that also trigger this action?

.github/workflows/bump-bundle-versions.yml
```diff
@@ -0,0 +1,66 @@
...
+          go get "github.com/redpanda-data/connect/v4@${{ steps.version.outputs.version }}"
+          go mod tidy
+
+      - name: Update enterprise go.mod
+        working-directory: public/bundle/enterprise
+        run: |
+          go get "github.com/redpanda-data/connect/v4@${{ steps.version.outputs.version }}"
+          go mod tidy
+
+      - name: Create Pull Request
```
mmatczuk, line 41: I wonder could we use the gh cli tool https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-github-cli

mmatczuk:
It would be beneficial to concentrate the whole build process in goreleaser, then possibly we could use the gh cli as a hook.

rockwotj:
@mmatczuk I'll let you do this in goreleaser then

> When we tag the bundle releases would that also trigger this action?

not because of the prefix not starting with `v`

PR #3649, opened 2025-10-01
Adds Cassandra exponential reconnection

### LINKED ISSUE #3641: Support Cassandra Exponential Reconnection
Hi @mmatczuk 

Would this be any use to you? 

The default Go one is `Constant`, a lot of the other drivers, like Java for instance it's Exponential.

jonny7:
Adds Exponential Reconnection Policy To Cassandra component. If accepted, would close #3641

### DIFF
--- a/docs/modules/components/pages/outputs/cassandra.adoc
+++ b/docs/modules/components/pages/outputs/cassandra.adoc
@@ -38,6 +38,7 @@ output:
   cassandra:
     addresses: [] # No default (required)
     timeout: 600ms
+    reconnect_interval: 60s
     query: "" # No default (required)
     args_mapping: "" # No default (optional)
     max_in_flight: 64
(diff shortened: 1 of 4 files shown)


docs/modules/components/pages/outputs/cassandra.adoc
```diff
@@ -79,6 +80,11 @@ output:
     host_selection_policy:
       local_dc: "" # No default (optional)
       local_rack: "" # No default (optional)
+    reconnect_interval: 60s
+    exponential_reconnection:
+      reconnection_max_retries: 0 # No default (required)
+      reconnection_initial_interval: "" # No default (required)
+      reconnection_max_interval: "" # No default (required)
```
mmatczuk: I wonder if we have `exponential_reconnection` perhaps the `reconnection` prefix in the fields could be skipped.

jonny7: Sure, I can drop that

jonny7: OH, I know why I did it, there is already a `max_retries` property in the cassandra output. If there's no issue reusing the label, which seems like no, I can do that

PR #3653, opened 2025-10-01
adding tcp_user_timeout setting for aws components

alextreichler:
In certain situations it is possible where a connection is ESTABLISHED, but for some reason, such as a network disconnect, the data is still sending to an aws service such as `sqs` or `dynamodb`, but those services are not responding back. Typically in this case, it will continue to retry sending the data up to the number set by the kernel setting for `tcp_retries2`, typically 15, which means it takes ~ 15-16mins before the connection is terminated.

This can cause the `input`, such as consumers to stop consuming from a topic/partition from Redpanda until that connection is killed and redpanda connect re-establishes a new connection.

Instead of modifying a systemwide kernel setting, it is possible to pass a value on the application level using `TCP_USER_TIMEOUT`. This will take president over the `tcp_retries2`.  This will make sure users don't have to modify a systemwide kernel setting and have more control specifically for redpanda connect in these cases.

### DIFF
--- a/internal/impl/aws/config/config.go
+++ b/internal/impl/aws/config/config.go
@@ -14,7 +14,10 @@
 
 package config
 
-import "github.com/redpanda-data/benthos/v4/public/service"
+import (
+	"github.com/redpanda-data/benthos/v4/public/netclient"
+	"github.com/redpanda-data/benthos/v4/public/service"
+)
 
 // SessionFields defines a re-usable set of config fields for an AWS session
 // that is compatible with the public service APIs and avoids importing the full
@@ -29,6 +32,12 @@ func SessionFields() []*service.ConfigField {
 			Description("Allows you to specify a custom endpoint for the AWS API.").
 			Optional().
 			Advanced(),
+		netclient.ConfigSpec(),
+		// service.NewDurationField("tcp_user_timeout").
+		// 	Description("Linux-specific TCP_USER_TIMEOUT for more fine grained connection stalling detection. How long to wait for acknowledgment of transmitted data on an established conenction before killing the connection. Only applies to connections in ESTAB state. Set to 0 (default) to disable.").
+		// 	Default("0s").
+		// 	Optional().
+		// 	Advanced(),
 		service.NewObjectField("credentials",
 			service.NewStringField("profile").
 				Description("A profile from `~/.aws/credentials` to use.").
@@ -52,7 +61,6 @@ func SessionFields() []*service.ConfigField {
 				Description("An external ID to provide when assuming a role.").
 				Optional().Advanced()).
 			Advanced().
-			Optional().
 			Description("Optional manual configuration of AWS credentials to use. More information can be found in xref:guides:cloud/aws.adoc[]."),
 	}
 }
--- a/internal/impl/aws/session.go
+++ b/internal/impl/aws/session.go
@@ -16,6 +16,7 @@ package aws
 
 import (
 	"context"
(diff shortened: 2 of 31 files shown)


internal/impl/aws/session.go
```diff
@@ -35,7 +35,7 @@ func int64Field(conf *service.ParsedConfig, path ...string) (int64, error) {
 	return int64(i), nil
 }
 
-// GetSession constructs an AWS session from a parsed config and provided options.
+// GetSession contstructs an AWS session from a parsed.config and provided options.
```
Jeffail:
```suggestion
// GetSession constructs an AWS session from a parsed config and provided options.
```

Guessing this was the IDE

internal/impl/aws/session.go
```diff
@@ -51,6 +51,20 @@ func GetSession(ctx context.Context, parsedConf *service.ParsedConfig, opts ...f
 			id, secret, token,
 		)))
 	}
+	// check if tcp_user_timeout is set.
+	// Note: this setting only applies to linux. It will build
+	// with the http_client_linux.go file, otherwise, if another OS is being used
+	// then this setting won't do anything.
+	logger := service.Logger{}
```
Jeffail: We can't do this as it'd spawn a logger that ignores the users settings. We can either return an explicit error or quietly ignore the setting. I think in this context it's probably fine to quietly ignore the field on non-linux OSes as the behavior is so indirect.

internal/impl/aws/config/config.go
```diff
@@ -29,6 +29,11 @@ func SessionFields() []*service.ConfigField {
 			Description("Allows you to specify a custom endpoint for the AWS API.").
 			Optional().
 			Advanced(),
+		service.NewDurationField("tcp_user_timeout").
+			Description("Linux-specific TCP_USER_TIMEOUT for more fine grained connection stalling detection. How long to wait for acknowledgment of transmitted data on an established conenction before killing the connection. Only applies to connections in ESTAB state. Set to 0 to disable.").
+			Default("0s").
+			Optional().
```
Jeffail: Remove this, the `Default` value overrides it.

Jeffail (changes requested):
Thanks @alextreichler, just a few comments, the docs look great. One thing we'd need to be careful of is the possibility of other fields in future also wanting to override the HTTP client. We could be proactive here and refactor this setting to be more of a general purpose HTTP customization function that other fields could contribute to, something like `httpClientFromConfig(p *service.ParsedConfig)` so it's more clear that all HTTP client fields go there.

alextreichler:
Based on Ash's earlier mention, I create an http_client.go file that is used to create the custom http client, which is called within the session.go file.

mmatczuk:
Could we have it not limited to AWS? I wonder could we perhaps have a `tcp` section in the main config and then `tcp` config object in components that support this for overwriting? This would allow to group and add more fields in the future instead of adding ad-hoc fields in multiple places.

We should have a dedicated library component that decorates / creates net.Dialer.

Possibly event this should be part of benthos?

alextreichler:
Based on Michal M, comment - I ended up changing this PR to "draft" and I added a PR in benthos framwork to add a new package called "netclient" in order for other components to use that within their project.  So, this is why I switched it to draft as the other one needs to be reviewed first. 

https://github.com/redpanda-data/benthos/pull/291

mmatczuk:
Closing this as work moved to Benthos.

PR #3666, opened 2025-10-06
chore(release): migrate Docker builds to GoReleaser dockers_v2

mmatczuk:
Consolidated Docker image building from dedicated GitHub Actions workflows into GoReleaser's dockers_v2 [1] feature. This way we compile each binary once, and Docker is another distribution channel not separate entity. This allows for fast local Docker builds, and faster releases.

- Removed docker_edge.yml GHA and separate docker job from release workflow
- Added dockers_v2 configuration to .goreleaser.yml for connect, connect-cloud, and connect-ai
- Adjusted Dockerfiles, changed to use prebuilt binaries, added $TARGETPLATFORM see [1], changed naming scheme to preserve Dockerfile suffix - fixes syntax highlighting
- Created `.goreleaser.local/` directory with split configs for local development, the reason for those separate files is that --id and --single-taget flags are a GoReleaser Pro feature and is not available
- Added `timetzdata` build tag to all builds for timezone support, moved it from Docker file
- Updated taskfiles/docker.yml to use GoReleaser, now this serves to build local docker images

[1] https://goreleaser.com/customization/dockers_v2/


.github/workflows/release.yml
```diff
@@ -84,116 +90,28 @@ jobs:
...
-          type=semver,suffix=${{ matrix.suffix }},pattern={{version}}
-          type=semver,suffix=${{ matrix.suffix }},pattern={{major}}.{{minor}}
-          type=semver,suffix=${{ matrix.suffix }},pattern={{major}}
-
-    - uses: aws-actions/configure-aws-credentials@v4
-      if: ${{ startsWith(github.ref, 'refs/tags/v') }}
-      with:
-        aws-region: us-east-1
-        role-to-assume: arn:aws:iam::${{ secrets.RP_AWS_CRED_ACCOUNT_ID }}:role/${{ vars.RP_AWS_CRED_BASE_ROLE_NAME }}${{ github.event.repository.name }}
+# TODO(mmt): what happens if we have cloud image with key
```
mmatczuk: We need to nicely handle that

PR #3667, opened 2025-10-06
Move redis_hash To Use Non-Deprecated HSET

jonny7:
Hi @mmatczuk 

Again, let me know if this is of interest to you guys. I think I got all the changes required

### DIFF
--- a/internal/impl/redis/output_hash.go
+++ b/internal/impl/redis/output_hash.go
@@ -191,7 +191,7 @@ func (r *redisHashWriter) Write(ctx context.Context, msg *service.Message) error
 	if r.walkJSON {
 		if err := walkForHashFields(msg, fields); err != nil {
 			err = fmt.Errorf("failed to walk JSON object: %v", err)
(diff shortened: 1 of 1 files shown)


internal/impl/redis/output_hash.go
```diff
@@ -191,7 +191,7 @@ func (r *redisHashWriter) Write(ctx context.Context, msg *service.Message) error
 	if r.walkJSON {
 		if err := walkForHashFields(msg, fields); err != nil {
 			err = fmt.Errorf("failed to walk JSON object: %v", err)
-			r.log.Errorf("HMSET error: %v\n", err)
+			r.log.Errorf("HSet error: %v\n", err)
```
mmatczuk: Shouldn't this be `HSET` too?

jonny7: Yes! Let me fix

jonny7: Updated

PR #3670, opened 2025-10-06
mongo/cdc: fix setup with sharded clusters

rockwotj:
Sharded clusters don't by definition have a consistent oplog between the
shards, so we just do a hack of starting a change stream and immediately
stopping it to get a resume token so we can start streaming after the
snapshot phase finishes. This ensures we don't loose any data between
the snapshot and streaming

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ Changelog
 
 All notable changes to this project will be documented in this file.
 
+## 4.67.0 - TBD
+
+### Fixed
+
+- (MongoDB CDC) Fixed an issue with connecting to sharded databases. (@rockwotj)
+
 ## 4.66.0 - 2025-10-03
 
 ### Added


internal/impl/mongodb/cdc/input.go
```diff
@@ -669,6 +680,27 @@ func (m *mongoCDC) readSnapshotRange(
 	return nil
 }
 
+func (m *mongoCDC) getCurrentResumeToken(ctx context.Context) (bson.Raw, error) {
+	filter := []bson.M{{"$match": bson.M{
+		"ns.coll": bson.M{"$in": slices.Clone(m.collections)},
```
mmatczuk, line 685: Why do we need clone here?

internal/impl/mongodb/cdc/integration_test.go
```diff
@@ -791,3 +808,79 @@ mongodb_cdc:
 	wait()
 	require.JSONEq(t, `[{"_id":1,"data":"hello"}, {"_id":2,"data":"hello"}, {"_id":3,"data":"hello"}]`, output.MessagesJSON(t))
 }
+
+func TestIntegrationMongoCDCWithSnapshotShardedCluster(t *testing.T) {
+	integration.CheckSkipExact(t)
+	// You can setup a sharded cluster with https://github.com/pkdone/sharded-mongodb-docker
```
mmatczuk, line 814: I'd consider adding it as a submodule here and running compose in the test.
Dockertest does not support compose directly.

mmatczuk (approved):

internal/impl/mongodb/cdc/input.go
rockwotj: It's copied from below. I don't remember why we make a copy.

internal/impl/mongodb/cdc/integration_test.go
rockwotj: Yeah there is a lot in the compose stuff and it's kind of a pain to setup, I'm not sure it's worth the effort right now.

internal/impl/mongodb/cdc/input.go
mmatczuk: My take is that if it can be modified we may have a race, if not it's probably not needed.

rockwotj: it cannot be modified l - it can be removed

PR #3671, opened 2025-10-07
.goreleaser.local: use fixed dev tags

mmatczuk:
This is more convenient and works around bundle tags being incompatible with docker tags.

For example after tagging bundles we get version like public/bundle/enterprise/v4.66.1-SNAPSHOT-255ed50c4

### DIFF
--- a/.goreleaser.local/connect-ai.yaml
+++ b/.goreleaser.local/connect-ai.yaml
@@ -25,7 +25,7 @@ dockers_v2:
     images:
       - redpandadata/connect
     tags:
-      - "{{ .Version }}-ai"
+      - dev-ai
     platforms:
       - linux/arm64
     extra_files:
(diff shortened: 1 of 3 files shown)


Jeffail:
What would happen during an actual release? would we need to manually push the version tags?

mmatczuk:
The release is always on tag so it does not show up. It did work yesterday i.e. until I pulled the new tags...

Jeffail:
@mmatczuk I mean during a release, with this PR would we not end up with only `dev*` tags uploaded to our docker repo? Ideally we want to reproduce the old ones so from a `v4.32.1` example we'd get `4`, `4.32`, `4.32.1` and `latest` for the regular image.

mmatczuk:
@Jeffail it is a different workflow, we have separate goreleaser files for local builds due to lack of features available in Pro version.

PR #3673, opened 2025-10-08
feat(microsoft_sql_server_cdc): add default sql based cache

josephwoodward:
Adds a default Microsoft SQL Server based cache for the `microsoft_sql_server_cdc` component which includes:

- If no cache component is specified it default to this
- Creates default table under `rpcn.CdcCheckpointCache` which can be configureable (including schema) via `checkpoint_cache_table_name` field.
- Creates stored proc under `rpcn.CdcCheckpointCacheUpdate` (or whatever schema is set for the cache table)

### DIFF
--- a/internal/impl/mssqlserver/checkpoint_cache.go
+++ b/internal/impl/mssqlserver/checkpoint_cache.go
@@ -1,3 +1,11 @@
+// Copyright 2025 Redpanda Data, Inc.
+//
+// Licensed as a Redpanda Enterprise file under the Redpanda Community
+// License (the "License"); you may not use this file except in compliance with
+// the License. You may obtain a copy of the License at
+//
+// https://github.com/redpanda-data/connect/blob/main/licenses/rcl.md
+
 package mssqlserver
 
 import (
@@ -15,10 +23,17 @@ import (
 )
 
 const (
-	defaultCacheKey        = "max_lsn"
-	defaultCheckpointCache = "rpcn.cdc_checkpoint_cache"
-	storedProcName         = "CdcCheckpointUpsert"
+	// cache updates a single row so we use a fixed key
+	defaultCacheKey = "max_lsn"
+	// defaultCheckpointCache can be configured by the user
+	defaultCheckpointCache = "rpcn.CdcCheckpointCache"
+	// defaultStoredProcName schema is inferred from the provided checkpoint cache config
+	// the stored procedure name cannot be configured by the user
+	defaultStoredProcName = "CdcCheckpointCacheUpdate"
 )
+
+// allowedTableIdentifiers is used for validating cache table names
+var allowedTableIdentifiers = regexp.MustCompile(`^[A-Za-z_][A-Za-z0-9_$]{0,127}$`)
 
 // cacheTable represents a formatted cache table name provided by the user configuration
 type cacheTable struct{ schema, name string }
@@ -81,7 +96,7 @@ func newCheckpointCache(
 	}
 
 	// create a prepared statement for calling the stored proc (created in same schema as cache table) during Set operations to remove avoidable overhead
-	if cacheSetStmt, err = db.PrepareContext(ctx, fmt.Sprintf("EXEC [%s].[%s] @Key=?, @Value=?", cacheTable.schema, storedProcName)); err != nil {
+	if cacheSetStmt, err = db.PrepareContext(ctx, fmt.Sprintf("EXEC [%s].[%s] @Key=?, @Value=?", cacheTable.schema, defaultStoredProcName)); err != nil {
 		_ = db.Close()
 		return nil, fmt.Errorf("preparing checkpoint cache statement: %w", err)
 	}
@@ -101,11 +116,10 @@ func newCheckpointCache(
 		_ = c.db.Close()
 		c.shutSig.TriggerHasStopped()
 	}()
-
 	return c, nil
 }
 
-// Get a cache item, we only do this at start up
+// Get a cache item, we only do this at start up, key can be ignored as we only ever store one entry
 func (c *checkpointCache) Get(ctx context.Context, _ string) ([]byte, error) {
 	if c.db == nil {
 		return nil, fmt.Errorf("checkpoint cache not initialised for get operation: %w", service.ErrNotConnected)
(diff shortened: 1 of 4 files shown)


internal/impl/mssqlserver/checkpoint_cache.go
```diff
@@ -0,0 +1,159 @@
...
+	"time"
+
+	"github.com/Jeffail/shutdown"
+
+	"github.com/redpanda-data/benthos/v4/public/service"
+)
+
+const (
+	defaultCacheSchema = "dbo"
+	defaultCachekey    = "max_lsn"
```
mmatczuk: Key - not the case

internal/impl/mssqlserver/checkpoint_cache.go
```diff
@@ -0,0 +1,159 @@
...
+
+	// on the hot path
+	storedProcName     = "__rp_connect_upsertCacheEntry"
+	storedProcFullName = defaultCacheSchema + "." + storedProcName
+)
+
+// sqlServerCheckpointCache is a Microsoft SQL Server specific cache created for the CDC component.
+// We have a custom cache because the cache_sql component doesn't support SQL Server due to its
+// inability to support upserting (meaning it can't be expressed in the cache_sql configs).
+type sqlServerCheckpointCache struct {
```
mmatczuk: In the context of this package the `sqlServer` is most likely not needed.

internal/impl/mssqlserver/checkpoint_cache.go
```diff
@@ -0,0 +1,159 @@
...
+const (
+	defaultCacheSchema = "dbo"
+	defaultCachekey    = "max_lsn"
+
+	cacheTableName     = "__rp_connect_checkpoint_cache"
+	cacheTableFullName = defaultCacheSchema + "." + cacheTableName
+
+	// on the hot path
+	storedProcName     = "__rp_connect_upsertCacheEntry"
+	storedProcFullName = defaultCacheSchema + "." + storedProcName
```
mmatczuk: As a user can I decide where I store the data - different table?

internal/impl/mssqlserver/checkpoint_cache.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/Jeffail/shutdown"
+
+	"github.com/redpanda-data/benthos/v4/public/service"
+)
+
+const (
+	defaultCacheSchema = "dbo"
+	defaultCachekey    = "max_lsn"
+
+	cacheTableName     = "__rp_connect_checkpoint_cache"
```
mmatczuk: Generally we use RPCN as an abbreviation. I think that it would be easier to comprehend if we used redpanda namespace to clearly indicate where it is coming from then the table can be name cdc_checkpoint_cache. Using a double underscore (__) prefix is not common practice in Microsoft SQL Server - at least in my 1m research...

internal/impl/mssqlserver/checkpoint_cache.go
```diff
@@ -0,0 +1,159 @@
...
+	storedProcFullName = defaultCacheSchema + "." + storedProcName
+)
+
+// sqlServerCheckpointCache is a Microsoft SQL Server specific cache created for the CDC component.
+// We have a custom cache because the cache_sql component doesn't support SQL Server due to its
+// inability to support upserting (meaning it can't be expressed in the cache_sql configs).
+type sqlServerCheckpointCache struct {
+	conn *sql.Conn
+
+	connMaxLifetime   time.Duration
```
mmatczuk: Why would we'd need to limit the max connection time?

internal/impl/mssqlserver/checkpoint_cache.go
```diff
@@ -0,0 +1,159 @@
...
+	connMaxIdleTime   time.Duration
+	shutSig           *shutdown.Signaller
+	cacheSetStatement *sql.Stmt
```
mmatczuk: consistency: you have Sig - abbreviation in one line and then Statement the full word in another 🤔

internal/impl/mssqlserver/checkpoint_cache.go
```diff
@@ -0,0 +1,159 @@
...
+	BEGIN
+		SET NOCOUNT ON;
+		IF EXISTS (SELECT 1 FROM %s WHERE cache_key = @Key)
+			UPDATE %s SET cache_val = @Value WHERE cache_key = @Key;
+		ELSE
+			INSERT INTO %s (cache_key, cache_val) VALUES (@Key, @Value);
+	END;`
+	if _, err = conn.ExecContext(ctx, fmt.Sprintf(q, storedProcFullName, cacheTableFullName, cacheTableFullName, cacheTableFullName)); err != nil {
+		return nil, fmt.Errorf("creating checkpoint cache write stored procedure: %w", err)
+	}
```
mmatczuk: Conider extracting to func.

internal/impl/mssqlserver/checkpoint_cache.go
```diff
@@ -0,0 +1,159 @@
...
+
+// Get a cache item, we only do this at start up
+func (c *sqlServerCheckpointCache) Get(ctx context.Context, _ string) ([]byte, error) {
+	if c.conn == nil {
+		return nil, fmt.Errorf("checkpoint cache not initialised for get operation: %w", service.ErrNotConnected)
+	}
+
+	var val []byte
+	q := "SELECT cache_val FROM %s WHERE cache_key = ?;"
+	if err := c.conn.QueryRowContext(ctx, fmt.Sprintf(q, cacheTableFullName), defaultCachekey).Scan(&val); err != nil {
```
mmatczuk: Even if we have default values we should host parameters as fields.

mmatczuk:
I think you can rebase commits squashing all the cleanup / improvement commits.
It would be nice to see unit tests (with docker) of the cache itself in separation.

internal/impl/mssqlserver/checkpoint_cache.go
josephwoodward: Unfortunately you can't use named parameters on table names 😞

internal/impl/mssqlserver/checkpoint_cache.go
josephwoodward: I've reworked this to use `sql.DB` which will manage the connection for us (reconnecting if a connection is closed due to being idle). Given we only create the one connection there won't be any contention over connection access.

internal/impl/mssqlserver/checkpoint_cache.go
josephwoodward: I was in the process of making this user configurable, which it is now with a default of `rpcn.cdc_checkpoint_cache`

internal/impl/mssqlserver/checkpoint_cache.go
josephwoodward: Yes, they can now. It was something I was working on.

emaxerrno: Why is this both snake_case and camelCase why not pick one.

josephwoodward: This was a WIP change, they're now SnakeCase. I recalled a convention of `sp_Name` for stored proc and used that initially. Now they live under a custom `rpcn` schema (which can be customised by the user).

PR #2200, opened 2023-11-02
Add pgx postgres driver option

madskrogh:
This pr addresses such issues as https://github.com/benthosdev/benthos/issues/1987 by adding support for the **pgx** postgres driver. It does so by importing and exposing the `pgx/stdlib` package, a faster and more feature rich drop-in replacement for `lib/pq`. The old postgres driver (pq) has been kept around for backwards compatibility.

Issue #1987, opened 2023-06-30
sql_insert output not compatible with Postgres array columns


mihaitodor:
If you run this config:

```yaml
input:
  generate:
    count: 1
    interval: 0s
    mapping: |
      root.arr = ["2"]

output:
  sql_insert:
    driver: postgres
    dsn: postgres://testuser:testpass@localhost:5432/testdb?sslmode=disable
    table: FOOBAR
    columns: [arr]
    args_mapping: root = [this.arr]
    init_statement: |
      CREATE TABLE IF NOT EXISTS FOOBAR (
        arr text[]
      );

shutdown_timeout: 2s
```

it will fail with the following error:

```
Failed to send message to sql_insert: sql: converting argument $1 type: unsupported type []interface {}, a slice of interface
```

This is due to Postgres requiring that array args are wrapped using `pq.Array()` before being sent to the sql driver as a query argument. However, this is Postgres-specific and I don't think it would be a good approach to sprinkle the Benthos codebase with such driver-specific args wrappers and maybe there's some generic way in the library that's used under the hood to build the query, but that will require some investigation...

As a workaround, one can use the `sql_raw` output and either use one of the custom JSON -> Array conversion functions described [here](https://dba.stackexchange.com/questions/54283/how-to-turn-json-array-into-postgres-array) or craft the entire sql string manually and set `unsafe_dynamic_query: true`. Not the best workaround, but it should do the trick...

madskrogh:
@mihaitodor what do think about swapping the standard postgres driver **pq** for **pgx** which has this functionality built in. You can try out your example on my fork here: https://github.com/madskrogh/benthos/tree/pgx

All the needed changes are in this commit: https://github.com/benthosdev/benthos/commit/7443b2d71ceb2944513a9e78c1877247db86ea08 

If you like the idea I will submit a pr.

mihaitodor:
@madskrogh Sounds like that might come in handy, but I assume there are various caveats to it, like maybe that `pgx` library has different performance characteristics. Also, it might do certain things differently which could break stuff for existing users. Would it be possible for the two drivers to be used at the same time? If yes, then it might be worth importing it as a separate driver and users can decide which one they want to use.

madskrogh:
[...]

PR #3674, opened 2025-10-08
feat(sql): add pgx as alternate postgres driver

maxtheaxe:
Draws on [this issue](https://github.com/redpanda-data/connect/issues/1987) (which I ran into myself), in which `sql_insert` doesn't actually support Postgres array columns. I was left trying to manually escape convoluted strings using regex, and handcraft array literals, which is really brittle.

Thus I am basically just rebasing and updating [this older PR](https://github.com/redpanda-data/connect/pull/2200). It adds **pgx** (which is intended to be a drop-in replacement for **lib/pq**) as an _alternate_ postgres driver.

It doesn't redefine any settings, just reuses everything from the original postgres driver.

---

I've run `go test` on `internal/impl/sql` as @mihaitodor suggested and built a docker image, and everything seems to be fully working.

That being said, I would like to state for the record that I leaned on codex a bit to refactor the two postgres integration tests (though basically all I'm doing is looping over the existing test for the now two driver options—super straightforward), so I would appreciate a proper review from someone who knows what they're doing better than I do.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -208,6 +208,7 @@ require (
 	github.com/go-git/go-billy/v5 v5.6.2 // indirect
 	github.com/go-quicktest/qt v1.101.1-0.20240301121107-c6c8733fa1e6 // indirect
 	github.com/invopop/jsonschema v0.13.0 // indirect
+	github.com/jackc/puddle/v2 v2.2.2 // indirect
 	github.com/jbenet/go-context v0.0.0-20150711004518-d14ea06fba99 // indirect
 	github.com/jcmturner/goidentity/v6 v6.0.1 // indirect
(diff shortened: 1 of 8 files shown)


maxtheaxe:
@madskrogh will need your signoff on the CLA

maxtheaxe:
@madskrogh really need you to sign off on this

josephwoodward:
@maxtheaxe Are you able to update docs when you get a moment please? (run `task docs`). If not I can do it and push them to your branch.

maxtheaxe:
@josephwoodward done and rebased on the latest main. Let me know if you need anything else.

PR #3677, opened 2025-10-10
Kafka components consolidation

Jeffail:
Deprecates all surplus kafka components such that only the `redpanda` input and output remain. The `_common` suffixed components behavior has been moved into `redpanda` by making the `seed_brokers` field optional, when omitted we use the common `redpanda` config block (and error if it does not exist).

In order to move the old `kafka_franz` behavior over to `redpanda` there is a new `unordered_processing` field block that allows enabling and further customizing the unordered processing behavior to match it.

Note: I've had to disable our unit test for detecting deprecated components in our docs examples because there is an example in the benthos repo (the `read_until` input) that references `kafka`, which will need to be removed first.

Leaving this as a draft for now because I need to move the integration tests over.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ All notable changes to this project will be documented in this file.
 - New `a2a_message` processor. (@birdayz)
 - New `jira` processor. (@zoltancsontosness, @atudose-ness)
 
+### Changed
+
+- The `kafka`, `kafka_franz` and `redpanda_common` inputs and outputs are now deprecated as their respective functionality has been rolled into the `redpanda` input and output. (@Jeffail)
+
 ## 4.67.0 - 2025-10-13
 
 ### Changed


docs/modules/components/pages/inputs/redpanda.adoc
```diff
@@ -86,6 +86,15 @@ input:
     partition_buffer_bytes: 1MB
     topic_lag_refresh_period: 5s
     max_yield_batch_bytes: 32KB
+    unordered_processing:
+      enabled: false
+      checkpoint_limit: 1024
+      batching:
```
rockwotj, line 92: Why not add this to ordered processing too, we need it in some scenarios like the snowflake connector where we want very specific output batch sizes?

Jeffail: Ordered processing currently has an advantage of being able to cut the batch _before_ reaching the limit because it doesn't have these fields. The generic batcher in the unordered version can't do this as a message that triggers the flush needs to be included for delivery guarantees (it's all within a single transaction).

My long term plan was going to be to improve the API for the generic batcher and then we can use it everywhere but that's a time sink and I don't think it's worth blocking this merge on it any longer. We can do improvements later and then deprecate these fields afterwards for a single mechanism.

internal/impl/kafka/franz_client.go
```diff
@@ -69,6 +78,8 @@ func FranzConnectionFields() []*service.ConfigField {
 // FranzConnectionDetails describes information required to create a kafka
 // connection.
 type FranzConnectionDetails struct {
+	isConfigured bool
```
mmatczuk: It would be better to have the method check the validity.

internal/impl/kafka/input_redpanda.go
```diff
@@ -100,22 +101,47 @@ func redpandaInputConfigFields() []*service.ConfigField {
...
+				}
+			} else {
+				mgr.Logger().Info("Connection fields omitted, falling back to common redpanda config.")
+
+				// We're using a common redpanda block to determine the connection.
+				if rdr, err = NewFranzReaderToggledFromConfig(conf, mgr, func() (clientOpts []kgo.Opt, err error) {
+					// Make multiple attempts here just to allow the redpanda logger
+					// to initialise in the background. Otherwise we get an annoying
+					// log.
+					for range 20 {
```
mmatczuk, line 131: Is this really needed, can't we wait in main for it?

internal/impl/kafka/franz_client.go
```diff
@@ -38,16 +38,25 @@ const (
 
 // FranzConnectionFields returns a slice of fields specifically for establishing
 // connections to kafka brokers via the franz-go library.
-func FranzConnectionFields() []*service.ConfigField {
+func FranzConnectionFields(optionalBrokers bool) []*service.ConfigField {
```
mmatczuk: Could we have `FranzConnectionFieldsWithBrokers` instead of bool arg?

cmd/tools/docs_gen/schema_test.go
```diff
@@ -31,7 +31,7 @@ func TestComponentExamples(t *testing.T) {
 	env := sch.Environment()
 
 	linter := sch.NewStreamConfigLinter()
-	linter.SetRejectDeprecated(true)
+	// linter.SetRejectDeprecated(true) TODO: Re-enable this once kafka is out of the benthos repo examples
```
mmatczuk: Maybe we could limit the scope here, say testComponent could detect kafka and exit early.

mmatczuk:
I think the update docs comments could be squashed to the relevant commits changing things.

mmatczuk:
CL needs to be updated. Also, it would be nice if we could have a clear indication in the commit messages it affects kafka.

mmatczuk:
I was thinking that we could move all the not deprecated (redpanda) components to redpanda package as noted earlier by Alex in the context of migrator. Doing so would allow for easy removal of legacy components just by full package removal. WDYT @Jeffail?

Jeffail:
@mmatczuk I might do that as a follow up PR as it'll mess up the diff quite a bit as there's a lot of shared code to divvy about.

internal/impl/kafka/input_redpanda.go
Jeffail: I can't remember the original context as to why it was implemented like this, 'm pretty sure we could but I'd need to retest it, maybe worth doing as a follow up PR.

mmatczuk (approved):

internal/impl/kafka/input_redpanda.go
mihaitodor: We had this circular dependency between the input and output and both of them were started by the manager in parallel, so you came up with this approach instead of pushing more complexity to the manager. Guess it's no longer needed...

PR #3688, opened 2025-10-14
Add A2A protocol processor

birdayz:
## Summary
Adds `a2a_message` processor for communicating with A2A (Agent-to-Agent) protocol agents via JSON-RPC transport. OAuth2 credentials are now passed via hidden CLI flags with built-in secret store resolution.

# Note: this is absolutely not feature complete, but a starting point. Having it in place will make further iteration a lot simpler.

### A2A Processor Implementation
- HTTP/JSON-RPC 2.0 transport for A2A protocol
- OAuth2 client credentials authentication
- Agent card resolution from `/.well-known/agent.json`
- Task polling: 5min timeout, 2s interval, context cancellation support
- Cloud-only build (registered in `info.csv`)

### Secret Resolution Infrastructure
- Added hidden CLI flags: `--x-redpanda-cloud-{token-url,client-id,client-secret,audience}`
- Implemented secret resolution for `${secrets.KEY}` references in CLI flags
- CLI layer resolves secrets via `--secrets` lookup system (supports env, aws, gcp, az, redis)
- Resolved values bridge to environment variables for processor consumption

## Authentication Flow

1. Platform passes CLI flags with secret references:
   ```bash
   --x-redpanda-cloud-client-secret="${secrets.OAUTH_SECRET}"
   ```

2. CLI resolves secrets using configured backend (default: `--secrets env:`)

3. CLI sets resolved values as environment variables:
   - `REDPANDA_CLOUD_TOKEN_URL`
   - `REDPANDA_CLOUD_CLIENT_ID`
   - `REDPANDA_CLOUD_CLIENT_SECRET`
   - `REDPANDA_CLOUD_AUDIENCE` (optional)

4. Processor reads env vars and authenticates via OAuth2 client credentials grant

## Configuration Example

```yaml
pipeline:
  processors:
    - a2a_message:
        agent_card_url: "https://agent.example.com"
        prompt: "${content()}"
        final_message_only: true
```

## Testing

Tested with cloud binary and secret resolution.

## TODO
- [x] Consider removing debug logging before merge
- [x] Revisit agent_card_url to accept full URL instead of base URL

### DIFF
--- a/internal/cli/flags_redpanda.go
+++ b/internal/cli/flags_redpanda.go
@@ -25,6 +25,7 @@ import (
 	"github.com/redpanda-data/connect/v4/internal/impl/kafka"
 	"github.com/redpanda-data/connect/v4/internal/license"
 	"github.com/redpanda-data/connect/v4/internal/secrets"
+	"github.com/redpanda-data/connect/v4/internal/serviceaccount"
 )
 
 const (
@@ -275,33 +276,17 @@ func resolveSecret(ctx context.Context, value string, lookupFn secrets.LookupFn)
 }
 
 // parseCloudAuthFlags parses the OAuth2/cloud authentication CLI flags,
-// resolves any secret references, and sets them as environment variables
-// for the a2a processor to use.
+// resolves any secret references, and initializes the global service account configuration.
 func parseCloudAuthFlags(ctx context.Context, c *cli.Context, secretLookupFn secrets.LookupFn) error {
 	tokenURL := resolveSecret(ctx, c.String(rfCloudTokenURL), secretLookupFn)
 	clientID := resolveSecret(ctx, c.String(rfCloudClientID), secretLookupFn)
 	clientSecret := resolveSecret(ctx, c.String(rfCloudClientSecret), secretLookupFn)
 	audience := resolveSecret(ctx, c.String(rfCloudAudience), secretLookupFn)
 
-	// Set resolved values as environment variables if provided
-	if tokenURL != "" {
-		if err := os.Setenv("REDPANDA_CLOUD_TOKEN_URL", tokenURL); err != nil {
-			return fmt.Errorf("failed to set REDPANDA_CLOUD_TOKEN_URL: %w", err)
-		}
-	}
-	if clientID != "" {
-		if err := os.Setenv("REDPANDA_CLOUD_CLIENT_ID", clientID); err != nil {
-			return fmt.Errorf("failed to set REDPANDA_CLOUD_CLIENT_ID: %w", err)
-		}
-	}
-	if clientSecret != "" {
-		if err := os.Setenv("REDPANDA_CLOUD_CLIENT_SECRET", clientSecret); err != nil {
-			return fmt.Errorf("failed to set REDPANDA_CLOUD_CLIENT_SECRET: %w", err)
-		}
-	}
-	if audience != "" {
-		if err := os.Setenv("REDPANDA_CLOUD_AUDIENCE", audience); err != nil {
-			return fmt.Errorf("failed to set REDPANDA_CLOUD_AUDIENCE: %w", err)
+	// Initialize global service account config if credentials are provided
+	if tokenURL != "" && clientID != "" && clientSecret != "" {
+		if err := serviceaccount.InitGlobal(ctx, tokenURL, clientID, clientSecret, audience); err != nil {
+			return fmt.Errorf("failed to initialize service account authentication: %w", err)
 		}
 	}
 
--- a/internal/impl/a2a/interceptor.go
+++ b/internal/impl/a2a/interceptor.go
@@ -6,7 +6,7 @@
 //
 // https://github.com/redpanda-data/connect/blob/main/licenses/rcl.md
 
-package serviceaccount
+package a2a
 
 import (
 	"context"
@@ -16,15 +16,15 @@ import (
 	"golang.org/x/oauth2"
 )
 
-// OAuth2BearerInterceptor adds OAuth2 Bearer tokens to outgoing A2A requests.
-type OAuth2BearerInterceptor struct {
+// oauth2BearerInterceptor adds OAuth2 Bearer tokens to outgoing A2A requests.
+type oauth2BearerInterceptor struct {
 	a2aclient.PassthroughInterceptor
-	TokenSource oauth2.TokenSource
+	tokenSource oauth2.TokenSource
 }
 
 // Before adds the OAuth2 Bearer token to the request metadata.
-func (i *OAuth2BearerInterceptor) Before(ctx context.Context, req *a2aclient.Request) (context.Context, error) {
-	token, err := i.TokenSource.Token()
+func (i *oauth2BearerInterceptor) Before(ctx context.Context, req *a2aclient.Request) (context.Context, error) {
+	token, err := i.tokenSource.Token()
 	if err != nil {
 		return ctx, fmt.Errorf("failed to get OAuth2 token: %w", err)
 	}
--- a/internal/impl/a2a/processor_message.go
+++ b/internal/impl/a2a/processor_message.go
@@ -120,17 +120,17 @@ func makeProcessor(conf *service.ParsedConfig, mgr *service.Resources) (service.
 		return nil, err
 	}
 
-	oauth2Cfg, err := serviceaccount.NewOAuth2ConfigFromEnv()
-	if err != nil {
-		return nil, err
-	}
-
 	ctx := context.Background()
 
-	// Create authenticated HTTP client
-	httpClient, tokenSource, err := oauth2Cfg.CreateHTTPClient(ctx)
+	// Get authenticated HTTP client and token source from global service account config
+	httpClient, err := serviceaccount.GetHTTPClient()
 	if err != nil {
-		return nil, fmt.Errorf("failed to create OAuth2 HTTP client: %w", err)
+		return nil, fmt.Errorf("failed to get service account HTTP client: %w", err)
+	}
+
+	tokenSource, err := serviceaccount.GetTokenSource()
+	if err != nil {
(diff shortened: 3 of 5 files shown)


internal/plugins/info.csv
```diff
@@ -1,4 +1,5 @@
 name                      ,type      ,commercial_name           ,version ,support    ,deprecated ,cloud ,cloud_with_gpu
+a2a_message               ,processor ,a2a_message               ,4.66.0  ,certified  ,n          ,y     ,y
```
birdayz, line 2: @Jeffail how should we configure it to be more or less hidden / cloud only?
we can open it up later for non-cloud, but for now it's only working there

internal/impl/a2a/transport_http.go
```diff
@@ -0,0 +1,444 @@
...
+)
+
+// httpTransport implements a2aclient.Transport using HTTP/JSON-RPC 2.0.
+type httpTransport struct {
+	baseURL    string
+	httpClient *http.Client
+}
+
+// NewHTTPTransport creates a new HTTP transport for A2A protocol.
+func NewHTTPTransport(baseURL string, httpClient *http.Client) a2aclient.Transport {
```
birdayz, line 34: this file is only needed until official a2a-go merges support for JSONRPC transport. so long, we run our own, but have no desire to stick to it longer than needed.

internal/impl/a2a/processor_message.go
```diff
@@ -0,0 +1,409 @@
...
+	} else {
+		// Return the complete Message as a structured object
+		outMsg.SetStructuredMut(msg)
+		p.logger.Infof("Returning full message object (%d parts)", len(msg.Parts))
+	}
+
+	return service.MessageBatch{outMsg}, nil
+}
+
+func (p *messageProcessor) pollTaskUntilComplete(ctx context.Context, taskID a2a.TaskID) (*a2a.Task, error) {
```
birdayz, line 362: might change this to use the stream endpoint. but keeping it simple for now, and will follow up with a stream option.

internal/serviceaccount/oauth2.go
```diff
@@ -0,0 +1,100 @@
...
+	ClientSecret string
+}
+
+// NewOAuth2ConfigFromEnv creates OAuth2 configuration from environment variables.
+// Returns nil if required env vars are not set (allowing processor to work without auth).
+func NewOAuth2ConfigFromEnv() (*OAuth2Config, error) {
+	tokenURL := os.Getenv(RPEnvTokenURL)
+	clientID := os.Getenv(RPEnvClientID)
+	clientSecret := os.Getenv(RPEnvClientSecret)
+	audience := os.Getenv(RPEnvAudience)
```
rockwotj: I think we should move this stuff to some internal package like we did for gateway.

internal/impl/a2a/processor_message.go
```diff
@@ -0,0 +1,404 @@
...
+For more information about the A2A protocol, see https://a2a-protocol.org/latest/specification`).
+		Version("4.40.0").
+		Fields(
+			service.NewURLField(ampFieldAgentCardURL).
+				Description("URL for the A2A agent card. Can be either a base URL (e.g., `https://example.com`) or a full path to the agent card (e.g., `https://example.com/.well-known/agent.json`). If no path is provided, defaults to `/.well-known/agent.json`. Authentication uses OAuth2 from environment variables."),
+			service.NewInterpolatedStringField(ampFieldPrompt).
+				Description("The user prompt to send to the agent. By default, the processor submits the entire payload as a string.").
+				Default(""),
+			service.NewBoolField(ampFieldFinalMessageOnly).
+				Description("If true, returns only the text from the final agent message (concatenated from all text parts). If false, returns the complete Message or Task object as structured data with full history, artifacts, and metadata.").
```
rockwotj: An example would be helpful

internal/impl/a2a/processor_message.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+
+	// If prompt is empty, use message payload as string
+	if promptText == "" {
+		payloadBytes, err := msg.AsBytes()
+		if err != nil {
+			return nil, fmt.Errorf("failed to get message payload: %w", err)
+		}
+		promptText = string(payloadBytes)
+	}
```
rockwotj: I think generally we do something like this:

```
if p.prompt != nil {
  promptText = p.prompt.TryString(msg)
} else {
  promptText = string(msg.AsBytes())
}
```

Then remove the default value from the config and make it optional using conf.Contains when parsing the config

internal/impl/a2a/transport_http.go
```diff
@@ -0,0 +1,445 @@
+// Copyright 2024 Redpanda Data, Inc.
```
rockwotj: Image

internal/plugins/info.csv
rockwotj: Just only add it to the cloud package and not community. I think this here is fine.

internal/cli/flags_redpanda.go
```diff
@@ -227,3 +251,55 @@ client_id: rpcn
...
+// for the a2a processor to use.
+func parseCloudAuthFlags(ctx context.Context, c *cli.Context, secretLookupFn secrets.LookupFn) error {
+	tokenURL := resolveSecret(ctx, c.String(rfCloudTokenURL), secretLookupFn)
+	clientID := resolveSecret(ctx, c.String(rfCloudClientID), secretLookupFn)
+	clientSecret := resolveSecret(ctx, c.String(rfCloudClientSecret), secretLookupFn)
+	audience := resolveSecret(ctx, c.String(rfCloudAudience), secretLookupFn)
+
+	// Set resolved values as environment variables if provided
+	if tokenURL != "" {
+		if err := os.Setenv("REDPANDA_CLOUD_TOKEN_URL", tokenURL); err != nil {
```
rockwotj: This feels very dirty. is there another way? Can we smuggle it in the context we pass to connect or something?

birdayz: it's definitely dirty.
i'm not sure what is a better way to wire up things, without touching too much code (i would leave that rather to rpcn folks). i didn't find a much nicer way. happy to differently if there's a better way to get things into the "processor constructor function"

internal/serviceaccount/oauth2.go
birdayz: Done. Moved OAuth2 auth to internal/serviceaccount package.

internal/impl/a2a/processor_message.go
birdayz: Added examples showing both output modes.

internal/impl/a2a/transport_http.go
birdayz: Fixed.

internal/impl/a2a/processor_message.go
birdayz: Done. Made prompt field truly optional using conf.Contains().

internal/cli/flags_redpanda.go
rockwotj: I would prefer an explicit singleton myself over abusing envvars, but @Jeffail might have some magic

internal/impl/a2a/interceptor.go
```diff
@@ -0,0 +1,38 @@
...
+//
+// https://github.com/redpanda-data/connect/blob/main/licenses/rcl.md
+
+package serviceaccount
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/a2aproject/a2a-go/a2aclient"
```
rockwotj, line 15: Since this bit is a2a specific let's move this into `impl/a2a`

internal/cli/flags_redpanda.go
```diff
@@ -38,6 +38,10 @@ const (
 	rfSASLMechanism     = "x-redpanda-sasl-mechanism"
 	rfSASLUsername      = "x-redpanda-sasl-username"
 	rfSASLPassword      = "x-redpanda-sasl-password"
+	rfCloudTokenURL     = "x-redpanda-cloud-token-url"
+	rfCloudClientID     = "x-redpanda-cloud-client-id"
+	rfCloudClientSecret = "x-redpanda-cloud-client-secret"
+	rfCloudAudience     = "x-redpanda-cloud-audience"
```
rockwotj: nit: It would be good to put `service-account` in there too?

internal/impl/a2a/interceptor.go
rockwotj: The reasoning being if we make a service account for each pipeline, then the kafka stuff might be able to use that by default in cloud too someday.

birdayz: Ok . Plans are to make this work with rp and maybe the Kafka ones in the next weeks. Should not be too much work.

internal/cli/flags_redpanda.go
birdayz: Added usage documentation mentioning service-account authentication.

birdayz: just to clarify: do you mean, we should turn it into x-redpanda-cloud-service-account-* ?

rockwotj: Yeah? I don't know a bit verbose, so fine not too

birdayz: Renamed to include service-account: x-redpanda-cloud-service-account-{token-url,client-id,client-secret,audience}

internal/cli/flags_redpanda.go
birdayz: Replaced environment variables with explicit singleton pattern. CLI calls serviceaccount.InitGlobal(), processors get clients via serviceaccount.GetHTTPClient() and GetTokenSource(). Added tests.

internal/impl/a2a/interceptor.go
birdayz: Moved to internal/impl/a2a/interceptor.go and made it unexported.

PR #3691, opened 2025-10-14
gha: improve release.yml workflow

andrewhsu:
jira: [DEVPROD-3463]

- reduce release time by using beefier runner
- adds `workflow_dispatch` to enable testing `goreleaser` before a release is tagged 
- reverts #3689 to get `cgo` builds back

[DEVPROD-3463]: https://redpandadata.atlassian.net/browse/DEVPROD-3463?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -98,7 +98,7 @@ jobs:
       if: ${{ github.event_name == 'push' }}
       uses: goreleaser/goreleaser-action@v6
       with:
-        args: release --release-notes=./release_notes.md --timeout 120m
+        args: release --release-notes=./release_notes.md --timeout 120m --verbose
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         CLOUDSMITH_API_KEY: ${{ env.CLOUDSMITH_API_KEY }}
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -20,13 +20,32 @@ builds:
             - QUILL_LOG_FILE=target/dist/quill-{{ .Target }}.log
     ignore:
       - goos: windows
-        goarch: arm64
+        goarch: arm
+      - goos: darwin
+        goarch: arm
     env:
       - CGO_ENABLED=0
     tags:
       - timetzdata
     ldflags: >
       -s -w
+      -X main.Version={{.Version}}
+      -X main.DateBuilt={{.Date}}
+      -X main.BinaryName=redpanda-connect
+      -X github.com/redpanda-data/connect/v4/internal/telemetry.ExportHost={{ if index .Env "CONNECT_TELEMETRY_HOST"  }}{{ .Env.CONNECT_TELEMETRY_HOST }}{{ else }}{{ end }}
+      -X github.com/redpanda-data/connect/v4/internal/telemetry.ExportDelay={{ if index .Env "CONNECT_TELEMETRY_DELAY"  }}{{ .Env.CONNECT_TELEMETRY_DELAY }}{{ else }}{{ end }}
+      -X github.com/redpanda-data/connect/v4/internal/telemetry.ExportPeriod={{ if index .Env "CONNECT_TELEMETRY_PERIOD"  }}{{ .Env.CONNECT_TELEMETRY_PERIOD }}{{ else }}{{ end }}
+
+  - id: connect-cgo
+    main: cmd/redpanda-connect/main.go
+    binary: redpanda-connect
+    goos: [ linux ]
+    goarch: [ amd64 ]
+    env:
+      - CGO_ENABLED=1
+    tags:
+      - timetzdata
+    ldflags: >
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -4,10 +4,12 @@ on:
   push:
     tags:
       - 'v*'
+  workflow_dispatch:
 
 jobs:
   goreleaser:
-    runs-on: ubuntu-latest
+    # need larger than standard 4-CPU 16-GB storage runner
```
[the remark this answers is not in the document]
andrewhsu: it is storage, but i put incorrect number according to [ref](https://docs.github.com/en/enterprise-cloud@latest/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories). i'll fix.

.github/workflows/release.yml
```diff
@@ -113,10 +94,20 @@ jobs:
...
         args: release --release-notes=./release_notes.md --timeout 120m
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         CLOUDSMITH_API_KEY: ${{ env.CLOUDSMITH_API_KEY }}
+
+    - name: GoReleaser release snapshot skip publish
+      if: ${{ github.event_name == 'workflow_dispatch' }}
+      uses: goreleaser/goreleaser-action@v6
+      with:
+        args: release --timeout 120m --snapshot --skip publish
```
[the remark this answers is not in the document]
andrewhsu: [doc](https://goreleaser.com/cmd/goreleaser_release/#options) show option without `-`

.goreleaser.yml
```diff
@@ -20,7 +20,9 @@ builds:
             - QUILL_LOG_FILE=target/dist/quill-{{ .Target }}.log
     ignore:
       - goos: windows
-        goarch: arm64
+        goarch: arm
+      - goos: darwin
+        goarch: arm
```
mmatczuk, line 25: Given that `goarch: [ amd64, arm64 ]` what happens here with the arm change?
Besides that I do not know why we would like not to build arm image @Jeffail?

.github/workflows/release.yml
```diff
@@ -4,10 +4,12 @@ on:
   push:
     tags:
       - 'v*'
+  workflow_dispatch:
```
mmatczuk, line 7: I'm not sure about this, we should only build from git tags. @andrewhsu can you share the motivation for this?

.github/workflows/release.yml
```diff
@@ -4,10 +4,12 @@ on:
...
 
 jobs:
   goreleaser:
-    runs-on: ubuntu-latest
+    # need larger than standard 4-CPU 14-GB storage runner
+    runs-on: ubuntu-latest-32
```
mmatczuk, line 12: Why 32? I fail to see anything like it in https://github.com/actions/runner-images?tab=readme-ov-file

.github/workflows/release.yml
andrewhsu: i can drop this commit if not desired.

motivation is to reduce tagging as a mechanism to verify `goreleaser release` will work. this on-demand event will run the `GoReleaser release snapshot skip publish` step which is most of the things that `goreleaser release` will do similar to what docs mention on [test locally](https://goreleaser.com/customization/dockers_v2/#testing-locally). it won't test push to docker hub, but it will test docker image building which is my main concern because `docker_v2` is still in [alpha](https://goreleaser.com/blog/goreleaser-v2.12/#docker-v2).

an alternative to having `workflow_dispatch` could be to change the release process to do `-rc1` or `-beta1` tags so users of connect would know that some tags are possibly not suitable for GA release.

another alternative could be nightly or weekly releases, but this would be slower feedback.

.github/workflows/release.yml
andrewhsu: this is a custom hosted large runner we have configured in our github redpanda-data org. a list can be found on our [wiki](https://redpandadata.atlassian.net/wiki/x/NwCYV).

i chose 32 because it is already used by [upload_plugin.yml](https://github.com/redpanda-data/connect/blob/main/.github/workflows/upload_plugin.yml) but could change it if desired.

andrewhsu:
@mmatczuk i noticed PR #3693 just merged which addresses some of the limitation of the release workflow. feel free to close this PR if not desired or let me know if i should rebase to address conflicts.

mmatczuk:
I'm stoked to learn about the 32 core runner! cc @Jeffail

Jeffail:
Thanks @andrewhsu, sorry I missed this PR during all the release chaos. This is awesome, I've created a new PR with just the runner update and the workflow dispatch option on the release job, can you take a look? https://github.com/redpanda-data/connect/pull/3702

PR #3693, opened 2025-10-15
Split release assets across goreleaser configs

Jeffail:
We're hitting performance and disk space issues with our builds, mostly because we have a large number of assets being built all within the same runner. In order to improve both I've broken our individual "flavors" of RPCN into individual goreleaser configs, and we can now execute them individually using a matrix in our workflows.

I've tested this out locally using act but that only gets us so far, ideally after merging we need to kick off a release candidate to test it properly.

### DIFF
--- a/internal/impl/redpanda/migrator/bench/docker-compose.yml
+++ b/internal/impl/redpanda/migrator/bench/docker-compose.yml
@@ -55,7 +55,7 @@ services:
           --topic-config flush.ms=1000
 
   loader:
-    image: redpandadata/connect:dev-arm64
+    image: redpandadata/connect:edge-arm64
     depends_on:
       setup:
         condition: service_completed_successfully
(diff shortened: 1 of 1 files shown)


mmatczuk:
Given that the differences are minor, how about we used some templating to generate the files. We still can host them generated but it would be easier to maintain. We can use something as plain as envsubst or use Go templates with embed or sth else...

Jeffail:
> Given that the differences are minor, how about we used some templating to generate the files. We still can host them generated but it would be easier to maintain. We can use something as plain as envsubst or use Go templates with embed or sth else...

I'm a little bit hesitant to add another layer of tooling, we also need to get this ready ASAP so if we do it I'd suggest taking it on later.

mmatczuk:
Here we use dev it shall be replaced with edge now https://github.com/redpanda-data/connect/blob/25217935688e5784c6613734fd686105ba0fab47/internal/impl/redpanda/migrator/bench/docker-compose.yml#L75

mmatczuk (approved):

PR #3715, opened 2025-10-22
Some minor fixes around migrator

mmatczuk:


internal/impl/kafka/redpandatest/redpanda.go
```diff
@@ -103,9 +103,7 @@ func StartRedpanda(t *testing.T, pool *dockertest.Pool, exposeBroker, autocreate
 		ctx, done := context.WithTimeout(t.Context(), 3*time.Second)
 		defer done()
 
-		req, err := http.NewRequestWithContext(ctx, http.MethodGet,
-			fmt.Sprintf("http://127.0.0.1:%s/v1/cluster/health_overview", resource.GetPort("9644/tcp")),
-			nil)
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("http://localhost:%s/v1/cluster/health_overview", resource.GetPort("9644/tcp")), nil)
```
Jeffail, line 106: Is this revert intentional?

mmatczuk: Yes it's not that, I can reproduce that locally, it appears to me that sometimes the containers are not properly cleaned when test fails leading to more random failures like this.

PR #3716, opened 2025-10-23
mssqlservercdc: support snapshotting tables in parallel

josephwoodward:
This change add the config `max_parallel_snapshot_tables` (following the same config name in the [PostgreSQL CDC connector](https://docs.redpanda.com/redpanda-connect/components/inputs/postgres_cdc/#max_parallel_snapshot_tables), the MySQL CDC connector doesn't currently support snapshotting tables in parallel), this change allows users to specify the number of workers (aka go routines) available when snapshotting tables, with each worker having its own dedicated connection and transaction.

Naturally this means specifying a value greater than the default value of `1` will result in a stream of mixed table rows (which can be filtered via the `table` metadata).

In terms of performance, this means the snapshotting process will take as long as the largest table. Further options to reduce the amount of time snapshotting takes are to read and process batches from tables concurrently (similar to PostgreSQL CDC) or look at [incremental snapshotting](https://debezium.io/blog/2021/10/07/incremental-snapshots/) (like Debezium does).

Left is before, right is after:

image

### DIFF
--- a/internal/impl/mssqlserver/replication/snapshot.go
+++ b/internal/impl/mssqlserver/replication/snapshot.go
@@ -88,17 +88,17 @@ func (s *Snapshot) snapshotTable(ctx context.Context, table UserDefinedTable, ma
 
 		// BeginTx opens/reuses a dedicated connection for the given table-based transaction, using context.Background()
 		// because we want the transaction to be long lived. We explicitly rollback/commit it on function exit
-		if tx, err = s.db.BeginTx(context.Background(), &sql.TxOptions{Isolation: sql.LevelSnapshot}); err != nil {
+		if tx, err = s.db.BeginTx(ctx, &sql.TxOptions{Isolation: sql.LevelSnapshot}); err != nil {
 			return fmt.Errorf("starting snapshot transaction: %w", err)
 		}
 		defer func() {
 			if err != nil {
-				if rbErr := tx.Rollback(); rbErr != nil {
-					l.Errorf("Failed to rollback snapshot transaction: %v", rbErr)
-				}
-			} else {
-				if cmErr := tx.Commit(); cmErr != nil {
-					l.Errorf("Failed to commit snapshot transaction: %v", cmErr)
+				// sql package automatically rolls back transaction if context is cancelled
+				if !errors.Is(err, context.Canceled) {
+					if rbErr := tx.Rollback(); rbErr != nil {
+						l.Errorf("Failed to rollback snapshot transaction: %v", rbErr)
+					}
+					return
 				}
 			}
 		}()
@@ -183,6 +183,9 @@ func (s *Snapshot) snapshotTable(ctx context.Context, table UserDefinedTable, ma
 			}
 		}
 
+		if err := tx.Commit(); err != nil {
+			l.Errorf("Failed to commit snapshot transaction: %v", err)
+		}
 		s.snapshotStatusMetric.Set(1, tableName)
 		l.Infof("Table snapshot completed, %d rows processed", numRowsProcessed)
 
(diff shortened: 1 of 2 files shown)


docs/modules/components/pages/inputs/microsoft_sql_server_cdc.adoc
```diff
@@ -129,6 +131,15 @@ If set to true, the connector will query all the existing data as a part of snap
 *Type*: `bool`
 
 
+=== `max_parallel_snapshot_tables`
```
mmatczuk, line 134: Could this be automatically calculated bases on available SQL server CPUs and RPCN CPUs?

josephwoodward: Potentially, however I feel it's best left as it is for a few reasons:

- It aligns with existing SQL based CDC components (PostgreSQL also defaults to `1`).
- With a default value of 1, the message stream is serialised/in order which feels like a sensible default. If more than 1 reader is configured then the user will need to handle the out of order message stream by filtering on table name.

mmatczuk:
There is some naming discrepancy in the diff `input_mssqlserver_cdc.go` renames connectionString to connString while `snapshot.go` adds connectionString.

internal/impl/mssqlserver/replication/snapshot.go
```diff
@@ -84,103 +78,141 @@ func (s *Snapshot) Prepare(ctx context.Context) (LSN, error) {
 	return maxLSN, nil
 }
 
-// Read starts the process of iterating through each table, reading rows based on maxBatchSize, sending the row as a
+// Read launches N number of go routines (based on s.maxReaders) and starts the process of
```
mmatczuk: I might suggest API change to Read(ctx context.Context, maxWorkers, maxBatchSize int) to make it more explicit and flexible.

internal/impl/mssqlserver/replication/snapshot.go
```diff
@@ -84,103 +78,141 @@ func (s *Snapshot) Prepare(ctx context.Context) (LSN, error) {
...
+				return err
+			}
+			l.Tracef("Primary keys for table '%s': %v", table, tablePks)
+			lastSeenPksValues := map[string]any{}
+			for _, pk := range tablePks {
+				lastSeenPksValues[pk] = nil
 			}
 
-			values, mappers := prepSnapshotScannerAndMappers(types)
+			var totalRowsProcessed int
```
mmatczuk: Variable renames obfuscate the diff

internal/impl/mssqlserver/replication/snapshot.go
```diff
@@ -84,103 +78,141 @@ func (s *Snapshot) Prepare(ctx context.Context) (LSN, error) {
...
-			var batchRows *sql.Rows
-			if numRowsProcessed == 0 {
-				batchRows, err = s.querySnapshotTable(ctx, table, tablePks, nil, maxBatchSize)
-			} else {
-				batchRows, err = s.querySnapshotTable(ctx, table, tablePks, lastSeenPksValues, maxBatchSize)
-			}
-			if err != nil {
-				return fmt.Errorf("failed to execute snapshot table query: %s", err)
+	for _, table := range s.tables {
+		wg.Go(func() error {
```
mmatczuk: Could this be a named method with extra arguments?

internal/impl/mssqlserver/replication/snapshot.go
josephwoodward: That's a good suggestion 👍 Done in https://github.com/redpanda-data/connect/pull/3716/commits/0ba5d054bda1b6005d0d4cc3dab64af89abef66f

internal/impl/mssqlserver/replication/snapshot.go
josephwoodward: Updated in https://github.com/redpanda-data/connect/pull/3716/commits/c234e9af454711b182f7d40da2ff914f34e8c31b

internal/impl/mssqlserver/replication/snapshot.go
josephwoodward: Reverted this and the connString rename as to reduce scope of this change. Can always follow up with another change in some general refactoring as a separate PR.

mmatczuk:
Given that this is API change would be worth adding a CL entry.
Otherwise LGTM, please squash.

PR #3723, opened 2025-10-28
ffi: introduce ffi processor

rockwotj:
The wildest golang I've ever written.

This introduces a new FFI processor that can at runtime [dlopen](https://man7.org/linux/man-pages/man3/dlopen.3.html) and invoke functions that adhere to the C ABI.

Right now we only really support 3 types: `int32`, `int64` and `byte*` (which is `void*` in C land). This processor can be useful if you already have code in C/C++/Rust and either the code will be called in a high performance scenario or it's a simple enough function that wrapping it in gRPC is overkill.

### DIFF
--- a/internal/impl/ffi/impl/impl.go
+++ b/internal/impl/ffi/impl/impl.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package ffi
+package impl
 
 import (
 	"fmt"
@@ -23,45 +23,52 @@ import (
 	"github.com/redpanda-data/benthos/v4/public/bloblang"
 )
 
-// processorImpl takes a set of executors (one for each parameter)
-// and the message being operated on, and returns the output values
-// or an error.
-type processorImpl func(args []any) ([]any, error)
+// ForeignFunc invokes a C ABI function and returns the result.
+type ForeignFunc func(args []any) ([]any, error)
 
-type returnType string
-
-const (
-	returnTypeVoid  returnType = "void"
-	returnTypeInt32 returnType = "int32"
-	returnTypeInt64 returnType = "int64"
-)
-
-type paramType string
+// ReturnType the result of an FFI function
+type ReturnType string
 
 const (
-	paramTypeBytePtr paramType = "byte*"
-	paramTypeInt32   paramType = "int32"
-	paramTypeInt64   paramType = "int64"
+	// ReturnTypeVoid is a void return type in C
+	ReturnTypeVoid ReturnType = "void"
+	// ReturnTypeInt32 is a int32_t in C
+	ReturnTypeInt32 ReturnType = "int32"
+	// ReturnTypeInt64 is a int64_t in C
+	ReturnTypeInt64 ReturnType = "int64"
 )
 
-type parameterSpec struct {
-	Type paramType
+// ParamType is the type of a FFI function parameter
+type ParamType string
+
+const (
+	// ParamTypeBytePtr is a void* type in C
+	ParamTypeBytePtr ParamType = "byte*"
+	// ParamTypeInt32 is a int32_t in C
+	ParamTypeInt32 ParamType = "int32"
+	// ParamTypeInt64 is a int64_t in C
+	ParamTypeInt64 ParamType = "int64"
+)
+
+// ParameterSpec is a specification for a parameter of an FFI function.
+type ParameterSpec struct {
+	Type ParamType
 	Out  bool
 }
 
-// signature is a string that represents a specific ABI that is supported.
-type signature struct {
-	Return returnType
-	Params []parameterSpec
+// Signature is a string that represents a specific ABI that is supported.
+type Signature struct {
+	Return ReturnType
+	Params []ParameterSpec
 }
 
-// signatureImpl is an implementation of given FFI signature
+// specialization is an implementation of given FFI signature
 // from Bloblang/Connect to the foreign function and back.
-type signatureImpl struct {
-	signature signature
+type specialization struct {
+	signature Signature
 	// Given a symbol for a function that implements the signature,
 	// return a processor that uses the function
-	impl func(addr uintptr) processorImpl
+	impl func(addr uintptr) ForeignFunc
 }
 
 // The reflection based fallback approach is very slow.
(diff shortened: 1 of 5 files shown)


taskfiles/test.yml
```diff
@@ -42,3 +42,13 @@ tasks:
...
+
+  ffi:
+    desc: Run FFI tests
+    dir: internal/impl/ffi
+    vars:
+      TIMEOUT: '{{if .CI}}15m{{else}}5m{{end}}'
+    cmds:
+      # g++ is aliased to clang++ on darwin
+      - g++ -shared -fPIC testdata/plugin.cc -o plugin.{{if eq OS "darwin"}}dylib{{else}}so{{end}}
+      - go test {{.GO_FLAGS}} -timeout {{.TIMEOUT}} {{if .CI}}{{else}}-v{{end}} .
```
mmatczuk: Could we move it to where plugin.cc is?

We can invoke it automatically in the test - build it if not present.

internal/impl/ffi/testdata/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.so
```
mmatczuk, line 1: Would it make sense to build it and keep it in testdata, I think it belongs there. Then this file could be moved to testdata  too.

internal/impl/ffi/impl.go
```diff
@@ -0,0 +1,223 @@
...
+// processorImpl takes a set of executors (one for each parameter)
+// and the message being operated on, and returns the output values
+// or an error.
+type processorImpl func(args []any) ([]any, error)
+
+// signature is a string that represents a specific ABI that is supported.
+//
+// The syntax is as follows:
+// `<return type>(out? <input type>*?,+)`
+type signature string
```
mmatczuk: Could this be a struct? Say sth like  

```
type paramSignature string

type functionSignature struct {
   ret paramSignature
   in []paramSignature
   out []bool
}
```

Alternatively we could have a builder but it would be nice to separate the string operations from the actual implementation.

internal/impl/ffi/impl/impl.go
```diff
@@ -0,0 +1,223 @@
...
+// signatureImpl is an implementation of given FFI signature
+// from Bloblang/Connect to the foreign function and back.
+type signatureImpl struct {
+	signature signature
+	// Given a symbol for a function that implements the signature,
+	// return a processor that uses the function
+	impl func(addr uintptr) processorImpl
+}
+
+// The reflection based fallback approach is very slow.
```
mmatczuk, line 74: Is speed a factor here? It looks like this is called once in init.

internal/impl/ffi/impl.go
```diff
@@ -0,0 +1,223 @@
...
+		if supported.signature == sig {
+			return supported.impl(addr), nil
+		}
+	}
+	// The fallback processor is slower, but works with all our supported types
+	return makeFallbackProcessorImpl(sig, addr)
+}
+
+func makeFallbackProcessorImpl(sig signature, addr uintptr) (processorImpl, error) {
+	pos := strings.IndexByte(string(sig), '(')
```
mmatczuk: This looks quite fragile, as noted in my earlier comment would be better to put more information into type system.

internal/impl/ffi/impl.go
```diff
@@ -0,0 +1,223 @@
...
+	case "int32":
+		returnTypes = append(returnTypes, reflect.TypeFor[int32]())
+	case "int64":
+		returnTypes = append(returnTypes, reflect.TypeFor[int64]())
+	default:
+		return nil, fmt.Errorf("unexpected return type: %q", returnType)
+	}
+	params := strings.Split(strings.TrimSuffix(string(sig)[pos+1:], ")"), ",")
+	paramTypes := []reflect.Type{}
+	paramConverter := []func(any) (any, error){}
```
mmatczuk: Empty slice declaration should be with var.

internal/impl/ffi/impl/impl.go
```diff
@@ -0,0 +1,223 @@
...
+		returnTypes = append(returnTypes, reflect.TypeFor[int32]())
+	case "int64":
+		returnTypes = append(returnTypes, reflect.TypeFor[int64]())
+	default:
+		return nil, fmt.Errorf("unexpected return type: %q", returnType)
+	}
+	params := strings.Split(strings.TrimSuffix(string(sig)[pos+1:], ")"), ",")
+	paramTypes := []reflect.Type{}
+	paramConverter := []func(any) (any, error){}
+	outParameters := map[int]bool{}
```
mmatczuk, line 201: Set is typically implemented as `map[int]struct{}`

internal/impl/ffi/shlib_others.go
```diff
@@ -0,0 +1,40 @@
...
+
+package ffi
+
+import (
+	"errors"
+	"fmt"
+	"runtime"
+)
+
+type sharedLibrary struct{}
```
mmatczuk: Usually those files are named `XXX_others.go`

internal/impl/ffi/impl/impl.go
rockwotj: The returned callback does a bunch of reflection and `runtime.Pinner` stuff etc.

internal/impl/ffi/impl.go
rockwotj: done

internal/impl/ffi/shlib_others.go
rockwotj: done

internal/impl/ffi/impl/impl.go
rockwotj: meh, leaving.

internal/impl/ffi/impl.go
rockwotj: done

internal/impl/ffi/impl.go
rockwotj: done

rockwotj:
Thanks for the review @mmatczuk! Pushed a commit to address the feedback

internal/impl/ffi/impl.go
mmatczuk: We could also have type constants like https://github.com/JupiterRider/ffi/blob/main/types.go

internal/impl/ffi/impl/impl.go
mmatczuk: I see

mmatczuk:
I wonder would it make sense to create internal package with the real implementation to separate the processor logic from the actual ffi. Then we could have some more flexibility with public vs private members.

mmatczuk:
This is actually beyond awesome, now we can panic like a pro.

rockwotj:
> I wonder would it make sense to create internal package with the real implementation to separate the processor logic from the actual ffi. Then we could have some more flexibility with public vs private members.

Done.

rockwotj:
> This is actually beyond awesome, now we can panic like a pro.

Haha yeah, things can crash in glorious ways if folks get the signature wrong too.

PR #3724, opened 2025-10-28
Fix gateway input reload bug

vuldin:
_Relies on https://github.com/redpanda-data/benthos/pull/306_

### Problem
When using `--watcher` for config reloads with the `gateway` input, the service would often get stuck returning 503 errors and fail to process requests until manually restarted.

This occurred due to a race condition during reload:
1. The old server's TCP port wasn't fully released before the new server tried to bind
2. Port binding happened asynchronously in a goroutine, so `Connect()` returned success even when binding failed
3. Once in this state, the server was marked as "running but not ready" with no recovery mechanism

### Solution
Made these changes to `internal/impl/gateway/input.go`:

1. **Enable SO_REUSEADDR socket option** - This feature is provided by https://github.com/redpanda-data/benthos/pull/306
2. **Synchronous port binding** - `Connect()` now creates the TCP listener before returning, immediately surfacing any binding errors

### Testing
- Added `TestHTTPServerReload` that verifies the server can be closed and recreated on the same port without errors

### DIFF
--- a/internal/impl/gateway/input.go
+++ b/internal/impl/gateway/input.go
@@ -31,6 +31,7 @@ import (
 	"github.com/Jeffail/shutdown"
 
 	"github.com/redpanda-data/benthos/v4/public/service"
+	"github.com/redpanda-data/benthos/v4/public/utils/netutil"
 	"github.com/redpanda-data/connect/v4/internal/gateway"
 )
 
@@ -149,7 +150,8 @@ You can access these metadata fields using xref:configuration:interpolation.adoc
 					}),
 				service.NewMetadataFilterField(hsiFieldResponseExtractMetadata).
 					Description("Specify criteria for which metadata values are added to the response as headers."),
-			).
+			),
+			netutil.ListenerConfigSpec().
 				Description("Customize messages returned via xref:guides:sync_responses.adoc[synchronous responses].").
 				Advanced(),
 		)
@@ -176,6 +178,7 @@ type Input struct {
 	log  *service.Logger
 	mgr  *service.Resources
 
+	lc     netutil.ListenerConfig
 	mux    *mux.Router
 	server *http.Server
 
@@ -214,6 +217,10 @@ func InputFromParsed(pConf *service.ParsedConfig, mgr *service.Resources) (*Inpu
 		}
 	}
 
+	if h.lc, err = netutil.ListenerConfigFromParsed(pConf.Namespace("tcp")); err != nil {
+		return nil, fmt.Errorf("parse tcp config: %w", err)
+	}
+
 	return &h, nil
 }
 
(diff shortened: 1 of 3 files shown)


vuldin:
Apologies for the multiple force pushes. One was to fix a failed check and the other two were related to removing unnecessary modifications not related to the fix.

internal/impl/gateway/input.go
```diff
@@ -244,14 +245,35 @@ func (ri *Input) Connect(_ context.Context) error {
...
 
+	// Create a custom listener with SO_REUSEADDR to allow fast port reuse during reloads
+	lc := net.ListenConfig{
+		Control: func(_, _ string, c syscall.RawConn) error {
+			var sockOptErr error
+			if err := c.Control(func(fd uintptr) {
+				// Enable SO_REUSEADDR to allow binding to ports in TIME_WAIT state
+				sockOptErr = syscall.SetsockoptInt(int(fd), syscall.SOL_SOCKET, syscall.SO_REUSEADDR, 1)
+			}); err != nil {
+				return err
```
josephwoodward: Would it be helpful to wrap `err` and/or `sockOptErr` with a little more context to help distinguish between them if one or the other does error? ie:

```go
fmt.Errorf("setting socket option to reuse address: %w", sockOptErr)
```

josephwoodward:
> Connection state tracking - Added isConnected atomic flag for observability

Has this been removed, I don't see it in the change?

vuldin:
> > Connection state tracking - Added isConnected atomic flag for observability
> 
> Has this been removed, I don't see it in the change?

Yes, I removed this since it wasn't critical to the fix.

internal/impl/gateway/input.go
vuldin: https://github.com/redpanda-data/connect/blob/350edfd809ce483bb6eb1f090631c2f7d0a2b456/internal/impl/gateway/input.go#L256-L261

internal/impl/gateway/input.go
```diff
@@ -244,14 +245,38 @@ func (ri *Input) Connect(_ context.Context) error {
...
+				return fmt.Errorf("failed to set SO_REUSEADDR socket option: %w", sockOptErr)
+			}
+			return nil
+		},
+	}
+
+	listener, err := lc.Listen(context.Background(), "tcp", ri.conf.Address)
+	if err != nil {
+		ri.log.With("error", err, "address", ri.conf.Address).Error("Failed to bind to address")
+		return fmt.Errorf("failed to bind to address %s: %w", ri.conf.Address, err)
```
josephwoodward, line 261: We should call `listener.Close()` before returning here just to be safe, checking `listener` for nil

vuldin: Hmm, what would be the reason?

1. If Listen() fails → listener is nil/invalid, nothing to close, we return error
2. If Listen() succeeds → goroutine starts immediately, takes ownership of listener

I'm not seeing any code path where a listener would be created but then not used. Is there another scenario? Or is this just standard practice?

josephwoodward: No other scenario, looking through the `Listen` function I couldn't see any code path where a listener is returned alongside an error (this isn't always the case in the Go standard lib) though we can't guarantee that will always be the case so more a preference to code defensively in these instances. Not a blocker, happy to leave as is if everyone else is.

josephwoodward:
LGTM

mmatczuk:
Adding `SO_REUSEADDR ` socket option can be useful on it's own, I'd rather see it added to https://github.com/redpanda-data/benthos/tree/main/public/utils/netutil.

If the port binding is synchronous do we still need `SO_REUSEADDR` hack?

vuldin:
> Adding `SO_REUSEADDR ` socket option can be useful on it's own, I'd rather see it added to https://github.com/redpanda-data/benthos/tree/main/public/utils/netutil.
> 
> If the port binding is synchronous do we still need `SO_REUSEADDR` hack?

Excellent feedback, I've created https://github.com/redpanda-data/benthos/pull/306 that adds this to the Benthos dependency directly. I'll updated this PR now to make use of that change, and then put this PR into draft mode until the Benthos side is merged and available here.

vuldin:
Hmm, `task docs` will need to be ran after the benthos PR is merged and available. It fails with:
```
❯ task docs
task: [target-dir] mkdir -p 'target'
task: [build:redpanda-connect] go build  -tags "" -ldflags "-w -s -X main.Version=v4.68.0-4-gb12dae055 -X main.DateBuilt=2025-10-29T12:07:55Z" -o target/redpanda-connect ./cmd/redpanda-connect
# github.com/redpanda-data/connect/v4/internal/impl/gateway
internal/impl/gateway/input.go:249:27: undefined: netutil.ListenWithReuseAddr
exit status 1
```

mmatczuk:
I updated the PR but there is some bug in how benthos parses the config leading to unexpected errors

```
=== RUN   TestHTTPServerReload
    input_test.go:257: 
        	Error Trace:	/Users/michal/RP/RPCN/connect/internal/impl/gateway/input_test.go:257
        	            	parse listener config: field 'tcp.reuse_addr' was not found in the config
        	Test:       	TestHTTPServerReload
--- FAIL: TestHTTPServerReload (0.10s)
```

mmatczuk:
My bad the field was set in wrong place

PR #3727, opened 2025-10-29
redpanda: add field for topic exclusion

mmatczuk:
Add support for excluding topics when using regex-based consumption in
franz-go based inputs (redpanda, redpanda_migrator). Deprecate `regexp_topics`
field in favor of more explicit include syntax.


mmatczuk:
The field names ergonomics is not great, shall we enable regex by default and change field names for the unified redpanda type @Jeffail?

mmatczuk:
Added test for listing rules.

rockwotj (approved):
LGTM

mmatczuk:
Rebased.

PR #3745, opened 2025-11-04
postgress_cdc: support inlining SSL certificates

alextreichler:
As a user, in certain situations, it is not always practical or even probable to have access to the operating systems file system in order to pass the file and path for ssl certs. This PR adds the ability to provide the contents of the certs as a raw string if say using something like aws secrets to store the secret.

### DIFF
--- a/internal/impl/postgresql/pglogicalstream/connection.go
+++ b/internal/impl/postgresql/pglogicalstream/connection.go
@@ -30,16 +30,15 @@ func openPgConnectionFromConfig(dbDSN string, tlsConfig *tls.Config) (*sql.DB, e
 	return stdlib.OpenDB(*config.ConnConfig), nil
 }
 
-func getPostgresVersion(dbDSN string) (int, error) {
-	conn, err := openPgConnectionFromConfig(dbDSN, nil)
+func getPostgresVersion(dbDSN string, tls *tls.Config) (int, error) {
+	conn, err := openPgConnectionFromConfig(dbDSN, tls)
 	if err != nil {
 		return 0, fmt.Errorf("failed to connect to the database: %w", err)
 	}
 	defer conn.Close()
 
 	var versionString string
-	err = conn.QueryRow("SHOW server_version").Scan(&versionString)
-	if err != nil {
+	if err = conn.QueryRow("SHOW server_version").Scan(&versionString); err != nil {
 		return 0, fmt.Errorf("failed to execute query: %w", err)
 	}
 
--- a/internal/impl/postgresql/pglogicalstream/logical_stream.go
+++ b/internal/impl/postgresql/pglogicalstream/logical_stream.go
@@ -152,8 +152,7 @@ func NewPgStream(ctx context.Context, config *Config) (*Stream, error) {
 	}
 
 	var version int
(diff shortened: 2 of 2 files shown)


alextreichler:
Switched it to use the `service.NewTLSField` , which makes more sense.

internal/impl/postgresql/pglogicalstream/connection.go
```diff
@@ -9,20 +9,29 @@
...
+	if err != nil {
+		return nil, err
+	}
+	config.ConnConfig.TLSConfig = tlsConfig
+	return stdlib.OpenDB(*config.ConnConfig), nil
 }
 
 func getPostgresVersion(dbDSN string) (int, error) {
-	conn, err := openPgConnectionFromConfig(dbDSN)
+	conn, err := openPgConnectionFromConfig(dbDSN, nil)
```
rockwotj: This is required to pass in the TLSConfig as well otherwise the version checks will fail

internal/impl/postgresql/pglogicalstream/logical_stream.go
```diff
@@ -136,7 +136,7 @@ func NewPgStream(ctx context.Context, config *Config) (*Stream, error) {
 
 	if config.HeartbeatInterval > 0 {
 		stream.heartbeat, err = newHeartbeat(
-			config.DBRawDSN,
+			config,
 			config.HeartbeatInterval,
 			config.Logger,
```
rockwotj: remove these if passing in the full config

internal/impl/postgresql/pglogicalstream/logical_stream.go
```diff
@@ -123,7 +123,7 @@ func NewPgStream(ctx context.Context, config *Config) (*Stream, error) {
 		unchangedToastValue:   config.UnchangedToastValue,
 	}
 
-	monitor, err := NewMonitor(ctx, config.DBRawDSN, stream.logger, tables, stream.slotName, config.WalMonitorInterval)
+	monitor, err := NewMonitor(ctx, config, stream.logger, tables, stream.slotName, config.WalMonitorInterval)
```
rockwotj: Remove the last parameter if passing in the full config

internal/impl/postgresql/input_pg_stream.go
```diff
@@ -125,6 +124,7 @@ This input adds the following metadata fields to each message:
 			Example("0s").
 			Example("24h").
 			Advanced()).
+		Field(service.NewTLSField("tls")).
```
rockwotj, line 127: Can we override the description here and mention this overrides the SSL* settings in the environment and in the DSN?

rockwotj (approved):
I am okay with this, but would prefer to see it in the DSN.

rockwotj:
Also tests are failing

internal/impl/postgresql/pglogicalstream/connection.go
josephwoodward: Addressed in https://github.com/redpanda-data/connect/pull/3745/commits/ce786908becefe8706456784eac851f8624e4fad

Jeffail (approved):

PR #3749, opened 2025-11-05
redpanda/migrator: grow partitions if topic already exists

mmatczuk:


internal/impl/redpanda/migrator/migrator_topic.go
```diff
@@ -268,13 +268,26 @@ func (m *topicMigrator) createTopicLocked(ctx context.Context, srcAdm, dstAdm *k
...
+			dstCount := len(dstInfo.Partitions)
+
+			if srcCount > dstCount {
+				_, err := dstAdm.CreatePartitions(ctx, srcCount-dstCount, dstTopic)
+				if err != nil {
+					m.metrics.IncCreateErrors()
+					return fmt.Errorf("increase partitions for topic %q from %d to %d: %w", dstTopic, dstCount, srcCount, err)
+				}
+
+				m.log.Infof("Topic migration: increased partitions for destination topic '%s' from %d to %d", dstTopic, dstCount, srcCount)
```
josephwoodward, line 286: question (non-blocking): what are your thoughts on using structured logging in these instances? (ie, having the topic as a field)

mmatczuk: I'm all for it.

PR #3753, opened 2025-11-06
Migrator: various minor improvements

mmatczuk:


internal/impl/redpanda/migrator/migrator_schema_registry.go
```diff
@@ -448,6 +455,9 @@ func (m *schemaRegistryMigrator) Sync(ctx context.Context) error {
 		return fmt.Errorf("list subject schemas: %w", err)
 	}
 	m.log.Debugf("Schema migration: found %d subject schemas", len(all))
+	for _, s := range all {
+		m.log.Debugf("Schema migration: found subject schema %s version=%d id=%d", s.Subject, s.Version, s.ID)
```
josephwoodward, line 459:
```suggestion
		m.log.Debugf("Schema migration: found subject schema '%s' version=%d id=%d", s.Subject, s.Version, s.ID)
```

internal/impl/redpanda/migrator/migrator_test.go
```diff
@@ -0,0 +1,203 @@
...
+			name: "key_field_set",
+			conf: `
+input:
+  redpanda_migrator:
+    seed_brokers: ["source:9092"]
+    topics: ["orders"]
+    consumer_group: "migration"
+
+output:
+  redpanda_migrator:
```
malinskibeniamin, line 114: Shall we have a test with a `redpanda_migrator` output `metadata` field as well?

mmatczuk: When linting we should report fields that are not registered cc:@Jeffail. I could add metadata to banned fields but it's not going anywhere. There is no such thing as metadata in this component https://docs.redpanda.com/redpanda-connect/components/outputs/redpanda_migrator/.

internal/impl/redpanda/migrator/migrator_schema_registry.go
mmatczuk: It's consistent with how it's logged elsewhere

josephwoodward: The other's place holders are preceded by an `=` or `:` though (where the schema name is easily distinguishable) whereas this isn't. Either way not a blocker.

mmatczuk: @josephwoodward I'd send a followup patch to unify all of that.

PR #3762, opened 2025-11-10
AMQP 1.0 Output - New Configuration Options

timo102:
This pull request adds three new configuration options to the AMQP 1.0 output, giving users more control.

All new configuration options are optional and do not affect existing setups. If not specified, the output will continue to function as before.

1. **persistent**
   - Type: `bool`
   - Description: When enabled, this option marks outgoing messages as persistent. Persistent messages should be are stored by the broker and survive restarts or crashes, ensuring that important data is not lost.
   - Default: `false`

In ActiveMQ this is easy visible (persistent: true):
activemq_persistent_true

persistent: false
activemq_persistent_false

2. **target_capabilities**
   - Type: `[]string`
   - Description: This option allows you to specify which advanced features or capabilities you expect from the AMQP broker for the target address. Capabilities can include support for queues, topics, durability, sharing, or temporary destinations. Some AMQP brokers ignore this like ActiveMQ, so it really depends on the AMQP broker implementation.
   - Example use case: Requesting a durable queue for reliable event processing, or a shared topic for pub/sub scenarios.
   - Example: `["queue"]`

3. **message_props_to**
   - Type: `string`
   - Description: This option sets the 'To' property on outgoing AMQP messages, which can be used to direct messages to a specific node or endpoint, even if the transfer is handled by an intermediary. This option is dependent on the AMQP broker implementation to properly function.
   - Example use case: Routing messages to a specific microservice or partition in a multi-tenant architecture.
   - Example: `"amqp://analytics-service/ingest"`

In ActiveMQ you can see this in the binary message details, example `message_properties_to: amqp://xx:5672/`:
activemq_message_properties_to

### DIFF
--- a/docs/modules/components/pages/outputs/amqp_1.adoc
+++ b/docs/modules/components/pages/outputs/amqp_1.adoc
@@ -72,6 +72,9 @@ output:
     metadata:
       exclude_prefixes: []
     content_type: opaque_binary
+    persistent: false
+    target_capabilities: [] # No default (optional)
+    message_properties_to: amqp://localhost:5672/ # No default (optional)
 ```
 
 --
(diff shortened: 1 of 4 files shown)


josephwoodward:
Thanks for this @timo102

PR #3776, opened 2025-11-14
chore: go.mod add ignore directive

mmatczuk:
https://go.dev/ref/mod#go-mod-file-ignore

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,18 @@ go 1.25.2
 
 replace github.com/99designs/keyring => github.com/Jeffail/keyring v1.2.3
 
+ignore (
+	./bin
+	./config
+	./dist
+	./docs
+	./licenses
+	./proto
+	./resources
+	./target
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -4,6 +4,21 @@ go 1.25.2
 
 replace github.com/99designs/keyring => github.com/Jeffail/keyring v1.2.3
 
+ignore (
+	./.github
```
josephwoodward: Directories that start with a `.` are already ignored by go build. https://pkg.go.dev/cmd/go#hdr-Package_lists_and_patterns:~:text=Directory%20and%20file%20names%20that%20begin%20with%20%22.%22%20or%20%22_%22%20are%20ignored%20by%20the%20go%20tool%2C%20as%20are%20directories%20named%20%22testdata%22.

mmatczuk: Good catch.

PR #3786, opened 2025-11-19
Migrator: novel method for exact offset translation of empty consumer groups

mmatczuk:
Implement exact offset translation using embedded source offset headers
to handle non-monotonic timestamps and sub-millisecond precision cases.

### DIFF
--- a/internal/impl/redpanda/migrator/migrator_groups.go
+++ b/internal/impl/redpanda/migrator/migrator_groups.go
@@ -192,13 +192,14 @@ type GroupOffset struct {
 //   - Runs in one-shot or continuous sync modes
 //   - Provides metrics and caching for performance
 type groupsMigrator struct {
-	conf    GroupsMigratorConfig
-	src     *kgo.Client
-	srcAdm  *kadm.Client
-	dst     *kgo.Client
-	dstAdm  *kadm.Client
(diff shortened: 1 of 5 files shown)


internal/impl/redpanda/migrator/migrator_groups.go
```diff
@@ -730,30 +768,82 @@ func (m *groupsMigrator) translateOffset(
...
 //
-// The function does not retry the request if it fails, as the caller should
-// handle retries according to their needs.
-func readRecordTimestamp(
+//   - the exact offset is found (returns the refined destination offset)
+//   - the computed offset reaches the destination end offset eo (returns eo)
+//   - the computed offset exceeds bounds (returns unknownOffset with error)
+//   - the maximum number of attempts is exhausted (returns unknownOffset with error)
+//
+// This method should only be called when m.offsetHeader is not empty.
```
josephwoodward, line 785: question (non-blocking): Is this something the function should verify itself ( check `m.offsetHeader` is not empty) rather than trusting the caller?

mmatczuk: Nah

rockwotj (commented):
Nice - I like it.

PR #3791, opened 2025-11-20
postgres_cdc: support IAM authenticated users

josephwoodward:
This change adds support for IAM based authentication for Aurora PostgreSQL, following the same approach of reducing AWS SDK dependencies as the [opensearch component](https://github.com/redpanda-data/connect/blob/main/internal/impl/opensearch/aws/aws.go).

**How it works:**

Upon connection, a short lived token (~15 mins) is requested for the user and used as a DB password. When the token expires connections will remain active until the next reconnect attempt (ie, lost connections) at which point the component will disconnect and request a new token before resuming as part of the normal reconnect flow.

If we wanted to be smart we could introduce a pool of active connections that we monitor and swap out as needed, but this feels like a good start.

image

### DIFF
--- a/internal/impl/postgresql/input_pg_stream.go
+++ b/internal/impl/postgresql/input_pg_stream.go
@@ -44,13 +44,13 @@ const (
 	fieldUnchangedToastValue       = "unchanged_toast_value"
 	fieldHeartbeatInterval         = "heartbeat_interval"
 	fieldAWSIAMAuth                = "aws"
-	// FieldAWSEnabled enabled field.
-	FieldAWSEnabled = "enabled"
-	shutdownTimeout = 5 * time.Second
+	// FieldAWSIAMAuthEnabled enabled field.
+	FieldAWSIAMAuthEnabled = "enabled"
+	shutdownTimeout        = 5 * time.Second
 )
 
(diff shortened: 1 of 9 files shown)


mmatczuk:
@josephwoodward do we need to document the prerequisites to use it including the misinformation in Aurora docs?

internal/impl/postgresql/input_pg_stream.go
```diff
@@ -315,9 +357,19 @@ type pgStreamInput struct {
...
+	iamAuthEnabled bool
 }
 
 func (p *pgStreamInput) Connect(ctx context.Context) error {
+	// If IAM authentication is enabled, generate a new token
+	if p.iamAuthEnabled && p.streamConfig.RefreshAuthToken != nil {
+		if err := p.streamConfig.RefreshAuthToken(ctx); err != nil {
+			return fmt.Errorf("unable to generate IAM auth token: %w", err)
+		}
+	}
```
rockwotj, line 371: Just so I understand, every 15 minutes with AWS we will "reconnect" the whole component right? Are we properly surfacing "service.NotConnected" errors when our auth token expires? I just am wondering if we get noisy error logs, and maybe there are some performance implications of this too (or will we get duplicates when we reconnect and there is stuff in flight?).

This might be ok for now, but I suspect we will need to embed the token refresh and stream reconnection down deeper into the implementation.

PR #3802, opened 2025-11-24
redpanda/tracer: add cloud service account authentication support

birdayz:
## What

Integrate Redpanda Cloud service account authentication with the redpanda tracer.

## Why

PR #3792 added the redpanda tracer, but it didn't support the built-in cloud service account authentication introduced in PR #3688. This meant tracer configs in Redpanda Cloud still required explicit SASL credentials.

## Implementation details

Added `use_redpanda_cloud_service_account` config flag to the tracer. When enabled:

1. Fetches OAuth2 token source from `serviceaccount.GetTokenSource()` singleton
2. Wraps it as franz-go OAUTHBEARER SASL mechanism with automatic token refresh
3. Prepends OAuth mechanism to kafka connection options before creating exporter

Validation ensures cloud SA and explicit SASL config aren't used together, with clear error messages when the singleton isn't initialized (missing CLI flags).

This matches the pattern used by A2A processors and makes tracer configuration cleaner in cloud environments.

**Example config:**
```yaml
tracer:
  redpanda:
    seed_brokers: ["cloud-broker.redpanda.cloud:9092"]
    topic: "otel-traces"
    use_redpanda_cloud_service_account: true
```

**Run with:**
```bash
redpanda-connect run config.yaml \
  --x-redpanda-cloud-service-account-token-url="..." \
  --x-redpanda-cloud-service-account-client-id="..." \
  --x-redpanda-cloud-service-account-client-secret="..."
```

## References

- PR #3688: Built-in cloud service account config
- PR #3792: Redpanda tracer implementation

### DIFF
--- a/docs/modules/components/pages/tracers/redpanda.adoc
+++ b/docs/modules/components/pages/tracers/redpanda.adoc
@@ -83,6 +83,7 @@ tracer:
     format: json
     service: redpanda-connect
     tags: {}
+    use_redpanda_cloud_service_account: false
     sampling:
       enabled: false
       ratio: 0.05 # No default (optional)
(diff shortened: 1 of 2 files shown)


rockwotj:
Here is what I would propose instead of this: https://github.com/redpanda-data/connect/pull/3806

PR #3805, opened 2025-11-24
redpanda_migrator: add client timeout config for schema registry client

josephwoodward:
This change adds a timeout configuration for the Schema Registry HTTP Client, ensuring the default behaviour remains the same as Franz's default client (5 second timeout and setting the UA header). https://github.com/twmb/franz-go/blob/master/pkg/sr/client.go#L87-L88

### DIFF
--- a/internal/impl/redpanda/migrator/migrator_schema_registry.go
+++ b/internal/impl/redpanda/migrator/migrator_schema_registry.go
@@ -66,8 +66,9 @@ const (
 	srObjectField = "schema_registry"
 
 	// Schema registry fields
-	srFieldURL = "url"
-	srFieldTLS = "tls"
+	srFieldURL     = "url"
+	srFieldTimeout = "timeout"
+	srFieldTLS     = "tls"
 
 	// Schema registry migrator fields
 	srFieldEnabled        = "enabled"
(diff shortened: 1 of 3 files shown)


internal/impl/redpanda/migrator/migrator_schema_registry.go
```diff
@@ -66,8 +66,9 @@ const (
 	srObjectField = "schema_registry"
 
 	// Schema registry fields
-	srFieldURL = "url"
-	srFieldTLS = "tls"
+	srFieldURL     = "url"
+	srFieldTimeout = "timeout"
```
mmatczuk, line 70: IMO this should go below TLS

mmatczuk (approved):

internal/impl/redpanda/migrator/migrator_schema_registry.go
josephwoodward: The ordering follows that of the configuration/docs ([here](https://github.com/redpanda-data/connect/blob/jw/migrator_registry_timeout_support/internal/impl/redpanda/migrator/migrator_schema_registry.go?plain=1#L88-L98))

PR #3806, opened 2025-11-24
kafka: add redpanda cloud service account as authn mechanism

rockwotj:


internal/impl/kafka/sasl.go
```diff
@@ -198,6 +203,20 @@ func scram512SaslFromConfig(c *service.ParsedConfig) (sasl.Mechanism, error) {
 	}), nil
 }
 
+func redpandaCloudSaslFromConfig(_ *service.ParsedConfig) (sasl.Mechanism, error) {
+	tokenSource, err := serviceaccount.GetTokenSource()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get Redpanda Cloud service account token source: %w (this mechanism only works in Redpanda Cloud)", err)
```
mmatczuk: * `(this mechanism only works in Redpanda Cloud)` would be hard to read and relate to the 1st part of the message
* ideally we do not add `failed to` prefixes to error messages because when wrapping errors they add little value

rockwotj: simplified the error message. thanks!

PR #3813, opened 2025-11-27
add support to elasticsearch/v9

peczenyj:
The main intent of this PR is allow this internal improvement:

https://github.com/elastic/go-elasticsearch/blob/main/CHANGELOG.md#typed-api-1

```
This release introduces a new MethodAPI used by the TypedClient which makes the client friendlier for dead code elimination. Reducing the size of the client when only a subset of the APIs are used. The old API structure remains available for backward compatibility, but it is now deprecated.
```

to use it, we can just change the yaml tag and use the output `elasticsearch_v9`, should be backward compatible with the v8 tag

however, since we may keep elasticsearch v8 and v9 for a while, we may not benefit immediately from the dead code elimination but the main development today is on v9, it support earlier versions of go and I think we should consider add support on it.

since both codes are very similar, I did not try create like a "base" package to reuse the logic, but I can do if needed (or in another PR)

Regards

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -26,6 +26,7 @@ jobs:
           - ./internal/impl/cockroachdb
           # - ./internal/impl/couchbase
           - ./internal/impl/elasticsearch/v8
+          - ./internal/impl/elasticsearch/v9
           # - ./internal/impl/gcp
           - ./internal/impl/gcp/enterprise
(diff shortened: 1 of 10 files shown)


peczenyj:
No news :/

mmatczuk:
Rebased and updated.

mmatczuk:
Thank you @peczenyj for your awesome contributions!

PR #3814, opened 2025-11-27
postgresql_cdc: Support IAM assume role chaining

josephwoodward:
This PR copies the functionality for supporting chained IAM roles from the [recently merged MySQL CDC](https://github.com/redpanda-data/connect/pull/3801) change. It also adds support for specifying access keys in config for both the MySQL and PostgreSQL IAM features.

Once the dust settles on this, I'll look at moving some of this into the `internal/impl/aws` directory where it can be shared between PostgreSQL and MySQL as it's mostly identical.

### Initial Connection

image

### Reconnection:

image

### Integration Tests

image

### DIFF
--- a/internal/impl/mysql/aws/aws.go
+++ b/internal/impl/mysql/aws/aws.go
@@ -16,11 +16,11 @@ package aws
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	awsconfig "github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
 	"github.com/aws/aws-sdk-go-v2/feature/rds/auth"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
@@ -46,12 +46,13 @@ func awsIAMAuth(ctx context.Context, awsConf *service.ParsedConfig, dbConf *mysq
 	}
 
 	var (
-		err    error
-		awsCfg aws.Config
-
+		err         error
+		awsCfg      aws.Config
 		endpoint    string
 		region      string
 		roleConfigs []roleConfig
+
+		opts []func(*awsconfig.LoadOptions) error
 	)
 	if awsCfg, err = awsconfig.LoadDefaultConfig(ctx); err != nil {
 		return nil, fmt.Errorf("unable to load AWS config: %w", err)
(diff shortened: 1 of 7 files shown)


internal/impl/mysql/input_mysql_stream.go
```diff
@@ -100,6 +101,10 @@ This input adds the following metadata fields to each message:
 		service.NewIntField(fieldSnapshotMaxBatchSize).
 			Description("The maximum number of rows to be streamed in a single batch when taking a snapshot.").
 			Default(1000),
+		service.NewIntField(fieldMaxReconnectAttempts).
+			Description("The maximum number of attempts the MySQL driver will try to re-establish a broken connection before Connect attempts reconnection. A zero or negative number means infinite retry attempts.").
+			Advanced().
+			Default(0),
```
josephwoodward: The canal driver currently defaults to 0 so will retry indefinitely (which is wrong imo). I think we should set this to a much lower value (10 maybe?) and let Connect attempt to reconnect, otherwise MySQL could be retrying indefinitely and still report that it's healthy.

mmatczuk: I suggest to leave it exported set default to 10.

internal/impl/mysql/aws/aws.go
```diff
@@ -46,26 +46,35 @@ func awsIAMAuth(ctx context.Context, awsConf *service.ParsedConfig, dbConf *mysq
...
 		return nil, fmt.Errorf("unable to load AWS config: %w", err)
 	}
 	if endpoint, err = awsConf.FieldString("endpoint"); err != nil {
 		return nil, err
 	}
-	if region, err = awsConf.FieldString("region"); err != nil {
-		return nil, err
-	} else if region != "" {
-		awsCfg.Region = region
+	if region, _ = awsConf.FieldString("region"); region != "" {
```
mmatczuk, line 63: I'm not a huge fan of errors swallowing.

josephwoodward: Me neither, this appears to be the established approach throughout Connect for expressing optional strings without setting a default value (or having to handle the error to verify the string's it's not empty).

internal/impl/mysql/input_mysql_stream.go
josephwoodward: Reduced to 10, though I'm somewhat hesitant to even allow it to be configurable, I'm not sure how helpful it is to users versus letting Connect perform reconnections and reporting the correct state of the pipeline. It also adds confusion.

internal/impl/mysql/aws/aws.go
Jeffail: It's fairly harmless on config parse for fields that are optional, just avoids a bit of bloat in the parsing logic.

Jeffail (approved):

internal/impl/postgresql/aws/aws.go
```diff
@@ -16,56 +16,148 @@ package aws
...
-		region, _ = awsConf.FieldString("region")
-		if region != "" {
-			awsCfg.Region = region
+		password, err := auth.BuildAuthToken(ctx, endpoint, cfg.Region, dbConf.User, cfg.Credentials)
+		if err != nil {
+			return fmt.Errorf("building IAM auth token: %w", err)
 		}
-		if awsCfg.Region == "" {
-			return nil, errors.New("aws.region is required for IAM authentication")
+		dbConf.Password = password
```
mmatczuk, line 113: It feels like it's a data race, it is a race condition? Worth at lease adding a comment.

josephwoodward: I don't _think_ so (given it's only performed during connection before we assign the values to the canal river) though worth verifying. 

Holding a reference to the config could probably be removed (and the password returned at point of use) now I know what the MySQL and PostgreSQL versions look like. The two versions are practically identical so I'm going to look to refactor into a shared location in the aws package.

PR #3818, opened 2025-12-01
couchbase: add counter operations increment and decrement

sapk:
This PR add atomic counter operations support for Couchbase processor.

This is backward compatible as they are new operations.

Note: This might conflict with #2709 but I can rebase over the first to be merged.

### DIFF
--- a/docs/modules/components/pages/outputs/couchbase.adoc
+++ b/docs/modules/components/pages/outputs/couchbase.adoc
@@ -235,6 +235,10 @@ Couchbase operation to perform.
 |===
 | Option | Summary
 
+| `decrement`
+| decrement a counter by the value in content.
+| `increment`
+| increment a counter by the value in content.
 | `insert`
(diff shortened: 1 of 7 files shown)


sapk:
I updated the PR to have the ttl support also on increment and decrement operations since #3825 was merged. 

I also change to require and enforce that content is a valid integer. Previously any invalid or empty value set the delta to zero meaning a noop but it had the bad design of silencing error. 
If wanted we can set and document a default value for delta if content is empty. Either 1 to increment or decrement by 1 by default or 0 to be an noop by default (simply retrieve the existing value).

Also I simply return the error in process batch but maybe I could be more subtle and attach the error on the message and skip adding it to the ops list.

peczenyj:
At least this feature is available in the _other_ implementation

twmb:
Please rebase on main which will pick up the Claude review skill

PR #3823, opened 2025-12-04
Fix: address migrator integration tests that were failing due to new Redpanda version

josephwoodward:
Fixes the migrator integration tests that were failing due to either a new version of Redpanda or/and that it was caching the `latest` tagged container.

image

### DIFF
--- a/internal/impl/redpanda/migrator/integration_test.go
+++ b/internal/impl/redpanda/migrator/integration_test.go
@@ -219,6 +219,14 @@ func TestIntegrationMigratorSinglePartitionMalformedSchemaID(t *testing.T) {
 		require.NoError(t, err)
 	}
 
+	t.Log("And: Destination schema registry subject is set to import mode")
+	{
+		srDst, err := sr.NewClient(sr.URLs(dst.SchemaRegistryURL))
+		require.NoError(t, err)
+		modeRes := srDst.SetMode(t.Context(), sr.ModeImport, subj)
+		require.NoError(t, modeRes[0].Err)
+	}
+
 	pfx := []byte{0x00, 0x01, 0x02, 0x03, 0x04}
 
 	t.Log("When: Messages with malformed schema ID headers are written to source cluster")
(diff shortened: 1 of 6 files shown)


internal/impl/redpanda/migrator/integration_test.go
```diff
@@ -209,7 +210,9 @@ func TestIntegrationMigratorSinglePartitionMalformedSchemaID(t *testing.T) {
 	)
 
 	t.Log("Given: Redpanda clusters")
-	src, dst := startRedpandaSourceAndDestination(t)
+	src, dst := startRedpandaSourceAndDestination(t, func(rcok redpandatestConfigOptKind, c *redpandatest.Config) {
+		c.Nightly = true
```
mmatczuk: As 25.3 is released we do not need nightly anymore

josephwoodward: Yeah, this is just a stop gap as it's not pulling 25.3

internal/impl/redpanda/redpandatest/redpandatest.go
```diff
@@ -186,3 +194,38 @@ func StartRedpanda(t *testing.T, pool *dockertest.Pool, exposeBroker, autocreate
 	endpoints, _, err := StartSingleBrokerWithConfig(t, pool, cfg)
 	return endpoints, err
 }
+
+// getLatestRedpandaTag queries the GitHub API to get the latest Redpanda release tag.
+// This helps to avoid using 'latest' which will get cached
+func getLatestRedpandaTag() (string, error) {
```
mmatczuk: This is smart, but I'm not sure if it's very practical. How about a plain docker pull before starting the tests? If using the custom runner, I think we may observe docker image caching, especially with labels such as latest. You could argue that Redpanda is special in that sense, but then I think we need more comments/docs here.

josephwoodward: Based on our conversation I've moved pull of the latest image to the task file, so we'll keep an eye on it to see if it pulls the latest next version.

PR #3825, opened 2025-12-04
couchbase: add ttl (expiry) support

sapk:
The ttl (called expiry in couchbase) was implemented in couchbase cache implementation but was missing in processor and output.
This PR add a new field ttl to set this value (like it is done in cache implementation).


josephwoodward:
Thank you for your contribution @sapk

PR #3827, opened 2025-12-07
Update CI to use `gotestfmt` for clearer test output

josephwoodward:
This change adds [gotestfmt](https://github.com/GoTestTools/gotestfmt) to the test output for local and CI test context and providers clearer test output, making it easier to navigate failed tests.

You can check the output out here under `Test`: 
https://github.com/redpanda-data/connect/actions/runs/20029086927/job/57433439325?pr=3827

image

image

### DIFF
--- a/taskfiles/test.yml
+++ b/taskfiles/test.yml
@@ -7,6 +7,9 @@ tasks:
       - ut
     vars:
       TIMEOUT: '{{if .CI}}5m{{else}}1m{{end}}'
+    preconditions:
+      - sh: command -v {{.TOOLS_BIN_DIR}}/gotestfmt
+        msg: "gotestfmt is not installed. Run 'task tools:install-gotestfmt' first."
     cmds:
       - go test {{.GO_FLAGS}} -json -timeout {{.TIMEOUT}} -shuffle=on {{if .CI}}{{else}}-v{{end}} ./... 2>&1 | tee /tmp/gotest.log | {{.TOOLS_BIN_DIR}}/gotestfmt
 
@@ -14,6 +17,9 @@ tasks:
     desc: Run unit tests with race detection
     aliases:
       - ut-race
+    preconditions:
+      - sh: command -v {{.TOOLS_BIN_DIR}}/gotestfmt
+        msg: "gotestfmt is not installed. Run 'task tools:install-gotestfmt' first."
     cmds:
       - go test {{.GO_FLAGS}} -json -timeout 3m -shuffle=on -race {{if .CI}}{{else}}-v{{end}} ./... 2>&1 | tee /tmp/gotest.log | {{.TOOLS_BIN_DIR}}/gotestfmt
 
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -36,8 +36,11 @@ jobs:
     - name: Docs
       run: CGO_ENABLED=1 TAGS=x_benthos_extra task docs && test -z "$(git ls-files --others --modified --exclude-standard)" || { >&2 echo "Stale docs detected. This can be fixed with 'task docs'."; exit 1; }
 
+    - name: Install gotestfmt
```
mmatczuk: Since this needs to be built I wonder could we cache it?

.github/workflows/test.yml
```diff
@@ -36,8 +36,11 @@ jobs:
...
+      run: go install github.com/gotesttools/gotestfmt/v2/cmd/gotestfmt@v2.5.0
+
     - name: Test
-      run: task test
+      run: task test GO_FLAGS=-json
```
mmatczuk: We could also put that to taskfile directly using if CI statements

josephwoodward: As talked about earlier, this is now part of the tasks so it runs locally as well as in CI.

taskfiles/test.yml
```diff
@@ -5,29 +5,35 @@ tasks:
...
     cmds:
-      - go test {{.GO_FLAGS}} -timeout 3m -shuffle=on -race {{if .CI}}{{else}}-v{{end}} ./...
+      - go test {{.GO_FLAGS}} -json -timeout 3m -shuffle=on -race {{if .CI}}{{else}}-v{{end}} ./... 2>&1 | tee /tmp/gotest.log | {{.TOOLS_BIN_DIR}}/gotestfmt
 
   integration-package:
     desc: Run integration tests for package PKG
     aliases:
       - it
+    deps:
+      - :tools:install-gotestfmt
```
mmatczuk: This would cause compilation on every test run. Overall it feels like the tools file needs a refactor from install-X to just X with caching. Alternatively, for simplicity, we could drop the deps.

josephwoodward: Yes it would. Alternatively we could check to see if the binary exists and if not return a message informing the user to install dependencies.

josephwoodward: Looking into this further, I've added a precondition step that will verify if `gotestfmt` binary is available in the `bin` directory, otherwise it'll return a notice informing the user to install it first.

taskfiles/test.yml
```diff
@@ -7,15 +7,21 @@ tasks:
       - ut
     vars:
       TIMEOUT: '{{if .CI}}5m{{else}}1m{{end}}'
+    preconditions:
```
mmatczuk, line 10: 💯

mmatczuk:
Close

PR #3828, opened 2025-12-08
add mcp metrics

alenkacz:


alenkacz:
@rockwotj claude made the updates, can you check it out? :)

internal/mcp/metrics/metrics.go
```diff
@@ -0,0 +1,131 @@
+// Copyright 2024 Redpanda Data, Inc.
```
rockwotj: Image

internal/mcp/metrics/metrics.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		m.toolInvocations.Incr(1, toolName, "error")
+		m.toolErrors.Incr(1, toolName, "invocation_error")
+	} else {
+		m.toolInvocations.Incr(1, toolName, "success")
+
+		// Track response size
+		if respBytes, err := json.Marshal(result); err == nil {
+			m.toolResponseSize.Timing(int64(len(respBytes)), toolName)
+		}
```
rockwotj: feels a bit heavy to marshal things twice, do we really need this metric here?

internal/mcp/metrics/metrics.go
```diff
@@ -0,0 +1,131 @@
...
+			return m.handleToolCall(ctx, next, req, start)
+		}
+
+		// Call the next handler
+		result, err = next(ctx, method, req)
+
+		// Track response metrics
+		m.messagesSent.Incr(1, method)
+		if err != nil {
+			m.toolErrors.Incr(1, method, "method_error")
```
rockwotj: This is not a tool

internal/mcp/metrics/metrics.go
```diff
@@ -0,0 +1,131 @@
...
+	result, err = next(ctx, "tools/call", req)
+
+	// Track execution duration
+	m.toolExecutionDuration.Timing(time.Since(start).Nanoseconds(), toolName)
+
+	// Track response
+	m.messagesSent.Incr(1, "tools/call")
+	if err != nil {
+		m.toolInvocations.Incr(1, toolName, "error")
+		m.toolErrors.Incr(1, toolName, "invocation_error")
```
rockwotj: I don't understand why we need both metrics here

alenkacz:
@rockwotj addressed

internal/mcp/metrics/metrics.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/redpanda-data/benthos/v4/public/service"
+)
+
+// Metrics contains counters, gauges, and timers for tracking MCP operations.
+type Metrics struct {
+	// Tool metrics
+	toolInvocations          *service.MetricCounter
+	toolExecutionDuration    *service.MetricTimer
+	toolConcurrentExecutions *service.MetricGauge
+	toolRequestSize          *service.MetricTimer
```
rockwotj: Request size should not be a timer

internal/mcp/metrics/metrics.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// Track concurrent executions
+	m.toolConcurrentExecutions.Incr(1, toolName)
+	defer m.toolConcurrentExecutions.Decr(1, toolName)
+
+	// Track request size
+	if reqBytes, err := json.Marshal(req); err == nil {
+		m.toolRequestSize.Timing(int64(len(reqBytes)), toolName)
+	}
+
```
rockwotj: If we're willing to give up response size, do we need request size?

internal/mcp/mcp.go
```diff
@@ -156,6 +157,10 @@ func NewServer(
 
 	license.RegisterService(resources, licenseConfig)
 
+	// Add metrics middleware to track all MCP method calls
+	mcpMetrics := metrics.NewMetrics(resources.Metrics())
+	s.AddReceivingMiddleware(mcpMetrics.Middleware)
```
rockwotj: You need to handle SendingMiddleware too for server initiated RPCs

alenkacz:
@rockwotj okay, 🤞 now :)

internal/mcp/metrics/metrics.go
```diff
@@ -0,0 +1,125 @@
...
+		// Message metrics
+		messagesReceived: m.NewCounter("mcp_messages_received_total", "method"),
+		messagesSent:     m.NewCounter("mcp_messages_sent_total", "method"),
+	}
+}
+
+// ReceivingMiddleware returns an MCP method handler that tracks metrics for client-initiated RPC calls.
+func (m *Metrics) ReceivingMiddleware(next mcp.MethodHandler) mcp.MethodHandler {
+	return func(ctx context.Context, method string, req mcp.Request) (result mcp.Result, err error) {
+		start := time.Now()
```
rockwotj: nit: move this into handle tool call

internal/mcp/metrics/metrics.go
```diff
@@ -0,0 +1,125 @@
...
+		return result, err
+	}
+}
+
+// handleToolCall handles metrics for tool invocations specifically.
+func (m *Metrics) handleToolCall(ctx context.Context, next mcp.MethodHandler, req mcp.Request, start time.Time) (result mcp.Result, err error) {
+	// Extract tool name from request
+	toolName := extractToolName(req)
+	if toolName == "" {
+		toolName = "unknown"
```
rockwotj: mega nit: probably could move this into the method below instead of translating empty string

rockwotj (approved):

internal/mcp/metrics/metrics.go
```diff
@@ -0,0 +1,125 @@
...
+// SendingMiddleware returns an MCP method handler that tracks metrics for server-initiated RPC calls.
+func (m *Metrics) SendingMiddleware(next mcp.MethodHandler) mcp.MethodHandler {
+	return func(ctx context.Context, method string, req mcp.Request) (result mcp.Result, err error) {
+		m.messagesSent.Incr(1, method)
+
+		// Call the next handler
+		result, err = next(ctx, method, req)
+
+		// Track response metrics
+		m.messagesReceived.Incr(1, method)
```
rockwotj: I this is not really a response, but if it is then so we need both send and recv metrics that are always basically the same?

PR #3848, opened 2025-12-15
changelog: sort results by Path

mmatczuk:
Without this patch the order of changes may change between runs.

### DIFF
--- a/internal/impl/changelog/bloblang.go
+++ b/internal/impl/changelog/bloblang.go
@@ -15,6 +15,10 @@
 package changelog
 
 import (
+	"fmt"
+	"slices"
+	"strings"
+
 	"github.com/go-viper/mapstructure/v2"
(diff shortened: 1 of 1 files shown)


internal/impl/changelog/bloblang.go
```diff
@@ -62,6 +66,29 @@ func init() {
...
 			}
 
+			// Sort the result by Path for stable output
+			pathAsString := func(m map[string]any) string {
+				pathVal, ok := m["Path"]
+				if !ok {
+					return ""
+				}
+				switch p := pathVal.(type) {
+				case []interface{}:
```
rockwotj, line 76: someone needs to tell claude about `any`

mmatczuk: Eh, good catch

PR #3854, opened 2025-12-16
License metric

mmatczuk:


internal/license/service.go
```diff
@@ -130,10 +130,19 @@ func InjectCustomLicenseBytes(res *service.Resources, conf Config, licenseBytes
...
+
 	setSharedService(res, s)
 	return nil
 }
 
+func setLicenseMetrics(res *service.Resources, license RedpandaLicense) {
+	if !license.AllowsEnterpriseFeatures() {
+		return
+	}
+	res.Metrics().NewGauge("license_expiry_timestamp").Set(license.Expiry)
```
josephwoodward: There's a Prometheus naming convention of having the unit at the end of the metric (in this case `_seconds`).

See the example `data_pipeline_last_record_processed_timestamp_seconds ` in their practices docs here: https://prometheus.io/docs/practices/naming/

mmatczuk: Thanks!

internal/license/service.go
```diff
@@ -130,10 +130,19 @@ func InjectCustomLicenseBytes(res *service.Resources, conf Config, licenseBytes
...
+
 	setSharedService(res, s)
 	return nil
 }
 
+func setLicenseMetrics(res *service.Resources, license RedpandaLicense) {
+	if !license.AllowsEnterpriseFeatures() {
+		return
+	}
+	res.Metrics().NewGauge("license_expiry_timestamp_seconds").Set(license.Expiry)
```
rockwotj: would be nice to align with the broker: https://docs.redpanda.com/current/get-started/licensing/check-status/metrics/ (so make this the delta between now and expiry and update it every ~hour or something)

Basically it makes it easier to write an alert to fire if this becomes less than 1 week in seconds

mmatczuk: While this is true this metric is out of date every second - how about using `now()` in alerts?

rockwotj: It works in some observability platforms, but not others I think. Some like datadog I think are notoriously inexpessive in terms of defining alerts

mmatczuk: I'll redo it to spawn update loop and change it accordingly

internal/license/service.go
```diff
@@ -129,11 +132,54 @@ func InjectCustomLicenseBytes(res *service.Resources, conf Config, licenseBytes
...
+		return
+	}
+
+	if s.expiryMetric == nil {
+		s.expiryMetric = res.Metrics().NewGauge("redpanda_cluster_features_enterprise_license_expiry_sec",
+			"Seconds remaining until the enterprise license expires.",
+		)
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	s.cancel = cancel
```
rockwotj, line 157: I guess this doesn't get cleaned up, but I guess that's fine for now (and would take a bunch of wiring to cleanup for unit tests)?

Alternatively, there is a seperate method to start the monitoring loop that we can run in the CLI main method and then tests aren't effected.

PR #3858, opened 2025-12-17
add separate observability endpoint for mcp

alenkacz:
this is needed so you can run authenticated endopoint for mcp and then unauthenticated observability one. This should copy what pipelines are doing too

### DIFF
--- a/internal/cli/mcp_server.go
+++ b/internal/cli/mcp_server.go
@@ -30,6 +30,10 @@ func mcpServerCli(rpMgr *enterprise.GlobalRedpandaManager) *cli.Command {
 			Name:  "address",
 			Usage: "An optional address to bind the MCP server to instead of running in stdio mode.",
 		},
+		&cli.StringFlag{
+			Name:  "observability-address",
+			Usage: "Address to bind the observability server (metrics, pprof) to. If not set, observability server is disabled. Only applies when --address is set.",
+		},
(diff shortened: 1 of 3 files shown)


internal/cli/mcp_server.go
```diff
@@ -30,6 +30,11 @@ func mcpServerCli(rpMgr *enterprise.GlobalRedpandaManager) *cli.Command {
 			Name:  "address",
 			Usage: "An optional address to bind the MCP server to instead of running in stdio mode.",
 		},
+		&cli.StringFlag{
```
birdayz, line 33: should we use the same port as in normal pipeline mode? 4195 afaik

alenkacz: since it's metrics only I find 6060 the more standardized number but no strong opinion

rockwotj:
Why do we need this? can't we use the standard o11y/metrics.yaml file to configure metrics like we do for pipelines?

alenkacz:
@rockwotj  but how do we expose it? we currently expose only one port and that one is the outside facing MCP port that's authenticated 🤔

internal/cli/mcp_server.go
rockwotj: Can we default to empty string which means off?

rockwotj (approved):

PR #3874, opened 2025-12-23
Extract elasticsearch common code between v8 and v9

peczenyj:
I think I can do better with some feedback


internal/impl/elasticsearch/common/interface.go
```diff
@@ -0,0 +1,85 @@
...
+// ToV9Configuration convert to elasticsearch v9 configuration.
+func (c *Config) ToV9Configuration() elasticsearch_v9.Config {
+	return elasticsearch_v9.Config{
+		Addresses: c.Addresses,
+		Username:  c.Username,
+		Password:  c.Password,
+		Transport: c.Transport,
+		Logger:    c.Logger,
+	}
+}
```
mmatczuk, line 82: Consider moving that to v8 and v9 packages respectively as converting constructors.

internal/impl/elasticsearch/v8/output.go
```diff
@@ -13,409 +13,103 @@
...
-	"github.com/elastic/elastic-transport-go/v8/elastictransport"
-	"github.com/elastic/go-elasticsearch/v8"
-	"github.com/elastic/go-elasticsearch/v8/typedapi/core/bulk"
-	"github.com/elastic/go-elasticsearch/v8/typedapi/types"
+	elasticsearch_v8 "github.com/elastic/go-elasticsearch/v8"
+	elasticsearch_v8_bulk "github.com/elastic/go-elasticsearch/v8/typedapi/core/bulk"
+	elasticsearch_v8_types "github.com/elastic/go-elasticsearch/v8/typedapi/types"
 
 	"github.com/redpanda-data/benthos/v4/public/service"
+	elasticsearch_common "github.com/redpanda-data/connect/v4/internal/impl/elasticsearch/common"
```
mmatczuk, line 28: Consider moving to elasticsearch directly for more ergonomic imports.

mmatczuk:
Thanks but keeping them separate is actually a good thing for now.

PR #3882, opened 2026-01-05
Add SQS Visibility timeout field for S3 input

squiidz:
Set queue-level visibility timeout for nack handling (not per-message)
The default is set to 30s as per the AWS [doc](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html)

### DIFF
--- a/internal/impl/aws/input_s3.go
+++ b/internal/impl/aws/input_s3.go
@@ -41,15 +41,15 @@ import (
 
 const (
 	// S3 Input SQS Fields
-	s3iSQSFieldURL             = "url"
-	s3iSQSFieldEndpoint        = "endpoint"
-	s3iSQSFieldEnvelopePath    = "envelope_path"
-	s3iSQSFieldKeyPath         = "key_path"
-	s3iSQSFieldBucketPath      = "bucket_path"
-	s3iSQSFieldDelayPeriod     = "delay_period"
-	s3iSQSFieldMaxMessages     = "max_messages"
-	s3iSQSFieldWaitTimeSeconds = "wait_time_seconds"
-	s3iSQSVisibilityTimeout    = "visibility_timeout"
+	s3iSQSFieldURL              = "url"
+	s3iSQSFieldEndpoint         = "endpoint"
+	s3iSQSFieldEnvelopePath     = "envelope_path"
+	s3iSQSFieldKeyPath          = "key_path"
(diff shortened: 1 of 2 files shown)


internal/impl/aws/input_s3.go
```diff
@@ -221,6 +226,10 @@ You can access these metadata fields using xref:configuration:interpolation.adoc
 					Description("Whether to set the wait time. Enabling this activates long-polling. Valid values: 0 to 20.").
 					Default(0).
 					Advanced(),
+				service.NewIntField(s3iSQSVisibilityTimeout).
+					Description("Custom SQS Visibility timeout in seconds. Default is 30s").
+					Default(30).
```
josephwoodward: Thought (non-blocking): Given the [client's default visibility is 30 seconds](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html#:~:text=default%20visibility%20timeout%20for%20a%20queue%20is%2030%20seconds), it might be better to fallback to the client's default value instead of setting it explicitly.

Jeffail: Because this is for the nack path explicitly we need to default to 0 in order to preserve compatibility, we should also alter the description somewhat to make it clear that this visibility timeout is set when messages are nacked/rejected as I can imagine people confusing this as a timeout reset when messages are consumed and are being processed.

internal/impl/aws/input_s3.go
```diff
@@ -49,6 +49,7 @@ const (
 	s3iSQSFieldDelayPeriod     = "delay_period"
 	s3iSQSFieldMaxMessages     = "max_messages"
 	s3iSQSFieldWaitTimeSeconds = "wait_time_seconds"
+	s3iSQSVisibilityTimeout    = "visibility_timeout"
```
Jeffail: I'd suggest we maybe call this `nack_visibility_timeout`. Although I'm also not sure of any scenario where we would set a visibility timeout on a message that _isn't_ being nacked.

squiidz: I renamed it nack_visibility_timeout and set the default value to 0

internal/impl/aws/input_s3.go
squiidz: I've set it to 0 as per jeffail comment

PR #3893, opened 2026-01-09
otlp/otlpconv: bidirectional conversion between OpenTelemetry Collector OTLP format and Redpanda OTEL v1 protobuf format

mmatczuk:
# Format Differences

OTLP Format (OpenTelemetry Collector):
  - Batched structure: ResourceSpans → ScopeSpans → []Span
  - Resource and Scope metadata shared at batch level
  - Efficient for network transmission (reduced redundancy)
  - Types: ptraceotlp.ExportRequest, plogotlp.ExportRequest, pmetricotlp.ExportRequest
  - Package: go.opentelemetry.io/collector/pdata

Redpanda OTEL Format:
  - Individual records: Each signal is self-contained
  - Resource and Scope embedded in every message
  - Optimized for Kafka partitioning (one signal per record)
  - Types: pb.Span, pb.LogRecord, pb.Metric
  - Package: buf.build/gen/go/redpandadata/otel/protocolbuffers/go/redpanda/otel/v1

# Conversion Directions

Direction 1: OTLP → Redpanda (Unbatching)
  - Extracts individual signals from batched OTLP format
  - Embeds Resource and Scope metadata into each signal
  - Use cases: OTLP input → Kafka output, pipeline processing

Direction 2: Redpanda → OTLP (Batching)
  - Groups individual signals by Resource and Scope
  - Creates efficient batched OTLP structure
  - Use cases: Kafka input → OTLP output, aggregation

# Data Preservation

All conversions preserve complete telemetry data:
  - Trace IDs, Span IDs (16-byte and 8-byte arrays)
  - Timestamps (nanosecond precision)
  - Attributes (all AnyValue types including nested structures)
  - Metadata (schema URLs, dropped counts, flags)
  - Span events, links, status
  - Metric data points, exemplars, aggregation types
  - Log severity, body, trace context

PR #3895, opened 2026-01-12
otlp: add gRPC and HTTP input components

mmatczuk:
Add OpenTelemetry Protocol (OTLP) input components supporting both gRPC
and HTTP transports for ingesting traces, logs, and metrics.

- Add otlp_grpc input on default port 4317
- Add otlp_http input on default port 4318 with /v1/traces, /v1/logs, and
  /v1/metrics endpoints
- Convert OTLP protobuf format to Redpanda OTEL v1 protobuf messages
- Unbatch export requests into individual messages (one per span/log/metric)
- Support optional rate limiting via rate limit resources
- Add TLS configuration for both transports
- Add TCP socket configuration (SO_REUSEADDR, SO_REUSEPORT)

### DIFF
--- a/internal/impl/otlp/input_grpc.go
+++ b/internal/impl/otlp/input_grpc.go
@@ -16,6 +16,7 @@ package otlp
 
 import (
 	"context"
+	"crypto/subtle"
 	"crypto/tls"
 	"errors"
 	"fmt"
@@ -28,6 +29,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 
 	pb "github.com/redpanda-data/common-go/redpanda-otel-exporter/proto"
@@ -40,6 +42,7 @@ import (
 const (
 	giFieldAddress        = "address"
 	giFieldTLS            = "tls"
+	giFieldAuthToken      = "auth_token"
 	giFieldMaxRecvMsgSize = "max_recv_msg_size"
 	giFieldRateLimit      = "rate_limit"
 
@@ -50,6 +53,7 @@ const (
 type grpcInputConfig struct {
 	Address        string
 	TLS            tlsConfig
+	AuthToken      string
 	MaxRecvMsgSize int
 	RateLimit      string
 	ListenerConfig netutil.ListenerConfig
@@ -82,6 +86,30 @@ Each OTLP export request is unbatched into individual messages:
 Messages are encoded in Redpanda OTEL v1 protobuf format with metadata:
 - `+"`signal_type`"+`: "trace", "log", or "metric"
 
+## Authentication
+
+When `+"`auth_token`"+` is configured, clients must include the token in the gRPC metadata:
+
+**Go Client Example:**
+`+"```go"+`
+import (
+    "go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+)
+
+exporter, err := otlptracegrpc.New(ctx,
+    otlptracegrpc.WithEndpoint("localhost:4317"),
+    otlptracegrpc.WithInsecure(), // or WithTLSCredentials() for TLS
+    otlptracegrpc.WithHeaders(map[string]string{
+        "authorization": "Bearer your-token-here",
+    }),
+)
+`+"```"+`
+
+**Environment Variable:**
+`+"```bash"+`
+export OTEL_EXPORTER_OTLP_HEADERS="authorization=Bearer your-token-here"
+`+"```"+`
+
 ## Rate Limiting
 
 An optional rate limit resource can be specified to throttle incoming requests. When the rate limit is breached, requests will receive a ResourceExhausted gRPC status code.
@@ -93,6 +121,11 @@ An optional rate limit resource can be specified to throttle incoming requests. 
 			service.NewObjectField(giFieldTLS,
 				tlsFields()...,
 			).Description("TLS configuration for gRPC.").
+				Advanced(),
+			service.NewStringField(giFieldAuthToken).
+				Description("Optional bearer token for authentication. When set, requests must include 'authorization: Bearer <token>' metadata.").
+				Default("").
+				Secret().
 				Advanced(),
 			service.NewIntField(giFieldMaxRecvMsgSize).
 				Description("Maximum size of gRPC messages to receive in bytes.").
@@ -137,6 +170,11 @@ func GRPCInputFromParsed(pConf *service.ParsedConfig, mgr *service.Resources) (s
 		if conf.TLS, err = parseTLSConfig(pConf.Namespace(giFieldTLS)); err != nil {
 			return nil, err
 		}
+	}
+
+	// Parse auth token
+	if conf.AuthToken, err = pConf.FieldString(giFieldAuthToken); err != nil {
+		return nil, err
 	}
 
 	// Parse netutil listener config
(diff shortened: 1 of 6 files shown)


mmatczuk:
Requires https://github.com/redpanda-data/connect/pull/3893 merged first.

internal/impl/otlp/input.go
```diff
@@ -0,0 +1,129 @@
...
+
+type otlpInput struct {
+	log       *service.Logger
+	mgr       *service.Resources
+	rateLimit string
+	resCh     chan asyncMessage
+	shutSig   *shutdown.Signaller
+}
+
+func makeOTLPInput(mgr *service.Resources, rateLimit string) otlpInput {
```
josephwoodward: Why `makeOTLPInput` instead of following the Go convention of `newOTLPInput`?

internal/impl/otlp/input.go
```diff
@@ -0,0 +1,129 @@
...
+		mgr:       mgr,
+		rateLimit: rateLimit,
+		resCh:     make(chan asyncMessage),
+		shutSig:   shutdown.NewSignaller(),
+	}
+}
+
+// maybeWaitForAccess blocks until the rate limiter grants access or the
+// context/shutdown signals. If no rate limit is configured, it returns
+// immediately. It must be called before calling [sendMessage].
```
josephwoodward: Should this be `sendMessageBatch`?
```suggestion
// immediately. It must be called before calling [sendMessageBatch].
```

internal/impl/otlp/input.go
```diff
@@ -0,0 +1,129 @@
...
+
+type asyncMessage struct {
+	msg   service.MessageBatch
+	ackFn service.AckFunc
+}
+
+type otlpInput struct {
+	log       *service.Logger
+	mgr       *service.Resources
+	rateLimit string
```
josephwoodward, line 34: Minor (non-blocking): Given this value is the `name` param on the rate limit component, would `rateLimitName` be a clearer name?

internal/impl/otlp/input.go
```diff
@@ -0,0 +1,129 @@
...
+			d   time.Duration
+			err error
+		)
+		if rerr := o.mgr.AccessRateLimit(ctx, o.rateLimit, func(rl service.RateLimit) {
+			d, err = rl.Access(ctx)
+		}); rerr != nil {
+			err = rerr
+		}
+		if err != nil {
+			o.log.Errorf("Rate limit error: %v\n", err)
```
josephwoodward:
```suggestion
			o.log.Errorf("Rate limit error: %v", err)
```

I don't think we need the newline here, or was this intentional?

internal/impl/otlp/input.go
```diff
@@ -0,0 +1,129 @@
...
+	case <-ctx.Done():
+		return nil, nil, ctx.Err()
+	case <-o.shutSig.HasStoppedChan():
+		return nil, nil, service.ErrEndOfInput
+	case am := <-o.resCh:
+		return am.msg, am.ackFn, nil
+	}
+}
+
+const gracefulShutdownTimeout = 5 * time.Second
```
josephwoodward: minor (non-blocking): given this is only used in the `input_grpc.go` file, would it be worth moving it to the top of that file as opposed to in here at the bottom?

internal/impl/otlp/input_grpc.go
```diff
@@ -0,0 +1,501 @@
...
+	giFieldAuthToken      = "auth_token"
+	giFieldMaxRecvMsgSize = "max_recv_msg_size"
+	giFieldRateLimit      = "rate_limit"
+
+	defaultGRPCAddress    = "0.0.0.0:4317"
+	defaultMaxRecvMsgSize = 4 * 1024 * 1024 // 4MB
+)
+
+type grpcInputConfig struct {
+	Address        string
```
josephwoodward, line 54: Do these fields need to be exported?

internal/impl/otlp/input_grpc.go
```diff
@@ -0,0 +1,501 @@
...
+
+// GRPCInputFromParsed creates an OTLP gRPC input from a parsed config.
+func GRPCInputFromParsed(pConf *service.ParsedConfig, mgr *service.Resources) (service.BatchInput, error) {
+	var (
+		conf grpcInputConfig
+		err  error
+	)
+
+	// Parse gRPC-specific config
+	conf.Address, err = pConf.FieldString(giFieldAddress)
```
josephwoodward: Minor (non-blocking): Very subjective so feel free to ignore, but these seem like a good candidate for inlining, ie:

```go
if conf.Address, err = pConf.FieldString(giFieldAddress); err != nil {
		return nil, err
}
```

internal/impl/otlp/input_grpc.go
```diff
@@ -0,0 +1,501 @@
...
+
+// Export implements the gRPC Export method for traces
+func (s *traceServiceServer) Export(ctx context.Context, req ptraceotlp.ExportRequest) (ptraceotlp.ExportResponse, error) {
+	if err := s.validateAuth(ctx); err != nil {
+		return ptraceotlp.NewExportResponse(), err
+	}
+
+	s.maybeWaitForAccess(ctx)
+
+	traces := req.Traces()
```
josephwoodward: Minor (non-blocking): I don't see us using `traces` variable later, perhaps just call `req.Traces().SpanCount()` instead of assigning to a variable?

internal/impl/otlp/input.go
josephwoodward: Also, given the type uses pointer receivers for its functions, would it be better to follow the convention of making this function return a pointer to indicate that it has pointer semantics?

mmatczuk: The thing is this is only for embedding - the pointer receiver works on addresses embedded in another struct.
I can sure rename it to new.

internal/impl/otlp/input.go
mmatczuk: nah, it's called like that everywhere

internal/impl/otlp/input.go
mmatczuk: Moved above Close.

internal/impl/otlp/input_grpc.go
mmatczuk: Fair point, those are exported fields of not exported type but it's on purpose for config type. It can make your life easier with tests or in separate package otlp_test.

internal/impl/otlp/input_grpc.go
```diff
@@ -0,0 +1,499 @@
...
+				Default(defaultGRPCAddress),
+			service.NewObjectField(giFieldTLS,
+				tlsFields()...,
+			).Description("TLS configuration for gRPC.").
+				Advanced(),
+			service.NewStringField(giFieldAuthToken).
+				Description("Optional bearer token for authentication. When set, requests must include 'authorization: Bearer <token>' metadata.").
+				Default("").
+				Secret().
+				Advanced(),
```
rockwotj, line 129: Is this just a static token for auth? I think this needs to have the same auth stuff and JWT middleware as the gateway component

internal/impl/otlp/input_http.go
```diff
@@ -0,0 +1,560 @@
...
+				Default(defaultHTTPAddress),
+			service.NewObjectField(hiFieldTLS,
+				tlsFields()...,
+			).Description("TLS configuration for HTTP.").
+				Advanced(),
+			service.NewStringField(hiFieldAuthToken).
+				Description("Optional bearer token for authentication. When set, requests must include 'Authorization: Bearer <token>' header.").
+				Default("").
+				Secret().
+				Advanced(),
```
rockwotj, line 149: Same as the comment for gRPC

PR #3896, opened 2026-01-12
Use common-go library and support v1 licensing

Jeffail:


internal/license/service.go
```diff
@@ -78,16 +58,17 @@ func (c Config) defaultLicenseFilepath() string {
...
 	s := &Service{
 		logger:        res.Logger(),
-		loadedLicense: &atomic.Pointer[RedpandaLicense]{},
+		loadedLicense: &atomic.Pointer[license.RedpandaLicense]{},
 		conf:          conf,
 	}
 
 	license, err := s.readAndValidateLicense()
 	if err != nil {
 		res.Logger().With("error", err).Error("Failed to read Redpanda License")
```
josephwoodward, line 67: Would it be useful to log that it's falling back to the open source license?

Jeffail: From a users perspective the open source license is basically no license at all, I'm a bit worried about making things even more confusing by mentioning it

PR #3902, opened 2026-01-14
otlp: add gRPC and HTTP output components

mmatczuk:


internal/impl/otlp/output_grpc.go
```diff
@@ -0,0 +1,350 @@
...
+The entire batch is converted to a single OTLP export request and sent via gRPC.
+`).
+		Fields(
+			service.NewStringField(goFieldEndpoint).
+				Description("The gRPC endpoint of the remote OTLP collector."),
+			service.NewStringField(goFieldAuthToken).
+				Description("Optional bearer token for authentication. When set, the token is sent in the 'authorization: Bearer <token>' metadata.").
+				Default("").
+				Secret().
+				Advanced(),
```
rockwotj: I think we want all the auth options that the `http` component has here

internal/impl/otlp/output_http.go
```diff
@@ -0,0 +1,358 @@
...
+				Description("The HTTP endpoint of the remote OTLP collector (without the signal path)."),
+			service.NewStringEnumField(hoFieldContentType, "protobuf", "json").
+				Description("Content type for HTTP requests. Options: 'protobuf' or 'json'.").
+				Default(defaultContentType).
+				Advanced(),
+			service.NewStringField(hoFieldAuthToken).
+				Description("Optional bearer token for authentication. When set, the token is sent in the 'Authorization: Bearer <token>' header.").
+				Default("").
+				Secret().
+				Advanced(),
```
rockwotj: Same comment as gRPC, and I also wonder if we should let people add their own custom headers here too

internal/impl/otlp/output_http.go
```diff
@@ -0,0 +1,358 @@
...
+	resp, err := h.client.Do(req)
+	if err != nil {
+		return fmt.Errorf("send HTTP request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	// Read and discard response body to allow connection reuse
+	if _, err := io.Copy(io.Discard, resp.Body); err != nil {
+		h.log.Warnf("Failed to read HTTP response body: %v", err)
+	}
```
rockwotj: https://opentelemetry.io/docs/specs/otlp/#partial-success-1

There can be partial success, and those failures I think we want to return errors for? There is a BatchError thing that allows you fail individual messages in a batch

mmatczuk: Fair point, the problem is that it contains only the count of failed items and a message plus we'd have to remap that to Redpanda OTEL messages. I'll unify with gRPC here and guarantee at least once delivery i.e. fail.

internal/impl/otlp/output_grpc.go
mmatczuk: Looks like we can do that https://github.com/grpc/grpc-go/tree/master/examples/features/authentication#authentication

mmatczuk:
- Handle partial success and return error both in gRPC and HTTP 
- Replace auth_token with auth options present in http
- Port and modernize OAuth2 from benthos and integrate in gRPC and HTTP

internal/impl/otlp/output.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+
+		ptr := P(&results[i])
+		if err := proto.Unmarshal(msgBytes, ptr); err != nil {
+			return nil, fmt.Errorf("message %d: failed to unmarshal %s: %w", i, typeName, err)
+		}
+	}
+
+	return results, nil
+}
```
rockwotj, line 78: so I think we have to give some options here... I believe cloud is defaulting to using protobuf with schema registry serdes, so we probably want to accept something that is output from the `schema_registry_decode` processor

mmatczuk: What is the use case? It feels to me that the output is something you hook to redpanda input to export to 3rd party system like this

redpanda input -> otel output-> Open Telemetry Collector with Exporter -> Some System like DataDog [examples](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter).

rockwotj: the use case is that we are emitting data with the schema registry encoding right now for the adp components and this cannot work with those

PR #3696, opened 2025-10-15
(sqlservercdc): Add benchmarking suite

josephwoodward:
Benchmarks from my laptop:

Runtime: ~4m 30s
Dataset: 1.4kb * 21198489 rows = 24.1GB

```
INFO rolling stats: 81000 msg/sec, 108 MB/sec      @service=redpanda-connect bytes/sec=1.0850761e+08 label="" msg/sec=81000 path=root.output.processors.0
INFO rolling stats: 71000 msg/sec, 95 MB/sec       @service=redpanda-connect bytes/sec=9.5111492e+07 label="" msg/sec=71000 path=root.output.processors.0
INFO rolling stats: 89000 msg/sec, 119 MB/sec      @service=redpanda-connect bytes/sec=1.19224274e+08 label="" msg/sec=89000 path=root.output.processors.0
INFO rolling stats: 101000 msg/sec, 135 MB/sec     @service=redpanda-connect bytes/sec=1.352995e+08 label="" msg/sec=101000 path=root.output.processors.0
INFO rolling stats: 104000 msg/sec, 139 MB/sec     @service=redpanda-connect bytes/sec=1.39318186e+08 label="" msg/sec=104000 path=root.output.processors.0
INFO rolling stats: 99000 msg/sec, 133 MB/sec      @service=redpanda-connect bytes/sec=1.32620246e+08 label="" msg/sec=99000 path=root.output.processors.0
INFO rolling stats: 65000 msg/sec, 87 MB/sec       @service=redpanda-connect bytes/sec=8.7073866e+07 label="" msg/sec=65000 path=root.output.processors.0
INFO rolling stats: 78000 msg/sec, 104 MB/sec      @service=redpanda-connect bytes/sec=1.04488712e+08 label="" msg/sec=78000 path=root.output.processors.0
INFO rolling stats: 100000 msg/sec, 134 MB/sec     @service=redpanda-connect bytes/sec=1.33959894e+08 label="" msg/sec=100000 path=root.output.processors.0
INFO rolling stats: 105000 msg/sec, 141 MB/sec     @service=redpanda-connect bytes/sec=1.40657538e+08 label="" msg/sec=105000 path=root.output.processors.0
INFO rolling stats: 103000 msg/sec, 138 MB/sec     @service=redpanda-connect bytes/sec=1.37978776e+08 label="" msg/sec=103000 path=root.output.processors.0
[...]

PR #3908, opened 2026-01-15
Add dynamodb streams input

squiidz:


david-yu:
Out of curiosity is there a reason why this can't be named amazon_dynamodb_cdc instead of amazon_dyanmodb_streams?

squiidz:
> Out of curiosity is there a reason why this can't be named amazon_dynamodb_cdc instead of amazon_dyanmodb_streams?

We can rename it, but it's referring to how aws calls this feature, so It might be more consistent to call it like that. In the end I don't mind renaming it if you think it's better, just let me know.

david-yu:
My personal opinion is that we are leveraging dynamo db streams as documented but `amazon_dyanmodb_cdc` will make it clear to folks that are looking for a CDC use case this is the right connector. Docs: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.html. Since we name all of our cdc connectors with a CDC suffix my recommendation would be to make it consistent with RPCN and document in the connector that this leverages a Dynamo DB streams API.

docs/modules/components/pages/inputs/aws_dynamodb_cdc.adoc
```diff
@@ -0,0 +1,313 @@
...
+
+=== `credentials.role_external_id`
+
+An external ID to provide when assuming a role.
+
+
+*Type*: `string`
+
+
+
```
david-yu, line 389: Could we also add an example of using this connector?

josephwoodward:
> My personal opinion is that we are leveraging dynamo db streams as documented but `amazon_dyanmodb_cdc` will make it clear to folks that are looking for a CDC use case this is the right connector. Docs: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.html. Since we name all of our cdc connectors with a CDC suffix my recommendation would be to make it consistent with RPCN and document in the connector that this leverages a Dynamo DB streams API.

Yeah, I agree.

We have a convention of having `cdc` in the suffix of CDC  specific components to help discoverability. See the below screenshot from [here](https://docs.redpanda.com/redpanda-connect/components/about/?search=cdc&type=input%2Coutput%2Cprocessor%2Cscanner%2Cmetric%2Ccache%2Ctracer%2Crate_limit%2Cbuffer&support=certified%2Ccommunity&cloud=yes%2Cno&enterprise=yes%2Cno).

Image

internal/impl/aws/input_dynamodb_cdc_batcher.go
```diff
@@ -0,0 +1,130 @@
...
+		// Enforce memory bounds on checkpoint map
+		if len(b.lastCheckpoints) > b.maxTrackedShards {
+			return fmt.Errorf("checkpoint map exceeded maximum size (%d shards) - possible memory leak", b.maxTrackedShards)
+		}
+
+		// Check if we should checkpoint BEFORE decrementing
+		// pendingCount tracks how many messages have been acked since last checkpoint
+		if b.pendingCount[shardID] >= checkpointer.checkpointLimit {
+			if err := checkpointer.Set(ctx, shardID, seq); err != nil {
+				b.log.Warnf("Failed to checkpoint shard %s: %w", shardID, err)
```
josephwoodward:
```suggestion
				b.log.Warnf("Failed to checkpoint shard %s: %v", shardID, err)
```
As this is a log entry we don't need to wrap the error. I notice there are a few other locations of this.

josephwoodward: Actually, given we return an error below won't this lead to double logging? Would it be better to remove this log, return the error and log it in the caller further up the call stack?

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,500 @@
...
+		startFrom:        startFrom,
+		checkpointLimit:  checkpointLimit,
+		maxTrackedShards: maxTrackedShards,
+		shardReaders:     make(map[string]*dynamoDBShardReader),
+		log:              mgr.Logger(),
+	}, nil
+}
+
+func (d *dynamoDBCDCInput) Connect(ctx context.Context) error {
+	if d.dynamoClient != nil {
```
josephwoodward: With this check, if the dynamo client is initialised but the streamsClient initialisation, checkpointer etc failed (ie, Connect function partially fails) then won't the connector get stuck in a broken state as this will continue to exit early and never reinitialise the broken client?

squiidz: You're right ! I'll remove that check.

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,496 @@
...
+	exhausted bool
+}
+
+func newDynamoDBCDCInputFromConfig(conf *service.ParsedConfig, mgr *service.Resources) (*dynamoDBCDCInput, error) {
+	awsConf, err := GetSession(context.Background(), conf)
+	if err != nil {
+		return nil, err
+	}
+
+	table, err := conf.FieldString("table")
```
josephwoodward: Is the requirement for just a single table or multiple tables?

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,496 @@
...
+func (d *dynamoDBCDCInput) Connect(ctx context.Context) error {
+	d.dynamoClient = dynamodb.NewFromConfig(d.awsConf)
+	d.streamsClient = dynamodbstreams.NewFromConfig(d.awsConf)
+
+	// Get stream ARN
+	descTable, err := d.dynamoClient.DescribeTable(ctx, &dynamodb.DescribeTableInput{
+		TableName: &d.table,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to describe table %s: %w", d.table, err)
```
josephwoodward: Thought: this error handling could probably be clearer in instances of the resource not being found? IIRC the AWS SDK includes such type.

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,496 @@
...
+
+func (d *dynamoDBCDCInput) ReadBatch(ctx context.Context) (service.MessageBatch, service.AckFunc, error) {
+	d.mu.Lock()
+	if d.closed || d.dynamoClient == nil {
+		d.mu.Unlock()
+		return nil, nil, service.ErrNotConnected
+	}
+
+	activeReaders := make([]*dynamoDBShardReader, 0, len(d.shardReaders))
+	for _, reader := range d.shardReaders {
```
josephwoodward: This looks racey, does it need a mutex over access to the map?

internal/impl/aws/input_dynamodb_cdc.go
squiidz: Good idea, I'll add it

internal/impl/aws/input_dynamodb_cdc.go
squiidz: From what I understand, it's one table with multiple shards.

internal/impl/aws/input_dynamodb_cdc.go
josephwoodward: Looking into this and it doesn't appear to be an issue. `-race` didn't return any problems.

internal/impl/aws/input_dynamodb_cdc.go
josephwoodward: Okay, great 👍

docs/modules/components/pages/inputs/aws_dynamodb_cdc.adoc
squiidz: From what I see in the other inputs, the closest from an example we have are the common and advanced config at the top of the files.

internal/impl/aws/input_dynamodb_cdc_batcher.go
josephwoodward: Looks like this has been resolved, marking as done.

internal/impl/aws/input_dynamodb_cdc.go
josephwoodward: Looks like this has been addressed, will resolve the conversation.

josephwoodward:
As per our call it'd be good to:
- Add the table name to the message metadata and add docs highlighting what metadata exists (Operation codes etc, see  the [PostgreSQL component as an example](https://docs.redpanda.com/redpanda-connect/components/inputs/postgres_cdc/#metadata))
- Update the integration tests to use more data to exercise batching
- Follow up PR of a `/bench/` suite demonstrating the performance characteristics (similar to [sql server](https://github.com/redpanda-data/connect/pull/3696) connector and migrator) 
Otherwise LGTM, would be good to get eyes from @mmatczuk and/or @Jeffail before merging.

internal/impl/aws/input_dynamodb_cdc_test.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	for _, tt := range tests {
+		result := min(tt.a, tt.b)
+		assert.Equal(t, tt.expected, result)
+	}
+}
+
+// Test shard reader structure
+func TestShardReaderStructure(t *testing.T) {
```
josephwoodward: Thought (non-blocking): Given this test is verifying we can add data to a map in a struct then read from it, it feels redundant (and no doubt covered by the integration tests) and can be removed. WDYT?

internal/impl/aws/input_dynamodb_cdc_integration_test.go
```diff
@@ -0,0 +1,433 @@
...
+	// Wait for DynamoDB to be ready and create table with streams
+	require.NoError(t, pool.Retry(func() error {
+		var err error
+		client, err = createTableWithStreams(context.Background(), t, resource.GetPort("8000/tcp"), tableName)
+		return err
+	}))
+
+	port := resource.GetPort("8000/tcp")
+
+	t.Run("ReadInsertEvents", func(t *testing.T) {
```
josephwoodward, line 177: It'd be good to verify the number of records we receive as part of the test (ie, if there's 100 records in the table then we see 100 objects in the output.

squiidz: done

internal/plugins/info.csv
```diff
@@ -13,6 +13,7 @@ aws_bedrock_embeddings    ,processor ,aws_bedrock_embeddings    ,4.37.0  ,certif
 aws_cloudwatch            ,metric    ,aws_cloudwatch            ,3.36.0  ,community  ,n          ,n     ,n
 aws_dynamodb              ,cache     ,AWS DynamoDB              ,3.36.0  ,community  ,n          ,y     ,y
 aws_dynamodb              ,output    ,AWS DynamoDB              ,3.36.0  ,community  ,n          ,y     ,y
+aws_dynamodb_cdc          ,input     ,aws_dynamodb_cdc          ,0.0.0   ,enterprise ,n          ,y     ,n
```
mmatczuk: Please update the version

docs/modules/components/pages/inputs/aws_dynamodb_cdc.adoc
```diff
@@ -0,0 +1,313 @@
...
+// © 2024 Redpanda Data Inc.
+
+
+component_type_dropdown::[]
+
+
+Reads change data capture (CDC) events from DynamoDB Streams
+
+Introduced in version 1.0.0.
```
mmatczuk: We need to set version

go.mod
```diff
@@ -158,7 +158,7 @@ require (
 	github.com/tetratelabs/wazero v1.9.0
 	github.com/tigerbeetle/tigerbeetle-go v0.16.61
 	github.com/timeplus-io/proton-go-driver/v2 v2.1.2
-	github.com/tmc/langchaingo v0.1.13
+	github.com/tmc/langchaingo v0.1.14
```
mmatczuk: Why we update langchain?

squiidz: I had issues with 0.1.13 missing a dependency.

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,501 @@
...
+}
+
+type dynamoDBCDCInput struct {
+	table            string
+	checkpointTable  string
+	batchSize        int32
+	pollInterval     time.Duration
+	startFrom        string
+	checkpointLimit  int
+	maxTrackedShards int
```
mmatczuk, line 164: nit: I like to embed config objects for clear separation

mmatczuk:
### Code review

Found 5 issues:

1. **pendingCount never decremented causing incorrect checkpointing** (bug: pendingCount incremented but never decremented, only reset)

The `pendingCount` map in the batcher is incremented when messages are added but never decremented when messages are acknowledged. It's only reset to 0 after checkpointing occurs. This causes the checkpoint logic to malfunction - once the count reaches the checkpoint limit, every subsequent ack will trigger a checkpoint even if only 1 message was processed.

https://github.com/redpanda-data/connect/blob/94512e4591229c8218d6f44f75d8113cd7091b48/internal/impl/aws/input_dynamodb_cdc_batcher.go#L105-L118

2. **Nil pointer dereference in ack function after Close()** (bug: race condition between Close() and pending acks)

The ack closure captures `d.recordBatcher` and `d.checkpointer` by reference through the receiver, but these are set to nil in `Close()`. If `Close()` is called while messages are pending acknowledgment, the ack function will panic with a nil pointer dereference when trying to call methods on these nil pointers.

https://github.com/redpanda-data/connect/blob/94512e4591229c8218d6f44f75d8113cd7091b48/internal/impl/aws/input_dynamodb_cdc.go#L409-L418

3. **Missing metrics/observability** (AGENTS.md says "Observability: Metrics, logs (warnings/errors only during issues), tracing hooks" and CONTRIBUTING.md says "Exposes useful metrics for debugging that avoid excessive cardinality")

The implementation lacks metrics for monitoring component health and performance. The comparable Kinesis input has metrics like `clientShardsMetric` and `shardsStolenMetric`. This component should have similar metrics for shard count, checkpoint operations, and other key performance indicators.

https://github.com/redpanda-data/connect/blob/94512e4591229c8218d6f44f75d8113cd7091b48/internal/impl/aws/input_dynamodb_cdc.go#L103-L125

4. **Iterator not preserved on GetRecords error, violating Kinesis pattern** (input_kinesis.go says "IMPORTANT TO NOTE: The returned shard iterator will always be the input iterator when the error parameter is nil" and "Do NOT modify this method without preserving this behaviour")

When `GetRecords` fails, the code marks the shard as exhausted without preserving the iterator. The Kinesis implementation explicitly documents that the iterator should be returned unchanged on error to allow retries from the same position. This can cause data loss on transient errors.

https://github.com/redpanda-data/connect/blob/94512e4591229c8218d6f44f75d8113cd7091b48/internal/impl/aws/input_dynamodb_cdc.go#L323-L332

Reference pattern: https://github.com/redpanda-data/connect/blob/94512e4591229c8218d6f44f75d8113cd7091b48/internal/impl/aws/input_kinesis.go#L433-L453

5. **Missing example configuration with metadata documentation** (PR reviewer requested: "Could we also add an example of using this connector?" and "Add the table name to the message metadata and add docs highlighting what metadata exists")

The documentation lacks a practical usage example and metadata documentation. CONTRIBUTING.md requires "examples, troubleshooting guidance, and known pitfalls" for certified connectors.

https://github.com/redpanda-data/connect/blob/94512e4591229c8218d6f44f75d8113cd7091b48/docs/modules/components/pages/inputs/aws_dynamodb_cdc.adoc#L1-L10

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,501 @@
...
+	awsConf       aws.Config
+	dynamoClient  *dynamodb.Client
+	streamsClient *dynamodbstreams.Client
+	streamArn     *string
+
+	checkpointer  *dynamoDBCDCCheckpointer
+	recordBatcher *dynamoDBCDCRecordBatcher
+
+	shardReaders map[string]*dynamoDBShardReader
+	mu           sync.Mutex
```
mmatczuk: Worth mentioning what it protects, the convention is to put mu on top of var group or use XMu for protection of X

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,501 @@
...
+
+	table, err := conf.FieldString("table")
+	if err != nil {
+		return nil, err
+	}
+
+	checkpointTable, err := conf.FieldString("checkpoint_table")
+	if err != nil {
+		return nil, err
+	}
```
mmatczuk: nit: we try to put that into if like if X.checkpointTable, err = conf.FieldString("checkpoint_table"); err != nil

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,501 @@
...
+	shardID   string
+	iterator  *string
+	exhausted bool
+}
+
+func newDynamoDBCDCInputFromConfig(conf *service.ParsedConfig, mgr *service.Resources) (*dynamoDBCDCInput, error) {
+	awsConf, err := GetSession(context.Background(), conf)
+	if err != nil {
+		return nil, err
+	}
```
mmatczuk: Maybe move that to Connect and remove the awsConf field

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,501 @@
...
+		}
+
+		// Check checkpoint
+		checkpoint, err := d.checkpointer.Get(ctx, shardID)
+		if err != nil {
+			d.log.Warnf("Failed to get checkpoint for shard %s: %w", shardID, err)
+		}
+
+		var iteratorType types.ShardIteratorType
+		var sequenceNumber *string
```
mmatczuk: Someone should tell Claude to group vars

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,501 @@
...
+func (d *dynamoDBCDCInput) refreshShards(ctx context.Context) error {
+	streamDesc, err := d.streamsClient.DescribeStream(ctx, &dynamodbstreams.DescribeStreamInput{
+		StreamArn: d.streamArn,
+	})
+	if err != nil {
+		return err
+	}
+
+	d.mu.Lock()
+	defer d.mu.Unlock()
```
mmatczuk: Could we just lock d.shardReaders access not i/o with Amazon

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,501 @@
...
+
+	// Read from all shards and aggregate results for better throughput
+	allMessages := make(service.MessageBatch, 0)
+
+	for _, reader := range activeReaders {
+		// Skip if iterator is nil (shouldn't happen but safety check)
+		if reader.iterator == nil {
+			d.mu.Lock()
+			reader.exhausted = true
+			d.mu.Unlock()
```
mmatczuk: We really need better / more targeted sync primitives unless we very explicitly explain that mu protects map and its contents. Then why iterator is not protected?

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,501 @@
...
+		return fmt.Errorf("failed to describe table %s: %w", d.table, err)
+	}
+
+	d.streamArn = descTable.Table.LatestStreamArn
+	if d.streamArn == nil {
+		return fmt.Errorf("no stream enabled on table %s", d.table)
+	}
+
+	// Initialize checkpointer
+	d.checkpointer, err = newDynamoDBCDCCheckpointer(ctx, d.dynamoClient, d.checkpointTable, *d.streamArn, d.checkpointLimit, d.log)
```
mmatczuk: Do not use ctx passed to Connect, search for shutdown.Signaller

internal/impl/aws/input_dynamodb_cdc_batcher.go
```diff
@@ -0,0 +1,129 @@
...
+import (
+	"context"
+	"fmt"
+	"maps"
+	"sync"
+
+	"github.com/redpanda-data/benthos/v4/public/service"
+)
+
+// dynamoDBCDCRecordBatcher tracks messages and their checkpoints for DynamoDB CDC.
```
mmatczuk: I'd appreciate if we could add more docs here about what it does and why.

internal/impl/aws/input_dynamodb_cdc_batcher.go
```diff
@@ -0,0 +1,129 @@
...
+type dynamoDBCDCRecordBatcher struct {
+	mu               sync.Mutex
+	messageTracker   map[*service.Message]*messageCheckpoint
+	log              *service.Logger
```
mmatczuk: Log in the middle, use some grouping or move it down.

internal/impl/aws/input_dynamodb_cdc_test.go
```diff
@@ -0,0 +1,220 @@
+// Copyright 2026 Redpanda Data, Inc.
+//
+// Licensed as a Redpanda Enterprise file under the Redpanda Community
+// License (the "License"); you may not use this file except in compliance with
+// the License. You may obtain a copy of the License at
+//
+// https://github.com/redpanda-data/connect/blob/main/licenses/rcl.md
```
mmatczuk, line 7: This may be treated as pkg doc - no separation

internal/impl/aws/input_dynamodb_cdc_test.go
```diff
@@ -0,0 +1,220 @@
...
+			"shard-001": {shardID: "shard-001", iterator: aws.String("iter-001"), exhausted: false},
+			"shard-002": {shardID: "shard-002", iterator: aws.String("iter-002"), exhausted: false},
+		},
+		log: logger,
+	}
+
+	// Multiple goroutines should be able to read concurrently
+	done := make(chan bool, 3)
+
+	for i := 0; i < 3; i++ {
```
mmatczuk: nit: it's generally better to use range

mmatczuk:
@squiidz I think we should work on locking granularity, usually we use a channel internally for sending batches, it also gives you data preloading.

go.mod
mmatczuk: If the problem is real please send a separate PR

squiidz: My bad it was on my side, I reverted it to 0.1.13

docs/modules/components/pages/inputs/aws_dynamodb_cdc.adoc
david-yu: Is it possible to include an example with working IAM auth? Postgres has an example pipeline like follows and I think any examples are always helpful: https://docs.redpanda.com/redpanda-connect/components/inputs/postgres_cdc/#example-pipeline

docs/modules/components/pages/inputs/aws_dynamodb_cdc.adoc
mmatczuk: Ping.

mmatczuk:
I'd consider moving all the helper types to a dedicated sub package ex. instead if `dynamoDBCDCRecordBatcher` use `dynamocdc.RecordBatcher`. In my view it would limit the use of shared aws namespace.

internal/impl/aws/input_dynamodb_cdc_batcher.go
```diff
@@ -0,0 +1,175 @@
...
+	mu             sync.Mutex
+	messageTracker map[*service.Message]*messageCheckpoint
+
+	// Checkpoint state per shard
+	pendingCount    map[string]int    // Count of acked but not-yet-checkpointed messages
+	lastCheckpoints map[string]string // Most recent sequence number per shard
+
+	// Configuration
+	maxTrackedShards   int // Memory safety limit for number of unique shards
+	maxTrackedMessages int // Memory safety limit for in-flight messages
```
mmatczuk: nit: the general practice is to put the configuration first then it do not need any introduction

internal/impl/aws/input_dynamodb_cdc_batcher.go
```diff
@@ -0,0 +1,175 @@
...
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	for _, msg := range batch {
+		delete(b.messageTracker, msg)
+	}
+}
+
+// checkpointerInterface defines the interface for checkpointing operations.
+type checkpointerInterface interface {
```
mmatczuk: It should be just checkpointer. Since it's only used once please move it directly above AckMessages or you can inline the type it in the function args.

internal/impl/aws/input_dynamodb_cdc_batcher.go
```diff
@@ -0,0 +1,175 @@
...
+	GetCheckpointLimit() int
+}
+
+// GetCheckpointLimit returns the checkpoint limit for the checkpointer.
+func (c *dynamoDBCDCCheckpointer) GetCheckpointLimit() int {
+	return c.checkpointLimit
+}
+
+// AckMessages marks messages as acknowledged and checkpoints if threshold is reached.
+func (b *dynamoDBCDCRecordBatcher) AckMessages(ctx context.Context, checkpointer checkpointerInterface, batch service.MessageBatch) error {
```
mmatczuk: Consider one argument per line.

internal/impl/aws/dynamocdc/checkpoint.go
```diff
@@ -0,0 +1,150 @@
...
+		return err
+	}
+
+	// Table doesn't exist, create it
+	input := &dynamodb.CreateTableInput{
+		AttributeDefinitions: []types.AttributeDefinition{
+			{AttributeName: aws.String("StreamArn"), AttributeType: types.ScalarAttributeTypeS},
+			{AttributeName: aws.String("ShardID"), AttributeType: types.ScalarAttributeTypeS},
+		},
+		BillingMode: types.BillingModePayPerRequest,
```
mmatczuk, line 74: I wonder do we want to allow other billing types i.e. provisioned

.gitignore
```diff
@@ -10,4 +10,5 @@ release_notes.md
 .task
 .vscode
 .op
+.gomodcache
```
mmatczuk: Please send is separate commit if needed.

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,787 @@
...
+
+	// Pending acknowledgments tracking
+	pendingAcks sync.WaitGroup
+	closed      atomic.Bool
+
+	log *service.Logger
+
+	// Metrics
+	shardsTrackedMetric *service.MetricGauge
+	shardsActiveMetric  *service.MetricGauge
```
mmatczuk: Consider extracting metrics to separate type

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,787 @@
...
+	d.mu.Lock()
+	activeCount := len(d.shardReaders)
+	d.mu.Unlock()
+
+	if activeCount == 0 {
+		return errors.New("no active shard readers available - stream may have no shards or all failed to initialize")
+	}
+
+	// Start background goroutine to coordinate shard readers
+	go d.runShardCoordinator()
```
mmatczuk: How about startShardCoordinator

internal/impl/aws/dynamocdc/checkpoint.go
squiidz: I'm not sure about this, maybe we need to ask @david-yu about it.

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,787 @@
...
+}
+
+// runShardCoordinator spawns goroutines for each shard and manages shard refresh
+func (d *dynamoDBCDCInput) runShardCoordinator() {
+	defer func() {
+		close(d.msgChan)
+		d.shutSig.TriggerHasStopped()
+	}()
+
+	ctx, cancel := d.shutSig.SoftStopCtx(context.Background())
```
mmatczuk: I think it would be better to accept context as argument

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,787 @@
...
+			if err := d.refreshShards(refreshCtx); err != nil && !errors.Is(err, context.Canceled) {
+				d.log.Warnf("Failed to refresh shards: %v", err)
+			}
+			refreshCancel()
+		}
+	}
+}
+
+// runShardReader continuously reads from a single shard and sends batches to the channel
+func (d *dynamoDBCDCInput) runShardReader(ctx context.Context, shardID string) {
```
mmatczuk: Same here the usual naming convention is that run waits for result and returns result and start is asynchronous.

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,787 @@
...
+				return
+			default:
+			}
+
+			// Apply backpressure if too many messages are in flight
+			if d.recordBatcher != nil && d.recordBatcher.ShouldThrottle() {
+				d.log.Debugf("Throttling shard %s due to too many in-flight messages", shardID)
+				time.Sleep(100 * time.Millisecond)
+				continue
+			}
```
mmatczuk: Should that maybe wait in a loop? Also the timeout should be configurable

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,787 @@
...
+
+			// Get current reader state
+			d.mu.RLock()
+			reader, exists := d.shardReaders[shardID]
+			if !exists || reader.exhausted || reader.iterator == nil {
```
mmatczuk: I guess if it does not exists it needs some error logging or even panic as it shall never happen.

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,787 @@
...
+				d.mu.RUnlock()
+				return
+			}
+			iterator := reader.iterator
+			d.mu.RUnlock()
+
+			// Read records from the shard (I/O operation - no lock held)
```
mmatczuk, line 514: Claudy comment, I'd suggest to change `// Get current reader state` to method and avoid ad-hoc locking in functions in general.

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,787 @@
...
+
+func (d *dynamoDBCDCInput) Close(ctx context.Context) error {
+	// Mark as closed to reject new acks
+	d.closed.Store(true)
+
+	d.mu.RLock()
+	shutSig := d.shutSig
+	checkpointer := d.checkpointer
+	batcher := d.recordBatcher
+	d.mu.RUnlock()
```
mmatczuk, line 692: Put all protected vars in one group in dynamoDBCDCInput struct

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -0,0 +1,787 @@
...
+
+	// Trigger graceful shutdown
+	if shutSig != nil {
```
mmatczuk: init that in constructor and do not sync or check for nilness

internal/impl/aws/dynamocdc/batcher.go
```diff
@@ -0,0 +1,226 @@
...
+	maxTrackedShards   int
+	maxTrackedMessages int
+	log                *service.Logger
+
+	mu             sync.Mutex
+	messageTracker map[*service.Message]*messageCheckpoint
+
+	// Checkpoint state per shard
+	pendingCount    map[string]int    // Count of acked but not-yet-checkpointed messages
+	lastCheckpoints map[string]string // Most recent sequence number per shard
```
mmatczuk: Feels like those shall be grouped

internal/impl/aws/dynamocdc/batcher.go
```diff
@@ -0,0 +1,226 @@
...
+		maxTrackedMessages = 1000 // Minimum reasonable size
+	}
+
+	return &RecordBatcher{
+		messageTracker:     make(map[*service.Message]*messageCheckpoint),
+		log:                log,
+		pendingCount:       make(map[string]int),
+		lastCheckpoints:    make(map[string]string),
+		maxTrackedShards:   maxTrackedShards,
+		maxTrackedMessages: maxTrackedMessages,
```
mmatczuk: Could we preserve the field order.

PR #3916, opened 2026-01-16
Fix bundle updates

Jeffail:


resources/scripts/update_bundles.sh
```diff
@@ -6,6 +6,6 @@
 # community, etc) and upgrades all dependencies (go get -u).
 
 for dir in $(ls ./public/bundle); do
-    ( cd "./public/bundle/$dir" && go get -u . && go mod tidy )
+    ( cd "./public/bundle/$dir" && go get github.com/redpanda-data/connect/v4@latest && go mod tidy )
```
josephwoodward, line 9: Ah nice, was just looking into this and noticed we update everything instead of just the connect version 👍

squiidz (approved):

PR #3918, opened 2026-01-16
otlp: add Schema Registry support

mmatczuk:
Just inputs for now.
Needs support in outputs and testing.

### DIFF
--- a/internal/impl/otlp/input.go
+++ b/internal/impl/otlp/input.go
@@ -16,11 +16,16 @@ package otlp
 
 import (
 	"context"
+	"encoding/binary"
+	"slices"
 	"time"
 
 	"github.com/Jeffail/shutdown"
+	"github.com/twmb/franz-go/pkg/sr"
+	"google.golang.org/protobuf/proto"
 
 	"github.com/redpanda-data/benthos/v4/public/service"
+	"github.com/redpanda-data/connect/v4/internal/schemaregistry"
 )
 
 type asyncMessage struct {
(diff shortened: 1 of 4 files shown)


internal/impl/otlp/proto.go
```diff
@@ -0,0 +1,24 @@
...
+var commonProtoSchema string
+
+//go:embed proto/trace.proto
+var spanProtoSchema string
+
+//go:embed proto/log.proto
+var logRecordProtoSchema string
+
+//go:embed proto/metric.proto
+var metricProtoSchema string
```
rockwotj, line 24: our redpanda otel exporter package that RPCN already uses does this. Maybe we should just make those public

internal/impl/otlp/schema_registry.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+		ss, err := client.CreateSchema(ctx, subject, schema)
+		if err != nil {
+			return nil, fmt.Errorf("failed to register schema for subject %s: %w", subject, err)
+		}
+		schemaIDs[SignalTypeLog] = ss.ID
+	}
+
+	{
+		subject := "redpanda.otel.v1.Metric"
```
rockwotj, line 57: Do these match redpanda-otel-exporter package? Just so we reduce the number of subjects if possible

internal/impl/otlp/input.go
```diff
@@ -125,3 +140,50 @@ func (o *otlpInput) ReadBatch(ctx context.Context) (service.MessageBatch, servic
...
+	return svcMsg, nil
+}
+
+func insertSchemaRegistryHeader(schemaID int, payload []byte) []byte {
+	result := slices.Grow(payload, 5)[:len(payload)+5]
+	copy(result[5:], payload)
+	result[0] = 0x00 // Confluent magic byte
+	binary.BigEndian.PutUint32(result[1:5], uint32(schemaID))
+	return result
+}
```
rockwotj, line 189: franzgo's sr package as a help for this, you're missing the additional byte for proto offsets

internal/impl/otlp/schema_registry.go
mmatczuk: Exporter uses `cfg.schemaSubject = cfg.topic + "-value"` as the subject name, looks like it's by design subjects heavy.

internal/impl/otlp/proto.go
mmatczuk: Exported schemas https://github.com/redpanda-data/common-go/pull/122, once merged I'll switch to using that.

internal/impl/otlp/proto/gen.sh
```diff
@@ -0,0 +1,18 @@
...
+# License (the "License"); you may not use this file except in compliance with
+# the License. You may obtain a copy of the License at
+#
+# https://github.com/redpanda-data/connect/blob/main/licenses/rcl.md
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+TAG="redpanda-otel-exporter/$(go list -m -f '{{.Version}}' github.com/redpanda-data/common-go/redpanda-otel-exporter)"
```
birdayz, line 14: can't we push in common-go to BSR and here import the go package from buf.build ?
this script looks like a hack for a well solved problem

mmatczuk: It's easier to export it, note https://github.com/redpanda-data/common-go/pull/122

mmatczuk:
Work moved to #3924

PR #3922, opened 2026-01-19
Fix TRACE, ALL, OFF and NONE log levels

### LINKED ISSUE #3872: logger: unexpected error log when using TRACE level
when I am using the logger with `TRACE` mode to inspect what I am doing based on this [documentation](https://docs.redpanda.com/redpanda-connect/components/logger/about/#level) 

that give the following options

```console
Options: OFF , FATAL , ERROR , WARN , INFO , DEBUG , TRACE , ALL , NONE
```

if I set TRACE (instead DEBUG) I can see the following error log

```
time="2025-12-23T08:51:16Z" level=error msg="Log level 'TRACE' not recognized, using to default level INFO" @service=benthos
```

despite this message, the `TRACE` level works fine

I find the root cause: [this line](https://github.com/redpanda-data/connect/blob/main/internal/cli/enterprise.go#L124)

BTW it is possible add support to `TRACE` levels using a [dedicated code wrapper](https://pkg.go.dev/log/slog#example-HandlerOptions-CustomLevels) but it means changing all places where we create a new slog ( inside repo `github.com/redpanda-data/benthos/v4` too )

mihaitodor:
Sending trace logs on the topic logger probably doesn't make much sense, so we fall back to DEBUG for it if TRACE is requested.

Fixes #3872.

PS: I noticed that `benthos run --log.level=trace` doesn't error out. Guess the topic logger is only configured via the `logger` config section, which might be a bit confusing...

### DIFF
--- a/internal/cli/enterprise.go
+++ b/internal/cli/enterprise.go
@@ -127,7 +127,7 @@ func InitEnterpriseCLI(binaryName, version, dateBuilt string, schema *service.Co
 				// Logging disabled
 			default:
 				logsLevel = levelPtr(slog.LevelInfo)
-				fbLogger.Errorf("Log level '%s' not recognized, using to default level %s", logsLevelStr, logsLevel)
+				fbLogger.Errorf("Log level '%s' not recognized, using the default level %s", logsLevelStr, logsLevel)
 			}
 
 			rpMgr.SetTopicLoggerLevel(logsLevel)


internal/cli/enterprise.go
```diff
@@ -110,21 +110,27 @@ func InitEnterpriseCLI(binaryName, version, dateBuilt string, schema *service.Co
...
-				logsLevel = slog.LevelError
+				l := slog.LevelWarn
+				logsLevel = &l
+			case "error", "fatal":
+				l := slog.LevelError
+				logsLevel = &l
+			case "off", "none":
+				// Logging disabled
 			default:
 				fbLogger.Errorf("Log level '%s' not recognized, using to default level %s", logsLevelStr, logsLevel)
```
josephwoodward: Thought(non-blocker): Is it worth changing this error whilst we're at it?  I presume it's meant to read "using the default level".

josephwoodward:
Thanks @mihaitodor!

internal/cli/enterprise.go
mihaitodor: Good call-out, thank you! I looked through the code again and I think it's better now.

I also realised that `topicLogger.Enabled()` returned `true` when `level == nil`, but I don't think that ever happened in the original code because it was passed by value. I could be wrong, but I think this should return `false` when users set the log level to `OFF` / `NONE`, so I updated it. WDYT?

CHANGELOG.md
```diff
@@ -234,7 +240,7 @@ All notable changes to this project will be documented in this file.
 - (Benthos) New Bloblang method `infer_schema`. (@Jeffail)
 - Custom s3 endpoints support in `snowflake_streaming` output. (@josephwoodward)
 - Experimental field `timely_nacks_maximum_wait` added to all kafka protocol inputs. (@Jeffail)
-- Added `subject_compatibility_level` to the `schema_registry` output. (@mmatczuk) 
+- Added `subject_compatibility_level` to the `schema_registry` output. (@mmatczuk)
```
mmatczuk: @mihaitodor can you remove this change.

internal/cli/enterprise.go
```diff
@@ -110,21 +110,27 @@ func InitEnterpriseCLI(binaryName, version, dateBuilt string, schema *service.Co
...
-			case "error":
-				logsLevel = slog.LevelError
+				logsLevel = levelPtr(slog.LevelWarn)
+			case "error", "fatal":
+				logsLevel = levelPtr(slog.LevelError)
+			case "off", "none":
+				// Logging disabled
 			default:
+				logsLevel = levelPtr(slog.LevelInfo)
 				fbLogger.Errorf("Log level '%s' not recognized, using to default level %s", logsLevelStr, logsLevel)
```
josephwoodward:
```suggestion
				fbLogger.Errorf("Log level '%s' not recognized, using the default level %s", logsLevelStr, logsLevel)
```

josephwoodward: > I also realised that topicLogger.Enabled() returned true when level == nil, but I don't think that ever happened in the original code because it was passed by value. I could be wrong, but I think this should return false when users set the log level to OFF / NONE, so I updated it. WDYT?

Sounds reasonable, so the behaviour remains the same?

mihaitodor: It should, yeah

CHANGELOG.md
mihaitodor: Sorted, thanks!

josephwoodward:
Thanks for this @mihaitodor!

PR #3923, opened 2026-01-20
Change metadata processing for single-byte values of kafka headers

volkerhess2:
The current processing of single-byte values leads to different types and values of metadata. For example integers =< 9 are forwarded with their ascii value and of type number, values above 9 are forwarded as strings. A byte encoded 2 for example is forwarded as 50 because 50 is the ascii code of 2, while a 12 is forwarded as string of "12" like one would expect. We had to check this in the piplines and cast it to the correct value:

```
    mapping: |
      meta ce_e2eekeyversion = if metadata("ce_e2eekeyversion").type()  == "number" {
        "%c".format(metadata("ce_e2eekeyversion"))
      } else {
        metadata("ce_e2eekeyversion")
      }
```

This fix changes the kafka input to propagate every value as string.


mmatczuk:
Thanks @volkerhess2!

PR #3926, opened 2026-01-20
feat(aws_s3): add local_file_path configuration for uploading files

### LINKED ISSUE #3925: feature: support streaming local files to AWS S3
Support for streaming local files to AWS S3, to efficiently upload large files without memory overflow (i.e., send as a stream instead of loading all file bytes into memory). 

This improvement is important to enable handling of large files (such as logs or binaries) that would otherwise exhaust available memory during upload.

juniorrhis1:
closes #3925

This pull request adds support for uploading files to AWS S3 from a specified local file path, in addition to the existing functionality of uploading message content. The changes include updates to both the documentation and the implementation to introduce a new configuration field, `local_file_path`, which allows users to specify a local file to upload per message.

**Key changes:**

**Feature: Support for uploading from a local file path**
- Added a new configuration field `local_file_path` to the S3 output component, allowing users to specify the path of a local file to upload instead of message content. This field supports interpolation functions for dynamic file path resolution. [[1]](diffhunk://#diff-9671479c9893dc0203ff74ee7c6cc32b8e5d94cc4eac1d7df6fea07f90632836R45) [[2]](diffhunk://#diff-9671479c9893dc0203ff74ee7c6cc32b8e5d94cc4eac1d7df6fea07f90632836R73) [[3]](diffhunk://#diff-9671479c9893dc0203ff74ee7c6cc32b8e5d94cc4eac1d7df6fea07f90632836R256-R259) [[4]](diffhunk://#diff-5d653b95f5eac8fdf4ee79ff79194410ba74e7f6cb8b92bcd8dc9480cdb18ceeR44) [[5]](diffhunk://#diff-5d653b95f5eac8fdf4ee79ff79194410ba74e7f6cb8b92bcd8dc9480cdb18ceeR69) [[6]](diffhunk://#diff-5d653b95f5eac8fdf4ee79ff79194410ba74e7f6cb8b92bcd8dc9480cdb18ceeR209-R224)
- Updated the S3 writer logic to check if `local_file_path` is set for each message; if so, it reads the specified file and uploads its contents, otherwise it uploads the message content as before. [[1]](diffhunk://#diff-9671479c9893dc0203ff74ee7c6cc32b8e5d94cc4eac1d7df6fea07f90632836L473-R493) [[2]](diffhunk://#diff-9671479c9893dc0203ff74ee7c6cc32b8e5d94cc4eac1d7df6fea07f90632836R542-R563)
- Improved documentation in `aws_s3.adoc` to describe the new `local_file_path` field, its type, default value, and usage examples.

**Implementation details**
- Refactored the upload logic by introducing a helper method `getUploadBody` to encapsulate the logic for choosing between local file content and message content.
- Added necessary imports for file handling in Go (`os`, `io`).


mihaitodor:
Hi @juniorrhis1 👋 Could you please let us know why you'd prefer this approach over something like the config below?

```yaml
input:
  file:
    paths:
      - test.txt
    scanner:
      to_the_end: {}

output:
  aws_s3:
    ...
```

juniorrhis1:
Hi @mihaitodor, to avoid load the entire file in memory.

For small files or json files with one document per line (using lines scanner) this approach is prefered, but is not possible send large files without consumes a lot of memory.

Jeffail:
Hey @juniorrhis1, this is cool but I'm a bit worried about the security implications. In some environments this might be considered a vulnerability and would block deployment. I think a good solution might be to add this as a separate an alternative output, maybe called `aws_s3_from_local`. I'm not good at picking names so feel free to choose something else. This way the behavioral differences can be clearly documented and users are able to custom build without this plugin enabled which eliminates the security issues.

PR #3938, opened 2026-01-29
claude: go-expert updates

mmatczuk:


.claude/agents/go-expert.md
```diff
@@ -196,6 +196,24 @@ var i = 42
 var coords = Point{X: 0, Y: 0}
 ```
 
+**Use var groups for multiple declarations**:
```
josephwoodward: Would it be worth clarifying even when they're empty or default declarations? ie

```go
var (
    name string
    email string
    address string
)
```

josephwoodward (approved):

.claude/agents/go-expert.md
mmatczuk: Updated.

PR #3943, opened 2026-01-30
Bump benthos and add dry-run subcommand

Jeffail:


internal/cli/dry_run.go
```diff
@@ -0,0 +1,286 @@
+// Copyright 2024 Redpanda Data, Inc.
```
josephwoodward:
```suggestion
// Copyright 2026 Redpanda Data, Inc.
```

internal/cli/dry_run.go
```diff
@@ -0,0 +1,286 @@
...
+			}
+
+			for _, target := range targets {
+				if isDirectory(target) {
+					if err := r.dryRunDirectory(c, target); err != nil {
+						return err
+					}
+				} else {
+					if err := r.dryRunFile(c, target); err != nil {
+						return err
```
josephwoodward, line 92: question (non-blocking): does this error return sufficient context to help the user identify why it failed or should there be more context added?

Jeffail: It'll be fairly high level sweeping errors like "Failed to read file foo.yaml", or "Failed to parse YAML file ..."

PR #3949, opened 2026-02-01
feat(migrator): streaming batch approach for large schema migrations

vuldin:
Trying to replicate a large number of schemas with Migrator fails/hangs since it tries to initially get all schemas. This PR allows streaming schemas in batches, keeping order (and schema IDs) in the target cluster the same as the source cluster. Memory utilization does not grow as schemas are produced. This addresses memory, visibility, and ordering issues when migrating large schema counts (40K+).

- Add `batch_size` config to control subjects fetched per batch (default: 100)
- Add `workers` config for parallel sync within batches (default: 10)
- Stream schemas: fetch batch → sync batch → report progress → repeat
- Sample first batch to detect schema references
- Only sort subjects by schema ID when references are detected
- Force sequential processing when references exist to preserve ordering
- Real-time progress logging with synced/skipped counts

Ordering guarantees:
- If references detected: subjects sorted by schema ID, sequential processing
- If no references: parallel processing safe with fixed IDs (translate_ids=false)

Memory optimization:
- Only hold one batch in memory at a time
- Avoid 2x API calls in common case (no references)
- Tracking maps grow with synced schemas (unavoidable for deduplication)

### DIFF
--- a/internal/impl/redpanda/migrator/migrator_schema_registry.go
+++ b/internal/impl/redpanda/migrator/migrator_schema_registry.go
@@ -81,6 +81,12 @@ const (
 	srFieldTranslateIDs   = "translate_ids"
 	srFieldNormalize      = "normalize"
 	srFieldStrict         = "strict"
+	srFieldWorkers        = "workers"
+	srFieldBatchSize      = "batch_size"
+
+	defaultWorkers      = 10
+	defaultBatchSize    = 100
+	progressLogInterval = 1000
 )
 
 func schemaRegistryField(extraFields ...*service.ConfigField) *service.ConfigField {
@@ -152,6 +158,12 @@ func schemaRegistryMigratorFields() []*service.ConfigField {
 				"Note: messages with 0-byte prefixes (e.g., protobuf) cannot be distinguished from schema registry headers and may fail when strict is enabled.").
 			Default(false).
 			LintRule(`root = if this && !this.schema_registry.translate_ids { "strict is only relevant when translate_ids is true" }`),
+		service.NewIntField(srFieldWorkers).
+			Description("Number of parallel workers for schema sync operations. Higher values improve throughput for large schema counts.").
+			Default(defaultWorkers),
+		service.NewIntField(srFieldBatchSize).
+			Description("Number of subjects to fetch and sync per batch. Schemas are streamed in batches rather than fetched all at once, reducing memory usage and providing real-time progress for large migrations.").
+			Default(defaultBatchSize),
 	}
 }
 
@@ -238,6 +250,10 @@ type SchemaRegistryMigratorConfig struct {
 	// Serverless narrows the set of schema configuration keys to those
 	// supported by serverless clusters.
 	Serverless bool
+	// Workers is the number of parallel workers for schema sync operations.
+	Workers int
+	// BatchSize is the number of subjects to fetch and sync per batch.
+	BatchSize int
 }
 
 // initFromParsed initializes the schema registry migrator with configuration from parsed config.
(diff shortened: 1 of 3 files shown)


internal/impl/redpanda/migrator/migrator_schema_registry.go
```diff
@@ -469,46 +538,306 @@ func (m *schemaRegistryMigrator) Sync(ctx context.Context) error {
...
 		}
 
 		if err := m.syncSubjectCompatibility(ctx, s.Subject); err != nil {
-			return fmt.Errorf("sync subject compatibility %s: %w", s.Subject, err)
+			return synced, fmt.Errorf("sync subject compatibility %s: %w", s.Subject, err)
 		}
 
-		m.mu.Lock()
+		srcInfo := schemaInfoFromSubjectSchema(s)
 		m.knownSubjects[srcInfo] = struct{}{}
```
josephwoodward: Does this access to `knownSubjects` now require a lock, it looks like it's accessed concurrently in `m.sr.DestinationSchemaID(...)`?

internal/impl/redpanda/migrator/migrator_schema_registry.go
```diff
@@ -469,46 +538,306 @@ func (m *schemaRegistryMigrator) Sync(ctx context.Context) error {
...
+	}
+
+	work := make(chan sr.SubjectSchema, len(toSync))
+	results := make(chan syncResult, len(toSync))
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	var wg sync.WaitGroup
+	for i := 0; i < m.conf.Workers; i++ {
```
josephwoodward, line 785: Could this be simplified with a `sync.WaitGroup`?

ie:
```go
	for i := 0; i < m.conf.Workers; i++ {
		wg.Go(func() {
			for s := range work {
				select {
				case <-ctx.Done():
					return
				default:
				}
				info, err := m.syncSubjectSchema(ctx, s)
				if err == nil {
					err = m.syncSubjectCompatibility(ctx, s.Subject)
				}
				results <- syncResult{src: s, info: info, err: err}
			}
		})
	}
```

internal/impl/redpanda/migrator/migrator_schema_registry.go
```diff
@@ -469,46 +538,306 @@ func (m *schemaRegistryMigrator) Sync(ctx context.Context) error {
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
+
+		batchEnd := batchStart + m.conf.BatchSize
+		if batchEnd > len(subjects) {
+			batchEnd = len(subjects)
+		}
```
josephwoodward: These 4 lines could be replaced with:
```go
batchEnd := min(batchStart + m.conf.BatchSize, len(subjects))
```

internal/impl/redpanda/migrator/migrator_schema_registry.go
```diff
@@ -469,46 +538,306 @@ func (m *schemaRegistryMigrator) Sync(ctx context.Context) error {
...
 	}
 
-	for _, s := range all {
-		srcInfo := schemaInfoFromSubjectSchema(s)
+	// Sample first batch to detect references - if found, sort all subjects by ID
+	hasReferences := false
+	sampleSize := m.conf.BatchSize
+	if sampleSize > len(subjects) {
+		sampleSize = len(subjects)
+	}
```
josephwoodward: These 4 lines could be replaced by:
```go
sampleSize := min(m.conf.BatchSize, len(subjects))
```

vuldin:
Thanks for the quick review @josephwoodward , I've updated the branch based on your suggestions. Let me know if you have any questions or additional suggestions.

vuldin:
I was able to complete a test using a build of this branch for schema migration:
- count: 120K schemas
- duration: ~3 hours
- memory growth: stable (~50MB)
- schema IDs remained identical across clusters
- errors: none

mmatczuk:
Work moved to https://github.com/redpanda-data/connect/pull/3951

PR #3954, opened 2026-02-04
authz cli hardening for cloud

mmatczuk:


internal/cli/enterprise.go
```diff
@@ -145,6 +151,14 @@ func InitEnterpriseCLI(binaryName, version, dateBuilt string, schema *service.Co
 			}
 
 			// Store authorization configuration if present
+			if dist == DistCloud || dist == DistCloudAI {
```
rockwotj, line 154: I don't think we can roll this out so aggressively for the existing gateway component unless we know there is no usage in production

mmatczuk: Let's skip it for now and revisit it later.

PR #3964, opened 2026-02-05
Enable Google Drive processors to use Shared Drives

MikolajGagatekRP:
Enable Google Drive Shared Drives Support

  This PR adds support for accessing files stored in Google Drive Shared Drives (formerly Team Drives) for the gcp_drive_search and gcp_drive_download processors.

  Changes

  - Added shared_drives boolean field to both processors (defaults to false for backwards compatibility)
  - Updated gcp_drive_search to include items from shared drives when enabled (SupportsAllDrives + IncludeItemsFromAllDrives)
  - Updated gcp_drive_download to download files from shared drives when enabled (SupportsAllDrives)

  Usage

  # Search files in shared drives
```
  processor:
    gcp_drive_search:
      query: "name contains 'report'"
      shared_drives: true
```

  # Download files from shared drives
```
  processor:
    gcp_drive_download:
      file_id: "${!json(\"id\")}"
      shared_drives: true
```

```
  Files: internal/impl/google/drive_search.go, internal/impl/google/drive_download.go
```

### DIFF
--- a/internal/impl/google/drive_search.go
+++ b/internal/impl/google/drive_search.go
@@ -156,11 +156,16 @@ func (g *googleDriveSearchProcessor) Process(ctx context.Context, msg *service.M
 		Context(ctx).
 		Q(q).
 		PageSize(min(int64(g.maxResults), 100)).
-		SupportsAllDrives(g.sharedDrives).
-		IncludeItemsFromAllDrives(g.sharedDrives).
 		Fields("nextPageToken", googleapi.Field("files("+strings.Join(g.fields, ",")+")"))
 	if l != "" {
 		call = call.IncludeLabels(l)
+	}
+	if g.sharedDrives {
+		// all of those flags are needed to look into shared drives
+		call.
+			SupportsAllDrives(g.sharedDrives).         // Flag 1: Tells API you know about Shared Drives
+			IncludeItemsFromAllDrives(g.sharedDrives). // Flag 2: Tells API to actually look in them
+			Corpora("allDrives")                       // Flag 3: Look everywhere the SA has access
 	}
 	var files []*drive.File
 	err = call.Pages(ctx, func(page *drive.FileList) error {


internal/impl/google/drive_download.go
```diff
@@ -27,6 +27,7 @@ const (
 	driveDownloadFieldFileID          = "file_id"
 	driveDownloadFieldMimeType        = "mime_type"
 	driveDownloadFieldExportMimeTypes = "export_mime_types"
+	driveDownloadSupportSharedDrives  = "shared_drives"
```
mmatczuk: Please add "Field"

internal/impl/google/drive_download.go
```diff
@@ -73,6 +74,9 @@ Can download a file from Google Drive based on a file ID.
 					"application/vnd.google-apps.drawing":      "image/svg+xml",
 				}).
 				Advanced(),
+			service.NewBoolField(driveDownloadSupportSharedDrives).
+				Description("Whether or not to include shared drives.").
+				Default(false),
```
mmatczuk, line 79: Is that needed?

internal/impl/google/drive_download.go
```diff
@@ -163,8 +174,8 @@ func (g *googleDriveDownloadProcessor) Process(ctx context.Context, msg *service
 	return service.MessageBatch{msg}, nil
 }
 
-func downloadFile(ctx context.Context, srv *drive.Service, fileID string) ([]byte, error) {
-	resp, err := srv.Files.Get(fileID).Context(ctx).Download()
+func downloadFile(ctx context.Context, srv *drive.Service, fileID string, sharedDrives bool) ([]byte, error) {
```
mmatczuk: Maybe we could make it a method?

internal/impl/google/drive_download.go
MikolajGagatekRP: ahhh didnt notice that, fixing 🫡

internal/impl/google/drive_download.go
MikolajGagatekRP: @birdayz suggested adding it.
We are preparing pipelines for CKO ADP Workshop and using our RP shared docs would be very helpful for that. Hence to achieve it we need to turn on that function to search/download files including SharedDrives

PR #3965, opened 2026-02-05
Auto schema for mysql cdc

Jeffail:
This PR attempts to extract the table schema whilst consuming rows and adds them as metadata to each consumed message. The schemas are currently cached but there is no concept of a TTL, we should probably consider adding this as the table schema could always change between snapshots.


internal/impl/confluent/serde_goavro_test.go
```diff
@@ -184,6 +184,39 @@ func TestAvroReferences(t *testing.T) {
 	}
 }
 
+// assertSchemaFieldsMatch checks that all expected fields in the expected schema match
+// the actual schema, while ignoring any extra fields (like "fingerprint") in the actual schema.
+// This allows tests to be resilient to future schema format extensions.
+func assertSchemaFieldsMatch(t *testing.T, expected, actual any) {
```
mmatczuk, line 190: How about cmp.Diff

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -198,6 +199,10 @@ type mysqlStreamInput struct {
 	// IAM authentication fields
 	iamAuthEnabled      bool
 	iamAuthTokenBuilder TokenBuilder
+
+	// Table schemas - stored as serialized format (map[string]any) for metadata
+	tableSchemas map[string]any
+	schemaMutex  sync.RWMutex
```
mmatczuk: Typically tableSchemasMu

internal/impl/mysql/input_mysql_stream.go
```diff
@@ -833,18 +849,18 @@ func (i *mysqlStreamInput) onMessage(e *canal.RowsEvent, initValue, incrementVal
 	return nil
 }
 
-func mapMessageColumn(v any, col schema.TableColumn) (any, error) {
+func mapMessageColumn(v any, col gomysqlschema.TableColumn) (any, error) {
```
mmatczuk: Why do new need to change it.

internal/impl/mysql/integration_test.go
```diff
@@ -749,3 +749,222 @@ file:
...
+	// Insert CDC rows
+	db.Exec("INSERT INTO test_schema VALUES (3, 'cdc1', '2024-01-03 12:00:00', 92.1, '{\"key\":\"value3\"}', 'tag2')")
+	db.Exec("INSERT INTO test_schema VALUES (4, 'cdc2', '2024-01-04 12:00:00', 88.7, '{\"key\":\"value4\"}', 'tag2,tag3')")
+
+	// Wait for CDC events
+	assert.Eventually(t, func() bool {
+		msgMut.Lock()
+		defer msgMut.Unlock()
+		return len(messages) == 4
+	}, time.Minute, time.Millisecond*100)
```
mmatczuk, line 864: We could use a channel instead

internal/impl/mysql/input_mysql_stream.go
Jeffail: Yeah this was an older change, fixed

internal/impl/confluent/serde_goavro_test.go
Jeffail: it looks like it'd be awkward, we can use cmp.Option to try and avoid erroring on extra fields but it would also make the tests weaker

PR #3966, opened 2026-02-05
Dynamodb cdc snapshot

squiidz:


mmatczuk:
### Code review

Found 5 issues:

1. **`snapshot_only` mode hangs after scan completes** (bugs: "snapshot goroutine lifecycle")

   In `snapshot_only` mode, no shard coordinator is started (`startShardCoordinator` only runs for `snapshot_and_cdc`). When the snapshot goroutine finishes, it calls `TriggerSoftStop()` (line 682), but nobody listens on `SoftStopChan()` to close `msgChan` and call `TriggerHasStopped()` — that logic lives in `startShardCoordinator`'s defer (lines 830-831). `ReadBatch` selects on `HasStoppedChan()` (line 1292), `ctx.Done()`, and `msgChan`. Once all snapshot batches drain from `msgChan`, `ReadBatch` blocks forever: `HasStoppedChan()` never fires because `TriggerHasStopped()` is never called, and `msgChan` is never closed. The input hangs until the framework's external context times out.

   https://github.com/redpanda-data/connect/blob/b00dcc8ba7ba9591eec0671585323d9f07b98910/internal/impl/aws/input_dynamodb_cdc.go#L678-L686

2. **Deduplication is dead code: `ShouldSkipCDCEvent` is never called** (bugs: "dedup not wired")

   The snapshot path populates the dedup buffer via `RecordSnapshotItem` (line 1097), but the CDC path (`startShardReader` / `convertRecordsToBatch`) never calls `ShouldSkipCDCEvent`. The buffer is built at memory/CPU cost but never consulted, so `snapshot_deduplicate: true` (the default) has no effect. Additionally, even if wired up, the comparison would be broken: snapshot items store `startTime.Format(time.RFC3339Nano)` (a timestamp like `"2026-02-18T..."`) as the sequence number, while CDC events use DynamoDB stream sequence numbers (21-digit decimal strings). The lexicographic comparison `sequenceNum <= snapshotSeq` has no semantic validity across these two formats.

   https://github.com/redpanda-data/connect/blob/b00dcc8ba7ba9591eec0671585323d9f07b98910/internal/impl/aws/input_dynamodb_cdc.go#L1091-L1099

3. **`ErrEndOfInput` returned from `Connect()` causes retry loop** (bugs: "wrong return site for ErrEndOfInput")

   When `snapshot_only` mode detects a previously completed snapshot, `connectWithSnapshot` returns `service.ErrEndOfInput` from `Connect()` (line 589). The benthos framework treats any error from `Connect()` as a connection failure and retries with backoff. `ErrEndOfInput` is intended to be returned from `ReadBatch()`, not `Connect()`. This path creates an infinite retry loop where the connector repeatedly connects, finds the snapshot complete, returns `ErrEndOfInput`, and reconnects.

   https://github.com/redpanda-data/connect/blob/b00dcc8ba7ba9591eec0671585323d9f07b98910/internal/impl/aws/input_dynamodb_cdc.go#L586-L590

4. **`time.NewTicker` panics if `snapshot_throttle: 0`** (bugs: "missing validation")

   `scanSegment` calls `time.NewTicker(s.throttle)` (line 147). `time.NewTicker` panics on zero or negative duration. The `snapshot_throttle` config field has `Default("100ms")` but no `LintRule` preventing `0`. Other snapshot fields like `snapshot_segments` and `snapshot_batch_size` have explicit lint rules, but this one is missing. A user setting `snapshot_throttle: 0` (a reasonable attempt to disable throttling) crashes the process.

   https://github.com/redpanda-data/connect/blob/b00dcc8ba7ba9591eec0671585323d9f07b98910/internal/impl/aws/dynamocdc/snapshot.go#L145-L149

5. **`buildItemKeyString` produces non-deterministic keys for most DynamoDB schemas** (bugs: "dedup key instability")

   The function checks only 5 hardcoded key names (`id`, `Id`, `ID`, `pk`, `PK`). Tables using any other key naming (e.g., `userId`, `order_id`, `hash_key`) fall through to the fallback that iterates all item attributes. Go map iteration order is non-deterministic, so the same item produces different key strings across calls. Even when matched, the snapshot path builds keys from the full item map while the CDC path would need to build keys from `record.Dynamodb.Keys` — these produce different strings for the same item, making shard lookups always miss. This is moot while issue 2 exists (dedup is dead code), but the implementation is fundamentally broken for when it gets wired up.

   https://github.com/redpanda-data/connect/blob/b00dcc8ba7ba9591eec0671585323d9f07b98910/internal/impl/aws/input_dynamodb_cdc.go#L1159-L1195

Generated with Claude Code

mmatczuk:
### Code review

Found 4 issues:

1. **Deduplication buffer populated but never consulted** (bugs: incomplete feature)

`ShouldSkipCDCEvent` is defined but never called anywhere. The CDC reader `startShardReader` is not modified to check the sequence buffer, so `snapshot_deduplicate: true` allocates and populates a buffer that has no effect. Users enabling this option get no deduplication.

https://github.com/redpanda-data/connect/blob/b00dcc8ba7ba9591eec0671585323d9f07b98910/internal/impl/aws/input_dynamodb_cdc.go#L361-L380

https://github.com/redpanda-data/connect/blob/b00dcc8ba7ba9591eec0671585323d9f07b98910/internal/impl/aws/input_dynamodb_cdc.go#L1094-L1098

2. **`buildItemKeyString` fallback produces non-deterministic keys** (bugs: logic error)

When an item's attributes don't match the hardcoded key name list (`id`, `Id`, `ID`, `pk`, `PK`), the fallback iterates `map[string]AttributeValue` in Go's non-deterministic order. The same item produces different key strings across calls, so deduplication lookups would always miss. Most real DynamoDB tables use key names like `user_id`, `order_id`, `customerId`, etc., making this the common path.

https://github.com/redpanda-data/connect/blob/b00dcc8ba7ba9591eec0671585323d9f07b98910/internal/impl/aws/input_dynamodb_cdc.go#L1178-L1192

3. **`time.Sleep` during throttle backoff ignores context cancellation** (bugs: shutdown delay)

`scanSegment` calls `time.Sleep(wait)` with up to 5s backoff (MaxInterval), blocking regardless of context cancellation. During shutdown, this stalls `Close()` by up to 5s per throttled segment. Should use `select` with `ctx.Done()` and `time.After(wait)`.

https://github.com/redpanda-data/connect/blob/b00dcc8ba7ba9591eec0671585323d9f07b98910/internal/impl/aws/dynamocdc/snapshot.go#L190-L194

4. **`snapshot_only` mode hangs after scan completes on first run** (bugs: logic error)

When the snapshot finishes, the background goroutine calls `d.shutSig.TriggerSoftStop()` (line 682). But `ReadBatch` listens on `shutSig.HasStoppedChan()` (hard stop, line 1292), not soft stop. After the last message is consumed from `msgChan`, `ReadBatch` blocks indefinitely — it never returns `ErrEndOfInput`. The `ErrEndOfInput` path at line 589 only fires on restart when the snapshot is already marked complete.

https://github.com/redpanda-data/connect/blob/b00dcc8ba7ba9591eec0671585323d9f07b98910/internal/impl/aws/input_dynamodb_cdc.go#L679-L683

https://github.com/redpanda-data/connect/blob/b00dcc8ba7ba9591eec0671585323d9f07b98910/internal/impl/aws/input_dynamodb_cdc.go#L1289-L1299

### Areas for human review

1. **`ShouldSkipCDCEvent` compares incompatible string types** — currently dead code (issue 1 above), but if wired in, it compares DynamoDB sequence numbers (e.g. `"4961760..."`) against RFC3339 timestamps (e.g. `"2026-02-18T..."`) via string `<=`. These are lexicographically incompatible; deduplication would never skip anything.

https://github.com/redpanda-data/connect/blob/b00dcc8ba7ba9591eec0671585323d9f07b98910/internal/impl/aws/input_dynamodb_cdc.go#L377-L379

https://github.com/redpanda-data/connect/blob/b00dcc8ba7ba9591eec0671585323d9f07b98910/internal/impl/aws/input_dynamodb_cdc.go#L1096-L1097

2. **TOCTOU race on overflow detection in `snapshotSequenceBuffer`** — multiple scan segment goroutines can all pass the `overflow.Load()` check at line 333 before any of them detects the limit at line 348, allowing `totalCount` to exceed `maxSize` by up to `segments * batchSize`. May be acceptable as best-effort, but the documented guarantee ("If exceeded, deduplication is disabled") is not precise.

https://github.com/redpanda-data/connect/blob/b00dcc8ba7ba9591eec0671585323d9f07b98910/internal/impl/aws/input_dynamodb_cdc.go#L331-L357

Generated with Claude Code

internal/impl/aws/dynamocdc/snapshot.go
```diff
@@ -0,0 +1,327 @@
...
+}
+
+// isThrottlingError checks if an error is due to AWS throttling
+func isThrottlingError(err error) bool {
+	if err == nil {
+		return false
+	}
+	var limitErr *types.LimitExceededException
+	var provisionedErr *types.ProvisionedThroughputExceededException
+	return errors.As(err, &limitErr) || errors.As(err, &provisionedErr)
```
josephwoodward: Seeing as we don't return the error here, after rebasing on main you'll be able to use the new [AsType](https://pkg.go.dev/errors#AsType) function. For instance:

```go
if pathError, ok := errors.AsType[*fs.PathError](err); ok {
    fmt.Println("Failed at path:", pathError.Path)
} 
```

internal/impl/aws/dynamocdc/checkpoint.go
```diff
@@ -152,3 +153,130 @@ func (c *Checkpointer) FlushCheckpoints(ctx context.Context, checkpoints map[str
 	}
 	return nil
 }
+
+// GetSnapshotProgress retrieves the snapshot checkpoint
```
mmatczuk: Docs shall have a dot at the end. If it's Claude update md files.

internal/impl/aws/dynamocdc/snapshot.go
```diff
@@ -0,0 +1,327 @@
...
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
+	"github.com/cenkalti/backoff/v4"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/redpanda-data/benthos/v4/public/service"
+)
+
+// SnapshotScanner performs parallel scanning of a DynamoDB table.
```
mmatczuk: It would be good to document the method here.

internal/impl/aws/dynamocdc/snapshot.go
```diff
@@ -0,0 +1,327 @@
...
+	table              string
+	segments           int
+	batchSize          int
+	throttle           time.Duration
+	checkpointer       *Checkpointer
+	checkpointInterval int // Checkpoint every N batches (0 = every batch)
+	log                *service.Logger
+
+	// Callbacks
+	onBatch            func(ctx context.Context, items []map[string]types.AttributeValue, segment int) error
```
mmatczuk: Can we have a type for that `[]map[string]types.AttributeValue`

internal/impl/aws/dynamocdc/snapshot.go
```diff
@@ -0,0 +1,327 @@
...
+	onProgress         func(segment, totalSegments int, recordsRead int64)
+	onCheckpointFailed func(segment int, err error)
+	onSegmentComplete  func(segment int, duration time.Duration, recordsRead int64)
+
+	// State tracking
+	segmentsActive atomic.Int32
```
mmatczuk: nit: segmentsActive or activeSegments

internal/impl/aws/dynamocdc/snapshot.go
```diff
@@ -0,0 +1,327 @@
...
+	s.onCheckpointFailed = fn
+}
+
+// SetSegmentCompleteCallback sets the callback for segment completion with duration tracking.
+func (s *SnapshotScanner) SetSegmentCompleteCallback(fn func(segment int, duration time.Duration, recordsRead int64)) {
+	s.onSegmentComplete = fn
+}
+
+// GetActiveSegments returns the current number of active scan segments.
+func (s *SnapshotScanner) GetActiveSegments() int {
```
mmatczuk: Look for methods starting with Get and remove the prefix to - that's the idiomatic Go.

internal/impl/aws/dynamocdc/snapshot.go
```diff
@@ -0,0 +1,327 @@
...
+		throttleTicker   = time.NewTicker(s.throttle)
+		firstRequest     = true
+	)
+	defer throttleTicker.Stop()
+
+	// Initialize backoff for throttling errors
+	boff := backoff.NewExponentialBackOff()
+	boff.InitialInterval = 200 * time.Millisecond
+	boff.MaxInterval = 5 * time.Second
+	boff.MaxElapsedTime = 0 // Never give up
```
mmatczuk: Shall be controllable by user

internal/impl/aws/dynamocdc/snapshot.go
```diff
@@ -0,0 +1,327 @@
...
+			TableName:         aws.String(s.table),
+			Limit:             aws.Int32(int32(s.batchSize)),
+			Segment:           aws.Int32(int32(segment)),
+			TotalSegments:     aws.Int32(int32(s.segments)),
+			ExclusiveStartKey: lastEvaluatedKey,
+			// Use eventually consistent reads (cheaper and faster)
+			ConsistentRead: aws.Bool(false),
+		})
+		if err != nil {
+			// Check if throttled
```
mmatczuk: This comment adds little value

internal/impl/aws/dynamocdc/snapshot.go
```diff
@@ -0,0 +1,327 @@
...
+			if isThrottlingError(err) {
+				wait := boff.NextBackOff()
+				s.log.Warnf("Segment %d throttled, backing off for %v", segment, wait)
+				// Respect context cancellation during backoff
+				select {
+				case <-ctx.Done():
+					return ctx.Err()
+				case <-time.After(wait):
+					// Backoff complete, retry
+				}
```
mmatczuk: We could use [SleepWithContext](https://pkg.go.dev/github.com/aws/smithy-go/time#SleepWithContext)

internal/impl/aws/dynamodb/snapshot.go
```diff
@@ -0,0 +1,327 @@
...
+			TotalSegments:     aws.Int32(int32(s.segments)),
+			ExclusiveStartKey: lastEvaluatedKey,
+			// Use eventually consistent reads (cheaper and faster)
+			ConsistentRead: aws.Bool(false),
+		})
+		if err != nil {
+			// Check if throttled
+			if isThrottlingError(err) {
+				wait := boff.NextBackOff()
+				s.log.Warnf("Segment %d throttled, backing off for %v", segment, wait)
```
mmatczuk, line 205: Can this be 0?

internal/impl/aws/dynamocdc/snapshot.go
```diff
@@ -0,0 +1,327 @@
...
+				continue
+			}
+			return fmt.Errorf("scan failed for segment %d: %w", segment, err)
+		}
+
+		// Success - reset backoff
+		boff.Reset()
+
+		// Process batch if we got items
+		if len(result.Items) > 0 {
```
mmatczuk: Good practice is to handle special cases return early

internal/impl/aws/dynamocdc/snapshot.go
```diff
@@ -0,0 +1,327 @@
...
+			if err := s.onBatch(ctx, result.Items, segment); err != nil {
+				return fmt.Errorf("failed to process batch for segment %d: %w", segment, err)
```
mmatczuk: failed to process/processing

internal/impl/aws/dynamocdc/snapshot.go
```diff
@@ -0,0 +1,327 @@
...
+			}
+			recordsRead += int64(len(result.Items))
+			batchCount++
+		}
+
+		// Update checkpoint periodically (not after every batch)
+		shouldCheckpoint := s.checkpointer != nil && (batchCount%s.checkpointInterval == 0 || result.LastEvaluatedKey == nil)
```
mmatczuk: Could this be a function - shouldCheckpoint()

internal/impl/aws/dynamodb/snapshot.go
```diff
@@ -0,0 +1,327 @@
...
+}
+
+// isThrottlingError checks if an error is due to AWS throttling.
+func isThrottlingError(err error) bool {
+	if err == nil {
+		return false
+	}
+	_, isLimit := errors.AsType[*types.LimitExceededException](err)
+	_, isProvisioned := errors.AsType[*types.ProvisionedThroughputExceededException](err)
+	return isLimit || isProvisioned
```
mmatczuk: Cool

internal/impl/aws/dynamocdc/checkpoint.go
```diff
@@ -152,3 +153,130 @@ func (c *Checkpointer) FlushCheckpoints(ctx context.Context, checkpoints map[str
 	}
 	return nil
 }
+
+// GetSnapshotProgress retrieves the snapshot checkpoint.
+func (c *Checkpointer) GetSnapshotProgress(ctx context.Context) (*SnapshotCheckpoint, error) {
+	checkpoint := NewSnapshotCheckpoint()
+
+	// Check if snapshot is complete
+	result, err := c.svc.GetItem(ctx, &dynamodb.GetItemInput{
```
mmatczuk: Standard naming for request is req and respone res

internal/impl/aws/dynamocdc/checkpoint.go
```diff
@@ -152,3 +153,130 @@ func (c *Checkpointer) FlushCheckpoints(ctx context.Context, checkpoints map[str
...
+		TableName: aws.String(c.tableName),
+		Key: map[string]types.AttributeValue{
+			"StreamArn": &types.AttributeValueMemberS{Value: c.streamArn},
+			"ShardID":   &types.AttributeValueMemberS{Value: "snapshot#complete"},
+		},
+	})
+	if err != nil {
+		var aerr *types.ResourceNotFoundException
+		if !errors.As(err, &aerr) {
```
mmatczuk: Use AsType

internal/impl/aws/dynamocdc/checkpoint.go
```diff
@@ -152,3 +153,130 @@ func (c *Checkpointer) FlushCheckpoints(ctx context.Context, checkpoints map[str
...
+		TableName:              aws.String(c.tableName),
+		KeyConditionExpression: aws.String("StreamArn = :stream_arn AND begins_with(ShardID, :snapshot_prefix)"),
+		ExpressionAttributeValues: map[string]types.AttributeValue{
+			":stream_arn":      &types.AttributeValueMemberS{Value: c.streamArn},
+			":snapshot_prefix": &types.AttributeValueMemberS{Value: "snapshot#segment#"},
+		},
+	})
+	if err != nil {
+		var aerr *types.ResourceNotFoundException
+		if !errors.As(err, &aerr) {
```
mmatczuk: Same

internal/impl/aws/dynamodb/checkpoint.go
```diff
@@ -152,3 +153,130 @@ func (c *Checkpointer) FlushCheckpoints(ctx context.Context, checkpoints map[str
...
+		}
+		// Table doesn't exist yet, return empty checkpoint
+		return checkpoint, nil
+	}
+
+	// Parse segment progress from items
+	for _, item := range queryResult.Items {
+		shardID, ok := item["ShardID"].(*types.AttributeValueMemberS)
+		if !ok {
+			continue
```
mmatczuk, line 198: Log it or panic.

internal/impl/aws/dynamodb/checkpoint.go
```diff
@@ -152,3 +153,130 @@ func (c *Checkpointer) FlushCheckpoints(ctx context.Context, checkpoints map[str
...
+		}
+
+		var segmentID int
+		if _, err := fmt.Sscanf(shardID.Value, "snapshot#segment#%d", &segmentID); err != nil {
```
mmatczuk, line 202: So cool

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -118,6 +144,32 @@ This input emits the following metrics:
 				Description("Time to wait when applying backpressure due to too many in-flight messages.").
 				Default(defaultDynamoDBThrottleBackoff).
 				Advanced(),
+			service.NewStringEnumField("snapshot_mode", "none", "snapshot_only", "snapshot_and_cdc").
+				Description("Snapshot behavior. `none`: CDC only (default). `snapshot_only`: one-time table scan, no streaming. `snapshot_and_cdc`: scan entire table then stream changes.").
+				Default("none"),
+			service.NewIntField("snapshot_segments").
+				Description("Number of parallel scan segments (1-10). Higher parallelism scans faster but consumes more RCUs. Start with 1 for safety.").
```
mmatczuk: What does it mean here 1-10?

mmatczuk: This is AWS limitation, right?

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -196,6 +278,127 @@ type dynamoDBShardReader struct {
...
+	startTime     time.Time
+	endTime       time.Time
+	seqBuffer     *snapshotSequenceBuffer
+	scanner       *dynamocdc.SnapshotScanner
+	recordsRead   atomic.Int64
+	segmentsTotal int
+}
+
+// snapshotSequenceBuffer tracks sequence numbers seen during snapshot for deduplication
+// Uses sharded locks for better concurrency (10-30x less contention on high-core machines)
```
mmatczuk: We do not really know that.

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -196,6 +278,127 @@ type dynamoDBShardReader struct {
...
+type snapshotSequenceBuffer struct {
+	shards     [32]bufferShard // 32 shards for good distribution
```
mmatczuk: Why 32? Do the maths pls.

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -196,6 +278,127 @@ type dynamoDBShardReader struct {
...
+	sequences map[string]string // item key -> sequence number seen in snapshot
+}
+
+func newSnapshotSequenceBuffer(maxSize int) *snapshotSequenceBuffer {
+	buf := &snapshotSequenceBuffer{
+		maxSize: maxSize,
+	}
+	// Initialize each shard
+	for i := range buf.shards {
+		buf.shards[i].sequences = make(map[string]string, maxSize/32)
```
mmatczuk: Magic number, use const for the blessed 32 shards.

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -196,6 +278,127 @@ type dynamoDBShardReader struct {
...
+	}
+	return buf
+}
+
+// getShard returns the shard for a given key using FNV-1a hash
+func (s *snapshotSequenceBuffer) getShard(key string) *bufferShard {
```
mmatczuk: Let's not roll our own hash functions here please. We have murmur to start with.

mmatczuk: There is also hash mod now.

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -196,6 +278,127 @@ type dynamoDBShardReader struct {
...
+// snapshotSequenceBuffer tracks sequence numbers seen during snapshot for deduplication
+// Uses sharded locks for better concurrency (10-30x less contention on high-core machines)
+type snapshotSequenceBuffer struct {
+	shards     [32]bufferShard // 32 shards for good distribution
+	maxSize    int
+	totalCount atomic.Int64 // Track total size across all shards
+	overflow   atomic.Bool  // true if buffer exceeded maxSize
+}
+
+// bufferShard is a single shard of the buffer with its own lock
```
mmatczuk: nit(global): Every function comment shall be a sentence and end with a dot.

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -196,6 +278,127 @@ type dynamoDBShardReader struct {
...
+	if s.overflow.Load() {
+		return
+	}
+
+	shard := s.getShard(key)
+	shard.mu.Lock()
+	defer shard.mu.Unlock()
+
+	// Check if key already exists (update, not insert)
+	if _, exists := shard.sequences[key]; exists {
```
mmatczuk: nit: standard naming is ok

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -196,6 +278,127 @@ type dynamoDBShardReader struct {
...
+		shard.sequences[key] = sequenceNum
+		return
+	}
+
+	// Atomic check-and-increment using compare-and-swap to prevent TOCTOU race
```
mmatczuk: Dot.

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -196,6 +278,127 @@ type dynamoDBShardReader struct {
...
+			}
+			return
+		}
+		// Try to increment atomically
+		if s.totalCount.CompareAndSwap(current, current+1) {
+			// Successfully incremented, proceed with insertion
+			break
+		}
+		// CAS failed (another goroutine modified count), retry the loop
+	}
```
mmatczuk: Worth extracting to function and documenting

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -196,6 +278,127 @@ type dynamoDBShardReader struct {
...
+
+	// Atomic check-and-increment using compare-and-swap to prevent TOCTOU race
+	// This ensures we never exceed maxSize even with concurrent goroutines
+	for {
+		current := s.totalCount.Load()
+		if current >= int64(s.maxSize) {
+			// Buffer full - set overflow flag
+			if !s.overflow.Load() {
+				s.overflow.Store(true)
```
mmatczuk: We should have logs around it!

mmatczuk: Actually metrics too but can be volatile

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -278,11 +528,27 @@ func (d *dynamoDBCDCInput) Connect(ctx context.Context) error {
...
-	// Initialize message channel
-	d.msgChan = make(chan asyncMessage)
+	// Initialize message channel with buffer to reduce blocking between scanner and processor
+	// Buffer size of 1000 allows scanner to work ahead without blocking
+	d.msgChan = make(chan asyncMessage, 1000)
 
 	d.log.Infof("Connected to DynamoDB stream: %s", *d.streamArn)
 
+	// Handle snapshot mode
+	if d.conf.snapshot.mode != "none" {
```
mmatczuk: Can we have enum for that, or at least string constants.

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -278,11 +528,27 @@ func (d *dynamoDBCDCInput) Connect(ctx context.Context) error {
...
+	}
+
+	// CDC-only mode (existing behavior)
+	return d.connectCDCOnly(ctx)
+}
+
+// connectCDCOnly starts CDC streaming without snapshot (original behavior)
+func (d *dynamoDBCDCInput) connectCDCOnly(ctx context.Context) error {
+	// Mark snapshot as complete (never started)
+	d.snapshot.state.Store(2)
```
mmatczuk: We must have enums for that.

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -307,6 +573,205 @@ func (d *dynamoDBCDCInput) Connect(ctx context.Context) error {
...
+				// Clear the snapshot completion marker to force re-snapshot
+				// Don't return here - fall through to run snapshot again
+				snapshotCheckpoint = dynamocdc.NewSnapshotCheckpoint() // Reset to empty
+			} else {
+				// CDC checkpoint is valid, safe to skip snapshot
+				d.snapshot.state.Store(2) // complete
+				d.metrics.snapshotState.Set(2)
+				return d.connectCDCOnly(ctx)
+			}
+		case "snapshot_only":
```
mmatczuk: Use constants

internal/impl/aws/dynamodb/input_cdc.go
```diff
@@ -307,6 +573,201 @@ func (d *dynamoDBCDCInput) Connect(ctx context.Context) error {
...
+		if err := d.snapshot.scanner.Scan(scanCtx, snapshotCheckpoint); err != nil {
+			if !errors.Is(err, context.Canceled) {
+				wrappedErr := fmt.Errorf("snapshot scan failed for table %s: %w", d.conf.table, err)
+				d.log.Errorf("%v", wrappedErr)
+				d.snapshot.errOnce.Do(func() {
+					d.snapshot.err = wrappedErr
+				})
+				d.snapshot.state.Store(3) // failed
+				d.metrics.snapshotState.Set(3)
+				return
```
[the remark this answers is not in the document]
mmatczuk: @squiidz ping?

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -603,11 +1100,236 @@ func (d *dynamoDBCDCInput) startShardReader(ctx context.Context, shardID string)
...
+		sb.WriteByte('=')
+		writeAttributeValueString(&sb, attr)
+	}
+
+	return sb.String()
+}
+
+// buildItemKeyStringFromStream creates a key string from DynamoDB Stream keys (for CDC deduplication)
+// Uses types.AttributeValue (from streams) instead of dynamodbtypes.AttributeValue (from table)
+func buildItemKeyStringFromStream(keys map[string]types.AttributeValue) string {
```
mmatczuk: I'd like to see a dedicated type here not string

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -603,11 +1100,236 @@ func (d *dynamoDBCDCInput) startShardReader(ctx context.Context, shardID string)
...
+		keyNames = append(keyNames, k)
+	}
+	// Simple bubble sort (fine for primary keys - usually 1-2 keys)
+	for i := 0; i < len(keyNames); i++ {
+		for j := i + 1; j < len(keyNames); j++ {
+			if keyNames[i] > keyNames[j] {
+				keyNames[i], keyNames[j] = keyNames[j], keyNames[i]
+			}
+		}
+	}
```
mmatczuk: C'mon slices.Sort

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -603,11 +1100,236 @@ func (d *dynamoDBCDCInput) startShardReader(ctx context.Context, shardID string)
...
+		sb.WriteString(k)
+		sb.WriteByte('=')
+		writeStreamAttributeValueString(&sb, keys[k])
+	}
+
+	return sb.String()
+}
+
+// writeStreamAttributeValueString writes a stream attribute value to a strings.Builder efficiently
+func writeStreamAttributeValueString(sb *strings.Builder, attr types.AttributeValue) {
```
mmatczuk: Maybe we want to wrap the strings.Builder and have a custom builder?

internal/impl/aws/input_dynamodb_cdc.go
```diff
@@ -797,3 +1554,45 @@ func convertAttributeValue(attr types.AttributeValue) any {
 		return nil
 	}
 }
+
+// convertDynamoDBAttributeMap converts DynamoDB table attribute values to Go types (for snapshot)
+func convertDynamoDBAttributeMap(attrs map[string]dynamodbtypes.AttributeValue) map[string]any {
```
mmatczuk: convert to what?

mmatczuk:
We need a larger scale testing of incremental snapshots. The test need to do reading and writing in parallel to test the cdc over snapshots logic.

mmatczuk:
Overall congratulations, keep it up.

internal/impl/aws/dynamodb/input_cdc.go
```diff
@@ -40,9 +46,26 @@ const (
...
+
+	// Metrics.
+	metricShardsTracked           = "dynamodb_cdc_shards_tracked"
+	metricShardsActive            = "dynamodb_cdc_shards_active"
+	metricSnapshotState           = "dynamodb_cdc_snapshot_state"
+	metricSnapshotRecordsRead     = "dynamodb_cdc_snapshot_records_read"
+	metricSnapshotSegmentsActive  = "dynamodb_cdc_snapshot_segments_active"
+	metricSnapshotBufferOverflow  = "dynamodb_cdc_snapshot_buffer_overflow"
+	metricCheckpointFailures      = "dynamodb_cdc_checkpoint_failures"
+	metricSnapshotSegmentDuration = "dynamodb_cdc_snapshot_segment_duration"
```
josephwoodward, line 68: We should document these 3 metrics as well.

internal/impl/aws/dynamodb/snapshot.go
```diff
@@ -0,0 +1,364 @@
...
+	onProgress         func(segment, totalSegments int, recordsRead int64)
+	onCheckpointFailed func(segment int, err error)
+	onSegmentComplete  func(segment int, duration time.Duration, recordsRead int64)
+
+	// State tracking
+	activeSegments atomic.Int32
+}
+
+// SnapshotScannerConfig holds configuration for snapshot scanning.
+type SnapshotScannerConfig struct {
```
josephwoodward: minor (non-blocking): consider moving the `SnapshotScannerConfig` above the `SnapshotScanner` so the scanner and its factory function live together. It's quite common to see the struct with the factory function below it so they're grouped in `go doc`.

internal/impl/aws/dynamodb/snapshot.go
```diff
@@ -0,0 +1,364 @@
...
+	BatchSize          int
+	Throttle           time.Duration
+	MaxBackoff         time.Duration // Maximum backoff on throttling errors (0 = no limit).
+	Checkpointer       *Checkpointer
+	CheckpointInterval int // Checkpoint every N batches (default: 10).
+	Logger             *service.Logger
+}
+
+// NewSnapshotScanner creates a new snapshot scanner.
+func NewSnapshotScanner(conf SnapshotScannerConfig) *SnapshotScanner {
```
josephwoodward, line 72: Does this need to be exported? (I know @mmatczuk like to export package configs to make them visible outside of the package).

internal/impl/aws/dynamodb/snapshot.go
```diff
@@ -0,0 +1,364 @@
...
+}
+
+// Scan performs the snapshot scan, optionally resuming from a checkpoint.
+func (s *SnapshotScanner) Scan(ctx context.Context, resume *SnapshotCheckpoint) error {
+	if s.onBatch == nil {
+		return errors.New("batch callback must be set before scanning")
+	}
+
+	g, ctx := errgroup.WithContext(ctx)
+	g.SetLimit(s.segments) // Limit concurrent goroutines
```
josephwoodward: thought: comment feels like of redundant.

internal/impl/aws/dynamodb/snapshot.go
```diff
@@ -0,0 +1,364 @@
...
+
+	s.log.Infof("Starting snapshot scan with %d segments", s.segments)
+
+	// Start a goroutine for each segment, skipping already-completed segments.
+	for segment := 0; segment < s.segments; segment++ {
+		segmentID := segment
```
josephwoodward, line 129: thought (non-blocking): not against this, but didn't Go 1.22 fix looped scoped variables?

internal/impl/aws/dynamodb/input_cdc.go
```diff
@@ -123,6 +162,36 @@ This input emits the following metrics:
 				Description("Time to wait when applying backpressure due to too many in-flight messages.").
 				Default(defaultDynamoDBThrottleBackoff).
 				Advanced(),
+			service.NewStringEnumField("snapshot_mode", "none", "snapshot_only", "snapshot_and_cdc").
+				Description("Snapshot behavior. `none`: CDC only (default). `snapshot_only`: one-time table scan, no streaming. `snapshot_and_cdc`: scan entire table then stream changes.").
+				Default("none"),
+			service.NewIntField("snapshot_segments").
+				Description("Number of parallel DynamoDB Scan segments (1-10, AWS API limitation). Each segment scans a portion of the table concurrently, increasing throughput at the cost of more provisioned read capacity.").
```
josephwoodward: The [docs indicate](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/dynamodb@v1.51.0#ScanInput.TotalSegments) that this value can be far greater than 10 (docs say 1000000). Am I looking at the wrong docs?

squiidz: You're right that's weird

internal/impl/aws/dynamodb/snapshot.go
```diff
@@ -0,0 +1,364 @@
...
+)
+
+// DynamoItems is a slice of DynamoDB attribute maps representing table items.
+type DynamoItems = []map[string]dynamodbtypes.AttributeValue
+
+// SnapshotScanner performs a parallel scan of a DynamoDB table using the
+// DynamoDB Scan API with configurable segment parallelism. It supports
+// resumable checkpointing, adaptive backoff on throttling, and reports
+// progress through user-supplied callbacks.
+type SnapshotScanner struct {
```
josephwoodward, line 50: Thought: Would it be worth moving some of these files into a module? I see batcher.go, cache.go, snapshot.go, checkpoint.go and their test files that all look like they could live in a separate dedicated cdc specific module (as opposed to living at the top level of the dynamodb folder).

Similar to the PostgreSQL (and others) CDC component https://github.com/redpanda-data/connect/tree/main/internal/impl/postgresql

squiidz: I can create a ticket to further split the aws package, I already merged one to split every product in it's own subpackage.

josephwoodward: I'm not necessarily thinking about splitting the AWS package, more moving he non AWS stuff into a sub-package (say `/aws/replication/` for instance) with a clearer API that the dynamodb input uses, so it's not living alongside the other non CDC AWS stuff such as the output and processors.

It'll make the integration with the input much clearer.

internal/impl/aws/dynamodb/input_cdc.go
```diff
@@ -226,6 +455,32 @@ func dynamoCDCInputConfigFromParsed(pConf *service.ParsedConfig) (conf dynamoDBC
 	if conf.throttleBackoff, err = pConf.FieldDuration("throttle_backoff"); err != nil {
 		return
 	}
+	if conf.snapshot.mode, err = pConf.FieldString("snapshot_mode"); err != nil {
+		return
+	}
+	if conf.snapshot.segments, err = pConf.FieldInt("snapshot_segments"); err != nil {
```
josephwoodward: Is it worth making these field names constants so they can be shared with the service definition (ie `service.NewIntField("snapshot_segments")`  and not risk them become misaligned?

PR #3967, opened 2026-02-05
Allow chroot directory to pre-exist for volume mount support

birdayz:
The chroot setup previously required the chroot directory to not exist, which prevented mounting ConfigMaps or other volumes directly inside the chroot path. This is a problem because the chroot-passthrough mechanism copies files (losing symlinks), which breaks fsnotify-based hot reload of authorization policy files mounted from ConfigMaps.

Allow the chroot directory to pre-exist so that Kubernetes volume mounts can place files directly inside it. Also handle EROFS errors in the makeReadOnly step, since read-only volume mounts (like ConfigMaps) cannot be chmod'd.

Pre-create /tmp/chroot in the cloud Dockerfile owned by the connect user so the process can populate the rest of the chroot structure at runtime.

Add integration tests covering hierarchical authz scope matching with production-like resource names and policy bindings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ Changelog
 
 All notable changes to this project will be documented in this file.
 
+## 4.80.1 - 2026-02-05
+
+### Changed
+
+- chroot: existing directories are now allowed. (@birdayz)
+
 ## 4.80.0 - 2026-02-04
 
 ### Added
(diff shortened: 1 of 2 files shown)


resources/docker/cloud.Dockerfile
```diff
@@ -27,6 +27,8 @@ COPY --from=build /etc/passwd /etc/passwd
 COPY --from=build /tmp/redpanda-connect /redpanda-connect
 COPY config/docker.yaml /connect.yaml
 
+RUN mkdir -p /tmp/chroot && chown 10001:10001 /tmp/chroot
```
rockwotj, line 34: A comment for this would be useful, otherwise LGTM

birdayz: done

rockwotj:
cc: @wreet @d3k4z @mmatczuk

Issue #3973, opened 2026-02-09
Custom headers in sign_jwt_*


Nicopfy:
Hello ! 

I'm trying to generate a signed JWT (RS256) with custom claims **and** custom headers (e. g. `kid`), and then store it in a variable to use in an HTTP body request.

I hit a wall with the standard `http` processor `jwt` block cause it force the JWT token generated to be in the `Authorization` header.

I looked into the Bloblang `sign_jwt_rs256` method, but it currently doesn't seem to support adding custom headers.

I was thinking of adding an optional second argument to the `sign_jwt_*` methods that accept an object of header. I've looked to the code and would be happy to submit a PR for this (or would love to hear from you if you have a workaround ;) )

Thanks !

Nicopfy:
Reopen until it's available on main :)

mmatczuk:
I think the PR has been merged.

PR #3976, opened 2026-02-10
feat: sign_jwt_* methods can take an optional param to add jwt headers

Nicopfy:
Related to this [issue](https://github.com/redpanda-data/connect/issues/3973)

### DIFF
--- a/docs/modules/guides/pages/bloblang/methods.adoc
+++ b/docs/modules/guides/pages/bloblang/methods.adoc
@@ -4175,7 +4175,7 @@ Introduced in version v4.20.0.
 ==== Parameters
 
 *`signing_secret`* &lt;string&gt; The secret to use for signing the token.  
-*`headers`* &lt;(optional) unknown&gt; Optional object of JWT header fields to include in the token. Keys "alg" and "typ" will be ignored if provided.  
+*`headers`* &lt;(optional) unknown&gt; Optional object of JWT header fields to include in the token. Keys "alg", "typ", "jku", "jwk", "x5u", "x5c", "x5t","x5t#S256" and "crit" will be ignored if provided.  
 
 ==== Examples
 
@@ -4208,7 +4208,7 @@ Introduced in version v4.20.0.
 ==== Parameters
 
 *`signing_secret`* &lt;string&gt; The secret to use for signing the token.  
-*`headers`* &lt;(optional) unknown&gt; Optional object of JWT header fields to include in the token. Keys "alg" and "typ" will be ignored if provided.  
+*`headers`* &lt;(optional) unknown&gt; Optional object of JWT header fields to include in the token. Keys "alg", "typ", "jku", "jwk", "x5u", "x5c", "x5t","x5t#S256" and "crit" will be ignored if provided.  
 
 ==== Examples
 
@@ -4241,7 +4241,7 @@ Introduced in version v4.20.0.
 ==== Parameters
 
 *`signing_secret`* &lt;string&gt; The secret to use for signing the token.  
-*`headers`* &lt;(optional) unknown&gt; Optional object of JWT header fields to include in the token. Keys "alg" and "typ" will be ignored if provided.  
+*`headers`* &lt;(optional) unknown&gt; Optional object of JWT header fields to include in the token. Keys "alg", "typ", "jku", "jwk", "x5u", "x5c", "x5t","x5t#S256" and "crit" will be ignored if provided.  
 
 ==== Examples
 
@@ -4274,7 +4274,7 @@ Introduced in version v4.12.0.
 ==== Parameters
 
 *`signing_secret`* &lt;string&gt; The secret to use for signing the token.  
-*`headers`* &lt;(optional) unknown&gt; Optional object of JWT header fields to include in the token. Keys "alg" and "typ" will be ignored if provided.  
+*`headers`* &lt;(optional) unknown&gt; Optional object of JWT header fields to include in the token. Keys "alg", "typ", "jku", "jwk", "x5u", "x5c", "x5t","x5t#S256" and "crit" will be ignored if provided.  
 
 ==== Examples
 
@@ -4303,7 +4303,7 @@ Introduced in version v4.12.0.
 ==== Parameters
 
 *`signing_secret`* &lt;string&gt; The secret to use for signing the token.  
-*`headers`* &lt;(optional) unknown&gt; Optional object of JWT header fields to include in the token. Keys "alg" and "typ" will be ignored if provided.  
+*`headers`* &lt;(optional) unknown&gt; Optional object of JWT header fields to include in the token. Keys "alg", "typ", "jku", "jwk", "x5u", "x5c", "x5t","x5t#S256" and "crit" will be ignored if provided.  
 
 ==== Examples
 
@@ -4332,7 +4332,7 @@ Introduced in version v4.12.0.
 ==== Parameters
 
 *`signing_secret`* &lt;string&gt; The secret to use for signing the token.  
-*`headers`* &lt;(optional) unknown&gt; Optional object of JWT header fields to include in the token. Keys "alg" and "typ" will be ignored if provided.  
+*`headers`* &lt;(optional) unknown&gt; Optional object of JWT header fields to include in the token. Keys "alg", "typ", "jku", "jwk", "x5u", "x5c", "x5t","x5t#S256" and "crit" will be ignored if provided.  
 
 ==== Examples
 
@@ -4361,7 +4361,7 @@ Introduced in version v4.18.0.
 ==== Parameters
 
 *`signing_secret`* &lt;string&gt; The secret to use for signing the token.  
-*`headers`* &lt;(optional) unknown&gt; Optional object of JWT header fields to include in the token. Keys "alg" and "typ" will be ignored if provided.  
+*`headers`* &lt;(optional) unknown&gt; Optional object of JWT header fields to include in the token. Keys "alg", "typ", "jku", "jwk", "x5u", "x5c", "x5t","x5t#S256" and "crit" will be ignored if provided.  
 
 ==== Examples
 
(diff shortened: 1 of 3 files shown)


Nicopfy:
Thank you! I have regenerated the documentation. Let me know if it needs something more to be approved :)

mmatczuk:
### Code review

Found 5 issues:

1. **Headers type assertion runs at call-time instead of constructor-time** (godev: "fail fast at parse time")

   The `switch htype := h.(type)` block runs inside the `ObjectMethod` closure on every invocation. Since `h` is bound once at parse time via `args.Get("headers")`, the type assertion should be moved to the constructor level (before `ObjectMethod` is returned). This would validate the type once, fail fast for invalid inputs, and avoid per-invocation overhead.

   https://github.com/redpanda-data/connect/blob/372f96ae025fec7e33c6518db2213caa4f872475/internal/impl/crypto/jwt_sign.go#L59-L71

2. **Unnecessary `tc := tc` loop variable capture** (tester: "Do not use `tc := tc` in loop bodies. Go 1.22+ fixed loop variable scoping.")

   This project uses Go 1.23. The loop variable capture is unnecessary and should be removed.

   https://github.com/redpanda-data/connect/blob/372f96ae025fec7e33c6518db2213caa4f872475/internal/impl/crypto/jwt_sign_test.go#L207-L208

3. **Detached error case should use table-driven `errContains` pattern** (tester: "single table with an `errContains` field for both success and error cases")

   The non-object headers error case is a standalone block after the for loop. It should be a row in the `testCases` table with an `errContains` field.

   https://github.com/redpanda-data/connect/blob/372f96ae025fec7e33c6518db2213caa4f872475/internal/impl/crypto/jwt_sign_test.go#L255-L261

4. **`require` used for independent header validations instead of `assert`** (tester: "`assert` for independent validations - test continues to report all failures")

   The "kid" and "x" header checks at lines 240-252 are independent validations. Using `require.Equal`/`require.Fail` stops at the first failure, preventing subsequent checks from reporting. These should use `assert` so all failures are reported.

   https://github.com/redpanda-data/connect/blob/372f96ae025fec7e33c6518db2213caa4f872475/internal/impl/crypto/jwt_sign_test.go#L240-L252

5. **Commit message "feat: adding docs" is vague and uses wrong format** (commit policy: must match `system: message` or `system(subsystem): message`)

   Both commits use "feat:" prefix instead of the required format (e.g., `crypto: add optional headers param to JWT signing methods`). The second commit message "feat: adding docs" doesn't describe what was documented. First commit has typo "paramter".

   Commits: a0dd0a559fdb985acbbd334378d7ad188f93752b, 372f96ae025fec7e33c6518db2213caa4f872475

### Areas for human review

1. **JWT header denylist may be incomplete for defense-in-depth** -- only `alg` and `typ` are blocked; RFC 7515 registered headers like `jku`, `jwk`, `x5u`, `x5c`, and `crit` could be security-relevant if downstream verifiers trust them for key resolution

   https://github.com/redpanda-data/connect/blob/372f96ae025fec7e33c6518db2213caa4f872475/internal/impl/crypto/jwt_sign.go#L62-L66

2. **Commit message format uses conventional commits ("feat:") instead of project format** -- unclear if this is intentional project evolution or oversight; project format requires `system: message` (e.g., `crypto: ...`)

   Commits: a0dd0a559fdb985acbbd334378d7ad188f93752b, 372f96ae025fec7e33c6518db2213caa4f872475

Generated with Claude Code

Nicopfy:
Thanks @mmatczuk for the review ! 

- Moved the header type assertion and filtering up to the constructor to fail fast and remove per-invocation overhead.

- Added a denylist for RFC 7515 key-resolution/security headers (alg, typ, jku, jwk, x5u, x5c, x5t, x5t#S256, crit).

- Moved the detached error case into the main table using errContains and updated the independent header validations to use assert instead of require.

- Squashed the commits and updated the message to match the repo convention.

Let me know if i missed something :)

mmatczuk (approved):

mmatczuk:
### Code review

Found 2 issues:

1. Description/doc mismatch with actual blocklist (godev + bugs: "document filtered keys accurately")

The param description on line 100 and all corresponding `.adoc` entries say `Keys "alg" and "typ" will be ignored if provided`, but the code on line 61 silently drops 9 keys: `alg`, `typ`, `jku`, `jwk`, `x5u`, `x5c`, `x5t`, `x5t#S256`, `crit`. Users providing `jku`, `crit`, etc. will experience silent data loss with no documentation or error indicating their header was dropped.

https://github.com/redpanda-data/connect/blob/79e0a3ff233c2a06382c5e96ec0ca29679c17de7/internal/impl/crypto/jwt_sign.go#L60-L63
https://github.com/redpanda-data/connect/blob/79e0a3ff233c2a06382c5e96ec0ca29679c17de7/internal/impl/crypto/jwt_sign.go#L99-L101

2. Test does not verify reserved header filtering (tester: "test must prove the filtering code ran")

The `sign_rs256_headers_ignored` case passes `alg: "none"` and `typ: "bar"`, but the only filtering assertion (`assert.NotEqual(t, "none", tok.Header["alg"])` on line 262) passes trivially because the JWT library always overrides `alg` with the signing method -- not because the filter on line 61 ran. There is no assertion that `typ: "bar"` was filtered out. Seven of nine reserved keys (`jku`, `jwk`, `x5u`, `x5c`, `x5t`, `x5t#S256`, `crit`) have zero test coverage. A test that includes both reserved and non-reserved keys in the same input and asserts both filtering and passthrough would properly verify the logic.

https://github.com/redpanda-data/connect/blob/79e0a3ff233c2a06382c5e96ec0ca29679c17de7/internal/impl/crypto/jwt_sign_test.go#L200-L202
https://github.com/redpanda-data/connect/blob/79e0a3ff233c2a06382c5e96ec0ca29679c17de7/internal/impl/crypto/jwt_sign_test.go#L255-L263

### Areas for human review

1. `kid` header allowed through the blocklist -- The code blocks all RFC 7515 Section 4.1 registered JOSE headers except `kid`. In systems where the verifier uses `kid` to look up the verification key, injecting an arbitrary `kid` value could cause key confusion. However, `kid` is commonly needed for key rotation, so this may be intentional.

https://github.com/redpanda-data/connect/blob/79e0a3ff233c2a06382c5e96ec0ca29679c17de7/internal/impl/crypto/jwt_sign.go#L60-L65

Generated with Claude Code

Nicopfy:
Thanks for the review @mmatczuk :)
- Added documentation about reserved headers being dropped
- Added tests for theses headers
- Mixed test with good headers and ignored headers (with a difference if it's a default header or not)

mmatczuk:
### Code review

Found 5 issues:

1. Commit message uses gerund instead of imperative (commit policy: "Message format (enforced)")

Commit `09c9b5d3f566f84a8c60b2948cc396c56a480b62`: `crypto: adding documentation and tests about jwt reserved headers being dropped` — should use imperative mood (`add`, not `adding`). All recent repo commits use imperative form.

2. Reserved header list duplicated between filter logic and param description (Go patterns: DRY)

The same 9-element reserved header list appears as an `if` chain in the filter logic and as prose in the `.Description()` string. A change to one without the other creates a desync. Consider extracting to a shared `var`.

https://github.com/redpanda-data/connect/blob/09c9b5d3f566f84a8c60b2948cc396c56a480b62/internal/impl/crypto/jwt_sign.go#L60-L65
https://github.com/redpanda-data/connect/blob/09c9b5d3f566f84a8c60b2948cc396c56a480b62/internal/impl/crypto/jwt_sign.go#L100

3. Test assertions use fragile `strings.Contains` pattern with hardcoded expected values (tests: assertion quality)

Assertions infer expected behavior by scanning the raw JSON string `tc.headerArg` for substrings like `"kid"` or `"foo"`. Expected values (`"1234"`, `"bar"`) are hardcoded in the assertion body rather than declared in the test struct. If a future test case has `"kid"` appearing as a value rather than a key, or a URL containing `"x5t"`, the assertions would fire incorrectly. A `wantHeaders map[string]any` / `absentHeaders []string` field on the test struct would be more robust and readable.

https://github.com/redpanda-data/connect/blob/09c9b5d3f566f84a8c60b2948cc396c56a480b62/internal/impl/crypto/jwt_sign_test.go#L247-L279

4. Missing edge case tests: empty headers map `{}` and all-reserved headers (tests: edge case coverage)

No test case passes an empty object `headers: {}`, and no test case has headers where every key is reserved (producing no custom headers after filtering). `sign_rs256_good_and_all_ignored_headers` includes `kid` which passes through, so it does not exercise the empty-after-filtering path.

5. Private key constants duplicated across test functions (tests: DRY)

`dummySecretHMAC`, `dummySecretRSA`, `dummySecretECDSA256/384/512` are copy-pasted between `TestBloblangSignJwt` and `TestBloblangSignJwt_WithHeaders`. Extract to package-level vars.

https://github.com/redpanda-data/connect/blob/09c9b5d3f566f84a8c60b2948cc396c56a480b62/internal/impl/crypto/jwt_sign_test.go#L144-L181

### Areas for human review

1. Silent `typ` header filtering may be a footgun — `typ` is commonly overridden by callers (e.g., `at+JWT` for OAuth access tokens per RFC 9068). Silently dropping it could confuse users expecting it in the output token. The behavior is documented, but a warning log or allowing `typ` through may be more user-friendly. Uncertainty: intentional design decision.

https://github.com/redpanda-data/connect/blob/09c9b5d3f566f84a8c60b2948cc396c56a480b62/internal/impl/crypto/jwt_sign.go#L60-L65

2. Type switch handles only `map[string]any` — if the bloblang runtime passes a different internal representation for the `NewAnyParam` value, the switch falls to the `default` error case. Likely works since JWT `MapClaims` uses the same type, but depends on bloblang internals. Uncertainty: unfamiliar with bloblang runtime type guarantees.

https://github.com/redpanda-data/connect/blob/09c9b5d3f566f84a8c60b2948cc396c56a480b62/internal/impl/crypto/jwt_sign.go#L57-L69

3. `args.Get("headers")` error returned without context wrapping — bare `return nil, err` instead of `fmt.Errorf("getting headers: %w", err)`. Follows the pre-existing pattern for `signing_secret` in the same function. Uncertainty: pre-existing pattern, not introduced by this PR.

https://github.com/redpanda-data/connect/blob/09c9b5d3f566f84a8c60b2948cc396c56a480b62/internal/impl/crypto/jwt_sign.go#L51-L54

Generated with Claude Code

mmatczuk:
Thanks for your contribution @Nicopfy

PR #3977, opened 2026-02-10
franz: remove output wide wait

rockwotj:
We were waiting for all records to be acked from the broker
(and maybe even later ones) before acking in connect.

### DIFF
--- a/internal/impl/kafka/franz_writer.go
+++ b/internal/impl/kafka/franz_writer.go
@@ -542,7 +542,7 @@ func (w *batchWriter) writeBatch(details *FranzSharedClientInfo) error {
 		}
 
 		wg.Add(1)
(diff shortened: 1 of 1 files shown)


internal/impl/kafka/franz_writer.go
```diff
@@ -524,6 +520,8 @@ func (w *batchWriter) writeBatch(details *FranzSharedClientInfo) error {
 	if len(records) != len(w.batch) {
 		return fmt.Errorf("record count mismatch: got %d records for %d messages", len(records), len(w.batch))
 	}
+	var errs []error
+	var wg sync.WaitGroup
```
mmatczuk, line 524: Maybe group vars.

mmatczuk:
Kudos for the test!

PR #3978, opened 2026-02-12
introduce iceberg output connector

rockwotj:
- **iceberg: add dependencies**
- **iceberg: add type inference and schema errors**
- **iceberg: add JSON-to-Parquet shredder**
- **iceberg: add Parquet helpers and stats**
- **iceberg: add partition key transforms**
- **iceberg: add catalog REST client**
- **iceberg: add output plugin config and entry point**
- **iceberg: add writer, router, and committer**
- **iceberg: add integration tests and demo**
- **iceberg: add component registration**
- **iceberg: add documentation**


rockwotj:
### Code review

Found 4 issues:

1. **Map value field ID uses KeyID instead of ValueID** (bugs: copy-paste error)

   `val = parquet.FieldID(val, t.KeyID)` should be `val = parquet.FieldID(val, t.ValueID)`. The key field correctly uses `t.KeyID` on line 176, but the value field on line 181 also uses `t.KeyID` instead of `t.ValueID`. This causes incorrect Parquet field IDs for map values, leading to wrong column statistics and potential query failures.

   https://github.com/redpanda-data/connect/blob/5abe13dff9ef92d2cb9fc69149885ac2c44571a3/internal/impl/iceberg/icebergx/parquet.go#L175-L184

2. **Catalog client leaked on success path in createWriter** (bugs: resource leak)

   `catalogx.NewCatalogClient` is created at line 467 and closed on error paths (lines 476, 483, 490), but on the success path (line 498 `return w, nil`), the client is never closed. The `writer` struct does not store the client, so it becomes unreachable. Each unique table written to leaks one catalog client (HTTP connections).

   https://github.com/redpanda-data/connect/blob/5abe13dff9ef92d2cb9fc69149885ac2c44571a3/internal/impl/iceberg/router.go#L460-L499

3. **Unchecked type assertions in compareLiteral will panic on type mismatch** (bugs: panic)

   Every branch does an unchecked assertion like `b.(iceberg.BoolLiteral)`. If `a` and `b` have different concrete Literal types (e.g., during schema evolution or from corrupt partition data), this panics and crashes the connector. Use comma-ok assertions or a type switch on both values.

   https://github.com/redpanda-data/connect/blob/5abe13dff9ef92d2cb9fc69149885ac2c44571a3/internal/impl/iceberg/icebergx/compare.go#L37-L79

4. **Error wrapping uses "failed to" prefix instead of gerund form** (godev: "Don't wrap errors with 'failed to'. Do use gerund form")

   65+ instances across all implementation files use `fmt.Errorf("failed to ...: %w", err)` instead of the required gerund form `fmt.Errorf("...-ing ...: %w", err)`. Examples: `"failed to create catalog client"` -> `"creating catalog client"`, `"failed to convert messages to parquet"` -> `"converting messages to parquet"`.

   Affected files: `router.go`, `writer.go`, `committer.go`, `output_iceberg.go`, `catalogx/catalog.go`, `type_inference.go`, `icebergx/partition_key.go`

   https://github.com/redpanda-data/connect/blob/5abe13dff9ef92d2cb9fc69149885ac2c44571a3/.claude/agents/godev.md#L446-L453

Generated with Claude Code

internal/plugins/info.csv
```diff
@@ -110,6 +110,7 @@ http_client               ,input     ,http_client               ,0.0.0   ,certif
 http_client               ,output    ,http_client               ,0.0.0   ,certified  ,n          ,y     ,y
 http_server               ,input     ,http_server               ,0.0.0   ,certified  ,n          ,y     ,y
 http_server               ,output    ,http_server               ,0.0.0   ,certified  ,n          ,n     ,n
+iceberg                   ,output    ,Apache Iceberg            ,4.80.0  ,enterprise ,n          ,y     ,y
```
david-yu, line 113: Thank you!

internal/impl/iceberg/config.go
```diff
@@ -0,0 +1,298 @@
...
+
+	// Table fields
+	ioFieldNamespace = "namespace"
+	ioFieldTable     = "table"
+
+	// Storage fields - common
+	ioFieldStorage = "storage"
+
+	// S3 storage fields
+	ioFieldStorageS3            = "s3"
```
prakhargarg105: Can we align with the current aws_s3 output naming convention so that users just have to wrap their existing output?

internal/impl/iceberg/config.go
```diff
@@ -0,0 +1,298 @@
...
+	ioFieldS3Bucket             = "bucket"
+	ioFieldS3Region             = "region"
+	ioFieldS3Endpoint           = "endpoint"
+	ioFieldS3ForcePathStyleURLs = "force_path_style_urls"
+
+	// GCS storage fields
+	ioFieldStorageGCS  = "gcs"
```
prakhargarg105: Same as above. Can we align with naming of gcp_cloud_storage

internal/impl/iceberg/config.go
```diff
@@ -0,0 +1,298 @@
...
+	ioFieldGCSBucket   = "bucket"
+	ioFieldGCSEndpoint = "endpoint"
+	ioFieldGCSCredType = "credentials_type"
+	ioFieldGCSKeyPath  = "credentials_file"
+	ioFieldGCSJSONKey  = "credentials_json"
+
+	// Azure storage fields
+	ioFieldStorageAzure          = "azure"
```
prakhargarg105: align to current output of azure_blob_storage

internal/impl/iceberg/config.go
rockwotj: Done for all fields

rockwotj:
Pushed a commit for e2e tests with AWS Glue

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/redpanda-data/connect/v4
 
 go 1.25.7
 
+replace github.com/apache/iceberg-go => github.com/rockwotj/iceberg-go v0.5.0-redpanda1
```
rockwotj: we can drop this when https://github.com/apache/iceberg-go/pull/723 lands upstream BTW

mmatczuk:
### Code review

Found 4 issues:

1. **Silent data loss on non-retryable commit errors** (bugs: error handling in retry loop)

When `txn.Commit(ctx)` returns a non-nil error that is NOT `rest.ErrCommitFailed` (e.g., network timeout, auth failure, context cancellation), the code falls through to `c.table = tbl` (nil) and `commitErr = nil`, then breaks. The function returns success without the commit having succeeded. Data files are written to storage but never registered in the catalog. Additionally, the next call to `currentSchemaID()` will nil-dereference `c.table`.

The same structural issue means `MaxRetries: 0` (which config allows) silently drops all commits: the loop body never executes, `commitErr` stays nil, and the function returns success.

Fix: after the `ErrCommitFailed` check, add an explicit branch for other errors:
```go
if errors.Is(err, rest.ErrCommitFailed) {
    commitErr = err
    continue
}
if err != nil {
    return nil, fmt.Errorf("committing transaction: %w", err)
}
```

https://github.com/redpanda-data/connect/blob/224f2b8330383376c0956754acb290cf7fd76b98/internal/impl/iceberg/committer.go#L112-L121

2. **`RegisterBatchOutput` instead of `MustRegisterBatchOutput`** (godev: component registration pattern)

All 43 existing batch outputs in the codebase use `service.MustRegisterBatchOutput`. This component uses `service.RegisterBatchOutput` with a manual `if err != nil { panic(err) }`. Functionally equivalent but inconsistent with the established pattern.

https://github.com/redpanda-data/connect/blob/224f2b8330383376c0956754acb290cf7fd76b98/internal/impl/iceberg/output_iceberg.go#L29-L38

3. **Unrelated commit in feature branch** (commit policy: granularity)

Commit `954df94c` (`gcp: suppress deprecated WithCredentialsJSON lint warnings`) modifies 3 GCP component files unrelated to the iceberg feature. Should be a separate PR or landed on main independently.

4. **Mixed concerns in Glue e2e commit** (commit policy: granularity)

Commit `80713a50` (`iceberg: add Glue e2e tests and improve schema evolution`) combines Glue e2e test infrastructure with feature changes: new `schema_evolution.table_location` config field, catalog refactoring, writer deduplication logic, and dependency updates across 17 files. The test additions and the feature changes should be separate commits.

Generated with Claude Code

mmatczuk:
### Code review

Found 4 issues:

1. **Commit retry loop operates on stale table state** (bugs: optimistic concurrency)

When `txn.Commit` returns `rest.ErrCommitFailed`, the loop retries with the same `c.table` without reloading it from the catalog. Each retry creates a transaction from an outdated snapshot, so under concurrent writers the commit will fail repeatedly until `MaxRetries` is exhausted. The table must be refreshed from the catalog between retries for optimistic concurrency to work.

https://github.com/redpanda-data/connect/blob/224f2b8330383376c0956754acb290cf7fd76b98/internal/impl/iceberg/committer.go#L102-L121

2. **Bare type assertions in `goValueToLiteral` can panic on unexpected parquet stats** (bugs: panic)

All type assertions use the bare form (`val.(bool)`, `val.(int32)`, etc.) without the comma-ok idiom. The `val` type is determined by the parquet physical type via `decodeParquetValue`, while the switch cases are keyed on the iceberg logical type. If a parquet file has malformed or unexpected statistics where the physical type doesn't match the logical type, this panics. Since this code processes external data files from object storage, the comma-ok form should be used with a descriptive error.

https://github.com/redpanda-data/connect/blob/224f2b8330383376c0956754acb290cf7fd76b98/internal/impl/iceberg/icebergx/stats.go#L259-L298

3. **Orphaned parquet files in object storage on mid-batch error** (bugs: resource leak)

The `Write` loop writes each partition's parquet file to object storage immediately (line 102), then builds the `DataFile` metadata and extracts stats. Only after all partitions succeed does it commit to the catalog (line 143). If `NewDataFileBuilder` or `ExtractParquetStats` fails for partition N, files for partitions 0..N-1 are already written to storage but never registered in the catalog. These orphaned files accumulate silently with no cleanup path.

https://github.com/redpanda-data/connect/blob/224f2b8330383376c0956754acb290cf7fd76b98/internal/impl/iceberg/writer.go#L87-L148

4. **Unrelated GCP commit included in iceberg PR** (commit policy: granularity)

Commit `954df94` ("gcp: suppress deprecated WithCredentialsJSON lint warnings") modifies three GCP components (Spanner CDC, Cloud Storage, Vertex AI) that are unrelated to the iceberg connector. This should be a separate PR.

### Areas for human review

1. **`NewCatalogClient` uses `context.Background()` for initial REST catalog connection** -- The callers (router methods like `createWriter`, `evolveSchema`) all have proper contexts from the write path, but `NewCatalogClient` doesn't accept a context parameter and creates the catalog with `context.Background()`. This means the initial connection and OAuth2 token fetch cannot be cancelled during shutdown. Uncertain whether iceberg-go's `rest.NewCatalog` uses the context only for the initial handshake or stores it for subsequent calls.

https://github.com/redpanda-data/connect/blob/224f2b8330383376c0956754acb290cf7fd76b98/internal/impl/iceberg/catalogx/catalog.go#L114-L124

Generated with Claude Code

rockwotj:
Push:

* Fixed the missing context parameter from the NewCatalogClient
* Fixed the commit loop error bug
* Use MustRegister for the output

The panic type assertion thing is not valid, yes there is an invariant here but the conditional check is not needed.

The GCP commit is needed because the new dep bumped our GCP library versions which make this deprecated.

I'm also upgrading to mainline iceberg-go as the upstream PR for fast commits has landed.

rockwotj:
Ready for another review when you have time @mmatczuk

rockwotj:
Force pushed to rebase on main

rockwotj:
Force push: fix the merge conflicts

internal/impl/iceberg/config.go
```diff
@@ -0,0 +1,326 @@
...
+[%header,format=dsv]
+|===
+Go type:Iceberg type
+string:string
+[]byte:binary
+bool:boolean
+int, int32:int
+int64:long
+float32:float
+float64:double
```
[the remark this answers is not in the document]
rockwotj: Fixed

Issue #3010, opened 2024-11-15
Authenticating with a token and user/pass in NATS JetStream resources 


ghstahl:
should be part of this old [one](https://github.com/redpanda-data/connect/issues/859)

Nats [connecting by token](https://docs.nats.io/using-nats/developer/connecting/token)  
```go
// Set a token
nc, err := nats.Connect("127.0.0.1", nats.Name("API Token Example"), nats.Token("mytoken"))
if err != nil {
    log.Fatal(err)
}
defer nc.Close()

// Do something with the connection
```
[Current](https://docs.redpanda.com/redpanda-connect/components/inputs/nats_jetstream/?tab=tabs-1-advanced)
```
 auth:
      nkey_file: ./seed.nk # No default (optional)
      nkey: "" # No default (optional)
      user_credentials_file: ./user.creds # No default (optional)
      user_jwt: "" # No default (optional)
      user_nkey_seed: "" # No default (optional)
```

I am currently using a token to connect which is handled by our AuthCallout.

ghstahl:
When using auth callout in nats I am able to pass my token as the password in the username/password option.
```go
usernameHanlder := nats.UserInfo("mapped", token)

nc, err := nats.Connect(request.NATSUrl, usernameHanlder)
```

Its a hack, but I have seen microsoft azure do the same thing where they do some basic auth where the user is some static thing, and the password is the encoded token.  

I am trying to get it to work where I pass a userJWT where it really isn't a jwt but mytoken.  I think nats intercepts this and doesn't pass it to the callout.  If this works then I can use existing benthos and just overload ``` user_jwt: "" # No default (optional)```

twmb:
Closing as the reporter found a workaround using username/password to pass a token for NATS auth callout. If you need more flexible NATS auth options, feel free to open a new issue or PR.

PR #3982, opened 2026-02-12
NATS Auth: user/password or token

ghstahl:
nats: add user/password and token authentication
https://github.com/redpanda-data/connect/issues/3010

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ release_notes.md
 .op
 __pycache__
 *.test
+*.test.exe
+compile_out.txt
+test_output.txt
--- a/docs/modules/components/pages/caches/nats_kv.adoc
+++ b/docs/modules/components/pages/caches/nats_kv.adoc
@@ -68,6 +68,9 @@ nats_kv:
     user_credentials_file: ./user.creds # No default (optional)
     user_jwt: "" # No default (optional)
     user_nkey_seed: "" # No default (optional)
+    user: "" # No default (optional)
+    password: "" # No default (optional)
+    token: "" # No default (optional)
 ```
 
 --
@@ -112,6 +115,14 @@ Alternatively, the `user_jwt` field can contain a plain text JWT and the `user_n
 the plain text NKey Seed.
 
 https://docs.nats.io/using-nats/developer/connecting/creds[More details^].
+
+=== Token
+
+The `token` field can contain a plain text token string for https://docs.nats.io/running-a-nats-service/configuration/securing_nats/auth_intro/tokens[token-based authentication^].
+
+=== User and password
+
+The `user` and `password` fields can be used for https://docs.nats.io/running-a-nats-service/configuration/securing_nats/auth_intro/username_password[username/password authentication^].
 
 == Fields
 
@@ -396,6 +407,40 @@ This field contains sensitive information that usually shouldn't be added to a c
 === `auth.user_nkey_seed`
 
 An optional plain text user NKey Seed (given along with the corresponding user JWT).
+[CAUTION]
+====
+This field contains sensitive information that usually shouldn't be added to a config directly, read our xref:configuration:secrets.adoc[secrets page for more info].
+====
+
+
+
+*Type*: `string`
+
+
+=== `auth.user`
+
+An optional plain text user name (given along with the corresponding user password).
+
+
+*Type*: `string`
+
+
+=== `auth.password`
+
+An optional plain text password (given along with the corresponding user name).
+[CAUTION]
+====
+This field contains sensitive information that usually shouldn't be added to a config directly, read our xref:configuration:secrets.adoc[secrets page for more info].
+====
+
+
+
+*Type*: `string`
+
+
+=== `auth.token`
+
+An optional plain text token.
 [CAUTION]
 ====
 This field contains sensitive information that usually shouldn't be added to a config directly, read our xref:configuration:secrets.adoc[secrets page for more info].
--- a/docs/modules/components/pages/inputs/nats.adoc
+++ b/docs/modules/components/pages/inputs/nats.adoc
@@ -76,6 +76,9 @@ input:
       user_credentials_file: ./user.creds # No default (optional)
       user_jwt: "" # No default (optional)
       user_nkey_seed: "" # No default (optional)
+      user: "" # No default (optional)
+      password: "" # No default (optional)
+      token: "" # No default (optional)
     extract_tracing_map: root = @ # No default (optional)
 ```
 
@@ -133,6 +136,14 @@ Alternatively, the `user_jwt` field can contain a plain text JWT and the `user_n
 the plain text NKey Seed.
 
 https://docs.nats.io/using-nats/developer/connecting/creds[More details^].
+
+=== Token
+
+The `token` field can contain a plain text token string for https://docs.nats.io/running-a-nats-service/configuration/securing_nats/auth_intro/tokens[token-based authentication^].
+
+=== User and password
+
+The `user` and `password` fields can be used for https://docs.nats.io/running-a-nats-service/configuration/securing_nats/auth_intro/username_password[username/password authentication^].
 
 == Fields
 
@@ -472,6 +483,40 @@ This field contains sensitive information that usually shouldn't be added to a c
 === `auth.user_nkey_seed`
 
 An optional plain text user NKey Seed (given along with the corresponding user JWT).
+[CAUTION]
+====
+This field contains sensitive information that usually shouldn't be added to a config directly, read our xref:configuration:secrets.adoc[secrets page for more info].
+====
+
+
+
+*Type*: `string`
+
+
+=== `auth.user`
+
+An optional plain text user name (given along with the corresponding user password).
+
+
+*Type*: `string`
+
+
+=== `auth.password`
+
+An optional plain text password (given along with the corresponding user name).
+[CAUTION]
+====
+This field contains sensitive information that usually shouldn't be added to a config directly, read our xref:configuration:secrets.adoc[secrets page for more info].
+====
+
+
+
+*Type*: `string`
+
+
+=== `auth.token`
+
+An optional plain text token.
 [CAUTION]
 ====
 This field contains sensitive information that usually shouldn't be added to a config directly, read our xref:configuration:secrets.adoc[secrets page for more info].
--- a/docs/modules/components/pages/inputs/nats_jetstream.adoc
+++ b/docs/modules/components/pages/inputs/nats_jetstream.adoc
@@ -82,6 +82,9 @@ input:
       user_credentials_file: ./user.creds # No default (optional)
       user_jwt: "" # No default (optional)
       user_nkey_seed: "" # No default (optional)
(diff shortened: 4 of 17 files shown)


mmatczuk:
### Code review

Found 6 issues:

1. **Output test uses input constructor** (tester: "Tests must exercise the code under test")

   The "Missing password" and "Missing user" subtests in `output_jetstream_test.go` call `newJetStreamReaderFromConfig` (input constructor) instead of `newJetStreamWriterFromConfig`. Copy-paste from `input_jetstream_test.go` — the variable is also named `inputConfig`. The auth validation is shared code so this accidentally passes, but the output code path is never tested.

   https://github.com/redpanda-data/connect/blob/c0d3830bcdbc00617c35712dc3b8cb8ccf6f78f9/internal/impl/nats/output_jetstream_test.go#L91-L93
   https://github.com/redpanda-data/connect/blob/c0d3830bcdbc00617c35712dc3b8cb8ccf6f78f9/internal/impl/nats/output_jetstream_test.go#L105-L107

2. **`user` field marked as `Secret()`** (godev: "[Certification Standards — Observability](https://github.com/redpanda-data/connect/blob/c0d3830bcdbc00617c35712dc3b8cb8ccf6f78f9/.claude/agents/godev.md#L419-L426)")

   Usernames are identifiers, not secrets. Marking `user` as `Secret()` causes it to be redacted in logs and config dumps, making auth debugging harder. The docs also show a misleading "sensitive information" warning for the username. Only `password` and `token` should be `Secret()`.

   https://github.com/redpanda-data/connect/blob/c0d3830bcdbc00617c35712dc3b8cb8ccf6f78f9/internal/impl/nats/auth.go#L90-L93

3. **No mutual exclusion between auth methods** (godev: "[Certification Standards — UX validation](https://github.com/redpanda-data/connect/blob/c0d3830bcdbc00617c35712dc3b8cb8ccf6f78f9/.claude/agents/godev.md#L419-L426)")

   A user can set `token` + `user`/`password` + `nkey_file` + `user_credentials_file` + `user_jwt`/`user_nkey_seed` simultaneously. All options are appended to the NATS connection, and the server silently picks one based on internal precedence. The existing auth methods already lack mutual exclusion (pre-existing), but this PR adds two more methods without validation, expanding the space of silent misconfigurations. A check at the end of `AuthFromParsedConfig` that counts configured auth methods and errors if >1 would fix this.

   https://github.com/redpanda-data/connect/blob/c0d3830bcdbc00617c35712dc3b8cb8ccf6f78f9/internal/impl/nats/auth.go#L159-L167
   https://github.com/redpanda-data/connect/blob/c0d3830bcdbc00617c35712dc3b8cb8ccf6f78f9/internal/impl/nats/auth.go#L204-L226

4. **Error assertions don't verify error message** (tester: "[`require.ErrorContains` preferred over `require.Error`](https://github.com/redpanda-data/connect/blob/c0d3830bcdbc00617c35712dc3b8cb8ccf6f78f9/.claude/agents/tester.md#L323-L333)")

   All 6 validation error tests across 3 files use bare `require.Error(t, err)`. These pass on any error rather than confirming the specific validation fired. Should use `require.ErrorContains(t, err, "missing auth.")`.

   Files: `input_jetstream_test.go`, `input_kv_test.go`, `output_jetstream_test.go` — all "Missing password" and "Missing user" subtests.

5. **Commit message is a bare URL** (commit policy: "Message format")

   Commit `67fdd35` message is `https://github.com/redpanda-data/connect/issues/3010`. Should follow `system: message` format, e.g., `nats: add user/password and token authentication`.

6. **Commit bundles unrelated changes** (commit policy: "Granularity")

   Commit `67fdd35` combines NATS auth additions, ockam platform-specific build files, and dependency bumps in a single commit. These are three unrelated changes and should be separate commits.

Generated with Claude Code

ghstahl:
Addressed all the Code Review Issues.

mmatczuk:
### Code review

Found 3 issues:

1. Unrelated change breaks `filepath_join`/`filepath_split` documentation examples (bugs/security: "invalid JSON in code examples")

   The diff replaces forward slashes with backslashes in the `filepath_join` and `filepath_split` examples. Backslashes in JSON are escape characters: `\f` becomes a form feed, and `\"` escapes the closing quote, producing syntactically invalid JSON. This looks like an accidental editor or OS-level substitution.

   https://github.com/redpanda-data/connect/blob/139e2e5269e9ed55b661014a4eef0c4a3af9d711/docs/modules/guides/pages/bloblang/methods.adoc#L180-L183

   https://github.com/redpanda-data/connect/blob/139e2e5269e9ed55b661014a4eef0c4a3af9d711/docs/modules/guides/pages/bloblang/methods.adoc#L194-L197

2. Commit history needs cleanup (commit policy: "granularity", "message format", "message quality")

   - `67fdd35`: Commit message is a bare URL (`https://github.com/redpanda-data/connect/issues/3010`). Must follow `system: message` format. Also mixes NATS auth, ockam platform files, and dependency bumps in one commit.
   - `ed7fdc5`: Message (`Commit bundles unrelated changes (commit policy: "Granularity")`) is meta-commentary about the previous commit, not a change description. Does not follow any accepted format.
   - `469e738` overlaps with `67fdd35` — both touch NATS auth files without the first being reverted.
   - `c0d3830`, `94da9da`, `139e2e5`: Three merge commits. Branch should be rebased onto main.

   The branch needs a rebase: squash the NATS auth work into one commit with a message like `nats: add user/password and token authentication`, put the ockam change in its own commit with a proper message, and remove merge commits.

3. `authConfToOptions` uses stricter check than `AuthFromParsedConfig`, creating a silent auth bypass path (bugs/security: "inconsistent validation between parsing and option construction")

   `AuthFromParsedConfig` validates that both `user` and `password` keys are present using `p.Contains()` (presence-based), but the mutual exclusion check and `authConfToOptions` use emptiness checks (`c.User != ""`, `auth.User != "" && auth.Password != ""`). A config with `user: "alice"` and `password: ""` (e.g., from unset env var `password: "${NATS_PASSWORD}"`) passes parsing and mutual exclusion but silently gets no authentication applied — `authConfToOptions` skips the `nats.UserInfo()` call because `auth.Password != ""` is false.

   https://github.com/redpanda-data/connect/blob/139e2e5269e9ed55b661014a4eef0c4a3af9d711/internal/impl/nats/auth.go#L160-L165

### Areas for human review

1. Test coverage of legacy auth methods reduced — field-mapping assertions removed, not augmented

   The existing happy-path tests in `input_jetstream_test.go`, `input_kv_test.go`, and `output_jetstream_test.go` that asserted `NKeyFile`, `UserCredentialsFile`, `UserJWT`, and `UserNkeySeed` struct values were replaced with tests that only assert `User` and `Password`. The mutual exclusion enforcement correctly makes the old multi-method config invalid, but no new per-method happy-path test was added to verify legacy field-mapping correctness. `auth_test.go` covers parsing success for each method but does not assert struct field values.

   https://github.com/redpanda-data/connect/blob/139e2e5269e9ed55b661014a4eef0c4a3af9d711/internal/impl/nats/input_jetstream_test.go#L35-L48

Generated with Claude Code

mmatczuk:
### Code review

Found 2 issues:

1. **Field ordering splits `user_jwt`/`user_nkey_seed` logical pair** (godev: "field name constants, ConfigSpec construction")

The new fields `user`, `password`, `token` are inserted between `user_jwt` (L89) and `user_nkey_seed` (L101) in `authFieldSpec()`. These two fields must be used together and appear as a pair in both the config spec and generated documentation. Splitting them with three unrelated fields makes the config harder to understand for users.

https://github.com/redpanda-data/connect/blob/36eec05fb328347f639fdc22437c2ad5e418a514/internal/impl/nats/auth.go#L86-L104

2. **Token and user/password docs appended inside "User credentials" section** (godev: "ConfigSpec construction")

The two new `Alternatively, ...` lines at L63-65 appear inside the `=== User credentials` section of `authDescription()`, after the `user_jwt`/`user_nkey_seed` alternative. Token auth and user/password auth are distinct auth methods, not alternatives to user credentials. They should be under their own `===` subsections, following the existing structure (`=== NKey file`, `=== User credentials`).

https://github.com/redpanda-data/connect/blob/36eec05fb328347f639fdc22437c2ad5e418a514/internal/impl/nats/auth.go#L62-L67

### Areas for human review

1. **Empty `user` with non-empty `password` silently drops auth** — follows the pre-existing value-emptiness pattern across all auth fields, but more likely to occur in practice for user/password (e.g., `user: ${NATS_USER}` where env var is empty string)

If `user: ""` is explicitly set alongside `password: "secret"`, the pairing validation at L204 passes (`p.Contains` checks presence, not value), both fields are extracted, but the mutual exclusion check at L238 (`c.User != ""`) and `authConfToOptions` at L157 (`auth.User != ""`) both skip it. The password is silently discarded and the connection proceeds unauthenticated. All other auth fields have the same structural pattern (value-emptiness gating), so this may be by design.

https://github.com/redpanda-data/connect/blob/36eec05fb328347f639fdc22437c2ad5e418a514/internal/impl/nats/auth.go#L204-L219

Generated with Claude Code

mmatczuk:
This looks good! Two things:

- Please fix the linting issues — you can check with `task lint`
- It would be great if you could address the docs colocation as well

ghstahl:
A small logic change was also added.  I have a custom NATS auth service where I decide what can be passed.  
Which means that user == "", password == "something" is allowed.  

Any combination as long as user or password has something it passes validation.  Not something I would do in practice but since NATS allows it then we should support it.

nats.test.exe
mmatczuk, line 1: Test/build artifacts accidentally committed. These three files should be removed from the PR:

- `nats.test.exe` — Windows test binary (inflates repo size permanently since git stores binary blobs)
- `compile_out.txt` — empty file
- `test_output.txt` — empty file

Consider adding `*.test.exe`, `compile_out.txt`, and `test_output.txt` to `.gitignore` to prevent future accidents.

mmatczuk:
**Commits**
1. `8b09a598` ("addressed the docs colocation as well") — format violation. Must match `system: message` pattern (e.g., `nats: fix docs colocation`). Message is also vague — doesn't describe what changed. This looks like a fixup to the first commit and should be squashed into `36eec05` (`nats: add user/password and token authentication`).

**Review**
Auth implementation is clean: mutual exclusion validation is correct, `authConfToOptions` switch is consistent with the validation guarantees, `password` and `token` fields are properly marked `.Secret()`. Test coverage for the new auth methods is thorough (field mapping, mutual exclusion, missing field errors). The bugfix in `output_jetstream_test.go` (was calling `newJetStreamReaderFromConfig` instead of `newJetStreamWriterFromConfig`) is a good catch.

1. Three test/build artifacts accidentally committed — `nats.test.exe` (binary), `compile_out.txt` (empty), `test_output.txt` (empty). See [inline comment](https://github.com/redpanda-data/connect/pull/3982/files#r2127989671).

twmb:
ping @ghstahl

ghstahl:
@twmb I see 2 things failing in the build but I don't think that is part of my contributions.   Am I missing something I need to complete?

PR #3986, opened 2026-02-16
Add updates_only field to nats kv

### LINKED ISSUE #3985: Feature: Add updates_only field to nats kv
**Problem**

The `nats_kv` input redelivers all keys on restart. Which can be problematic if we're reading from a bucket with a large number of keys. 

There is an `include_history` option. But this: 
- includes historical values for a given key when set to `true` 
- includes only the latest value for a given key when set to `false`

**Suggested fix**

Expose an `updates_only` toggle option which passes `jetstream.UpdatesOnly()` to the nats client when `true` . Only problem with this is that the nats client will throw an error when both `updates_only` and `include_history` are set to  `true`: 

`nats: invalid jetstream option: updates only can not be used with include history`

matthiasdv:
This would fix #3985

### DIFF
--- a/internal/impl/nats/input_kv.go
+++ b/internal/impl/nats/input_kv.go
@@ -30,6 +30,7 @@ const (
 	kviFieldKey            = "key"
 	kviFieldIgnoreDeletes  = "ignore_deletes"
 	kviFieldIncludeHistory = "include_history"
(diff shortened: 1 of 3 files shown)


twmb:
Ping on CLA^

PR #3988, opened 2026-02-16
added aws cloudwatch logs input

squiidz:


mmatczuk:
Can we squash the commits and add "aws" commit prefix, thanks!

mmatczuk:
### Code review

Found 5 issues:

1. **Checkpoint logic causes event loss on historical replay** (godev: component correctness, bugs: logic error)

When pagination ends (`nextToken == nil`), `c.startTime` is unconditionally set to `time.Now().UnixMilli()`, overwriting the per-event checkpoint that was carefully computed from `IngestionTime` at lines 360-366. This means: (a) if the user configured `start_time` in the past to replay history, all remaining historical events are skipped after the first non-paginated response; (b) even in live mode, any events with ingestion times between the last event and `time.Now()` are permanently skipped.

https://github.com/redpanda-data/connect/blob/d9c3b1a2944fd2b91d7c805a2e0d861b7d9e0787/internal/impl/aws/input_cloudwatch_logs.go#L372-L377

2. **Missing `AutoRetryNacksToggled` wrapper** (godev: "Other AWS inputs wrap with AutoRetryNacks")

All other AWS inputs in this package (`aws_kinesis` at L155/L186, `aws_s3` at L261) wrap their registration return value with `service.AutoRetryNacksToggled` and include `service.NewAutoRetryNacksToggleField()` in their spec. Without this, nacked messages are silently dropped instead of retried. The no-op ack function at line 447 compounds this: there is no retry or dead-letter path.

https://github.com/redpanda-data/connect/blob/d9c3b1a2944fd2b91d7c805a2e0d861b7d9e0787/internal/impl/aws/input_cloudwatch_logs.go#L116-L124

3. **`Connect()` spawns additional goroutines on reconnect** (bugs: race condition)

The Benthos framework calls `Connect()` again after errors. Each call executes `c.wg.Add(1); go c.pollLoop()` with no guard against re-entry. Two concurrent `pollLoop` goroutines race on `c.startTime`, `c.nextToken`, `c.endTime` (all unsynchronized), and write duplicate messages to `c.messages`. Other inputs in the package either guard against double-connect or are designed to be idempotent.

https://github.com/redpanda-data/connect/blob/d9c3b1a2944fd2b91d7c805a2e0d861b7d9e0787/internal/impl/aws/input_cloudwatch_logs.go#L272-L275

4. **`context.Background()` in `poll()` prevents prompt shutdown** (godev: "no context.Background() in methods")

`poll()` creates a context from `context.Background()` with a 30-second timeout. When `Close()` is called while `FilterLogEvents` is in-flight, the goroutine blocks until the API call completes or times out (up to 30s). `Close()` at line 461 calls `c.wg.Wait()`, so shutdown stalls for that duration. The fix is to derive from a cancellable context tied to `shutdownCh`.

https://github.com/redpanda-data/connect/blob/d9c3b1a2944fd2b91d7c805a2e0d861b7d9e0787/internal/impl/aws/input_cloudwatch_logs.go#L306-L308

5. **No behavioral or lifecycle tests** (tester: "Flag new components without integration tests")

The test file covers only config parsing and validation (226 lines). No tests exercise `Connect`, `Read`, `Close`, `poll()`, `eventToMessage()`, or the structured vs plain output modes. The most critical code paths (polling loop, event-to-message conversion, checkpoint advancement, shutdown behavior) are completely untested. The `client` field is a concrete `*cloudwatchlogs.Client`, so unit testing the poll path requires either an interface extraction or an integration test with LocalStack.

https://github.com/redpanda-data/connect/blob/d9c3b1a2944fd2b91d7c805a2e0d861b7d9e0787/internal/impl/aws/input_cloudwatch_logs_test.go#L1-L226

Generated with Claude Code

mmatczuk:
### Code review

Found 5 issues:

1. **Connect() has no reconnection support** (godev + bugs: "Connect() called multiple times spawns duplicate goroutines; shutdown lifecycle broken after Close()")

   Each `Connect()` call does `c.wg.Add(1); go c.pollLoop()` with no guard against duplicate goroutines. If the framework retries Connect() after a transient error, multiple pollLoop goroutines run concurrently, all writing `c.startTime`, `c.nextToken`, and `c.endTime` without synchronization -- a data race. Worse: after `Close()` + reconnect, `shutdownCh` is permanently closed via `closeOnce`, so `pollLoop` exits immediately and `Read()` returns `ErrEndOfInput` forever. Compare with `input_dynamodb_cdc.go` which uses a `shutdown.Signaller` and recreates channels on each Connect.

   https://github.com/redpanda-data/connect/blob/d9c3b1a2944fd2b91d7c805a2e0d861b7d9e0787/internal/impl/aws/input_cloudwatch_logs.go#L248-L278

2. **Time advancement logic loses events** (bugs: "per-event startTime tracking overwritten by time.Now()")

   Lines 360-366 advance `c.startTime` to `lastIngestionTime + 1` as events are processed. Then lines 372-377 overwrite `c.startTime` with `time.Now().UnixMilli()` when pagination ends. This creates a gap: events ingested by CloudWatch between the API call start and `time.Now()` are skipped. The per-event tracking is effectively dead code since line 375 always overwrites it. Fix: only set `c.startTime = time.Now()` when zero events were returned; otherwise keep the per-event value.

   https://github.com/redpanda-data/connect/blob/d9c3b1a2944fd2b91d7c805a2e0d861b7d9e0787/internal/impl/aws/input_cloudwatch_logs.go#L355-L377

3. **No unit tests for core logic** (tester: "flag changed code lacking tests")

   Tests only cover config parsing and validation. The `eventToMessage()` function (structured vs. unstructured paths, nil field handling) and `poll()` logic (pagination, time advancement, error handling) have zero test coverage. These contain the most consequential logic in the component.

   https://github.com/redpanda-data/connect/blob/d9c3b1a2944fd2b91d7c805a2e0d861b7d9e0787/internal/impl/aws/input_cloudwatch_logs_test.go#L1-L226

4. **Concrete AWS client type prevents testing** (godev: "component architecture -- interface boundary for testability")

   `client *cloudwatchlogs.Client` is a concrete SDK type. Without an interface, poll/fetch logic cannot be unit tested with mocks. A small interface covering `FilterLogEvents` and `DescribeLogGroups` would make issue 3 solvable. This is the standard pattern in this repo for testable AWS components.

   https://github.com/redpanda-data/connect/blob/d9c3b1a2944fd2b91d7c805a2e0d861b7d9e0787/internal/impl/aws/input_cloudwatch_logs.go#L205-L209

5. **No integration tests for new component** (tester: "new components without integration tests")

   New input components are expected to have integration tests. Localstack supports CloudWatch Logs and the existing AWS integration test infrastructure in this package uses it.

### Areas for human review

1. **`context.Background()` in `poll()` may delay shutdown up to 30s** -- practical impact depends on AWS API latency and framework shutdown timeout. Under normal conditions the API returns quickly, but on degraded networks the 30s timeout is the worst case. A cancellable context derived from the component lifecycle would allow interrupting in-flight API calls on shutdown.

   https://github.com/redpanda-data/connect/blob/d9c3b1a2944fd2b91d7c805a2e0d861b7d9e0787/internal/impl/aws/input_cloudwatch_logs.go#L306-L307

2. **Commit message "aws cloudwatch logs input" lacks a verb** -- unclear whether this adds, fixes, or modifies. Preferred format per commit policy: `aws: add cloudwatch logs input`. The message is understandable in context but technically vague per the message quality guideline.

Generated with Claude Code

mmatczuk:
@squiidz is there a specific reason you deleted CLAUDE.md?

config/examples/aws_cloudwatch_logs.yaml
```diff
@@ -0,0 +1,90 @@
+# AWS CloudWatch Logs Source Connector (Confluent-compatible)
```
mmatczuk, line 1: Kudos for examples

docs/modules/components/pages/inputs/aws_cloudwatch_logs.adoc
```diff
@@ -0,0 +1,374 @@
...
+--
+
+```yml
+# Common config fields, showing default values
+input:
+  label: ""
+  aws_cloudwatch_logs:
+    log_group_name: my-app-logs # No default (required)
+    log_stream_names: [] # No default (optional)
+    log_stream_prefix: prod- # No default (optional)
```
mmatczuk, line 44: I suggest we use regex for new code, note confer.RegexpFilter

mmatczuk:
Could you split SNS to separate commit

mmatczuk:
We normally do not update bundles during development

internal/plugins/info.csv
```diff
@@ -11,6 +11,7 @@ awk                       ,processor ,awk                       ,0.0.0   ,commun
 aws_bedrock_chat          ,processor ,aws_bedrock_chat          ,4.34.0  ,certified  ,n          ,y     ,y
 aws_bedrock_embeddings    ,processor ,aws_bedrock_embeddings    ,4.37.0  ,certified  ,n          ,y     ,y
 aws_cloudwatch            ,metric    ,aws_cloudwatch            ,3.36.0  ,community  ,n          ,n     ,n
+aws_cloudwatch_logs       ,input     ,AWS CloudWatch Logs       ,4.43.0  ,community  ,n          ,y     ,y
```
mmatczuk: Version

CLAUDE.md
mmatczuk, line 15: Nice addition, please send in separate PR.

internal/impl/aws/output_sns.go
mmatczuk, line 48: Please send me a separate commit and document why this change is needed.

internal/impl/aws/input_cloudwatch_logs.go
```diff
@@ -0,0 +1,506 @@
...
+	cwlFieldStartTime       = "start_time"
+	cwlFieldPollInterval    = "poll_interval"
+	cwlFieldLimit           = "limit"
+	cwlFieldStructuredLog   = "structured_log"
+)
+
+func cloudWatchLogsInputSpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Stable().
+		Version("4.43.0").
```
mmatczuk: Version

internal/impl/aws/input_cloudwatch_logs.go
```diff
@@ -0,0 +1,506 @@
...
+		Fields(config.SessionFields()...).
+		LintRule(`
+root = if this.log_stream_names.or([]).length() > 0 && this.exists("log_stream_prefix") {
+  "cannot specify both log_stream_names and log_stream_prefix"
+}
+`)
+}
+
+func init() {
+	err := service.RegisterInput("aws_cloudwatch_logs", cloudWatchLogsInputSpec(),
```
mmatczuk: Use MustRegister.

internal/impl/aws/input_cloudwatch_logs.go
```diff
@@ -0,0 +1,506 @@
...
+	}
+}
+
+// cloudWatchLogsAPI defines the CloudWatch Logs API operations used by this input.
+type cloudWatchLogsAPI interface {
+	FilterLogEvents(ctx context.Context, input *cloudwatchlogs.FilterLogEventsInput, opts ...func(*cloudwatchlogs.Options)) (*cloudwatchlogs.FilterLogEventsOutput, error)
+	DescribeLogGroups(ctx context.Context, input *cloudwatchlogs.DescribeLogGroupsInput, opts ...func(*cloudwatchlogs.Options)) (*cloudwatchlogs.DescribeLogGroupsOutput, error)
+}
+
+var _ cloudWatchLogsAPI = (*cloudwatchlogs.Client)(nil)
```
mmatczuk: Do we need this?

internal/impl/aws/input_cloudwatch_logs.go
```diff
@@ -0,0 +1,506 @@
...
+		}
+		conf.FilterPattern = &pattern
+	}
+
+	if pConf.Contains(cwlFieldStartTime) {
+		var startTimeStr string
+		if startTimeStr, err = pConf.FieldString(cwlFieldStartTime); err != nil {
+			return
+		}
+		if startTimeStr == "now" {
```
mmatczuk, line 187: Worth stripping spaces before comparison

internal/impl/aws/input_cloudwatch_logs.go
```diff
@@ -0,0 +1,506 @@
...
+
+	// Validate mutual exclusion
+	if len(conf.LogStreamNames) > 0 && conf.LogStreamPrefix != nil {
+		return conf, errors.New("cannot specify both log_stream_names and log_stream_prefix")
+	}
+
+	// Validate limit range
+	if conf.Limit < 1 || conf.Limit > 10000 {
+		return conf, errors.New("limit must be between 1 and 10000")
+	}
```
mmatczuk, line 223: Worth adding lint rule

internal/impl/aws/input_cloudwatch_logs.go
```diff
@@ -0,0 +1,506 @@
...
+		log:       mgr.Logger(),
+		client:    client,
+		startTime: startTime,
+		endTime:   0,
+		messages:  make(chan *service.Message, 1000), // Buffered to prevent blocking
+	}, nil
+}
+
+func (c *cloudWatchLogsInput) Connect(ctx context.Context) error {
+	c.connMut.Lock()
```
mmatczuk: Typical name would be connMu, feel free to send a patch that renames all xxxMut to xxxMu fields in the aws package.

internal/impl/aws/input_cloudwatch_logs.go
```diff
@@ -0,0 +1,506 @@
...
+	for _, lg := range output.LogGroups {
+		if lg.LogGroupName != nil && *lg.LogGroupName == c.conf.LogGroupName {
+			found = true
+			break
+		}
+	}
+
+	if !found {
+		return fmt.Errorf("log group %q not found", c.conf.LogGroupName)
+	}
```
mmatczuk: Refactor to function, watch for long lifetimes of variables that are no longer used. Also, use short variable names for short lifetimes.

internal/impl/aws/input_cloudwatch_logs.go
```diff
@@ -0,0 +1,506 @@
...
+
+	// Create fresh shutdown channels for this connection
+	c.shutdownCh = make(chan struct{})
+	c.shutdownCtx, c.cancelFunc = context.WithCancel(context.Background())
```
mmatczuk: Maybe we should be using stopSig         *shutdown.Signaller

internal/impl/aws/input_cloudwatch_logs.go
```diff
@@ -0,0 +1,506 @@
...
+	}, nil
+}
+
+func (c *cloudWatchLogsInput) Connect(ctx context.Context) error {
+	c.connMut.Lock()
+	defer c.connMut.Unlock()
+
+	if c.connected {
+		return nil
+	}
```
mmatczuk: @Jeffail what is our story around reconnecting? Do we have any guarantees?

mmatczuk: I have seen that in serval places but I think that we should allow to reconnect in principle.

internal/impl/aws/input_cloudwatch_logs.go
```diff
@@ -0,0 +1,506 @@
...
+}
+
+func (c *cloudWatchLogsInput) pollLoop() {
+	defer c.wg.Done()
+
+	ticker := time.NewTicker(c.conf.PollInterval)
+	defer ticker.Stop()
+
+	// Poll immediately on startup
+	hasMore := c.poll()
```
mmatczuk: Could we exit too early here?

internal/impl/aws/input_cloudwatch_logs.go
```diff
@@ -0,0 +1,506 @@
...
+
+func (c *cloudWatchLogsInput) eventToMessage(event types.FilteredLogEvent) (*service.Message, error) {
+	if event.Message == nil {
+		return nil, errors.New("log event has no message")
+	}
+
+	var msg *service.Message
+
+	if c.conf.StructuredLog {
+		// Confluent-style structured output
```
mmatczuk: ?

docs/modules/components/pages/inputs/aws_cloudwatch_logs.adoc
mmatczuk: @squiidz what about the log_stream selection with regex pattern

internal/impl/aws/input_cloudwatch_logs.go
```diff
@@ -0,0 +1,514 @@
...
+		Stable().
+		Version("4.81.0").
+		Categories("Services", "AWS").
+		Summary("Consumes log events from AWS CloudWatch Logs.").
+		Description(`
+Polls CloudWatch Log Groups for log events. Supports filtering by log streams, CloudWatch filter patterns, and configurable start times.
+
+Each log event becomes a separate message with metadata including the log group name, log stream name, timestamp, and ingestion time.
+
+IMPORTANT: This input tracks its position in memory only. If the process restarts, it will resume from the configured start_time (or the beginning if not set). For exactly-once processing, you should configure an appropriate start_time or implement idempotent downstream processing.
```
mmatczuk, line 60: We should implement a cache kv store here as we have in other places.

It would be best to have default built in durable cache and avoid relying on 3rd party systems - something to consider for the future @Jeffail.

internal/impl/aws/input_cloudwatch_logs.go
```diff
@@ -0,0 +1,514 @@
...
+			i, err := newCloudWatchLogsInputFromConfig(conf, mgr)
+			if err != nil {
+				return nil, err
+			}
+			return service.AutoRetryNacksToggled(conf, i)
+		})
+}
+
+// cloudWatchLogsAPI defines the CloudWatch Logs API operations used by this input.
+type cloudWatchLogsAPI interface {
```
mmatczuk, line 137: Do we need this?

internal/impl/aws/input_cloudwatch_logs.go
mmatczuk: Ping

internal/impl/aws/input_cloudwatch_logs.go
mmatczuk: Ping

internal/impl/aws/input_cloudwatch_logs.go
```diff
@@ -0,0 +1,514 @@
...
+		case <-c.shutSig.SoftStopChan():
+			return
+		case <-ticker.C:
+			hasMore = c.poll(ctx)
+		}
+	}
+}
+
+func (c *cloudWatchLogsInput) poll(ctx context.Context) bool {
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
mmatczuk: Worth exporting that as configuration option

docs/modules/components/pages/inputs/aws_cloudwatch_logs.adoc
mmatczuk: I see, this maps to the SDK options, still curious to hear your thoughts

internal/impl/aws/input_cloudwatch_logs.go
```diff
@@ -0,0 +1,514 @@
...
+	}
+
+	// Return true if there's more data to fetch (pagination)
+	return c.nextToken != nil
+}
+
+func (c *cloudWatchLogsInput) eventToMessage(event types.FilteredLogEvent) (*service.Message, error) {
+	if event.Message == nil {
+		return nil, errors.New("log event has no message")
+	}
```
mmatczuk: It's not worth the error in the API just to handle that case.

internal/impl/aws/input_cloudwatch_logs.go
```diff
@@ -0,0 +1,514 @@
...
+
+	output, err := c.client.FilterLogEvents(ctx, input)
+	if err != nil {
+		c.log.Errorf("Polling CloudWatch Logs: %v", err)
+		return false
+	}
+
+	// Process events
+	for _, event := range output.Events {
+		msg, err := c.eventToMessage(event)
```
mmatczuk: We MUST have batching here.

internal/impl/aws/input_cloudwatch_logs_test.go
```diff
@@ -0,0 +1,622 @@
...
+	if m.describeLogGroupsIndex >= len(m.describeLogGroupsResponses) {
+		return nil, errors.New("mock: no more DescribeLogGroups responses configured")
+	}
+
+	resp := m.describeLogGroupsResponses[m.describeLogGroupsIndex]
+	m.describeLogGroupsIndex++
+	return resp.output, resp.err
+}
+
+func TestCloudWatchLogsInput_EventToMessage(t *testing.T) {
```
mmatczuk: We do not use _ as separator in test names, please add that to tester.md in separate PR.

PR #3991, opened 2026-02-17
feat(amqp_1): support for anonymous terminus in amqp_1 and interpolated message_property_to (#3990)

### LINKED ISSUE #3990: amqp_1 does not support Anonymous Terminus use case
The `amqp_1` output connector does not allow an empty `target_address` which prevents the [Anonymous Terminus](https://docs.oasis-open.org/amqp/anonterm/v1.0/anonterm-v1.0.html) pattern specified in the AMQP protocol. 

Moreover, in the Anonymous Terminus pattern, destination is set at the message level using the property `to`. The rpc plugin `amqp_1` support this field via the `message_property_to` config but this config does not support interpolation via Bloblang, preventing one of the main use-casesof this pattern (a.k.a dynamic routing of messages based on message content).

AbelThorne:
* `target_address` can be left empty to support anonymous terminus
* `message_property_to`becomes an interpolated string to support per-message dynamic routing

(should close #3990)

### DIFF
--- a/internal/impl/amqp1/integration_test.go
+++ b/internal/impl/amqp1/integration_test.go
@@ -126,6 +126,21 @@ input:
     source_address: "queue:/$ID"
 `
 
+	templateWithAnonymousTerminusBloblang := `
+output:
+  amqp_1:
+    url: amqp://guest:guest@localhost:$PORT/
+    target_address: ""
+    message_properties_to: '${! meta("target_queue").or("queue:/$ID") }'
+    max_in_flight: $MAX_IN_FLIGHT
+    metadata:
+      exclude_prefixes: [ $OUTPUT_META_EXCLUDE_PREFIX ]
+input:
+  amqp_1:
+    url: amqp://guest:guest@localhost:$PORT/
+    source_address: "queue:/$ID"
+`
+
 	testcases := []struct {
 		label    string
 		template string
@@ -145,6 +160,10 @@ input:
 		{
 			label:    "should handle Anonymous Terminus pattern",
 			template: templateWithAnonymousTerminus,
+		},
+		{
+			label:    "should handle Anonymous Terminus with Bloblang interpolation",
+			template: templateWithAnonymousTerminusBloblang,
 		},
 	}
 


mmatczuk:
### Code review

Found 4 issues:

1. **Commit message format violation** (commit policy: "must match `system: message` or `system(subsystem): message`")

   All 3 commits use conventional-commits prefixes (`feat(amqp1):`, `test(amqp1):`) which is not the format used by this project. Should be `amqp_1: ...` or `amqp_1(output): ...`. Additionally, the first commit contains a typo: "et" instead of "and".

   - `11e7b46f57e62d8e1473a34fde5f3e56b11820f3` — `feat(amqp1): enhance message destination handling with Bloblang support et anonymous terminus support`
   - `af48ff005e068c7c7e571bde4121bbc222adcf09` — `test(amqp1): add tests for Anonymous Terminus...`
   - `6c5cf01c5e19a4e07af54ea623af48e959220ddc` — `test(amqp1): add support for Anonymous Terminus in integration tests...`

2. **Missing config-time validation for anonymous terminus** (godev: "UX validation: Strong config linting with clear error messages")

   The lint rule only validates that `urls` is set. With `target_address` now defaulting to `""`, a user can omit both `target_address` and `message_properties_to`, producing a config that parses successfully but will fail at runtime (messages sent with no destination). A lint rule should warn when `target_address` is empty and `message_properties_to` is not set.

   https://github.com/redpanda-data/connect/blob/6c5cf01c5e19a4e07af54ea623af48e959220ddc/internal/impl/amqp1/output.go#L104-L108

3. **Misaligned `.Example()` indentation** (godev: "Formatting and Linting")

   The second `.Example()` call on the `messagePropsTo` field is indented at the `service.New...` level (3 tabs) instead of the method chain level (4 tabs). `gofumpt` does not enforce method chain alignment, so this won't be caught by `task fmt`.

   https://github.com/redpanda-data/connect/blob/6c5cf01c5e19a4e07af54ea623af48e959220ddc/internal/impl/amqp1/output.go#L102-L103

4. **Error wrapping does not use gerund form** (godev: "Prefix with the action in gerund form")

   `"message_properties_to interpolation: %w"` should be `"interpolating message_properties_to: %w"`.

   https://github.com/redpanda-data/connect/blob/6c5cf01c5e19a4e07af54ea623af48e959220ddc/internal/impl/amqp1/output.go#L330

### Areas for human review

1. **Anonymous terminus may not produce a spec-compliant null Target** — In the go-amqp library, `session.NewSender(ctx, "", opts)` creates a `Target` with an omitted address rather than a null Target on the attach frame. The AMQP 1.0 spec (section 2.6.12) defines anonymous terminus as a null target. Most mainstream brokers (ActiveMQ, Azure Service Bus) accept either form, but strictly compliant brokers may reject it.

   https://github.com/redpanda-data/connect/blob/6c5cf01c5e19a4e07af54ea623af48e959220ddc/internal/impl/amqp1/output.go#L256-L261

2. **Dynamic per-message interpolation is not tested end-to-end** — The integration test uses `message_properties_to: "queue:/$ID"` where `$ID` is framework-level substitution, not Bloblang interpolation. The new capability of routing different messages to different destinations based on metadata (e.g., `${! meta("target_queue") }`) is tested at the config-parsing level but not at runtime. A unit test for the `Write()` interpolation path or an integration test with actual Bloblang interpolation would strengthen coverage.

   https://github.com/redpanda-data/connect/blob/6c5cf01c5e19a4e07af54ea623af48e959220ddc/internal/impl/amqp1/integration_test.go#L113-L125

Generated with Claude Code

AbelThorne:
@mmatczuk all review comments should hopefully have been addressed

PR #3995, opened 2026-02-19
Implement buffer_acquire/buffer_release Bloblang plugins

rockwotj:
## Summary

- Implement `buffer_acquire("pool", size)` Bloblang function that returns a `[]byte` from a named `sync.Pool`, allocating a new buffer if the pool is empty
- Implement `buffer_release("pool", $buffer)` Bloblang function that returns a buffer to its named pool for reuse
- Add tests for acquire, release, unknown pool error, and pool reuse

## Test plan

- [x] `TestBufferAcquire` — acquires a buffer and verifies correct size
- [x] `TestBufferAcquireAndRelease` — full acquire→release cycle, release returns nil
- [x] `TestBufferReleaseUnknownPool` — error when releasing to non-existent pool
- [x] `TestBufferPoolReuse` — acquire, release, re-acquire from same pool


internal/impl/lang/bloblang.go
```diff
@@ -195,6 +196,72 @@ root.data = this`)
...
 		panic(err)
 	}
+
+	bufferPools := &sync.Map{}
+
+	bloblang.MustRegisterFunctionV2(
+		"buffer_acquire",
+		bloblang.NewPluginSpec().
+			Category("General").
+			Description("Acquires a byte buffer from a named pool. Buffers are reused across invocations to reduce allocations. If no buffer is available in the pool, a new one of the specified size is created. For a given pool, the size must never change.").
```
josephwoodward: thought (non-blocking): I assume we want to always ensure a buffer is released, in which case do we want to make this clearer in the the documentation, linking to the `buffer_release` documentation as well?

internal/impl/lang/bloblang.go
```diff
@@ -195,6 +196,72 @@ root.data = this`)
 	if err := registerULID(); err != nil {
 		panic(err)
 	}
+
+	bufferPools := &sync.Map{}
```
mmatczuk, line 200: nit: maybe it's just me but this syntax is really weird and reminds me of #[non_exhaustive], new is cleaner

internal/impl/lang/bloblang.go
```diff
@@ -195,6 +196,72 @@ root.data = this`)
...
+			size, err := args.GetInt64("size")
+			if err != nil {
+				return nil, err
+			}
+			v, ok := bufferPools.Load(poolName)
+			if !ok {
+				v, _ = bufferPools.LoadOrStore(poolName, &sync.Pool{New: func() any {
+					return make([]byte, size)
+				}})
+			}
```
mmatczuk, line 225: Why not always use LoadOrStore, there is a race condition here.

internal/impl/lang/bloblang.go
```diff
@@ -195,6 +196,72 @@ root.data = this`)
...
+			Description("Returns a byte buffer to a named pool for reuse. The buffer should have been previously acquired with buffer_acquire from the same pool. Returns null.").
+			Param(bloblang.NewStringParam("pool").Description("The name of the buffer pool to return the buffer to.")).
+			Param(bloblang.NewAnyParam("buffer").Description("The byte buffer to return to the pool.")).
+			Impure(),
+		func(args *bloblang.ParsedParams) (bloblang.Function, error) {
+			poolName, err := args.GetString("pool")
+			if err != nil {
+				return nil, err
+			}
+			raw, err := args.Get("buffer")
```
mmatczuk, line 246: I think Claude was kind of correct, for starers it would be safer to set length to 0

mmatczuk: Also, worth adding test for that, and we could document it. It's better to return buffers

internal/impl/lang/bloblang.go
rockwotj: To prevent an allocation in the common path

internal/impl/lang/bloblang.go
rockwotj: No you can't mutate the length like that in bloblang so making the length zero would not allow for reuse?

rockwotj:
We ended up not going down this route for the customer I'm working with so I won't have time to finish this out. Feel free to resurrect if needed in the future.

PR #4009, opened 2026-02-23
kafka: prevent data loss on cooperative rebalance in unordered reader

TayPark:
Fixes #4010

---

We've been running Vector as our Kafka-to-S3 sink connector, but kept hitting two reliability problems: S3 uploads failing when IRSA tokens expired (Vector doesn't refresh tokens gracefully — it drops the batch), and OOM crashes from the adaptive concurrency controller under bursty load. We started a PoC with Redpanda Connect to replace it.

```
  Kafka ───► Redpanda Connect (unordered_processing) ───► S3
              per-partition input batcher (50MB / 1min)
              deterministic path: topic+partition+offset.jsonl.zst
```

Our pipeline uses input batching with a deterministic S3 path so that reprocessed data overwrites the same file — at-least-once with idempotent writes:

```yaml
input:
  redpanda:
    consumer_group: my-group
    unordered_processing:
      enabled: true
      checkpoint_limit: 1
      batching:
        byte_size: 52428800  # 50 MiB per partition
        period: "1m"

output:
  aws_s3:
    path: '${!meta("kafka_topic")}+${!meta("kafka_partition")}+${!meta("kafka_offset")}.jsonl.zst'
    max_in_flight: 10
```

We tested rebalancing by triggering HPA scale-ups. Redpanda Connect generally handled it better than Vector, but during one cooperative-sticky rebalance we saw that 7 revoked partitions each lost exactly one batcher window of messages (10,433 total). We traced through the source and found the root cause.

## The bug

```
 Timeline          Consumer A                        Kafka broker
 ───────────────────────────────────────────────────────────────────

 T0  normal        sendBatch()
                     Track(r)
                     batchChan ← batch
                     pipeline processing...

 T1  rebalance     OnPartitionsRevoked:
                     CommitMarkedOffsets()    ← in-flight batch NOT marked
                     removeTopicPartitions()
                       close() → SoftStop    ← batcher discards buffered msgs
                       delete from map
                     return                  ───► partition assigned to Consumer B

 T2  pipeline      onAck() → commitFn()
     completes       MarkCommitRecords(r)    ← no ownership check in franz-go
                                                re-inserts into g.uncommitted

 T3  auto-commit   commits offset           ───► broker accepts
                                                  Consumer B starts past the gap
                                                  data LOST
```

Two things go wrong:

1. **`MarkCommitRecords` after revoke**: franz-go doesn't validate partition ownership in `MarkCommitRecords`. After `revoke()` deletes the partition from `g.uncommitted`, a late call re-inserts it, and the next auto-commit cycle commits it.

2. **Batcher silent discard**: `loop()` returns immediately on SoftStop (L277) without flushing buffered messages. One batcher window of data is discarded.

## The fix

**Block MarkCommitRecords for revoked partitions.** `checkpointTracker` gets a `revoked` set. `removeTopicPartitions` marks partitions as revoked before deleting. `commitFn` checks the set and skips `MarkCommitRecords` for revoked partitions. `OnPartitionsAssigned` clears the set on re-assignment.

**Flush batcher on SoftStop.** When the partition tracker's loop receives SoftStop, it flushes remaining buffered messages into the pipeline before returning.

```
 With fix:

 T1  OnPartitionsRevoked:
       SoftStop → batcher.Flush() → sendBatch()    ← buffered msgs delivered
       revoked[partition] = marked

 T2  onAck() → commitFn()
       isRevoked() == true → skip MarkCommitRecords ← offset NOT advanced

 T3  Consumer B starts at last committed offset
       reprocesses same range → same S3 path        ← idempotent, no loss
```

## Test plan

- [x] `TestCheckpointTracker_RevokedPartitionBlocksCommit` — revoke partition, ack in-flight batch, verify commitFn skipped
- [x] `TestCheckpointTracker_ClearRevokedOnReassign` — revoke then re-assign, verify revoked state cleared
- [x] `TestCheckpointTracker_SoftStopFlushesBatcher` — buffer messages below threshold, SoftStop, verify all flushed
- [x] Existing `TestPartitionCacheOrdering` / `TestPartitionCacheBatching` pass (no regression)

### DIFF
--- a/internal/impl/kafka/franz_reader_unordered.go
+++ b/internal/impl/kafka/franz_reader_unordered.go
@@ -187,7 +187,8 @@ type partitionTracker struct {
 	outBatchChan chan<- batchWithAckFn
 	commitFn     func(r *kgo.Record)
 
-	shutSig *shutdown.Signaller
+	closeCtxCh chan context.Context
+	shutSig    *shutdown.Signaller
 }
 
 func newPartitionTracker(batcher *service.Batcher, batchChan chan<- batchWithAckFn, commitFn func(r *kgo.Record)) *partitionTracker {
@@ -196,6 +197,7 @@ func newPartitionTracker(batcher *service.Batcher, batchChan chan<- batchWithAck
 		checkpointer: checkpoint.NewUncapped[*kgo.Record](),
 		outBatchChan: batchChan,
 		commitFn:     commitFn,
+		closeCtxCh:   make(chan context.Context, 1),
 		shutSig:      shutdown.NewSignaller(),
 	}
 	go pt.loop()
@@ -275,6 +277,23 @@ func (p *partitionTracker) loop() {
 				}
 			}
 		case <-p.shutSig.SoftStopChan():
+			if p.batcher != nil {
+				closeCtx := context.Background()
+				select {
+				case ctx := <-p.closeCtxCh:
+					closeCtx = ctx
+				default:
+				}
+				p.batcherLock.Lock()
+				if batch, _ := p.batcher.Flush(closeCtx); len(batch) > 0 {
+					record := p.topBatchRecord
+					p.topBatchRecord = nil
+					p.batcherLock.Unlock()
+					_ = p.sendBatch(closeCtx, batch, record)
+				} else {
+					p.batcherLock.Unlock()
+				}
+			}
 			return
 		}
 	}
@@ -347,6 +366,10 @@ func (p *partitionTracker) pauseFetch(limit int) (pauseFetch bool) {
 }
 
 func (p *partitionTracker) close(ctx context.Context) error {
+	select {
+	case p.closeCtxCh <- ctx:
+	default:
+	}
 	p.shutSig.TriggerSoftStop()
 	select {
 	case <-ctx.Done():
@@ -359,8 +382,9 @@ func (p *partitionTracker) close(ctx context.Context) error {
 //------------------------------------------------------------------------------
 
 type checkpointTracker struct {
-	mut    sync.Mutex
-	topics map[string]map[int32]*partitionTracker
+	mut     sync.Mutex
+	topics  map[string]map[int32]*partitionTracker
+	revoked map[string]map[int32]struct{}
 
 	res       *service.Resources
 	batchChan chan<- batchWithAckFn
@@ -376,6 +400,7 @@ func newCheckpointTracker(
 ) *checkpointTracker {
 	return &checkpointTracker{
 		topics:    map[string]map[int32]*partitionTracker{},
+		revoked:   map[string]map[int32]struct{}{},
 		res:       res,
 		batchChan: batchChan,
 		commitFn:  releaseFn,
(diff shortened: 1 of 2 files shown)


mmatczuk:
### Code review

Found 2 issues:

1. **SoftStop flush uses `context.Background()` — can block indefinitely** (bugs: bounded blocking)

   The flush added in the SoftStop handler calls both `p.batcher.Flush(context.Background())` and `p.sendBatch(context.Background(), ...)`. `sendBatch` pushes to the unbuffered `outBatchChan`. With `context.Background()`, this has no deadline and blocks forever if nothing reads from the channel.

   Since `close()` waits on `HasStoppedChan()` (which fires when `loop()` returns), and `loop()` can't return while `sendBatch` is blocked, this can stall the `OnPartitionsRevoked` callback and the entire rebalance. The `close(ctx)` escape hatch lets the caller eventually proceed if its context expires, but the goroutine leaks.

   Consider threading the `ctx` from `close()` into the SoftStop handler (e.g., via a stored context or channel), or using a derived context with a timeout.

   https://github.com/redpanda-data/connect/blob/df8a197909ac321fddec195ce666fbe6e8fae998/internal/impl/kafka/franz_reader_unordered.go#L278-L288

2. **Commit message format** (commit policy: "`system: message`" or "`system(subsystem): message`")

   The commit `fix(kafka): prevent data loss on cooperative rebalance in unordered reader` uses conventional-commits style (`type(scope): message`). The repo convention is `system: message` or `system(subsystem): message`. Expected: `kafka: prevent data loss on cooperative rebalance in unordered reader` or `kafka(unordered-reader): prevent data loss on cooperative rebalance`.

### Areas for human review

1. **`commitFn` field mutated after construction without synchronization** — The field is written on L548 after `newCheckpointTracker` returns on L546, then read later from ack callbacks on arbitrary goroutines. The write happens before any goroutine that reads it starts, so it's benign in practice, but the Go memory model doesn't strictly guarantee visibility without explicit synchronization. The race detector may flag it.

   https://github.com/redpanda-data/connect/blob/df8a197909ac321fddec195ce666fbe6e8fae998/internal/impl/kafka/franz_reader_unordered.go#L546-L557

2. **`revoked` map grows monotonically for partitions reassigned to other consumers** — `clearRevoked` is called from `OnPartitionsAssigned` and only clears partitions assigned to this consumer. In cooperative-sticky rebalance, revoked partitions typically go to other consumers, so those entries are never cleared. Each entry is small (string key + int32 + empty struct), so memory impact is negligible, but the map never shrinks over the lifetime of the consumer.

   https://github.com/redpanda-data/connect/blob/df8a197909ac321fddec195ce666fbe6e8fae998/internal/impl/kafka/franz_reader_unordered.go#L466-L469

Generated with Claude Code

TayPark:
Hi @mmatczuk, thank you for the thorough review.

I've addressed both issues in the latest commit:

1. **Bounded context for SoftStop flush**: `partitionTracker` now receives the close context from `close(ctx)` and uses it in the SoftStop handler instead of `context.Background()`. The happens-before guarantee is provided by the channel close in `TriggerSoftStop()`.
2. **Lint fix**: Renamed unused parameter `r` to `_` in test callbacks.

Could you approve the CI workflow run when you get a chance? Looking forward to your feedback.

TayPark:
Rebased and squashed into a single commit. Fixed golangci-lint issues (containedctx + rangeint/modernize). All tests pass locally. Could you approve the CI workflow run as well? Thanks!

Jeffail:
Thank you, @TayPark, for the thorough investigation and write-up here — the timeline and reproduction notes made the underlying behaviour easy to follow.

We have opened #4477, which supersedes this PR. It addresses the same commit-after-revoke behaviour, with a few differences worth noting:

- It also fixes the **ordered** reader (`partitionCache`), which is the default code path for the `redpanda` input and carries the identical commit-after-revoke shape. This PR only covered the unordered reader.
- It attaches the guard to the per-partition tracker object rather than a shared revoked-set, which removes the need for the `clearRevoked` / `OnPartitionsAssigned` reset and is immune to the same-member reassignment case.
- It omits the SoftStop batcher-flush change, since buffered records are never committed and are re-read by the new owner after a revoke.

One point of genuine uncertainty we would like to flag respectfully: we have not yet been able to confirm a mechanism by which this behaviour actually **loses** data, as opposed to producing spurious commits for unowned partitions. We have expanded on the reasoning in #4010 and would very much welcome your input there, as your production observations are the strongest evidence available.

With your agreement, we will close this PR in favour of #4477. Thank you again for surfacing the issue.

TayPark:
Thank you @Jeffail.

PR #4028, opened 2026-02-26
feat(kafka): expose more kafka_franz parameters

dyurchanka:
Hi, I would like to introduce several changes to kafka_franz output (shared with redpanda output) to expose more go client parameters into redpanda-connect. My main use case is low throughput of kafka output (I tried all, sarama, franz, redpanda). I stopped on kafka_franz for now (I know it is deprecated, but actually redpanda uses franz under the hood) and main issue is default limitation of 10k buffer in kafka_franz which results in limited throughput over latency networks. As well as  minor concurrency bug in redpanda output. Also default values for all new params are same as before (equal to kafka_franz defaults) 

## 1: Add configurable `acks` field to redpanda/kafka_franz output

**File**: `internal/impl/kafka/franz_writer.go`

**Problem**: The `redpanda` and `kafka_franz` outputs hardcode `acks=all` (all ISR replicas) with no way to configure it. When `idempotent_write: false`, users may want `acks=leader` (acks=1) for higher throughput at the cost of durability, or `acks=none` (acks=0) for fire-and-forget scenarios. Previously, the franz-go default of `acks=all` was always used regardless of the `idempotent_write` setting.

**Fix**: Add a new `acks` field to `FranzProducerFields()` with three options:

| Value    | Kafka equivalent | Behavior |
|----------|-----------------|----------|
| `all`    | acks=-1         | Wait for all in-sync replicas (default, required for idempotent writes) |
| `leader` | acks=1          | Wait for leader only (available when `idempotent_write: false`) |
| `none`   | acks=0          | Fire-and-forget (available when `idempotent_write: false`) |

**Validation**: If `idempotent_write: true` and `acks` is set to anything other than `all`, the config fails with an error at both lint time and runtime. This matches the Kafka protocol requirement that idempotent producers must use `acks=all`.

**Config example**:
```yaml
output:
  redpanda:
    seed_brokers: ["localhost:9092"]
    topic: my-topic
    idempotent_write: false
    acks: leader  # only valid when idempotent_write is false
```

**Changes**:
1. Added `kfwFieldAcks` constant and `StringAnnotatedEnumField` with `all`, `leader`, `none` options (default: `all`)
2. Added validation in `FranzProducerOptsFromConfig`: error if `idempotent_write: true && acks != "all"`
3. Maps config values to franz-go options: `kgo.AllISRAcks()`, `kgo.LeaderAck()`, `kgo.NoAck()`
4. Added lint rule: `this.idempotent_write == true && this.acks.or("all") != "all"` for early config validation

**Backward compatible**: Default is `all`, matching the previous implicit behavior.

---

## 2: Expose producer buffer and in-flight tuning fields

**File**: `internal/impl/kafka/franz_writer.go`

**Problem**: The franz-go producer has several critical tuning parameters hardcoded to defaults and not exposed in the `redpanda`/`kafka_franz` output config:

| Parameter | franz-go default | Issue |
|-----------|-----------------|-------|
| `MaxBufferedRecords` | 10,000 | `Produce()` blocks when hit. For high-throughput pipelines, records arrive faster than they drain, causing the caller goroutine to stall. |
| `MaxBufferedBytes` | 0 (unlimited) | No byte-level memory guard. 10k large records could consume gigabytes. |
| `MaxProduceRequestsInflightPerBroker` | 1 | Only one produce request per broker TCP connection at a time. The pipeline idles during the full broker round-trip, wasting the ability to pipeline requests. |
| `RecordRetries` | unlimited | On persistent broker failure (i/o timeout), franz-go retries every record forever. `WriteBatch` hangs because callbacks never fire, the process becomes a zombie — alive but non-functional. |
| `RecordDeliveryTimeout` | unlimited | Same zombie problem as above but time-based. Equivalent to Kafka's `delivery.timeout.ms`. |

**Fix**: Add five new fields to `FranzProducerLimitsFields()`:

```yaml
output:
  redpanda:
    # New fields (shown with defaults):
    max_buffered_records: 10000       # max records buffered before Produce() blocks
    max_buffered_bytes: "0"           # max bytes buffered (0 = unlimited, only record limit applies)
    max_in_flight_requests: 1         # max produce requests per broker TCP connection
    record_retries: 0                 # max retries per record (0 = unlimited)
    record_delivery_timeout: "0s"     # max time a record can sit in buffer (0s = unlimited)
```

**Validation**:
- `max_buffered_records` must be >= 1
- `max_buffered_bytes` accepts human-readable sizes ("256MB", "50mib"), 0 disables
- `max_in_flight_requests` must be >= 1; with `idempotent_write: true`, franz-go internally caps at 5 (Kafka v1+) or 1 (Kafka < v1)
- `record_retries` 0 = unlimited (default), > 0 = fail after N retries
- `record_delivery_timeout` 0s = unlimited (default), > 0 = fail after duration
- With `idempotent_write: true`, both retry/timeout options are only enforced when safe (no invalid sequence numbers)

Thanks

### DIFF
--- a/internal/impl/kafka/franz_writer.go
+++ b/internal/impl/kafka/franz_writer.go
@@ -37,11 +37,17 @@ const (
 	// Producer fields
 	kfwFieldPartitioner            = "partitioner"
 	kfwFieldIdempotentWrite        = "idempotent_write"
+	kfwFieldAcks                   = "acks"
 	kfwFieldCompression            = "compression"
 	kfwFieldAllowAutoTopicCreation = "allow_auto_topic_creation"
-	kfwFieldTimeout                = "timeout"
-	kfwFieldMaxMessageBytes        = "max_message_bytes"
-	kfwFieldBrokerWriteMaxBytes    = "broker_write_max_bytes"
+	kfwFieldTimeout                     = "timeout"
+	kfwFieldMaxMessageBytes             = "max_message_bytes"
+	kfwFieldBrokerWriteMaxBytes         = "broker_write_max_bytes"
+	kfwFieldMaxBufferedRecords          = "max_buffered_records"
+	kfwFieldMaxBufferedBytes            = "max_buffered_bytes"
+	kfwFieldMaxInFlightRequestsPerBrkr  = "max_in_flight_requests"
+	kfwFieldRecordRetries               = "record_retries"
+	kfwFieldRecordDeliveryTimeout       = "record_delivery_timeout"
 )
 
 // FranzProducerLimitsFields returns a slice of fields specifically for
@@ -68,6 +74,42 @@ func FranzProducerLimitsFields() []*service.ConfigField {
 			Default("100MiB").
 			Example("128MB").
 			Example("50mib"),
+		service.NewIntField(kfwFieldMaxBufferedRecords).
+			Description("The maximum number of records the client will buffer in memory before blocking. " +
+				"When this limit is reached, `Produce()` calls will block until buffered records are delivered and space frees up. " +
+				"Increase this value for high-throughput pipelines to avoid back-pressure stalls.").
+			Advanced().
+			Default(10000),
+		service.NewStringField(kfwFieldMaxBufferedBytes).
+			Description("The maximum number of bytes the client will buffer in memory before blocking. " +
+				"When this limit is reached, `Produce()` calls will block until buffered records are delivered. " +
+				"Set to `0` to disable the byte-level limit (only `max_buffered_records` applies). " +
+				"This limit is checked after `max_buffered_records`.").
+			Advanced().
+			Default("0").
+			Example("256MB").
+			Example("50mib"),
+		service.NewIntField(kfwFieldMaxInFlightRequestsPerBrkr).
+			Description("The maximum number of produce requests in flight per broker connection. " +
+				"When `idempotent_write` is enabled, this is capped at 5 by the Kafka protocol (and at 1 for Kafka < v1.0.0). " +
+				"When `idempotent_write` is disabled, higher values improve throughput by pipelining requests but may cause out-of-order delivery.").
+			Advanced().
+			Default(1),
+		service.NewIntField(kfwFieldRecordRetries).
+			Description("The maximum number of times a record produce is retried on failure before the record is failed. " +
+				"When a record fails, all records buffered in the same partition are also failed to preserve gapless ordering. " +
+				"Set to `0` for unlimited retries (the default). " +
+				"With `idempotent_write` enabled, retries are only enforced when safe to do so without creating invalid sequence numbers.").
+			Advanced().
+			Default(0),
+		service.NewDurationField(kfwFieldRecordDeliveryTimeout).
+			Description("The maximum time a record can sit in the producer buffer before it is failed, roughly equivalent to Kafka's `delivery.timeout.ms`. " +
+				"This is evaluated before writing a request or after a produce response. " +
+				"When a record times out, all records in the same partition are also failed. " +
+				"Set to `0s` for no timeout (the default). " +
+				"With `idempotent_write` enabled, timeouts are only enforced when safe to do so without creating invalid sequence numbers.").
+			Advanced().
+			Default("0s"),
 	}
 }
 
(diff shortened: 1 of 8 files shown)


internal/impl/kafka/franz_writer.go
```diff
@@ -144,6 +195,52 @@ func FranzProducerLimitsOptsFromConfig(conf *service.ParsedConfig) ([]kgo.Opt, e
...
+	maxBufferedBytesStr, err := conf.FieldString(kfwFieldMaxBufferedBytes)
+	if err != nil {
+		return nil, err
+	}
+	maxBufferedBytes, err := humanize.ParseBytes(maxBufferedBytesStr)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse max_buffered_bytes: %w", err)
+	}
+	if maxBufferedBytes > 0 {
+		opts = append(opts, kgo.MaxBufferedBytes(int(maxBufferedBytes)))
```
mmatczuk, line 220: **Bug: Missing overflow check on `max_buffered_bytes` cast from `uint64` to `int`**

`humanize.ParseBytes` returns `uint64`, but `kgo.MaxBufferedBytes` takes `int`. The cast `int(maxBufferedBytes)` silently overflows for large values (>2GB on 32-bit, >2^63 on 64-bit).

The existing `max_message_bytes` and `broker_write_max_bytes` fields both validate against an upper bound before casting (e.g. `if maxMessageBytes > uint64(math.MaxInt32)`). Add an equivalent check here before the cast.

mmatczuk:
**Commits**
1. `feat(kafka): expose more kafka_franz parameters` — format violation: `feat(kafka):` uses a conventional commit prefix instead of a system name. Should be `kafka:` or `kafka(franz):` per commit policy (system name first, not `feat`/`fix`/etc. prefixes).
2. Granularity: the commit mixes two independent changes — new config fields (`acks`, `max_buffered_records`, `max_buffered_bytes`, `max_in_flight_requests`, `record_retries`, `record_delivery_timeout`) and a `sync.Mutex` → `sync.RWMutex` refactor with double-checked locking in `output_redpanda.go`. These should be separate commits.

**Review**
One bug found: missing overflow guard on `max_buffered_bytes`.

1. [Missing overflow check on `max_buffered_bytes` uint64→int cast](https://github.com/redpanda-data/connect/blob/3e8e30e920a8fe42af297f2b87de01412eaf71fe/internal/impl/kafka/franz_writer.go#L215-L217) — `humanize.ParseBytes` returns `uint64`, cast to `int` without upper-bound validation. Existing fields `max_message_bytes` and `broker_write_max_bytes` both validate before casting. Add an equivalent check.

internal/impl/kafka/franz_writer.go
dyurchanka: Thanks for a quick review. Must be fixed now

dyurchanka:
replaced commit with suggested semantic. Removed redpanda plugin changes out of this PR and edited PR description

dyurchanka:
ah, some other changed were pushed, give me a sec, will fix

dyurchanka:
Must be OK now for review) Sorry

internal/impl/kafka/output_kafka_franz_test.go
```diff
@@ -67,6 +67,72 @@ kafka_franz:
...
+  seed_brokers: [ foo:1234 ]
+  topic: foo
+  idempotent_write: false
+  max_buffered_records: 50000
+  max_buffered_bytes: "128MB"
+  max_in_flight_requests: 5
+  record_retries: 10
+  record_delivery_timeout: "30s"
+`,
+		},
```
mmatczuk, line 135: The new test cases only exercise config lint rules via `AddOutputYAML`. The runtime validation logic added in `FranzProducerLimitsOptsFromConfig` — specifically `maxBufferedRecords < 1`, `maxInFlightRequests < 1`, and `maxBufferedBytes > math.MaxInt` — has no test coverage.

Consider adding test cases that call `FranzProducerLimitsOptsFromConfig` directly (or through output instantiation) to verify these validation paths reject invalid values like `max_buffered_records: 0`, `max_in_flight_requests: 0`, etc.

mmatczuk:
**Commits**
1. `2c8c7e18a163` (`kafka: fix linting errors and regenerate docs`) mixes unrelated work (linting fix in `franz_writer.go` + doc regeneration across 6 files) and is effectively a fixup of `f977a0b0` since the linting errors were introduced by that commit. Should be squashed into commit 1, with the result being a single commit containing the implementation, tests, and regenerated docs.

**Review**
Clean implementation. Constants, enum field type, validation, and kgo option mapping are all correct. No bugs or security issues found.

1. [Missing test coverage for runtime validation](https://github.com/redpanda-data/connect/blob/2c8c7e18a163301bd1d1c618e6d39f46b3819e63/internal/impl/kafka/output_kafka_franz_test.go#L122-L135): The new runtime validation in `FranzProducerLimitsOptsFromConfig` (`max_buffered_records >= 1`, `max_in_flight_requests >= 1`, `max_buffered_bytes` overflow) is not exercised by the existing lint-level tests.

mmatczuk (approved):
LGTM

dyurchanka:
> LGTM

So I confused. It is approved, but do I need to push changes based your last comment?

mmatczuk:
@dyurchanka all good from my end, I'd like to get others to review that before merging

Jeffail (approved):

twmb:
Please sign CLA and we will merge.

dyurchanka:
It was signed already, I think there is an issue with CLA status. I just re-checked, https://cla-assistant.io/redpanda-data/connect?pullRequest=4028 shows me as signed

twmb:
Not sure what's up with the CLA status check, the prior message also is clear indication of signing imo, merging, thank you!

PR #4112, opened 2026-03-17
output/aws_dynamodb: chunk batch writes to 25 item limit

### LINKED ISSUE #992: DynamoDB writer should limit write batches to 25 items
DynamoDB batch writes are limited to 25 items, as per the [documentation here](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html). The benthos dynamo writer currently falls back to one-by-one writes in the case of any errors (such as exceeding this 25 item limit), which has a huge impact on throughput in the case of large output batch sizes.

It would be nice if the dynamo writer would internally limit each batch call to 25 items, making this invisible to the user and avoiding the need to restrict output batch settings to 25 or fewer in benthos config.

twmb:
## Summary

- Chunks `BatchWriteItem` calls into groups of 25 items per the [DynamoDB API limit](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html), instead of sending all items at once and falling back to expensive one-by-one writes on failure
- Resets backoff between chunks so each chunk gets a fresh retry budget
- Marks items in unattempted chunks as failed in `BatchError` when a chunk fails, preventing silent data loss

## Test plan

- [x] Existing unit tests pass (`go test ./internal/impl/aws/dynamodb/...`)
- [ ] Manual test with batch sizes > 25 against a real DynamoDB table
- [ ] Integration tests (require Docker)

Closes #992

### DIFF
--- a/internal/impl/aws/dynamodb/output.go
+++ b/internal/impl/aws/dynamodb/output.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"maps"
+	"reflect"
 	"strconv"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


mmatczuk:
Please add integration test

squiidz (approved):

PR #4204, opened 2026-04-01
azure: add blob index tags support for blob storage output

### LINKED ISSUE #2904: Add Metadata and Tags support for `azure_blob_storage`
Azure Blob Storage supports setting [Metadata](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-properties-metadata) and [Tags](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-manage-find-blobs) for each blob. However, the current implementation of `azure_blob_storage` does not support these features. It would be nice to have this functionality added to provide more flexibility when working with blobs.

Go SDK: [UploadStreamOptions](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob@v1.4.1/blockblob#UploadStreamOptions)

fredmeltwater:
## Summary 
  - Adds a new `tags` field to the `azure_blob_storage` output, allowing users to set blob index tags as key/value 
  pairs with interpolation support                                                                                   
  - Tags are applied atomically on block blob uploads and via `SetTags` for append blobs
  - Enforces Azure's 10-tag limit at config parse time
  - `review` skill used on this PR                                                              
                  
  Closes #2904                                                                                                       
                                                                                                                     
  ## Test plan
  - [x] `go build ./internal/impl/azure/...` passes
  - [x] `go test ./internal/impl/azure/... -short` passes
  - [x] Integration test `blob_storage_tags` validates tags are set correctly on block blobs via Azurite             
   
### DIFF
--- a/internal/impl/azure/output_blob_storage.go
+++ b/internal/impl/azure/output_blob_storage.go
@@ -19,10 +19,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sort"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/streaming"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/appendblob"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blockblob"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/bloberror"
 
 	"github.com/redpanda-data/benthos/v4/public/service"
@@ -32,14 +35,21 @@ const (
 	// Blob Storage Output Fields
 	bsoFieldContainer         = "container"
 	bsoFieldPath              = "path"
+	bsoFieldTags              = "tags"
 	bsoFieldBlobType          = "blob_type"
 	bsoFieldPublicAccessLevel = "public_access_level"
 )
+
+type bsoTagPair struct {
+	key   string
+	value *service.InterpolatedString
+}
 
 type bsoConfig struct {
 	client            *azblob.Client
 	Container         *service.InterpolatedString
 	Path              *service.InterpolatedString
+	Tags              []bsoTagPair
 	BlobType          *service.InterpolatedString
 	PublicAccessLevel *service.InterpolatedString
 }
@@ -59,6 +69,23 @@ func bsoConfigFromParsed(pConf *service.ParsedConfig) (conf bsoConfig, err error
 	if conf.Path, err = pConf.FieldInterpolatedString(bsoFieldPath); err != nil {
 		return
 	}
+
+	var tagMap map[string]*service.InterpolatedString
+	if tagMap, err = pConf.FieldInterpolatedStringMap(bsoFieldTags); err != nil {
+		return
+	}
+	if len(tagMap) > 10 {
+		err = fmt.Errorf("at most 10 blob index tags are permitted, got %d", len(tagMap))
+		return
+	}
+	conf.Tags = make([]bsoTagPair, 0, len(tagMap))
+	for k, v := range tagMap {
+		conf.Tags = append(conf.Tags, bsoTagPair{key: k, value: v})
+	}
+	sort.Slice(conf.Tags, func(i, j int) bool {
+		return conf.Tags[i].key < conf.Tags[j].key
+	})
+
 	if conf.BlobType, err = pConf.FieldInterpolatedString(bsoFieldBlobType); err != nil {
 		return
 	}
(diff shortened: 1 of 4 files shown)


meltwater-coreplatform (commented):
**Commits**

LGTM

**Review**

Clean implementation adding blob index tags support to the `azure_blob_storage` output. Tags are parsed at config time as an `InterpolatedStringMap` with a 10-tag limit (matching Azure's constraint), deterministically sorted, and resolved per-message. Block blobs pass tags via `UploadStreamOptions`; append blobs set tags at creation via `CreateOptions` and update them on subsequent appends via `SetTags`. Integration test validates tag roundtrip through Azurite.

LGTM

fredmeltwater:
Hi @Jeffail this has been open for a few weeks — would love a review when you get a       
  chance

josephwoodward:
Hi @fredmeltwater, are you able to run `task docs` on this to generate the new docs?

fredmeltwater:
Hi @josephwoodward were you able to take a look?

VictorErmakov:
Just a bump for this PR. We are really looking into it

fredmeltwater:
Just a bump to ask if anyone has had the chance to review? @VictorErmakov

internal/impl/azure/output_blob_storage.go
```diff
@@ -59,6 +69,23 @@ func bsoConfigFromParsed(pConf *service.ParsedConfig) (conf bsoConfig, err error
 	if conf.Path, err = pConf.FieldInterpolatedString(bsoFieldPath); err != nil {
 		return
 	}
+
+	var tagMap map[string]*service.InterpolatedString
+	if tagMap, err = pConf.FieldInterpolatedStringMap(bsoFieldTags); err != nil {
+		return
+	}
+	if len(tagMap) > 10 {
```
josephwoodward: minor (non-blocking): Given this 10 tag limit is set by Azure, it might be worth moving it to the `const` declaration block at line 31, naming it something like `MaxTagsPermitted` with a comment saying it's set by Azure and a ref to the docs. It can then be used anywhere else the limit is mentioned such as docs etc.

fredmeltwater: Done in dbb62a418 — added bsoMaxTagsPermitted = 10 to the const block with a comment noting it's an Azure-imposed limit plus a docs link, and referenced it from both the validation error and the field description (via fmt.Sprintf), so the number is stated in one place. Went with the bso prefix rather than an exported MaxTagsPermitted to match the file's existing naming. Rendered docs are unchanged.

josephwoodward (approved):

internal/impl/azure/output_blob_storage.go
```diff
@@ -167,8 +211,22 @@ func (a *azureBlobStorageWriter) uploadBlob(ctx context.Context, containerName,
...
 			}
 		}
+		// Tags must be set separately for append blobs since AppendBlock does not
+		// support tags. On creation they are set via CreateOptions, but when
+		// appending to an existing blob we need an explicit SetTags call to ensure
+		// tags reflect the latest configured values.
+		if len(tags) > 0 {
+			blobClient := containerClient.NewBlobClient(blobName)
+			if _, err = blobClient.SetTags(ctx, tags, nil); err != nil {
+				return fmt.Errorf("setting blob tags: %w", err)
```
josephwoodward: I think there's risk of data duplication here? Assuming the `appendBlobClient.AppendBlock` call above succeeds, if `blobClient.SetTags` fails then the event will get redelivered and `appendBlobClient.AppendBlock` could try to re-appends the same message content into the blob. I assume `AppendBlock` isn't idempotent?

fredmeltwater: Confirmed, and thanks for the precision on the mechanism — both AppendBlock call sites pass nil options so the AppendPosition precondition is unused, which means a SetTags failure after a committed append nacks a message that then gets appended twice on redelivery. Same applies after the create-then-retry path. Block blobs are unaffected since tags ride in UploadStreamOptions atomically.

Fixed in `d6ddfefdb` by writing the tags after a successful append and logging failures rather than returning them. The content is already committed by that point, so losing a tag update is the better trade.

Worth recording why it isn't creation-only, which was my first instinct: applying tags solely via CreateOptions does close the duplication window, but it makes the feature a silent no-op for any blob this output didn't create — which is the steady state for an append blob written across restarts. 

Verified on Azurite: a pre-created untagged blob stays untagged through repeated tagged writes, all acked. There's now an integration subtest (blob_storage_tags_append_preexisting_blob) guarding exactly that; it fails with the fix reverted.

Also in the same commit, from digging around the same code:

- Tag keys are validated at config time now. They're static map keys, so an empty, over-long or invalid-charset key is detectable at startup rather than 400-ing every message forever. Values stay Azure's to reject since they're interpolated.

- Tag prose moved into a == Tags section following aws_s3, leaving the field description to the hard limits. Documents interpolation being values-only, the SAS t permission requirement, and the best-effort tag writes.

- Unit tests for the config-time validation. Note this PR's integration-test and integration-test-filter checks both report SKIPPED, so integration coverage doesn't gate the change — hence covering the validation with tests that need no Docker.

Two things I've deliberately kept out:

- A pre-existing bug. appendBlobClient.Create is called without an If-None-Match precondition, so it can't return BlobAlreadyExists and that guard is unreachable. On a concurrent path it recreates and truncates an existing append blob. Reproduced on Azurite — emulator-vs-real-Azure fidelity is the open question, not whether Create clobbers. Predates this PR; happy to file it separately rather than grow this diff.

- The branch is ~4.5 months behind main. I've left CHANGELOG.md alone since the section my base offers has long shipped, and touching it would conflict — following #4644 and #4629 where entries were backfilled. Lmk if you'd rather I rebase and add it.

internal/impl/azure/output_blob_storage.go
```diff
@@ -167,8 +265,26 @@ func (a *azureBlobStorageWriter) uploadBlob(ctx context.Context, containerName,
 				return fmt.Errorf("appending block to blob: %w", err)
 			}
 		}
+		// AppendBlock does not support tags, so they are written separately. This
```
Leward, line 268: For append blobs, tags are a property of the blob, not of each message.

Setting them on `Create` is enough in my opinion. The `SetTags` after every `AppendBlock` doubles the request count on the hot path and only ever reflects the last message.

I would drop it and document that pre-existing blobs are not retagged.

If we do want to cover existing blobs, we should tag them much less often (once per process?), but not once per message.

PR #4216, opened 2026-04-06
add PostgreSQL benchmarking

ness-david-dedu:
Adds a benchmark suite for the postgres input under internal/impl/postgresql/bench/, following the same pattern as the existing Salesforce
   and MSSQL Server benchmarks.

  What's included:
  - Taskfile.yaml — tasks for Docker container lifecycle, data insertion, and parameterised benchmark runs (task bench:cart CORES=4 BATCH=5000,
  task bench:users CORES=2 BATCH=1000)
  - create.sql, users.sql, products.sql, cart.sql — test datasets: 150K large rows (~625 KB each) and 10M small rows (~600 B each)
  - benchmark_config.yaml — postgres_cdc → benchmark processor → drop
  - README.md — setup and run instructions
  - docs/benchmark-results/postgres.md — full results across CPU core counts (1, 2, 4, 8) and batch sizes (1000, 5000, 10000) for both datasets

  Key findings:
  - Small rows (cart): scales to ~318K msg/sec at 8 cores, batch=5000 is the sweet spot
  - Large rows (users): I/O bound, plateaus at ~1,166 msg/sec / 766 MB/sec at 2 cores regardless of additional cores or batch size

### DIFF
--- a/docs/benchmark-results/postgres.md
+++ b/docs/benchmark-results/postgres.md
@@ -1,16 +1,14 @@
-# PostgreSQL CDC Benchmark Results
+# PostgreSQL Benchmark Results
 
-Benchmarks for the `postgres_cdc` input using logical replication (snapshot mode).
+**Environment:** Intel Core i7-10850H @ 2.70GHz, 32 GB RAM, WSL2 (Linux 6.6.87.2), x86_64
 
-See [`internal/impl/postgresql/bench/`](../../internal/impl/postgresql/bench/) for the benchmark configs and run instructions.
+See [`internal/impl/postgresql/bench/`](../../internal/impl/postgresql/bench/) for configs and run instructions.
 
-## Snapshot — Small Rows (cart table), CPU & Batch Size Scaling
+---
 
-Full snapshot of the `public.cart` table: 10,000,000 rows, ~600 B per row. Varying `GOMAXPROCS` and `batching.count` to produce a throughput matrix.
+## CDC / Snapshot — Small Rows (cart table)
 
-**Environment:** Intel Core i7-10850H @ 2.70GHz, 32 GB RAM, WSL2 (Linux 6.6.87.2), x86_64, PostgreSQL 16 running in Docker (localhost)
-
-**Dataset:** 10,000,000 rows × ~600 B
+Full snapshot of `public.cart`: 10,000,000 rows × ~600 B. Varying `GOMAXPROCS` and `batching.count`.
 
 ### msg/sec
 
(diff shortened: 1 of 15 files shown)


squiidz:
1. Hardcoded credentials in plain text

  Every file references postgres://postgres:postgres@localhost:5432/testdb. While this is a local
  benchmark, hardcoding passwords in committed files is a bad habit and makes it harder to run
  against non-default setups.

  Suggestion: Use Taskfile variables with defaults at the top of Taskfile.yaml:
  vars:
    PG_DSN: '{{.PG_DSN | default
  "postgres://postgres:postgres@localhost:5432/testdb?sslmode=disable"}}'
  Then reference {{.PG_DSN}} everywhere. Same for benchmark_config.yaml — accept a DSN env var
  override.

  2. products table is a copy-paste of users

  create.sql defines products with columns like surname, email, date_of_birth, login_count — these
  are user fields, not product fields. The products.sql data script also generates surname-N,
  product@example.com, etc. This looks like a copy-paste of the users table/data that was never
  adapted.

  If the intent is just "another large-row table for benchmarking," that's fine, but it should be
  called out explicitly. Otherwise, give it realistic product columns (name, description, price,
  category, etc.).

  3. bench:run and bench:run:cdc are identical

  Both tasks do psql:drop-slot then run the same config. There's no differentiation for CDC mode
  (e.g., stream_snapshot: false). The README says to start the connector then insert data in a
  second terminal, but the task itself doesn't set stream_snapshot: false, so it will just do a
  snapshot like bench:run.

  Suggestion: Either create a separate benchmark_config_cdc.yaml with stream_snapshot: false, or use
   --set input.postgres_cdc.stream_snapshot=false in the bench:run:cdc task.

  4. Missing bench:users parameterised task

  There's bench:cart with CORES/BATCH parameters, but no equivalent bench:users. The README mentions
   task bench:users CORES=2 BATCH=1000 but this task doesn't exist in the Taskfile.

  5. GOMAXPROCS={{.CORES}} when CORES is empty

  In bench:cart, when CORES defaults to "", the command becomes GOMAXPROCS= go run .... This sets
  GOMAXPROCS to an empty string rather than leaving it unset. Depending on the Go runtime, this may
  or may not behave as "unbounded."

  Suggestion:
  cmds:
    - task psql:drop-slot
    - |
      {{if .CORES}}GOMAXPROCS={{.CORES}}{{end}} go run ...

  6. Large-row benchmark results are sparse (many N/A cells)

  The users table results have N/A for most batch/core combinations. The PR body explains this is
  because the workload is I/O bound, which is fair, but it would strengthen the results to have at
  least one or two more data points to confirm the plateau (e.g., batch=5000 at 2 cores).

  7. No .gitignore or cleanup for Docker/temp artifacts

  If someone runs these benchmarks, there's no postgres:down mentioned in the README's setup flow,
  and no mention of cleanup. The README should include a teardown section.

  8. sleep 5 in postgres:up is fragile

  The postgres:up task uses sleep 5 to wait for PostgreSQL readiness. This can fail on slower
  machines or CI.

  Suggestion: Use pg_isready in a loop:
  - until pg_isready -h localhost -p 5432; do sleep 1; done

squiidz (approved):

josephwoodward (approved):

PR #4218, opened 2026-04-07
doris: add stream load output

xylaaaaa:
## Summary
- add a new `doris_stream_load` output
- wire the component into community/default builds
- support JSON and CSV payloads, FE -> BE 307 redirect handling, `Expect: 100-continue`, query-port based connection tests, and FE failover via `fe_urls`
- promote common Doris Stream Load options into first-class config fields
- add unit coverage for encoding, redirect/failover behavior, promoted header mapping, and connection tests

## Validation
- `go test ./internal/impl/doris`
- `go build -o target/redpanda-connect ./cmd/redpanda-connect`

## Notes
- local demo configs and local docs were intentionally left out of this PR


internal/impl/doris/output_stream_load.go
```diff
@@ -0,0 +1,1077 @@
...
+  root = root.append("partitions and temporary_partitions cannot both be set")
+}
+if this.max_filter_ratio < 0 || this.max_filter_ratio > 1 {
+  root = root.append("max_filter_ratio must be between 0 and 1")
+}
+if this.group_commit != "" && this.group_commit != "sync_mode" && this.group_commit != "async_mode" && this.group_commit != "off_mode" {
+  root = root.append("group_commit must be one of sync_mode, async_mode, off_mode")
+}
+`).
+		Description(service.OutputPerformanceDocs(true, true))
```
josephwoodward: `.Description()` is called twice on this spec — once at [line 185](https://github.com/redpanda-data/connect/blob/ce413f85b2989fee98f13f6992d1689aae56a84d/internal/impl/doris/output_stream_load.go#L184-L186) with the primary user-facing description, and again here with service.OutputPerformanceDocs(true, true). The second call replaces the first, so the documented behavior of the sink (FE→BE redirect, ack-on-success, etc.) is lost in the generated docs.

Other components combine the two into a single .Description() call by string-concatenating the performance docs (see e.g. internal/impl/snowflake/output_snowflake_streaming.go). Either concatenate the strings or move the performance docs into the original Description block.

internal/plugins/info.csv
```diff
@@ -69,6 +69,7 @@ cypher                    ,output    ,cypher                    ,4.37.0  ,commun
 decompress                ,processor ,decompress                ,0.0.0   ,certified  ,n          ,y     ,y
 decompress                ,scanner   ,decompress                ,0.0.0   ,certified  ,n          ,y     ,y
 dedupe                    ,processor ,dedupe                    ,0.0.0   ,certified  ,n          ,y     ,y
+doris_stream_load         ,output    ,doris_stream_load         ,0.0.0   ,community  ,n          ,n     ,n
```
josephwoodward: Can this be moved below both `discords` but above `drop` so it stays alphabetically ordered?

internal/impl/doris/output_stream_load.go
```diff
@@ -0,0 +1,1077 @@
...
+	case "csv":
+		req.Header.Set("Content-Type", "text/csv")
+		req.Header.Set(dsHeaderColumnSeparator, escapeDorisHeaderValue(d.conf.ColumnSeparator))
+		req.Header.Set(dsHeaderLineDelimiter, escapeDorisHeaderValue(d.conf.LineDelimiter))
+	}
+
+	return req, nil
+}
+
+func (_ *dorisStreamLoadOutput) resolveRedirectURL(feEndpoint *url.URL, location string) (string, error) {
```
josephwoodward: Given this doesn't depend on the `dorisStreamLoadOutput`, could this instead be a package-level function?

internal/impl/doris/output_stream_load.go
```diff
@@ -0,0 +1,1077 @@
...
+		BackoffMaxRetries:      3,
+		Retry:                  httpclient.DefaultRetryConfig(),
+		MetricPrefix:           "doris_stream_load_http",
+	}
+	// Doris requires the Expect header, but waiting for a 100-continue
+	// response adds avoidable latency on servers that don't emit it.
+	cfg.Transport.ExpectContinueTimeout = 0
+	cfg.AuthSigner = httpclient.BasicAuthSigner(conf.Username, conf.Password)
+
+	if strings.HasPrefix(conf.FEURLs[0], "https://") {
```
josephwoodward: Should we check all `conf.FEURLs` values here instead of just the first? If the first element is `http://` but the second is `https://` then TLS would never be configured?

josephwoodward:
I can see there are a few linting violations that need clearing up. And are you also able to run `task docs` to generate the docs for this new connector please?

Would you also be able to sign the CLA when you have a moment please? https://github.com/redpanda-data/connect/pull/4218#issuecomment-4227383203

xylaaaaa:
Thanks, this should be addressed now.
I fixed the lint violations and generated the new connector docsd.
The CLA should also be resolved now.

internal/impl/doris/output_stream_load.go
```diff
@@ -0,0 +1,1089 @@
...
+	}
+	if loadResp.Status == dsStatusPublishTimeout {
+		d.log.Warnf(
+			"Doris stream load committed but data may become visible later label=%s table=%s.%s status=%s loaded_rows=%d filtered_rows=%d total_rows=%d",
+			logLabel, d.conf.Database, d.conf.Table, loadResp.Status, loadResp.NumberLoadedRows, loadResp.NumberFilteredRows, loadResp.NumberTotalRows,
+		)
+		return nil
+	}
+
+	d.log.Infof(
```
josephwoodward: This has potential to be quite noisy, would it be worth downgrading it to `Debug`?

internal/impl/doris/output_stream_load.go
```diff
@@ -0,0 +1,1089 @@
...
+		return err
+	}
+	defer resp.Body.Close()
+
+	loadResp, rawBody, err := parseDorisStreamLoadResponse(resp)
+	if err != nil {
+		if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+			return fmt.Errorf("doris returned HTTP %d: %s", resp.StatusCode, strings.TrimSpace(string(rawBody)))
+		}
+		return fmt.Errorf("parsing Doris response: %w", err)
```
josephwoodward:
```suggestion
		return fmt.Errorf("parsing doris response: %w", err)
```

Following the standard of lowercase error messages (similar to the error returned on line 707)

internal/impl/doris/output_stream_load.go
```diff
@@ -0,0 +1,1089 @@
...
+}
+
+func streamLoadURLs(baseURLs []string, database, table string) ([]*url.URL, error) {
+	endpoints := make([]*url.URL, 0, len(baseURLs))
+	for _, baseURL := range baseURLs {
+		u, err := url.Parse(baseURL)
+		if err != nil {
+			return nil, fmt.Errorf("parsing base url %q: %w", baseURL, err)
+		}
+		u.Path = path.Join(u.Path, "api", database, table, "_stream_load")
```
josephwoodward: If parsing URLs, would `url.JoinPath()` be better here?

internal/impl/doris/output_stream_load.go
```diff
@@ -0,0 +1,1089 @@
...
+		return d.connectionCheckFn(ctx, feURL, queryPort)
+	}
+
+	parsed, err := url.Parse(feURL)
+	if err != nil {
+		return fmt.Errorf("parsing Doris FE URL %q for query_port checks: %w", feURL, err)
+	}
+	host := parsed.Hostname()
+	if host == "" {
+		return fmt.Errorf("unable to determine host from Doris FE URL %q for query_port checks", feURL)
```
josephwoodward:
```suggestion
		return fmt.Errorf("unable to determine host from doris FE URL %q for query_port checks", feURL)
```

internal/impl/doris/output_stream_load.go
```diff
@@ -0,0 +1,1089 @@
...
+		if d.conf.QueryPort > 0 {
+			if err := d.connectionCheck(ctx, feURL, d.conf.QueryPort); err != nil {
+				lastErr = err
+				continue
+			}
+		}
+		return service.ConnectionTestSucceeded().AsList()
+	}
+	if lastErr == nil {
+		lastErr = errors.New("no Doris FE endpoints configured")
```
josephwoodward:
```suggestion
		lastErr = errors.New("no doris FE endpoints configured")
```

josephwoodward:
@xylaaaaa Just added a few more comments, otherwise it's looking good!

I see there's a [Doris Docker image](https://hub.docker.com/r/apache/doris), have you considered adding some integration tests utilising this container? You can follow the same pattern in a lot of our Integration tests (such as our [MySQL integration tests](https://github.com/redpanda-data/connect/blob/main/internal/impl/mysql/integration_test.go?plain=1#L112)).

xylaaaaa:
Addressed the follow-up comments and added a basic Docker-based integration test.
The new test starts Doris FE/BE using `apache/doris:fe-4.1.0` and `apache/doris:be-4.1.0`, creates a test table, writes a JSON batch through `doris_stream_load`, and verifies the loaded rows via Doris SQL.
Verified locally with:
`go test -mod=mod -timeout 20m ./internal/impl/doris -run '^TestIntegrationDorisStreamLoadOutput$' -count=1 -v`
Result: `PASS`

morningman:
hi @josephwoodward could you help to re-run "Calude Code Review" workflow?

internal/impl/doris/output_stream_load.go
```diff
@@ -0,0 +1,1087 @@
...
+	}
+	if err := classifyDorisStreamLoadResponse(resp.StatusCode, loadResp, rawBody); err != nil {
+		return err
+	}
+
+	logLabel := label
+	if loadResp.Label != "" {
+		logLabel = loadResp.Label
+	}
+	if loadResp.Status == dsStatusPublishTimeout {
```
josephwoodward, line 722: Would it be worth adding a comment here to explain that string matching against "Publish Timeout" is the only way here? Perhaps linking to the [documentation](https://doris.apache.org/docs/3.x/data-operate/import/import-way/stream-load-manual). And any other places where string matching like this is necessary?

internal/impl/doris/integration_test.go
```diff
@@ -0,0 +1,301 @@
...
+	"github.com/redpanda-data/benthos/v4/public/service"
+	"github.com/redpanda-data/benthos/v4/public/service/integration"
+)
+
+const dorisIntegrationVersion = "4.1.0"
+
+func TestIntegrationDorisStreamLoadOutput(t *testing.T) {
+	integration.CheckSkip(t)
+	if runtime.GOOS != "linux" {
+		t.Skip("Doris Stream Load redirects to the BE container IP, which requires Linux Docker bridge networking")
```
josephwoodward: Is there any way to get these tests passing on MacOS? I don't appear to be able to get them running in our CI pipeline. I believe this is due to resource constraints but I'd need more time to look into it.

josephwoodward:
> hi @josephwoodward could you help to re-run "Calude Code Review" workflow?

Hi @morningman, Claude Review is only triggered on PRs created by people with write permission to the Connect repo (to avoid abuse). I've pushed a separate branch [here](https://github.com/redpanda-data/connect/pull/4472) to trigger a review but it didn't find much I feel is of concern.

@xylaaaaa I've added a few minor comments/questions if you want to take a look? Once you've replied I think we're in a good point to merge this. Thanks for all of your work!

morningman:
> > hi @josephwoodward could you help to re-run "Calude Code Review" workflow?
> 
> Hi @morningman, Claude Review is only triggered on PRs created by people with write permission to the Connect repo (to avoid abuse). I've pushed a separate branch [here](https://github.com/redpanda-data/connect/pull/4472) to trigger a review but it didn't find much I feel is of concern.
> 
> @xylaaaaa I've added a few minor comments/questions if you want to take a look? Once you've replied I think we're in a good point to merge this. Thanks for all of your work!

hi @josephwoodward , looks like every comment is solved, PTAL

josephwoodward (approved):

josephwoodward:
Great, thank you for your contribution @xylaaaaa and @morningman

PR #4220, opened 2026-04-07
BigQuery: storage write API connector

squiidz:


mmatczuk:
Top 2 commits can be squashed I think

mmatczuk:
Note https://java.testcontainers.org/modules/gcloud/

mmatczuk:
Would be nice to redo test comments into given/when/then logs

mmatczuk:
Can we have a constructor for tests instead of

```
&bigQueryWriteAPIOutput{
		conf: bigQueryWriteAPIConfig{
			ProjectID: "my-project",
			DatasetID: "my_dataset",
		},
```

internal/impl/gcp/enterprise/output_bigquery_write_api.go
```diff
@@ -0,0 +1,704 @@
...
+)
+
+func init() {
+	service.MustRegisterBatchOutput("gcp_bigquery_write_api", bigQueryWriteAPISpec(),
+		func(conf *service.ParsedConfig, mgr *service.Resources) (
+			out service.BatchOutput, batchPolicy service.BatchPolicy, maxInFlight int, err error,
+		) {
+			if err = license.CheckRunningEnterprise(mgr); err != nil {
+				return
+			}
```
mmatczuk: Check license in constructor not here.

internal/impl/gcp/enterprise/bigquery/output.go
```diff
@@ -0,0 +1,704 @@
...
+		) {
+			if err = license.CheckRunningEnterprise(mgr); err != nil {
+				return
+			}
+			if maxInFlight, err = conf.FieldMaxInFlight(); err != nil {
+				return
+			}
+			if batchPolicy, err = conf.FieldBatchPolicy(bqwaFieldBatching); err != nil {
+				return
+			}
```
mmatczuk, line 66: Try preserving order where it makes sense

internal/impl/gcp/enterprise/bigquery/output.go
```diff
@@ -0,0 +1,704 @@
...
+JSON format the component automatically fetches the table schema and converts
+each message to the corresponding proto representation.
+
+WARNING: The proto3 JSON mapping encodes int64 and uint64 values as strings.
+JSON messages with integer fields must use string values (e.g. `+"`"+`"age": "30"`+"`"+`
+not `+"`"+`"age": 30`+"`"+`), otherwise the write will fail with an unmarshalling error.
+
+When batching is enabled the table name is resolved from the first message in
+each batch; all messages in the same batch are written to that table.
+`).
```
mmatczuk, line 90: nit: it's the best for maintenance to use sentence per line

mmatczuk:
It's handy to support `credentials_json`

mmatczuk:
I'd suggest doing what we did with AWS, extract that to separate package. Retroactive regrouping should follow. When we have a separate package please unpack utilities to separate files.

internal/impl/gcp/enterprise/bigquery/output.go
```diff
@@ -0,0 +1,715 @@
...
+				Default([]any{}),
+			service.NewStringField(bqwaFieldStreamIdleTimeout).
+				Description("How long a cached stream can remain unused before being closed."+
+					" Relevant when the table field uses interpolation to route to many tables.").
+				Advanced().
+				Default("5m"),
+			service.NewStringField(bqwaFieldStreamSweepInterval).
+				Description("How often to check for idle streams to close.").
+				Advanced().
+				Default("1m"),
```
Jeffail, line 133: These duration fields are declared as `NewStringField` and then parsed manually via `time.ParseDuration` in `bigQueryWriteAPIConfigFromParsed`. Benthos exposes `service.NewDurationField` + `pConf.FieldDuration`, which roughly fifteen other components in this tree already use (snowflake, oracledb, sftp, redpanda, salesforce, etc.). Switching over would drop the hand-rolled `"invalid %s: %w"` wrapping and bring this in line with the rest of the codebase.

internal/impl/gcp/enterprise/bigquery/output.go
```diff
@@ -0,0 +1,715 @@
...
+			return fmt.Errorf("unsupported message format: %s", o.conf.MessageFormat)
+		}
+		rows = append(rows, protoBytes)
+	}
+
+	result, err := swd.stream.AppendRows(ctx, rows)
+	if err != nil {
+		o.evictStream(cacheKey)
+		if classifyGRPCError(err) == grpcPermanent {
+			return fmt.Errorf("permanent error appending rows (will not retry): %w", err)
```
Jeffail: The `grpcPermanent` branch doesn't currently prevent retries: the function returns a plain `error`, which the benthos framework will retry, so the `"will not retry"` substring in the log message isn't quite accurate. The metric gating also only increments `retries` on transient errors, which means permanent errors (which are in fact retried by the framework) are invisible to that counter.

The kafka output in `output_sarama_kafka.go` has a pattern that does enforce the intent: build a `service.NewBatchError` and mark every message in the batch `Failed(i, err)` — benthos treats `BatchError` entries as permanent per-message failures and won't retry them. Happy either way, whether that's wiring `classifyGRPCError` through `BatchError.Failed`, or dropping the classification and relying on `managedwriter`'s internal retries plus benthos's batch retry. As it stands the classification is a bit of a half-measure.

internal/impl/gcp/enterprise/bigquery/output.go
```diff
@@ -0,0 +1,715 @@
...
+		return errors.New("could not determine GCP project ID; set the 'project' field explicitly")
+	}
+
+	storageClient, err := managedwriter.NewClient(ctx, resolvedProject, storageOpts...)
+	if err != nil {
+		_ = bqClient.Close()
+		return fmt.Errorf("creating storage write client: %w", err)
+	}
+
+	o.conf.ProjectID = resolvedProject
```
Jeffail: Minor: mutating the config struct after construction is a little unusual — a dedicated `resolvedProjectID` field on the output struct would keep `conf` immutable post-parse, matching the convention elsewhere in the codebase.

internal/impl/gcp/enterprise/bigquery/output.go
```diff
@@ -0,0 +1,715 @@
...
+	if endpointOverride != "" {
+		opts := []option.ClientOption{
+			option.WithoutAuthentication(),
+			option.WithEndpoint(endpointOverride),
+		}
+		if isGRPC {
+			opts = append(opts, option.WithGRPCDialOption(grpc.WithTransportCredentials(insecure.NewCredentials())))
+		}
+		return opts, nil
+	}
```
Jeffail, line 508: If both `endpoint` and `target_principal` are set, impersonation is silently dropped on this early-return path. Worth either logging a warning during construction or calling the interaction out in the field description — right now the behaviour is easy to miss.

internal/impl/gcp/enterprise/bigquery/output.go
```diff
@@ -0,0 +1,715 @@
...
+	}, nil
+}
+
+func (o *bigQueryWriteAPIOutput) Connect(ctx context.Context) error {
+	o.connMu.Lock()
+	defer o.connMu.Unlock()
+
+	if o.client != nil {
+		return nil
+	}
```
Jeffail, line 314: Nit: this short-circuit means a client that has gone bad (for instance a transport-level failure that wasn't surfaced through `Close`) can't be healed by benthos's reconnect loop — it'll stay broken until `Close` runs. `managedwriter` handles most of this internally in practice, so this is more of a design observation than a defect.

Jeffail (commented):
Did a foundation pass to sanity-check whether the new code sits on the existing primitives cleanly. Most of it does — flagging two spots where the implementation works around the foundation rather than leveraging it, plus a few smaller observations that are more judgment calls.

internal/impl/gcp/enterprise/bigquery/output.go
```diff
@@ -0,0 +1,720 @@
...
+	rowsSent     *service.MetricCounter
+	rowsFailed   *service.MetricCounter
+	batchesSent  *service.MetricCounter
+	batchLatency *service.MetricTimer
+	retries      *service.MetricCounter
+}
+
+func newBQWAMetrics(m *service.Metrics) *bqwaMetrics {
+	return &bqwaMetrics{
+		rowsSent:     m.NewCounter("bigquery_write_api_rows_sent"),
```
josephwoodward: It would be good to follow the [Prometheus naming conventions](https://prometheus.io/docs/practices/naming/) on these metrics. For instance:

> ...SHOULD have a suffix describing the unit, in plural form. Note that an accumulating count has total as a suffix, in addition to the unit if applicable. Also note that this applies to units in the narrow sense (like the units in the table below), but not to countable things in general. For example, connections or notifications are not considered units for this rule and do not have to be at the end of the metric name. (See also examples in the next paragraph.)
> - http_request_duration_seconds
> - node_memory_usage_bytes
> - http_requests_total (for a unit-less accumulating count)
> - process_cpu_seconds_total (for an accumulating count with unit)
> - foobar_build_info (for a pseudo-metric that provides [metadata ](https://www.robustperception.io/exposing-the-software-version-to-prometheus) about the running binary)
> - data_pipeline_last_record_processed_timestamp_seconds (for a timestamp that tracks the time of the latest record processed in a data processing pipeline)

internal/impl/gcp/enterprise/bigquery/output.go
```diff
@@ -0,0 +1,720 @@
...
+				return
+			}
+			out, err = bigQueryWriteAPIOutputFromConfig(conf, mgr)
+			return
+		})
+}
+
+func bigQueryWriteAPISpec() *service.ConfigSpec {
+	return service.NewConfigSpec().
+		Version("4.87.0").
```
josephwoodward: Update this to `4.90.0` if it's going out in this release?

josephwoodward (approved):

PR #4236, opened 2026-04-09
arc: add output plugin for Arc columnar database

xe-nvdk:
## Summary

- Add a new `arc` batch output plugin that sends data to Arc's MessagePack ingestion endpoint
- Arc is a columnar analytical database optimized for time-series and observability workloads, capable of ingesting millions of records per second
- This plugin enables Redpanda Connect users to stream data from any supported source directly into Arc without custom integration code

### Features

- Columnar and row payload formats (columnar default, 25-35% faster ingestion)
- zstd/gzip compression with pooled encoders for high throughput
- Interpolated measurement names for per-message routing
- Automatic timestamp unit detection (s, ms, us, ns)
- Bloblang-based tag extraction for row format
- Bearer token authentication and TLS support
- No new dependencies (uses existing msgpack/v5 and klauspost/compress)

### Example config

```yaml
output:
  arc:
    url: http://localhost:8000
    database: mydb
    measurement: server_metrics
    format: columnar
    compression: zstd
    batching:
      count: 1000
      period: 1s
```

## Test plan

- [x] Unit tests covering columnar, row, multi-measurement, compression, auth, errors, empty batch, and timestamp conversion (9 tests, all passing)
- [x] End-to-end test against a running Arc instance with generated data
- [x] Build verification with `go build` and `go vet`


internal/impl/arc/output.go
```diff
@@ -0,0 +1,551 @@
...
+			order = append(order, measName)
+		}
+
+		rec.columns["time"] = append(rec.columns["time"], ts)
+		for k, v := range dataMap {
+			if k == o.timestampField {
+				continue
+			}
+			rec.columns[k] = append(rec.columns[k], v)
+		}
```
mmatczuk, line 301: **Bug: `"time"` column collision when `timestamp_field` is empty**

When `timestamp_field` is `""` (the default), `o.timestampField == ""` means the guard `k == o.timestampField` is equivalent to `k == ""`. No real message field is named `""`, so the skip never triggers. If a message contains a field named `"time"` (extremely common in time-series data), two values get appended to `rec.columns["time"]`:

1. The auto-generated timestamp on line 313
2. The original `"time"` value from the message in the loop on line 318

This leaves the `"time"` column with twice as many entries as every other column, corrupting the columnar payload (Arc requires all columns have equal length).

Fix: when `timestampField` is empty, also exclude the key `"time"` from the loop:
```go
if k == o.timestampField || (o.timestampField == "" && k == "time") {
    continue
}
```
Or, cleaner: always use an explicit exclusion for the column that holds the generated timestamp regardless of the `timestampField` setting.

internal/impl/arc/output.go
```diff
@@ -0,0 +1,551 @@
...
+	}
+	return v
+}
+
+func (o *arcOutput) compress(data []byte) ([]byte, string, error) {
+	switch o.compression {
+	case "none":
+		return data, "", nil
+	case "zstd":
```
mmatczuk, line 461: **Bug: zstd encoder returned to pool in error state**

When `w.Close()` returns an error, the encoder is put back into the pool in an unknown internal state. Any subsequent caller that retrieves it and calls `w.Reset(&buf)` may silently produce malformed compressed output or panic.

The same issue exists in the `gzip` branch (~line 492-494).

Only return the writer to the pool on the success path. On error, let it be GC'd:
```go
if err := w.Close(); err != nil {
    // Don't Put — encoder state is unknown
    return nil, "", err
}
zstdEncoderPool.Put(w)
```

internal/impl/arc/output.go
```diff
@@ -0,0 +1,551 @@
...
+			return nil, err
+		}
+	}
+
+	if o.compression, err = conf.FieldString(fieldCompression); err != nil {
+		return nil, err
+	}
+
+	// Set up HTTP client with optional TLS
+	transport := http.DefaultTransport.(*http.Transport).Clone()
```
mmatczuk: **Bug: type assertion on `http.DefaultTransport` will panic if replaced**

`http.DefaultTransport` is a package-level `http.RoundTripper` variable. Any code in the process (test helpers, monitoring libraries, middleware) is allowed to replace it with a non-`*http.Transport` value. This type assertion panics with no recovery if that happens.

Construct a fresh transport instead of cloning the default:
```go
transport := &http.Transport{
    // copy only the fields you need, or use http.DefaultTransport values explicitly
}
```
Or use a comma-ok assertion and return an error.

internal/impl/arc/output_test.go
```diff
@@ -0,0 +1,383 @@
...
+	var headers http.Header
+
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		mu.Lock()
+		defer mu.Unlock()
+		headers = r.Header.Clone()
+		var err error
+		received, err = io.ReadAll(r.Body)
+		require.NoError(t, err)
+		w.WriteHeader(http.StatusNoContent)
```
mmatczuk, line 51: **Test: `require` inside HTTP handler goroutine causes incorrect failure behavior**

`httptest.NewServer` handlers run on a goroutine spawned by `net/http`. Calling `require.NoError(t, err)` inside the handler invokes `t.FailNow()`, which calls `runtime.Goexit()` on the wrong goroutine. This terminates the HTTP handler goroutine (not the test goroutine), causes the server to close the connection with no response, and leaves the test goroutine hanging waiting for a response that never arrives.

The same issue exists in every other test that calls `require.NoError(t, err)` inside the server handler (`TestArcOutput_ColumnarMultipleMeasurements`, `TestArcOutput_RowFormat`, `TestArcOutput_ZstdCompression`, `TestArcOutput_GzipCompression`).

Replace `require` with `assert` inside handlers, or capture the error and check it after `WriteBatch` returns:
```go
srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    mu.Lock()
    defer mu.Unlock()
    headers = r.Header.Clone()
    var readErr error
    received, readErr = io.ReadAll(r.Body)
    assert.NoError(t, readErr) // use assert, not require
    w.WriteHeader(http.StatusNoContent)
}))
```

internal/impl/arc/output_test.go
```diff
@@ -0,0 +1,383 @@
...
+	usages := columns["usage"].([]any)
+	assert.Len(t, usages, 2)
+
+	times := columns["time"].([]any)
+	assert.Len(t, times, 2)
+
+	require.NoError(t, out.Close(t.Context()))
+}
+
+func TestArcOutput_ColumnarMultipleMeasurements(t *testing.T) {
```
mmatczuk, line 111: **Test: data race on shared variable `received` (and others)**

`TestArcOutput_ColumnarSingleMeasurement` correctly protects `received` with `sync.Mutex`. The remaining tests that capture `received`, `contentEncoding`, and `authHeader` from the handler — `TestArcOutput_ColumnarMultipleMeasurements`, `TestArcOutput_RowFormat`, `TestArcOutput_ZstdCompression`, `TestArcOutput_GzipCompression`, `TestArcOutput_BearerToken` — do not.

The HTTP handler goroutine writes these variables, and the test goroutine reads them after `WriteBatch` returns. Without explicit synchronization, the Go race detector will flag this. Add a `sync.Mutex` (or `sync/atomic`) to each test, following the pattern established in `TestArcOutput_ColumnarSingleMeasurement`.

internal/impl/arc/output_test.go
mmatczuk, line 1: **Test: missing config linting test and invalid-config error-path tests**

Two standard test patterns are missing:

1. **Config linting test** — validates that the `ConfigSpec` definition itself is self-consistent (no duplicate field names, valid defaults, etc.). Example:
   ```go
   func TestArcOutputConfigLinting(t *testing.T) {
       env := service.NewEnvironment()
       // register arc output in env, then lint
   }
   ```

2. **Invalid config table test** — no tests exercise error paths in `newArcOutput`. The constructor validates the token (no `\r\n`) and the database name, but there are no tests for configs that should fail. A table-driven test covering invalid values (e.g., token with newline, missing required fields) is expected alongside the happy-path tests.

**Test: no integration test for new component**

New components are expected to include integration tests (see tester agent scope). This PR has unit tests only. An integration test using testcontainers-go (or `integration.CheckSkip(t)` if Arc doesn't have a public Docker image yet) should be added, or a follow-up ticket filed and linked here.

mmatczuk:
**Commits**
LGTM

**Review**
New `arc` batch output plugin. Well-structured implementation with good compression pooling and columnar/row format support. A few real bugs and test pattern gaps need attention before merge.

1. **[Data corruption] `"time"` column collision in columnar format** — [`output.go#L313-L319`](https://github.com/redpanda-data/connect/blob/b0ab5a4fd808021be76f659afeab113cd7f28109/internal/impl/arc/output.go#L313-L319): when `timestamp_field` is `""` (default), the guard `k == o.timestampField` never matches any real key. A message with a `"time"` field gets it appended both from the explicit timestamp line and from the loop, producing a column with twice as many entries as others and corrupting the payload.

2. **[Correctness] Poisoned encoder returned to pool** — [`output.go#L477-L479`](https://github.com/redpanda-data/connect/blob/b0ab5a4fd808021be76f659afeab113cd7f28109/internal/impl/arc/output.go#L477-L479): when `w.Close()` fails, the zstd encoder (and gzip writer, ~L492) is put back into the pool in an unknown state. A subsequent caller may silently produce malformed output. Only `Put` on the success path.

3. **[Panic] `http.DefaultTransport` type assertion** — [`output.go#L218`](https://github.com/redpanda-data/connect/blob/b0ab5a4fd808021be76f659afeab113cd7f28109/internal/impl/arc/output.go#L218): panics if any code in the process replaces `http.DefaultTransport` with a non-`*http.Transport`. Construct a fresh `http.Transport{}` instead.

4. **[Test] `require` inside HTTP handler goroutines** — [`output_test.go#L24-L31`](https://github.com/redpanda-data/connect/blob/b0ab5a4fd808021be76f659afeab113cd7f28109/internal/impl/arc/output_test.go#L24-L31): affects 5 tests. `require.NoError` calls `t.FailNow()` on the handler goroutine, not the test goroutine, causing the handler to exit without sending a response and leaving the test goroutine hung. Use `assert.NoError` inside handlers.

5. **[Test] Data race on shared variables** — [`output_test.go#L76-L84`](https://github.com/redpanda-data/connect/blob/b0ab5a4fd808021be76f659afeab113cd7f28109/internal/impl/arc/output_test.go#L76-L84): 5 tests write `received`/`contentEncoding`/`authHeader` from the handler goroutine and read them from the test goroutine without a mutex. `go test -race` will flag these. Follow the mutex pattern from `TestArcOutput_ColumnarSingleMeasurement`.

6. **[Test] Missing config linting test, invalid-config tests, and integration tests** — no test covers config spec self-consistency, error paths in `newArcOutput` (e.g., token/database with `\r\n`), or an end-to-end integration test for the new component.

xe-nvdk:
Awesome feedback, thank you. We are going to address these today.

xe-nvdk:
Thanks for the thorough review @mmatczuk. All six issues have been addressed in the latest push:

**Bug fixes:**
1. **Time column collision** - Fixed. When `timestamp_field` is empty (default), the `"time"` key from message data is now excluded from the column loop to prevent duplicate entries. Same fix applied to both columnar and row paths.
2. **Poisoned encoder pool** - Fixed. Compression writers are only returned to the pool on the success path. On error, they are left for GC.
3. **DefaultTransport type assertion** - Fixed. Now constructs a fresh `http.Transport` with explicit settings instead of asserting on `http.DefaultTransport`.

**Test fixes:**
4. **require inside handler goroutines** - Fixed. Removed all testify calls from HTTP handler goroutines. Handlers now use plain `if err != nil` checks and return 500.
5. **Data races on shared variables** - Fixed. All tests now use `sync.Mutex` for variables shared between handler and test goroutines. Extracted a shared `newTestServer` helper to enforce the pattern consistently.
6. **Missing tests** - Added:
   - `TestArcOutput_TimeColumnCollision` - verifies the time field deduplication fix
   - `TestArcOutput_InvalidConfig` - table-driven test covering CRLF rejection for token and database
   - `integration_test.go` - end-to-end test using `ghcr.io/basekick-labs/arc:latest` via testcontainers-go, covering columnar write+query and row format write+query

All unit tests pass with `-race`, integration tests pass against a real Arc container.

Ready for another look.

internal/impl/arc/output_test.go
```diff
@@ -0,0 +1,468 @@
+package arc
```
mmatczuk, line 15: Missing Apache 2.0 license header. Both `output.go` and `integration_test.go` include the standard header, but this file starts with bare `package arc`. CI license checks will fail.

Add the same license header used in `output.go`:
```
// Copyright 2024 Redpanda Data, Inc.
//
// Licensed under the Apache License, Version 2.0 (the "License");
// ...
```

internal/impl/arc/output.go
```diff
@@ -0,0 +1,554 @@
...
+			order = append(order, measName)
+		}
+
+		rec.columns["time"] = append(rec.columns["time"], ts)
+		for k, v := range dataMap {
+			if k == o.timestampField || (o.timestampField == "" && k == "time") {
+				continue
+			}
+			rec.columns[k] = append(rec.columns[k], v)
+		}
```
mmatczuk, line 301: Bug: columnar payload produces mismatched column lengths when messages have heterogeneous fields.

If messages within the same measurement group have different field sets (e.g., msg1 has `{a, b}`, msg2 has `{a, c}`), the resulting column arrays will have different lengths: column `a` has 2 entries, columns `b` and `c` each have 1. There is no way to determine which row a value belongs to — this produces invalid columnar data that Arc will either reject or misalign.

Fix: pad missing columns with a null/sentinel value for each message that lacks a given field, so all column arrays within a measurement group have equal length. Alternatively, collect the union of all field names first, then iterate messages again filling nulls for missing fields.

internal/impl/arc/output.go
```diff
@@ -0,0 +1,554 @@
...
+		w.Reset(&buf)
+		if _, err := w.Write(data); err != nil {
+			w.Close()
+			return nil, "", err
+		}
+		if err := w.Close(); err != nil {
+			return nil, "", err
+		}
+		gzipWriterPool.Put(w)
+		return buf.Bytes(), "gzip", nil
```
mmatczuk, line 484: Resource leak: compression writers obtained from `sync.Pool` are not returned on error paths.

When `w.Write()` fails, `w.Close()` is called but the writer is never returned to the pool. When `w.Close()` fails, the writer is also lost. This applies to both the zstd (lines 488-498) and gzip (lines 501-511) branches.

Under sustained write or close errors, the pool will be depleted and every subsequent call will allocate a new encoder — these are expensive objects (zstd encoders in particular).

Fix: use `defer` to unconditionally return the writer to the pool after `Reset()`, regardless of error path. Both `zstd.Encoder` and `gzip.Writer` support `Reset()` after `Close()`, so reuse after put-back is safe.

mmatczuk (commented):
Review of arc output plugin

mmatczuk:
**Commits**
1. `b7e491ce015e` "arc: address PR review feedback" — vague message. Does not describe what actually changed. Should be split or reworded to reflect the specific fixes (e.g., `arc: fix time column collision and pool leak in compress`, `arc: add integration tests`).

**Review**
New Arc batch output plugin with columnar/row formats, compression, auth, and TLS. Overall solid structure — correct registration pattern, proper use of `service.MustRegisterBatchOutput`, good test coverage. Three issues found:

1. **Missing license header** — [`output_test.go:1`](https://github.com/redpanda-data/connect/blob/b7e491ce015e1175abdc4e5b5fd5fe4db6a6fde2/internal/impl/arc/output_test.go#L1) — Missing Apache 2.0 license header present in the other two files.
2. **Columnar column length mismatch** — [`output.go:320-326`](https://github.com/redpanda-data/connect/blob/b7e491ce015e1175abdc4e5b5fd5fe4db6a6fde2/internal/impl/arc/output.go#L320-L326) — Messages with heterogeneous fields produce columns of different lengths, resulting in invalid columnar data.
3. **sync.Pool writer leak on error** — [`output.go:488-511`](https://github.com/redpanda-data/connect/blob/b7e491ce015e1175abdc4e5b5fd5fe4db6a6fde2/internal/impl/arc/output.go#L488-L511) — Compression writers are not returned to the pool on error paths in both zstd and gzip branches.

xe-nvdk:
Vague commit message? Did you even read the message, yeah, the title maybe is vague but the message says what was fixed. 

So, I heavily recommend you to not depend 100% of AI, doesn't look good when the responses are not accurate.

I hope that you take this comment in a good way. 

I will address the remaining issues that you found. 


Thank you

mmatczuk:
> Vague commit message? Did you even read the message, yeah, the title maybe is vague but the message says what was fixed.

LOL don't worry about it, it's a policy thing, we can squash it later

xe-nvdk:
Thanks for the second pass @mmatczuk. Addressed in the latest push:

**1. License header** - Added Apache 2.0 header to `output_test.go`.

**2. Columnar column length mismatch** - This is an inherent constraint of columnar format, not a plugin bug. Arc validates this server-side and returns a clear error:

```go
// arc/internal/ingest/msgpack.go:202-216
for colName, colData := range payload.Columns {
    if firstCol {
        numRecords = len(colData)
        firstCol = false
    } else if len(colData) != numRecords {
        return nil, fmt.Errorf("columnar format: array length mismatch (expected %d, got %d for '%s')",
            numRecords, len(colData), colName)
    }
}
```

Data is never silently misaligned. Arc rejects the batch with a descriptive error that the plugin surfaces to the caller.

Schema evolution across separate batches is fully supported. Batch 1 with columns `{a, b}` and batch 2 with columns `{a, c}` works fine. Arc appends new columns and fills nulls for previous rows.

Adding client-side null-padding would hide schema inconsistencies from the user. The row format already exists for heterogeneous schemas. Added a note in the config description documenting this constraint.

**3. Pool writer leak** - Fixed. Now uses `defer pool.Put(w)` immediately after `Get()` so writers are returned on all code paths. Both `zstd.Encoder` and `gzip.Writer` support `Reset()` after `Close()`, so reuse is safe.

internal/impl/arc/output_test.go
```diff
@@ -0,0 +1,482 @@
+// Copyright 2024 Redpanda Data, Inc.
```
mmatczuk: 2026

mmatczuk:
Please use https://github.com/redpanda-data/connect/tree/main/internal/httpclient for http client

xe-nvdk:
@mmatczuk all addressed in the latest push:

1. **License header year** - Updated copyright year from 2024 to 2026 across all arc package files.

2. **HTTP client** - Refactored to use \`internal/httpclient\`. Now uses \`httpclient.NewConfigFromParsed\`, \`httpclient.BearerTokenSigner\`, and \`httpclient.NewClient\`. This removes our manual \`http.Transport\` construction, retry logic, and timeout handling. Field renamed from \`url\` to \`base_url\` to match the shared convention.

Also, a recommendation for a future PR: the current \`github.com/vmihailenco/msgpack/v5\` dependency is effectively unmaintained (last commit is over 3 years old). We maintain a fork at \`github.com/Basekick-Labs/msgpack/v6\` that we use in Arc with the following improvements:

- **+8.6% throughput** and **-9.6% p50 latency** on real workloads (measured in Arc)
- Zero-alloc \`MarshalAppend\` API for encoding into pooled buffers
- \`*interface{}\` fast path that bypasses reflection for common decode patterns
- Pooled encoder/decoder buffers with \`GetEncoder/PutEncoder\` helpers
- Type-switch fast paths for \`map[string]interface{}\`, \`map[string]string\`, \`[]interface{}\`
- Bug fixes for OOM protection (capped slice/map allocations), float-to-int validation, non-addressable pointer receivers, and non-string map key support in interface decode

The API is fully compatible with vmihailenco/msgpack/v5 (same \`Marshal\`/\`Unmarshal\` signatures and interfaces), so it would be a drop-in replacement. Happy to open a separate PR for that if you're open to it.

mmatczuk:
Thanks for the heads up for `github.com/Basekick-Labs/msgpack/v6`

internal/impl/arc/output.go
```diff
@@ -0,0 +1,524 @@
...
+			order = append(order, measName)
+		}
+
+		rec.columns["time"] = append(rec.columns["time"], ts)
+		for k, v := range dataMap {
+			if k == o.timestampField || (o.timestampField == "" && k == "time") {
+				continue
+			}
+			rec.columns[k] = append(rec.columns[k], v)
+		}
```
mmatczuk, line 301: Bug: "time" column collision when `timestamp_field` is set to a value other than `"time"`.

L295 unconditionally appends the extracted timestamp to `rec.columns["time"]`. The skip condition at L297 only excludes `k == "time"` when `timestampField` is empty. When `timestampField` is e.g. `"ts"` and a message contains a field literally named `"time"`, that field passes the filter and gets appended to `rec.columns["time"]` via L300.

Result: the "time" column ends up with 2N entries while all other columns have N entries. Arc rejects the batch due to column length mismatch.

Fix: always skip `k == "time"` since the output reserves that column for the extracted timestamp:

```go
if k == o.timestampField || k == "time" {
```

internal/impl/arc/output.go
```diff
@@ -0,0 +1,524 @@
...
+		ts := o.extractTimestamp(dataMap)
+
+		// Remove timestamp field from fields if present
+		fields := make(map[string]any, len(dataMap))
+		for k, v := range dataMap {
+			if k == o.timestampField || (o.timestampField == "" && k == "time") {
+				continue
+			}
+			fields[k] = v
+		}
```
mmatczuk, line 354: Same "time" field collision issue as the columnar path. When `timestamp_field` is set to something other than `"time"` and a message has a `"time"` field, it leaks into `fields` as a regular data field alongside the extracted timestamp in `"t"`. Lower severity here (data duplication, not structural corruption), but same root cause and same fix:

```go
if k == o.timestampField || k == "time" {
```

mmatczuk (commented):
Bug: "time" field collision when `timestamp_field` is set to a non-"time" value. See inline comments.

mmatczuk:
Several `fmt.Errorf` calls use noun form instead of gerund form. Project pattern requires gerund-form prefixes. The file is inconsistent — `"creating request: %w"` and `"sending request: %w"` are correct, but others are not:
   - L217: `"create HTTP client: %w"` → `"creating HTTP client: %w"`
   - L251: `"msgpack encoding: %w"` → `"encoding msgpack: %w"`
   - L255: `"compression: %w"` → `"compressing: %w"`
   - L275, L335: `"measurement interpolation: %w"` → `"interpolating measurement: %w"`
   - L368: `"tags mapping: %w"` → `"mapping tags: %w"`
   - L373: `"tags mapping result: %w"` → `"reading tags mapping result: %w"`

mmatczuk:
Please fix lint and test

xe-nvdk:
@mmatczuk all addressed in the latest push:

**Error message forms** - Updated all 6 locations to gerund form:
- \`creating HTTP client\`
- \`encoding msgpack\`
- \`compressing\`
- \`interpolating measurement\`
- \`mapping tags\`
- \`reading tags mapping result\`

**"time" column collision (when timestamp_field is set to something other than "time")** - Good catch. Fixed in both columnar and row paths. The "time" column is now always reserved for the extracted timestamp; any literal "time" field in message data is skipped regardless of the \`timestamp_field\` configuration.

**Lint and tests** - Both clean:
- \`golangci-lint run internal/impl/arc/...\` - 0 issues
- \`go test ./internal/impl/arc/... -race\` - all 13 unit tests pass
- \`go test ./internal/impl/arc/... -tags integration\` - both integration tests pass against \`ghcr.io/basekick-labs/arc:latest\` (26.04.1), now including end-to-end query verification for row format as well

Ready for final look.

mmatczuk:
Could you run `task docs` and include the docs file for arc.

xe-nvdk:
@mmatczuk done, auto-generated `docs/modules/components/pages/outputs/arc.adoc` is included in the latest push.

mmatczuk:
Work moved to https://github.com/redpanda-data/connect/pull/4265

PR #4247, opened 2026-04-11
confluent: bump twmb/avro to v1.6.0, generalize avro reference hydration

twmb:
## Summary

Three changes:

### 1. Bump `github.com/twmb/avro` to v1.7.1

v1.5.0 decodes decimal logical types to `*big.Rat` instead of
`json.Number` for `*any` targets, matching hamba/avro and linkedin/goavro.
It also exports `RatFromBytes` for CustomType callbacks and fixes
EncodeJSON `json.Number` coercion bugs for int/long schemas.

### 2. Add decimal CustomType for preserve_logical_types path

v1.5.0's `*big.Rat` decimal type can't pass through `SetStructuredMut`
(which uses `json.Marshal`, and `*big.Rat` doesn't implement
`json.Marshaler`). Added a decimal `CustomType` in
`preserveLogicalTypeOpts` that converts the raw `[]byte` to
`json.Number` via `avro.RatFromBytes` at decode time.

### 3. Resolve avro references for arbitrary schema shapes

`resolveAvroReferences` previously only handled one specific schema
shape: a root that parses as a JSON array of reference names
(`["Ref1", "Ref2"]`), the Confluent Schema Registry convention for
"subject is a union of other subjects". Any other shape — a record
whose field type references another subject, an array whose items
reference another subject, etc. — produced the misleading error
`"parsing root schema as enum"`.

It also had two bugs:

- **Swallowed WalkReferences error.** On walk failure, the function
  returned `("", nil)`, silently dropping the error.
- **Transitive references not inlined.** If the root referenced Foo
  and Foo referenced Bar, Bar remained unresolved inside the inlined Foo.

Replaced the hand-rolled hydrator with a recursive walker
`hydrateAvroRefs` that traverses type positions throughout the schema
tree and inlines any reference name it finds. Each named type is
inlined at most once per walk; subsequent references are left as string
name references.

## Test plan

- [x] `go test ./internal/impl/confluent/...` passes
- [x] `go test ./internal/impl/avro/...` passes
- [x] Existing `TestAvroReferences` (legacy shape) still passes
- [x] 11 new `TestHydrateAvroRefs` subtests covering each supported shape

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -170,7 +170,7 @@ require (
 	github.com/timeplus-io/proton-go-driver/v2 v2.1.4
 	github.com/tmc/langchaingo v0.1.14
 	github.com/trinodb/trino-go-client v0.333.0
-	github.com/twmb/avro v1.3.4
+	github.com/twmb/avro v1.7.1
 	github.com/twmb/franz-go v1.20.7
 	github.com/twmb/franz-go/pkg/kadm v1.17.2
 	github.com/twmb/franz-go/pkg/kmsg v1.12.0
--- a/go.mod
+++ b/go.mod
@@ -169,7 +169,7 @@ require (
 	github.com/timeplus-io/proton-go-driver/v2 v2.1.4
 	github.com/tmc/langchaingo v0.1.14
 	github.com/trinodb/trino-go-client v0.333.0
-	github.com/twmb/avro v1.3.4
+	github.com/twmb/avro v1.6.0
 	github.com/twmb/franz-go v1.20.7
 	github.com/twmb/franz-go/pkg/kadm v1.17.2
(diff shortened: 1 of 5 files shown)


mmatczuk:
I wonder do we want to bump deps twice in one patch.

mmatczuk (approved):

twmb:
> I wonder do we want to bump deps twice in one patch.

I can squash this if preferable! I'm trying not to release so much; migrating iceberg-go also popped up a few small edge cases I needed to improve (and I only realized across a patch and then a minor)

twmb:
I'm fine with the commit message; bumped avro again, squashed the commit, will merge once passes given prior +1 and the only change is a dep bump again / squash.

PR #4262, opened 2026-04-15
mysql_cdc: ensure snapshot primary key order is consistent for composite primary keys

### LINKED ISSUE #4256: mysql_cdc Composite PK cursor wrong order (snapshot.go:197)
Composite PK cursor wrong order (snapshot.go:197)
Cursor values are pulled from a map[string]any via range iteration. Go map iteration is random, so values get bound in the wrong order → snapshot loops forever on composite PK tables. Fix: iterate usin

Bug 1 — Composite PK cursor wrong order snapshot.go

  Confirmed. In querySnapshotTable, the values for the WHERE-clause placeholders are built by iterating over *lastSeenPkVal which is a map[string]any:

  // snapshot.go – querySnapshotTable
  for _, pkCol := range *lastSeenPkVal {          // ← map iteration: random order
      lastSeenPkVals = append(lastSeenPkVals, pkCol)
      placeholders = append(placeholders, "?")
  }
  // Then:
  fmt.Sprintf("WHERE (%s) > (%s)", strings.Join(pk, ", "), strings.Join(placeholders, ", "))
  // column names come from pk []string (correct order)
  // bound VALUES come from map iteration (random order)

  Column names are in the right order (from the ordered pk []string), but the bound values are in random Go map order. For composite PKs this means e.g. WHERE (user_id, tenant_id) > (?, ?) gets values bound as (tenant_id_val, user_id_val) — wrong row boundary → cursor never advances →
  infinite loop.

  Your proposed fix is correct: iterate pk []string and do lastSeenPkVal[pkCol] to get values in declared key order.

josephwoodward:
Go's map do not guarantee ordering (the Go team even intentionally built randomisation into it to dissuade people from depending on its order), so switch to iterating ordered pk slice instead.

image

Closes: https://github.com/redpanda-data/connect/issues/4256

### DIFF
--- a/internal/impl/mysql/snapshot.go
+++ b/internal/impl/mysql/snapshot.go
@@ -196,8 +196,12 @@ func (s *Snapshot) querySnapshotTable(ctx context.Context, table string, pk []st
 
 	var lastSeenPkVals []any
 	var placeholders []string
-	for _, pkCol := range *lastSeenPkVal {
-		lastSeenPkVals = append(lastSeenPkVals, pkCol)
+	for _, pkCol := range pk {
+		val, ok := (*lastSeenPkVal)[pkCol]
+		if !ok {
+			return nil, fmt.Errorf("primary key column '%s' not found in last seen values", pkCol)
+		}
+		lastSeenPkVals = append(lastSeenPkVals, val)
 		placeholders = append(placeholders, "?")
 	}
 


internal/impl/mysql/snapshot.go
```diff
@@ -196,8 +196,8 @@ func (s *Snapshot) querySnapshotTable(ctx context.Context, table string, pk []st
 
 	var lastSeenPkVals []any
 	var placeholders []string
-	for _, pkCol := range *lastSeenPkVal {
-		lastSeenPkVals = append(lastSeenPkVals, pkCol)
+	for _, pkCol := range pk {
+		lastSeenPkVals = append(lastSeenPkVals, (*lastSeenPkVal)[pkCol])
```
squiidz: Should we check for nil on lastSeenPkVal before dereferencing?

josephwoodward: There's a nil check above (outside of the diff on line 188) that returns from the function early if nil.

josephwoodward: I have updated to put a check on the contents of the map though.

squiidz (approved):

mmatczuk (approved):
LGTM

PR #4265, opened 2026-04-16
arc: add output plugin for Arc columnar database

mmatczuk:


xe-nvdk:
Why doing this? Hard to track contributors, doesn't feel like a good practice. Can you clarify?

mmatczuk:
@xe-nvdk to streamline things on your end like failing CI and to be able to include it in a release. Cheers.

internal/impl/arc/output.go
```diff
@@ -0,0 +1,524 @@
...
+
+		rec, exists := groups[measName]
+		if !exists {
+			rec = &columnarRecord{columns: make(map[string][]any)}
+			groups[measName] = rec
+			order = append(order, measName)
+		}
+
+		rec.columns["time"] = append(rec.columns["time"], ts)
+		for k, v := range dataMap {
```
josephwoodward, line 296: Should we be concerned with map's random ordering here?

internal/impl/arc/output.go
```diff
@@ -0,0 +1,524 @@
...
+			if k == o.timestampField || k == "time" {
+				continue
```
josephwoodward, line 298: Should we document/log that any message field named "time" is dropped silently?

internal/impl/arc/output.go
```diff
@@ -0,0 +1,524 @@
...
+
+	switch o.format {
+	case "columnar":
+		payload, err = o.buildColumnarPayload(batch)
+	case "row":
+		payload, err = o.buildRowPayload(batch)
+	default:
+		return fmt.Errorf("unsupported format: %s", o.format)
+	}
+	if err != nil {
```
josephwoodward, line 242: Given we return on err here, would it not be better returning earlier (within the case)?

internal/impl/arc/output.go
```diff
@@ -0,0 +1,524 @@
...
+func (*arcOutput) Connect(_ context.Context) error {
+	return nil
+}
+
+func (o *arcOutput) WriteBatch(ctx context.Context, batch service.MessageBatch) error {
+	if len(batch) == 0 {
+		return nil
+	}
+
+	var payload any
```
josephwoodward, line 231: suggestion:
```go
var (
    payload any
    err error
)
```

internal/impl/arc/output.go
```diff
@@ -0,0 +1,524 @@
...
+	rows := make([]any, 0, len(batch))
+	for i, msg := range batch {
+		measName, err := measExec.TryString(i)
+		if err != nil {
+			return nil, fmt.Errorf("interpolating measurement: %w", err)
+		}
+
+		data, err := msg.AsStructuredMut()
+		if err != nil {
+			return nil, fmt.Errorf("message %d: %w", i, err)
```
josephwoodward, line 337: Not sure how helpful this error message prefix is (`message %d`), Is there a clearer prefix we could use?

josephwoodward (approved):

josephwoodward:
None of these comments are blockers and can be addressed post merge.

PR #4270, opened 2026-04-16
mongodb: fix ResumeWithSnapshot test expecting error on clean shutdown

mmatczuk:
stream.Run() returns nil on forced shutdown (benthos logs a warning
instead of returning an error). Changed RunAsyncWithErrors to RunAsync
and removed the unused helper method.

Fixes CON-419


mmatczuk:

PR #4271, opened 2026-04-16
mongodb: fix data race on ChangeStream between TryNext and ResumeToken

mmatczuk:
The ack callback called stream.ResumeToken() from a different goroutine
than the one running stream.TryNext(), racing on the ChangeStream's
internal resume token field. Use the value already resolved by the
checkpoint system instead, which is both race-free and semantically
correct.

Fixes CON-418


mmatczuk:

PR #4272, opened 2026-04-16
redpanda: fix flaky chaos tests after container restart

mmatczuk:
Pin the Kafka host port so it survives container stop/start and
kill/start cycles. Without pinning, Docker assigns a new random port on
restart, but the Redpanda config inside the container still advertises
the old port, causing Kafka clients to reconnect to a dead address.

Also replace fixed 2s sleep before chaos with require.Eventually that
waits for both producer and consumer to confirm message flow, and add
a TCP readiness check after every broker restart.

Extend redpandatest.Config with ExtraOpts to allow passing additional
testcontainers options, so the chaos tests reuse
StartSingleBrokerWithConfig instead of duplicating container setup.

Fixes CON-417

### DIFF
--- a/internal/impl/redpanda/integration_chaos_test.go
+++ b/internal/impl/redpanda/integration_chaos_test.go
@@ -19,13 +19,17 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"net"
 	"os"
+	"strconv"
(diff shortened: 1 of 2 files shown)


mmatczuk:

PR #4273, opened 2026-04-16
pulsar: disable StreamTestAtLeastOnceDelivery due to upstream data race

mmatczuk:
The benthos StreamTestAtLeastOnceDelivery test helper has an unsynchronized
concurrent map read/write at stream_test_definitions.go:571-584. A goroutine
iterates a shared map while another writes to it. Since this is in upstream
benthos code, disabling the test until the race is fixed upstream.

Fixes CON-415


mmatczuk:

PR #4274, opened 2026-04-16
pulsar: use native ARM image to fix container startup timeouts

mmatczuk:
Switch from apachepulsar/pulsar-standalone:2.8.3 (amd64-only) to
apachepulsar/pulsar:3.3.4 (multi-arch) so the container runs natively
on ARM instead of under x86 emulation. This eliminates the slow
startup that caused context deadline exceeded errors.

Fixes CON-387


mmatczuk:

PR #4275, opened 2026-04-16
redpanda_migrator: fix data race on groupsMigrator admin clients

mmatczuk:
groupsMigrator.enabled() reads srcAdm/dstAdm without synchronization
while onInputConnected/onOutputConnected write them concurrently under
Migrator.mu. Add a dedicated RWMutex to groupsMigrator and use setter
methods so writes are properly synchronized with the read in enabled().

Fixes CON-416


mmatczuk:

PR #4276, opened 2026-04-16
mssqlserver: fix flaky CDC checkpoint resume test under x86 emulation

mmatczuk:
Same CDC agent slowness issue as CON-414. The test inserts rows and
expects them via pure CDC streaming, but under emulation the capture
agent is too slow.

- Reduce rows per phase from 1000 to 100
- Wait for CDC to process inserts before starting the stream
- Remove fixed time.Sleep in favor of deterministic CDC polling

Fixes CON-382


mmatczuk:
Requires #4277

mmatczuk:

PR #4277, opened 2026-04-16
mssqlserver: fix flaky CDC streaming tests under x86 emulation

mmatczuk:
The CDC capture agent in SQL Server under x86 emulation on Apple Silicon
is too slow to process streaming inserts within the test timeout,
especially with 3 parallel subtests each spinning up their own container.

- Add WaitForCDCChanges helper that polls CDC change tables until the
  expected row count appears, replacing fixed time.Sleep(5s)
- Reduce streaming insert count from 3000 to 30 (snapshot already
  validates volume at 3000 rows)
- Wait for CDC to process streaming inserts before asserting

Fixes CON-414


mmatczuk:

PR #4279, opened 2026-04-16
integration: fix --unit flag skipping integration tests

mmatczuk:
Benthos CheckSkip guard requires test.run to be non-empty and match
the test name. With --unit we were omitting -run entirely, causing
all integration tests to skip. Use -run ^Test instead to match all
test functions while satisfying the guard.


mmatczuk:

PR #4280, opened 2026-04-16
confluent: fix data race on shutSig in schema registry encoder

mmatczuk:
The background goroutine started by newSchemaRegistryEncoder re-reads
s.shutSig on every loop iteration. When newSchemaRegistryEncoderFromConfig
replaces s.shutSig with a new signaller, there is a concurrent read/write
race. Capture shutSig in a local variable before launching the goroutine
so it never re-reads the struct field.

Fixes CON-412


mmatczuk:

PR #4281, opened 2026-04-16
postgresql: fix data race in Stream.Stop() on pgConn

mmatczuk:
Stop() was closing pgConn concurrently with streamMessages() still
using it for ReceiveMessage and SendStandbyStatusUpdate. Wait for the
stream goroutine to exit (HasStoppedChan) before closing the connection.

Fixes CON-406


mmatczuk:

PR #4282, opened 2026-04-16
sql: fix data race in TestIntegrationCosmosDB

mmatczuk:
Skip TestIntegrationCosmosDB when running with the race detector.

The gocosmos driver depends on gjrc v0.2.2 which has a data race in
GjrcResponse.Body: it uses broken double-checked locking where the
outer nil check reads a shared field without the mutex while a
background goroutine (spawned eagerly by gjrc.buildResponse) writes
it under the mutex. This is a bug in the third-party library that
affects every gocosmos REST call and cannot be worked around.

gjrc v0.2.2 is the latest available version as of this commit.

Fixes CON-409


mmatczuk:

PR #4283, opened 2026-04-16
redpanda: fix data race on TestingOnSetSubjectMode callback

mmatczuk:
importModeManager.Close() restores subject modes concurrently via
errgroup. Each goroutine calls setSubjectMode() which invokes the
TestingOnSetSubjectMode callback. Serialize callback invocations
with a dedicated mutex to prevent data races in test callbacks
that write to shared state.

Fixes CON-410


mmatczuk:

PR #4284, opened 2026-04-16
redis: increase pubsub channel buffer to prevent message drops

mmatczuk:
The go-redis pubsub.Channel() defaults to a buffer of 100 messages.
Under concurrent writes (e.g. max_in_flight=10 publishing batches of
100), the buffer overflows and messages are silently dropped since
Redis PubSub is fire-and-forget. Increase buffer to 2000.

Fixes CON-411


mmatczuk:

PR #4285, opened 2026-04-16
oracledb: handle LOB_TRIM events to establish LOB locator state

mmatczuk:
Oracle emits LOB_TRIM (op 11) before LOB_WRITE for certain LOB column
types (e.g. out-of-line VARCHAR MAX). Without processing LOB_TRIM, the
LOB locator state is never established and subsequent LOB_WRITE events
are dropped, causing the column to be silently absent from streaming
records.

Add OpLobTrim to the query filter and handle it alongside
OpSelectLobLocator since both share the same SQL structure.

Fixes CON-408


mmatczuk:

PR #4286, opened 2026-04-16
kafka: fix data race on msgChan in connectExplicitTopics

mmatczuk:
Partition consumer goroutines were spawned before k.msgChan was assigned,
causing a data race between the goroutine reads (line 104) and the
subsequent write (line 304). Move the assignment before the spawn loop.

Fixes CON-407


mmatczuk:

PR #4288, opened 2026-04-16
mssqlserver: guard Connect against concurrent calls

mmatczuk:
The benthos framework can call Connect() concurrently or re-enter it
before the background goroutine has stopped. Without a guard, two
goroutines share the same publisher and checkpoint tracker, corrupting
state and causing the streaming phase to hang indefinitely.

Add a mutex and a HasStoppedChan check so that a second Connect() is
a no-op while the previous goroutine is still running.

Fixes CON-389


mmatczuk:

PR #4289, opened 2026-04-16
cockroachdb: fix data race between Read() and closeConnection()

mmatczuk:
Read() copied c.rows under the mutex then called rows.Next() after
releasing it. Concurrently, the shutdown goroutine called
closeConnection() which called rows.Close() on the same pgx.Rows
object. pgx Rows is not safe for concurrent use.

Fix: hold dbMut for the entire Read() call (including the blocking
rows.Next()) and use SoftStopCtx to create a cancellable query context
that unblocks rows.Next() on shutdown. Extract closeQueryLocked() to
deduplicate cleanup between Read() and closeConnection().

Fixes CON-404


mmatczuk:

PR #4290, opened 2026-04-16
postgresql: filter context.Canceled from streamOut.Run in test goroutine

mmatczuk:
streamOut.Run correctly returns context.Canceled when the test context
is cancelled at shutdown. Use the same filtered error check pattern
already used elsewhere in the file.

Fixes CON-422


mmatczuk:

PR #4291, opened 2026-04-16
oracledb: use background context for deferred EndSession cleanup

mmatczuk:
The deferred EndSession call in ReadChanges used the parent context
which is already cancelled during shutdown. The go-ora driver's
ExecContext races internally when given a cancelled context:
StartContext spawns a goroutine that writes the session break flag
while the main goroutine reads it.

Fixes CON-421


mmatczuk:

PR #4294, opened 2026-04-17
parquet_encode: add default_timestamp_unit field

### LINKED ISSUE #3570: Add timestamp unit customization to parquet encode processor
Currently all parquet timestamps are serialized with nanosecond precision, we should expand the capabilities of the `parquet_encode` processor so that users can override this default in config.

ankit481:
Closes: https://github.com/redpanda-data/connect/issues/3570

## Summary

Adds a new `default_timestamp_unit` configuration field to the `parquet_encode` processor, accepting `NANOSECOND` (default, preserves existing behaviour), `MICROSECOND`, or `MILLISECOND`. The unit is applied to both the static schema path and the dynamic `schema_metadata` path (used by CDC inputs such as `mysql_cdc`).

Resolves the long-standing TODO referenced at #3570.

## Motivation

`parquet_encode` currently hardcodes `INT64 (TIMESTAMP(NANOS, true))` for every TIMESTAMP column, in both code paths (`processor_encode.go:159` and `:369`). This produces files that Apache Spark, Databricks, AWS Athena and DuckDB cannot read — they fail with:

```
[PARQUET_TYPE_ILLEGAL] Illegal Parquet type: INT64 (TIMESTAMP(NANOS,true))
```

Spark's legacy compatibility flags for this were removed in recent versions, so there is no reader-side workaround on modern Databricks runtimes. The dynamic-schema path is particularly impacted because `mysql_cdc` sets `schema_metadata` immutably, and Bloblang cannot mutate the schema node passed into `parquet_encode` — leaving affected users with no in-pipeline escape hatch.

Note: MySQL `DATETIME`/`TIMESTAMP` source precision is microseconds, so nanosecond encoding is structurally unnecessary for MySQL CDC pipelines — no data is lost by downcasting.

## Design

- Default remains `NANOSECOND` to preserve existing behaviour. Zero impact on existing users.
- New field is `Advanced()`-tagged and applies uniformly to both schema construction paths.
- The existing Iceberg integration (`internal/impl/iceberg/icebergx/parquet.go:146-148`) already uses `parquet.Timestamp(parquet.Microsecond)`, which this change brings parity for in the general encode processor.

## Test plan

- [x] Existing tests pass (`go test ./internal/impl/parquet/...`)
- [x] New `TestParquetEncodeTimestampUnit` verifies all three units on the static schema path, inspecting the encoded file's schema string.
- [x] New `TestParquetEncodeTimestampUnitDynamicSchema` verifies the unit flows through the `schema_metadata` path used by CDC inputs.
- [x] New `TestParquetEncodeTimestampUnitInvalid` verifies config validation rejects unknown values.
- [x] `go build ./...` succeeds.
- [x] `docs_gen` re-run; generated `.adoc` updated.
- [x] `CHANGELOG.md` updated under Unreleased.

### DIFF
--- a/internal/impl/parquet/processor_encode.go
+++ b/internal/impl/parquet/processor_encode.go
@@ -49,6 +49,12 @@ func parquetEncodeProcessorConfig() *service.ConfigSpec {
 				Default("DELTA_LENGTH_BYTE_ARRAY").
 				Advanced().
 				Version("4.11.0"),
+			service.NewStringEnumField("default_timestamp_unit",
+				"NANOSECOND", "MICROSECOND", "MILLISECOND",
+			).
+				Description("The precision used when encoding TIMESTAMP logical types. The default `NANOSECOND` matches historical behaviour, but `TIMESTAMP(NANOS)` is not readable by Apache Spark (Databricks), AWS Athena or DuckDB; set this to `MICROSECOND` (or `MILLISECOND`) when writing Parquet files intended for consumption by those engines.").
+				Default("NANOSECOND").
+				Advanced(),
 		).
 		Description(`
 This processor uses https://github.com/parquet-go/parquet-go[https://github.com/parquet-go/parquet-go^], which is itself experimental. Therefore changes could be made into how this processor functions outside of major version releases.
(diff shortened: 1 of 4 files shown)


josephwoodward (approved):

internal/impl/parquet/processor_encode.go
```diff
@@ -49,6 +49,12 @@ func parquetEncodeProcessorConfig() *service.ConfigSpec {
 				Default("DELTA_LENGTH_BYTE_ARRAY").
 				Advanced().
 				Version("4.11.0"),
+			service.NewStringEnumField("default_timestamp_unit",
+				"NANOSECOND", "MICROSECOND", "MILLISECOND",
+			).
+				Description("The precision used when encoding TIMESTAMP logical types. The default `NANOSECOND` matches historical behaviour, but `TIMESTAMP(NANOS)` is not readable by Apache Spark (Databricks), AWS Athena or DuckDB; set this to `MICROSECOND` (or `MILLISECOND`) when writing Parquet files intended for consumption by those engines.").
+				Default("NANOSECOND").
+				Advanced(),
```
josephwoodward: It would be good to add a `Version("4.89.0")` after this so it's clear in the documentation which version it was added in.

josephwoodward (approved):

PR #4296, opened 2026-04-20
cassandra: avoid gocql host-lookup against docker internal IP

mmatczuk:
## Commits

- avoid gocql host-lookup against docker internal IP

## Jira

- CON-451


Jeffail (approved):
LGTM — clean test-only fix. Uses the already-exposed `disable_initial_host_lookup` component option, which is the canonical workaround for single-node containerised Cassandra under testcontainers-go (the driver would otherwise chase `system.peers` to unreachable Docker-internal IPs). No production behaviour changed.

PR #4297, opened 2026-04-20
cockroachdb: fix data race in changefeed Read/closeConnection

mmatczuk:
## Commits

- fix data race between Read() and closeConnection()
- tolerate context.Canceled in test stream goroutine

## Jira

- CON-404
- CON-447

### DIFF
--- a/internal/impl/cockroachdb/input_changefeed.go
+++ b/internal/impl/cockroachdb/input_changefeed.go
@@ -76,6 +76,10 @@ type crdbChangefeedInput struct {
 	rows     pgx.Rows
 	dbMut    sync.Mutex
 
+	// queryCancel cancels the active changefeed query context, unblocking a
+	// blocking rows.Next() call in Read(). Protected by dbMut.
+	queryCancel context.CancelFunc
+
 	res     *service.Resources
 	logger  *service.Logger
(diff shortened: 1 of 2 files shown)


Jeffail (approved):
Approving with the title now reflecting the headline change. Commit `0496d37b1` is a real production fix — `Read()` previously released `dbMut` before calling `rows.Next()`, racing with `rows.Close()` from the shutdown path; pgx rows are explicitly not safe for concurrent use. The fix correctly holds `dbMut` across `Next()` and uses `shutdown.SoftStopCtx` to cancel the in-flight query on shutdown.

Two small non-blocking observations:

1. The `context.Canceled` tolerance in commit `4f6d2ef3c` is applied only to the second `streamOut.Run` goroutine in `TestIntegrationCRDB`. The first (around line 117) still does `_ = streamOut.Run(t.Context())` and silently swallows all errors — asymmetric, but harmless in practice.
2. The pre-fix `closeConnection` logged `rows.Err()` as a warning when non-nil. That diagnostic is now dropped on the soft-stop branch in `Read()` because `service.ErrNotConnected` is returned before `err` is consulted. Minor loss of log fidelity on shutdown.

Neither is a merge blocker.

PR #4298, opened 2026-04-20
confluent: fix data race on shutSig in schema registry encoder

mmatczuk:
## Commits

- fix data race on shutSig in schema registry encoder

## Jira

- CON-412

### DIFF
--- a/internal/impl/confluent/processor_schema_registry_encode.go
+++ b/internal/impl/confluent/processor_schema_registry_encode.go
@@ -303,12 +303,13 @@ func newSchemaRegistryEncoder(
 		return nil, err
 	}
 
+	shutSig := s.shutSig
 	go func() {
 		for {
 			select {
 			case <-time.After(schemaRefreshTicker):
 				s.refreshEncoders()
-			case <-s.shutSig.SoftStopChan():
+			case <-shutSig.SoftStopChan():
 				return
 			}
 		}


Jeffail (approved):
Approving — the race is real and the fix is minimal and correct. The refresh goroutine in `newSchemaRegistryEncoder` was reading `s.shutSig` on every loop iteration while `newSchemaRegistryEncoderFromConfig` reassigned the field (line 267) after triggering soft-stop on the prior signaller. Capturing into a local before goroutine spawn closes the race cleanly, and the captured signaller is the one that gets `TriggerSoftStop()` called on it, so the goroutine still terminates as expected.

Two non-blocking follow-up notes:

1. The metadata-mode goroutine at line 268-277 still reads `s.shutSig` from inside the closure (line 273). Safe today since no further reassignment occurs, but inconsistent with the fix style — applying the same local-capture there would future-proof against the same bug class.
2. The underlying construction shape (`newSchemaRegistryEncoder` starts a goroutine, then `newSchemaRegistryEncoderFromConfig` conditionally tears it down and starts a different one with a reassigned signaller) is what creates the race in the first place. A future refactor that defers goroutine launch until the mode is known, or threads the mode into the constructor so only one loop is ever started, would make this class of bug structurally impossible. Worth a follow-up but well out of scope here.

PR #4300, opened 2026-04-20
integration: add 10 missing packages and skip field for excluded ones

mmatczuk:
## Commits

- add 10 missing packages and skip field for excluded ones


mmatczuk:
Dropping — does not belong in this batch.

PR #4302, opened 2026-04-20
migrator: tolerate context.Canceled in consumer group test stream goroutine

mmatczuk:
## Commits

- tolerate context.Canceled in consumer group test stream goroutine

## Jira

- CON-446

### DIFF
--- a/internal/impl/redpanda/migrator/integration_test.go
+++ b/internal/impl/redpanda/migrator/integration_test.go
@@ -420,7 +420,9 @@ logger:
 		go func() {
 			ctx, cancel := context.WithTimeout(t.Context(), redpandaTestWaitTimeout)
 			defer cancel()
-			require.NoError(t, stream.Run(ctx))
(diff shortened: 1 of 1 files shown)


Jeffail (approved):
Approving — correct narrow fix. `context.Canceled` is genuinely expected here: the goroutine's deferred `cancel()` races `stream.StopWithin` on shutdown and either ordering is benign. The `require.NoError` → `t.Errorf` swap also fixes the latent `FailNow`-from-non-test-goroutine bug per the `testing` package contract. Matches the dominant repo idiom.

Worth a follow-up (out of scope for this PR): the `RunStreamInBackground` helper at `internal/impl/snowflake/integration_test.go:143` already implements this exact pattern (cleanup-driven cancel + wg.Wait + Canceled tolerance + non-FailNow assertion). Promoting it to a shared test helper would let us migrate the remaining `require.NoError(t, stream.Run(...))`-in-goroutine call sites across this file (e.g. L633, L688, L789, L901) and the sibling packages in one sweep instead of landing per-call-site one-line PRs.

PR #4321, opened 2026-04-21
fix for oracle non-leading zero

jamesjoegraham:


josephwoodward:
Thank you for the PR @jamesjoegraham, I'll close this one (and cherry-pick your commit) to move it over to https://github.com/redpanda-data/connect/pull/4322 as it'll get the Claudebot review as part of the PR process, I hope that's okay. I've also updated the integration test to verify behaviour is consistent between snapshotting and streaming as well as shared the normalisation function with the schema code as well.

PR #4326, opened 2026-04-23
postgres_cdc: continuous table discovery with mid-stream snapshot

binlogreader:
## Summary

Adds an opt-in `auto_discover_tables` flag to the `postgres_cdc` input. When enabled, a discovery loop polls the configured `schema` for new tables on `discovery_interval` (default `60s`) and incorporates them into the running CDC pipeline **without restarting the replication slot**.

This addresses the operational pain of having to restart pipelines (and redeploy config) every time a new table is added at the source — a common scenario at scale (we run 30+ source DBs) and a feature most competing CDC products (Debezium incremental snapshot, Fivetran, Airbyte, Estuary) already support.

> **Design tracking issue:** #4327 was filed alongside this PR per CONTRIBUTING.md §3.1.1. The issue captures the design and open questions for maintainer scoping. Happy to close this PR and work from the issue alone if that is preferred — the code is here for concreteness, not to bypass scoping.

## Design

For each newly-found table T, the discoverer:

1. `ALTER PUBLICATION pglog_stream_<slot> ADD TABLE T` — captures all future WAL events for T.
2. Opens a **temporary** logical replication slot with `EXPORT_SNAPSHOT` to anchor the snapshot at a known LSN (`L_snap`).
3. Snapshots T at the exported snapshot using a single reader (small/new tables don't need keyspace splitting).
4. Records `tableWatermark[T] = L_snap` on the parent stream.
5. Drops the temporary slot.
6. Registers T with the monitor and adds it to the live tables set.

A single branch in `processChange` suppresses WAL events for T where `event_LSN <= tableWatermark[T]` — these are duplicates already in the snapshot. The branch is a no-op when the watermark map is empty, so existing behavior is unchanged when the flag is off.

The main streaming slot is **never** restarted; LSN advances continuously. The implementation reuses existing primitives:

- `CreatePublication` already issues `ALTER PUBLICATION ... ADD TABLE` (pglogrepl.go:365)
- `EXPORT_SNAPSHOT` + `SET TRANSACTION SNAPSHOT` is the same mechanism the initial snapshotter uses (logical_stream.go:227)
- `scanTableRange` is reused for the per-table snapshot

## Why this shape

- **Additive.** Default `false`, no behavior change for existing pipelines.
- **No superuser required.** Avoids the `FOR ALL TABLES` trap on RDS.
- **Crash-safe.** If we crash between ALTER and snapshot completion, on restart the discovery loop re-detects T (still missing locally) and repeats steps 2–6 idempotently. ALTER PUBLICATION on an already-published table is a no-op.
- **Monotonic LSNs make dedup trivially correct.** No epoch, no replay queue, no buffering — one map read + one comparison on the hot path.
- **Tables without a primary key** are skipped with a warning and rolled out of the publication, so we never silently emit WAL for an unsnapshotted table.

## Open design questions for maintainers

1. **Config surface:** is `auto_discover_tables: bool` + `discovery_interval: duration` the right shape, or do you prefer a nested object like `auto_discovery: { enabled: true, interval: 60s }`?
2. **Tables without primary keys:** I currently skip + roll back the publication change. Alternatives: fall back to a sequential scan with no PK filtering, or fail loudly. Preference?
3. **Discovery scope:** I poll `pg_tables` for the configured schema only. Should we support cross-schema discovery via a list?
4. **Watermark persistence:** in-memory only. On restart the re-added table is also re-snapshotted (idempotent), but if persistence in the checkpoint cache is preferred I can add it.
5. **Metrics:** I did not add new metrics yet. Candidates: `postgres_discovered_tables_total` gauge, `postgres_discovery_lag_seconds`. Want them in this PR or a follow-up?
6. **Enterprise license gate:** `postgres_cdc` already gates on `license.CheckRunningEnterprise`. Auto-discovery inherits that gate at the input level. Is that the intended posture?

## Files changed (~760 LOC, mostly tests + comments)

- `input_pg_stream.go` — two new config fields, plumbing
- `pglogicalstream/config.go` — two new struct fields
- `pglogicalstream/logical_stream.go` — `tablesMu`, `tableWatermark` map, `installDiscoveredTable`, `shouldSuppressForWatermark`, dedup branch in `processChange`, `maybeStartDiscoverer`, lifecycle wiring in `Stop`
- `pglogicalstream/discovery.go` (new) — discoverer struct, polling loop, `addTable` orchestrator, `tempSlotName`
- `pglogicalstream/monitor.go` — `RegisterTable` + `sync.RWMutex` around mutating maps
- `pglogicalstream/discovery_test.go` (new) — unit tests for the watermark + slot-name helpers
- `integration_test.go` — `TestIntegrationPostgresAutoDiscoverTables` end-to-end test
- `CHANGELOG.md`

## Test plan

- [x] Unit tests pass: `go test ./internal/impl/postgresql/pglogicalstream/ -short`
- [x] Whole-repo build clean: `go build ./...`
- [x] `go vet ./internal/impl/postgresql/...` clean
- [x] **New integration test passes** against PG 16 via testcontainers: `TestIntegrationPostgresAutoDiscoverTables` — 4.18s. Covers: mid-stream `CREATE TABLE`, snapshot of pre-discovery rows emitted as `read`, post-discovery inserts arrive as `insert` exactly once, no duplicate reads (watermark dedup works).
- [x] **No regression in existing integration suite**: `TestIntegrationPostgresNoTxnMarkers`, `TestIntegrationPostgresIncludeTxnMarkers`, `TestIntegrationPostgresMetadata`, `TestIntegrationTOASTValues`, `TestIntegrationSnapshotConsistency`, `TestIntegrationSnapshotParallel` all pass (46.9s total, exit 0).
- [ ] Manual test against an RDS instance (no superuser required) — not yet done

### Bug caught during integration testing (fixed)

The first integration test run surfaced a real issue: primary-key lookup for discovered tables was routing through `s.pgConn` (the replication connection, concurrently owned by the streaming goroutine), causing `conn busy` errors. Fixed in commit `c4f0703` by moving the PK query to the discoverer's own SQL connection. All tests now green.

## Out of scope (separate issues if there is interest)

- Periodic full refresh of existing tables (uses the same watermark primitive but raises distinct upsert-vs-append semantics downstream)
- Auto-removal of dropped tables
- Generalizing the watermark abstraction to `mysql_cdc`, `mssqlserver_cdc`, `oracledb_cdc` (the same primitive applies; would be a follow-up after this lands and patterns are agreed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 - parquet_encode: Added `default_timestamp_unit` field (values `NANOSECOND`, `MICROSECOND`, `MILLISECOND`) controlling the precision of TIMESTAMP logical types. Default remains `NANOSECOND` for backwards compatibility. Use `MICROSECOND` when writing files for Apache Spark/Databricks, AWS Athena or DuckDB, which do not support `TIMESTAMP(NANOS)`. ([#3570](https://github.com/redpanda-data/connect/issues/3570))
+- postgres_cdc: Added `auto_discover_tables` and `discovery_interval` fields to enable continuous mid-stream discovery of new tables. When enabled, newly-created tables in the configured schema are added to the publication, snapshotted, and included in CDC without restarting the replication slot. Per-table snapshot LSN watermarks deduplicate WAL events that overlap with the snapshot window.
 
 ## 4.88.0 - 2026-04-16
 
--- a/internal/impl/postgresql/input_pg_stream.go
+++ b/internal/impl/postgresql/input_pg_stream.go
@@ -43,6 +43,8 @@ const (
 	fieldMaxParallelSnapshotTables = "max_parallel_snapshot_tables"
 	fieldUnchangedToastValue       = "unchanged_toast_value"
 	fieldHeartbeatInterval         = "heartbeat_interval"
+	fieldAutoDiscoverTables        = "auto_discover_tables"
+	fieldDiscoveryInterval         = "discovery_interval"
 	fieldAWSIAMAuth                = "aws"
 	// FieldAWSIAMAuthEnabled enabled field.
 	FieldAWSIAMAuthEnabled = "enabled"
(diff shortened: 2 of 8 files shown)


binlogreader:
Closing this PR after discussing with @josephwoodward

PR #4340, opened 2026-04-23
Iceberg: fix numeric type precision

squiidz:


internal/impl/iceberg/icebergx/parquet.go
```diff
@@ -148,6 +149,9 @@ func icebergTypeToParquet(t iceberg.Type) (parquet.Node, error) {
 		return parquet.Timestamp(parquet.Microsecond), nil
 	case iceberg.UUIDType:
 		return parquet.UUID(), nil
+	case iceberg.DecimalType:
+		byteWidth := int(math.Ceil((math.Log10(2) + float64(t.Precision())) / math.Log10(256)))
```
josephwoodward: Is it worth using `decimalByteWidth` here? They appear the same.

internal/impl/iceberg/shredder/shredder.go
```diff
@@ -457,3 +456,156 @@ func convertLeafValue(value any, typ iceberg.Type) (parquet.Value, error) {
...
+// Decimal values are stored as big-endian two's complement byte arrays.
+func convertToDecimal(value any, dt iceberg.DecimalType) (parquet.Value, error) {
+	scale := dt.Scale()
+
+	var (
+		unscaled *big.Int
+		err      error
+	)
+
+	switch v := value.(type) {
```
josephwoodward, line 470: How exhaustive does this need to be? What about `int8`, `int16`, `uint`, `uint8`, `uint16`, `uint32` and `uint64`?

Jeffail:
Hi @squiidz, thank you for the work here — flagging one concern on the `json.Number` → `Int64` inference path for discussion.

The doc comment at `internal/impl/iceberg/type_inference.go` correctly identifies the schema-evolution hazard: when the first value seen for a field is integer-parseable, the column is locked to `long`, and subsequent fractional values are coerced via `bloblang.ValueAsInt64`. Tracing that path into benthos, `value.IGetInt` falls back to `Float64()` and casts to `int64` on failure — so `json.Number("100.5")` silently lands as `100` in Parquet, with no error or warning surfaced to the operator.

A few options worth considering:

1. Default `json.Number` → `Float64Type` to preserve the pre-PR semantics for that specific case. This still benefits from the new `int` vs. `long` distinction for typed Go values, which is the main motivation of the PR, and avoids silent truncation as a default.
2. Detect mixed integer/fractional values within an inference window and prefer `double` whenever ambiguity is observed.
3. Keep the current inference but make the shredder fail loudly (or at least warn) when a fractional `json.Number` is coerced into an integer column.

My preference would be option 1 on safety grounds, but happy to defer to your judgement — you may have context on the typical caller shape that I do not.

Cow fact: cows form close friendships with specific herd members and become measurably stressed when separated from their preferred companions.

Jeffail:
Hi @squiidz, one further item on the change to `resolveFromSchemaMetadata` in `internal/impl/iceberg/type_resolver.go`.

The previous behaviour returned an explicit error when `schema_metadata` was configured but the key was absent on a message; the new behaviour logs at `Debugf` and falls back to inference. This is a meaningful change in failure semantics — a user who has opted in to `schema_metadata` has typically done so precisely because they want the metadata to drive types. A single message arriving without the key (e.g. due to an upstream bug, a transient routing issue, or operator misconfiguration) will now silently produce a column with an inferred type rather than the declared one, and that column type is then locked for the remainder of the table's life.

Two options that might address this:

1. Preserve the error by default and require users to opt in to the fallback behaviour explicitly via a config flag (something like `schema_metadata_fallback: infer`). This keeps the strict guarantee for existing users.
2. Keep the fallback as the default but raise the log level to `Warnf` so it surfaces at default verbosity rather than being effectively invisible.

I would lean toward option 1 on safety grounds — silently writing the wrong column type is hard to recover from once a table is established — but happy to defer if there is a use case I am missing.

Cow fact: cows have a near-360-degree panoramic field of vision, with only a small blind spot directly behind them.

Jeffail (approved):

PR #4358, opened 2026-04-28
schema: adopt Decimal and BigDecimal common types across sources and converters

Jeffail:
## Summary

- Threads benthos's new `Decimal` and `BigDecimal` common-schema types end-to-end across five CDC sources (postgres, mysql, mssqlserver, oracledb, mongodb) and four format converters (iceberg, parquet, avro, json schema), with values normalised to canonical decimal strings via a new `internal/sqlutil` canonicaliser.
- Adds Decimal logical-type detection in the Avro reverse reader and `*big.Rat` → canonical-string normalisation in the `schema_registry_decode` `store_schema_metadata` path so downstream metadata-driven encoders see consistent inputs.
- Extracts shared Parquet decimal-byte helpers into `internal/impl/parquet/parquetdecimal`, dedupes against `iceberg/icebergx`, and tightens the iceberg shredder to prefer `schema.ParseDecimal` for canonical inputs.

## Note for reviewers

The branch carries a temporary `replace` directive in `go.mod` pointing at a local benthos checkout while the matching benthos release is prepared. A follow-up commit will flip that directive to a tagged version once benthos publishes one — please flag if there's a preferred form for the interim.

## Test plan

- [x] `task fmt` clean.
- [x] `task lint` clean for every touched package (the pre-existing `internal/impl/tigerbeetle` typecheck failure is unrelated).
- [x] `task test:unit` green across iceberg, parquet, parquetdecimal, confluent, postgresql, mysql, mssqlserver, mongodb, oracledb, and the new `internal/sqlutil`.
- [x] New unit-test coverage for the `sqlutil` and `parquetdecimal` helpers, every converter Decimal/BigDecimal case, the Avro reverse reader's logical-type detection, the `normaliseAvroDecimals` walker (including tagged-union dispatch), the Mongo `Decimal128` walker (including scientific-notation inputs), the Postgres `atttypmod` parser, and the Postgres value-side decoder.
- [x] Parquet end-to-end smoke test exercising all three precision buckets (Int32, Int64, FixedLenByteArray) plus BigDecimal rejection.
- [x] Integration-test assertions in postgres, oracledb, and mssqlserver updated to match the new schema metadata and canonical-string value shape; mysql and mongodb integration tests already aligned.
- [x] Run integration suites (`task test:integration`) once Docker is available in the reviewer's environment, particularly `oracledb` and `mssqlserver` where the value-shape change is most visible.

## CHANGELOG

Per-source `Added` and `Changed` entries are included under `## Unreleased`.


internal/impl/confluent/ecs_avro.go
```diff
@@ -17,46 +17,73 @@ package confluent
...
+		return int32(x), nil
+	case json.Number:
+		n, err := x.Int64()
+		if err != nil {
+			return 0, fmt.Errorf("not an integer: %v", x)
+		}
+		return int32(n), nil
+	case nil:
+		return 0, fmt.Errorf("missing")
+	default:
```
josephwoodward, line 40: Avro's `scale` is [optional](https://avro.apache.org/docs/1.11.1/specification/#:~:text=scale%2C%20a%20JSON%20integer%20representing%20the%20scale%20(optional).%20If%20not%20specified%20the%20scale%20is%200.) isn't it but would fall into this default case?

internal/sqlutil/decimal.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// Fast path: input already matches the canonical-string contract at the
+	// declared scale. Short-circuits before any allocation when possible.
+	if unscaled, err := params.Parse(text); err == nil {
+		return params.Format(unscaled)
+	}
+
+	// Slower path: permit extended forms (scientific notation, leading +,
+	// fewer fractional digits than scale, etc.).
+	bf, _, err := new(big.Float).SetPrec(256).Parse(text, 10)
```
josephwoodward: The big.Float fallback adds ±0.5 before truncating to get an integer. For an input like 1.56789 against a NUMBER(10,2) column, it silently produces "1.57" rather than rejecting the value.

josephwoodward: Adding the following test case to `TestCanonicaliseDecimal` verifies this:

```go
{name: "more fractional digits than scale should error not round", input: "1.56789", precision: 10, scale: 2, wantErr: true},
```

go.mod
```diff
@@ -4,6 +4,10 @@ go 1.26.2
 
 replace github.com/99designs/keyring => github.com/Jeffail/keyring v1.2.3
 
+// TEMPORARY: replace benthos with local checkout while decimal common-type adoption is in flight.
+// Flip back to a tagged version once benthos releases with both Decimal and BigDecimal types.
+replace github.com/redpanda-data/benthos/v4 => /Users/ash/src/ai/benthos-schema-decimal-types
```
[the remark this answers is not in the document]
Jeffail: Addressed in ed9b74d6c — the local-path replace directive has been removed and benthos is now pinned to a pseudo-version of main HEAD (`v4.71.1-0.20260428154615-ddb2b041e022`) so the module graph resolves on every machine, including CI. The corresponding `go.sum` entries have been restored. The plan is to flip the pseudo-version to a tagged benthos release once one is cut, in a follow-up commit.

internal/sqlutil/decimal.go
```diff
@@ -0,0 +1,101 @@
+// Copyright 2026 Redpanda Data, Inc.
+//
+// Licensed as a Redpanda Enterprise file under the Redpanda Community
+// License (the "License"); you may not use this file except in compliance with
+// the License. You may obtain a copy of the License at
+//
+// https://github.com/redpanda-data/connect/v4/blob/main/licenses/rcl.md
```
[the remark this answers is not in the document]
Jeffail: Fixed in 8f2cc5b9f — both `internal/sqlutil/decimal.go:7` and `internal/sqlutil/decimal_test.go:7` now use the headless URL form (`https://github.com/redpanda-data/connect/blob/main/licenses/rcl.md`) matching the rest of the repo's RCL headers.

internal/impl/confluent/ecs_avro.go
Jeffail: Good catch — fixed in 8f2cc5b9f. The decimal logical-type handler now defaults `scale` to 0 when the field is absent, matching the Avro spec. `precision` remains required. Added `TestEcsAvroFromBytesDecimalScaleDefaultsToZero` covering the omitted-scale case.

internal/sqlutil/decimal.go
Jeffail: Good catch — fixed in 8f2cc5b9f. The `big.Float` fallback has been replaced with a `big.Rat` parse and an exact `IsInt` check on `value × 10^scale`. `"1.56789"` against `NUMBER(10, 2)` now returns `decimal "1.56789" has more fractional digits than the column's scale 2` rather than silently rounding to `"1.57"`. Scientific notation, leading `+`, and shorter-than-scale fractional inputs still canonicalise as before. Test cases covering both rejection paths (`"1.56789"` at `(10, 2)` and `"1.5e-5"` at `(10, 2)`) added.

josephwoodward (approved):

squiidz (approved):

josephwoodward:
Failing integration test appears to be due to timeout, ran locally to verify.

PR #4372, opened 2026-04-30
aws_s3: Don't set ACL by default since ACLs are disabled by default in S3

### LINKED ISSUE #4371: aws_s3: Output should not set ACL by default
Relatively recently in version 4.56.0, a [change](https://github.com/redpanda-data/connect/pull/3456) was added to add the `object_canned_acl` config to `aws_s3` and set the default to `private`.

Since 2023 the default setting for new S3 buckets has been to set the "bucket owner enforced" policy, which disables ACLs altogether. Thus I think the default behavior should be to not set an ACL or have the default be `bucket-owner-full-control`. 

Setting an ACL means that the upload will break in certain situations. I believe that the 'private' ACL works for buckets in the same account since the API basically ignores it. However for buckets in other accounts this fails. This change lead to an unfortunate situation where we were suddenly unable to push files to a client's bucket because we had upgraded the version of redpanda-connect from <4.56 to 4.86. 

From the doc string on the UploadObjectInput ACL field:

> // If the bucket that you're uploading objects to uses the bucket owner enforced
	// setting for S3 Object Ownership, ACLs are disabled and no longer affect
	// permissions. Buckets that use this setting only accept PUT requests that don't
	// specify an ACL or PUT requests that specify bucket owner full control ACLs, such
	// as the bucket-owner-full-control canned ACL or an equivalent form of this ACL
	// expressed in the XML format. PUT requests that contain other ACLs (for example,
	// custom grants to certain Amazon Web Services accounts) fail and return a 400
	// error with the error code AccessControlListNotSupported . For more information,
	// see [Controlling ownership of objects and disabling ACLs]in the Amazon S3 User Guide.

ttreptow:
Setting an ACL throws an error when uploading to a bucket with ACLs disabled (default since 2023) when the bucket is in another account. The former default 'private' is accepted by buckets in the same account but is a no-op.

This change should not affect cases where users are writing to a bucket with ACLs enabled, since the default ACL applied by AWS is 'private', which matches the expected behavior when `object_canned_acl` is not specified. If a user is writing to a bucket with ACLs enabled and is specifying a different canned ACL, that will continue to work with this change.

Fixes #4371

### DIFF
--- a/docs/modules/components/pages/outputs/aws_s3.adoc
+++ b/docs/modules/components/pages/outputs/aws_s3.adoc
@@ -81,7 +81,7 @@ output:
     force_path_style_urls: false
     max_in_flight: 64
     timeout: 5s
-    object_canned_acl: private
+    object_canned_acl: ""
     batching:
       count: 0
       byte_size: 0
@@ -387,7 +387,7 @@ The object canned ACL value.
 
 *Type*: `string`
 
-*Default*: `"private"`
+*Default*: `""`
 
 Options:
 `private`
--- a/internal/impl/aws/s3/output.go
+++ b/internal/impl/aws/s3/output.go
@@ -313,7 +313,7 @@ output:
 					}
 				})...).
 				Description("The object canned ACL value.").
-				Default(string(types.ObjectCannedACLPrivate)).
+				Default("").
 				Advanced(),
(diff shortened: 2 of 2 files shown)


mmatczuk (approved):
LGTM

PR #4380, opened 2026-05-01
avro: fix schema metadata extraction failure for nested records in nullable unions

josephwoodward:


internal/impl/confluent/ecs_avro.go
```diff
@@ -318,7 +322,9 @@ func ecsAvroFromAnyMap(cfg ecsAvroConfig, as map[string]any) (schema.Common, err
 	case schema.Object:
 		fields, exists := as["fields"].([]any)
 		if !exists {
-			return schema.Common{}, fmt.Errorf("expected `fields` field of type array, got %T", as["fields"])
+			// nil/absent fields: back-reference or schema with "fields":null.
+			// Return an opaque record rather than failing schema extraction.
+			return c, nil
 		}
```
Jeffail: Minor tightening: the current handler treats any non-`[]any` value as opaque, which silently swallows genuinely malformed input (e.g. `"fields": "oops"`). Narrowing the lenient path to nil/missing keeps the test case covered while preserving an error for wrong types. Also adjusted the comment, as Avro back-references are name strings rather than this shape.

```suggestion
		raw, present := as["fields"]
		if !present || raw == nil {
			// Some generators emit records without a fields key (or with
			// fields:null). Return an opaque record rather than failing
			// the entire schema extraction.
			return c, nil
		}
		fields, ok := raw.([]any)
		if !ok {
			return schema.Common{}, fmt.Errorf("expected `fields` field of type array, got %T", raw)
		}
```

Jeffail (approved):

PR #4382, opened 2026-05-01
fix(iceberg): apply schema_metadata for nil fields during table creation

jrivany:
Issue:
Using Iceberg with AWS Glue and an explicit schema from the registry, some columns were not created initially if their first observed value was null, even though they were present in the only schema version. Those columns were added later once non-null values appeared.

Changes:

* move the nil-skip check to run after schema_metadata override
* preserve existing behavior for fields with neither inferred type nor metadata type
* add a regression test covering metadata-defined type + nil record value


josephwoodward (approved):

josephwoodward:
Closing in favour of https://github.com/redpanda-data/connect/pull/4383

PR #4384, opened 2026-05-01
plugins: require cloud_unsupported_reason and reject placeholder versions

twmb:
Adds a cloud_unsupported_reason column to internal/plugins/info.csv so that every connector excluded from the cloud distribution carries a one-line rationale alongside the cloud=n flag, and drops the version column entirely (157 of 320 rows carried the 0.0.0 placeholder, and nothing in the runtime, docs, or cloud bundling consumed the value).

A new TestPluginCloudEnablement in internal/plugins/alltest enforces the invariant going forward: every plugin must either be cloud-enabled or carry a non-empty cloud_unsupported_reason.

The reasons for connectors deliberately gated on security review are bucketed into "security: arbitrary code execution", "security: local filesystem access", "security: raw socket access", "security: opens a listener unreachable from cloud", and "security: pipeline stdio not exposed in cloud". The remaining reasons cover deprecated upstream protocols, managed metrics/tracing in cloud, missing certification, and a few connector-specific limitations.

### DIFF
--- a/internal/plugins/alltest/plugins_test.go
+++ b/internal/plugins/alltest/plugins_test.go
@@ -67,3 +67,20 @@ func TestAllPluginsInInfoCSV(t *testing.T) {
 		check(name, plugins.TypeTracer)
 	})
 }
+
+// TestPluginCloudEnablement requires each plugin to either be enabled in the
+// cloud distribution or to document why it is not, via the
+// cloud_unsupported_reason column in internal/plugins/info.csv.
+func TestPluginCloudEnablement(t *testing.T) {
+	for key, info := range plugins.BaseInfo {
+		if info.Cloud {
+			if info.CloudUnsupportedReason != "" {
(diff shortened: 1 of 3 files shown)


mmatczuk:
I'd consider completely redoing this file or removing it completely. Having said that such a file comes in handy as a single source of truth. The problem is that data in component registration and cvs may and do diverge.

JakeSCahill:
having this data in `rpk connect list` would be useful too. We'd also like a `deprecated_reason` to be present when fields or connectors are deprecated.

twmb:
@JakeSCahill we'll follow up on this — exposing `cloud_unsupported_reason` (and a `deprecated_reason`) via `rpk connect list` requires a corresponding change in benthos to add the fields to `service.ConfigSpec` so they flow through the schema and into the `list` output. Tracking it as a follow-up rather than blocking this PR.

twmb:
(pr desc updated)

twmb:
merging, no changes since prior +1

josephwoodward (approved):

PR #4385, opened 2026-05-01
bigquery: add schema resolver, evolution, and table name sanitization

squiidz:
Schema resolution + evolution
- Add schemaResolver that fetches table metadata, builds a proto descriptor via the adapt pipeline, caches per table, and deduplicates concurrent cold-cache loads via singleflight.
- Add schemaEvolver that diffs proto descriptor vs table schema and adds missing columns. Uses ETag with CAS-on-412 retry (max 5 attempts) so concurrent additive evolutions don't clobber each other; on retry exhaustion or a benign "another writer added the columns" outcome, signal the caller to retry the write.
- Add grpcSchemaMismatch classification via StorageError details. In handleWriteError: on schema mismatch, evolve and retry; on evolution failure, return a permanent BatchError so messages go to DLQ instead of being retried forever.
- Add bigquery_write_api_schema_evolutions_total and bigquery_write_api_schema_evolution_failures_total counters.

Table name handling
- Sanitize interpolated table names (dots, hyphens, slashes, and whitespace become underscores; non-ASCII-alphanumerics are stripped; leading digits prefixed with _; cap at 1024 chars).
- Reject empty-after-sanitization names with a permanent BatchError.

Error classification
- Add isPermanentBQError that recognises both gRPC permanent codes and *googleapi.Error 4xx (excluding 408/429), so REST 4xx errors from the resolver path no longer retry forever.

Lifecycle and concurrency
- Drop bqClient/datasetID fields from resolver and evolver; pass client + dataset as method arguments so they can't outlive the client. Close drains the resolver cache.
- Track previously fire-and-forget stream closes (evictStream and the idle sweeper) on a closeWg so Close waits for them before tearing down the underlying clients.
- Snapshot client and resolvedProjectID together under one connMu RLock in WriteBatch; pass projectID through to tableCacheKey.
- Run client closes unconditionally in Close — the bigquery and managedwriter clients are non-blocking gRPC teardowns; gating them on ctx.Err() leaks connections.

Config validation and tuning
- Reject delegates without target_principal at parse time.
- Lower max_in_flight default from 64 to 4 to match snowflake_streaming and iceberg sibling outputs.
- Detach the impersonate token source from Connect's ctx with context.WithoutCancel so a cancelled connect doesn't break later token refreshes.
- Warn explicitly that endpoint overrides disable authentication.

Cleanup
- Drop dead fieldNameMapping plumbing (resolvedSchema field, jsonToProtoBytes parameter, streamWithDescriptor field) — never wired in production.
- Drop unused *service.Message argument from Resolve.
- Reject non-positive proto Kind in protoKindToBQFieldType instead of silently coercing to STRING; cap RECORD nesting at 15 (BigQuery's own limit) to fail fast on self-referential descriptors.
- Set Repeated on field schemas for repeated proto fields.

Tests
- Unit tests for resolver (cache, evict, no-client fallback), evolver helpers (kind mapping, schema diff, repeated fields), buildAuthOpts (endpoint override, credentials_json, no-auth), error classification (gRPC permanent, REST 4xx, wrapped), table-name sanitization, sweep goroutine, config validation (durations, delegates).
- Integration tests for schema evolution and table-name sanitization against the goccy bigquery emulator.

### DIFF
--- a/internal/impl/gcp/enterprise/bigquery/integration_test.go
+++ b/internal/impl/gcp/enterprise/bigquery/integration_test.go
@@ -16,12 +16,14 @@ import (
 	"time"
 
 	"cloud.google.com/go/bigquery"
+	"cloud.google.com/go/bigquery/storage/managedwriter/adapt"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/testcontainers/testcontainers-go"
 	"github.com/testcontainers/testcontainers-go/wait"
 	"google.golang.org/api/iterator"
 	"google.golang.org/api/option"
+	"google.golang.org/protobuf/reflect/protoreflect"
 
 	_ "github.com/redpanda-data/benthos/v4/public/components/io"
 	_ "github.com/redpanda-data/benthos/v4/public/components/pure"
@@ -31,14 +33,16 @@ import (
 	"github.com/redpanda-data/connect/v4/internal/license"
 )
 
-func TestIntegrationBigQueryWriteAPI(t *testing.T) {
-	integration.CheckSkip(t)
+// bqEmulator holds connection details for a running BigQuery emulator container.
+type bqEmulator struct {
+	httpEndpoint string
+	grpcEndpoint string
+	bqClient     *bigquery.Client
+}
 
-	const (
-		projectID = "test-project"
-		datasetID = "test_dataset"
-		tableID   = "test_table"
-	)
+// startEmulator launches a BigQuery emulator container and returns connection details.
+func startEmulator(t *testing.T, projectID, datasetID string) *bqEmulator {
+	t.Helper()
 
 	t.Log("Given a BigQuery emulator running with HTTP and gRPC ports")
 	ctr, err := testcontainers.Run(t.Context(),
@@ -54,7 +58,10 @@ func TestIntegrationBigQueryWriteAPI(t *testing.T) {
 			},
 		}),
 		testcontainers.WithWaitStrategy(
-			wait.ForListeningPort("9050/tcp").WithStartupTimeout(60*time.Second),
+			wait.ForAll(
+				wait.ForListeningPort("9050/tcp"),
+				wait.ForListeningPort("9060/tcp"),
+			).WithDeadline(60*time.Second),
 		),
 	)
 	testcontainers.CleanupContainer(t, ctr)
@@ -70,19 +77,50 @@ func TestIntegrationBigQueryWriteAPI(t *testing.T) {
 	httpEndpoint := fmt.Sprintf("http://%s:%s", host, httpPort.Port())
 	grpcEndpoint := fmt.Sprintf("%s:%s", host, grpcPort.Port())
 
-	t.Log("Given a table with name and age columns")
 	bqClient, err := bigquery.NewClient(t.Context(), projectID,
 		option.WithoutAuthentication(),
 		option.WithEndpoint(httpEndpoint),
 	)
 	require.NoError(t, err)
-	defer bqClient.Close()
+	t.Cleanup(func() { bqClient.Close() })
 
+	return &bqEmulator{
+		httpEndpoint: httpEndpoint,
+		grpcEndpoint: grpcEndpoint,
+		bqClient:     bqClient,
+	}
+}
+
+// bqSchemaToMessageDescriptor converts a BigQuery schema to a protoreflect.MessageDescriptor
+// via the adapt pipeline, for use in tests that need proto descriptors.
+func bqSchemaToMessageDescriptor(t *testing.T, schema bigquery.Schema) protoreflect.MessageDescriptor {
+	t.Helper()
+	tableSchema, err := adapt.BQSchemaToStorageTableSchema(schema)
+	require.NoError(t, err)
+	descriptor, err := adapt.StorageSchemaToProto2Descriptor(tableSchema, "root")
+	require.NoError(t, err)
+	md, ok := descriptor.(protoreflect.MessageDescriptor)
+	require.True(t, ok)
+	return md
+}
+
+func TestIntegrationBigQueryWriteAPI(t *testing.T) {
+	integration.CheckSkip(t)
+
+	const (
+		projectID = "test-project"
+		datasetID = "test_dataset"
+		tableID   = "test_table"
+	)
+
+	emu := startEmulator(t, projectID, datasetID)
+
+	t.Log("Given a table with name and age columns")
 	schema := bigquery.Schema{
 		{Name: "name", Type: bigquery.StringFieldType, Required: true},
 		{Name: "age", Type: bigquery.IntegerFieldType, Required: true},
 	}
-	err = bqClient.Dataset(datasetID).Table(tableID).Create(t.Context(), &bigquery.TableMetadata{
+	err := emu.bqClient.Dataset(datasetID).Table(tableID).Create(t.Context(), &bigquery.TableMetadata{
 		Schema: schema,
 	})
 	require.NoError(t, err)
@@ -102,7 +140,7 @@ gcp_bigquery_write_api:
   endpoint:
     http: %s
     grpc: %s
-`, projectID, datasetID, tableID, httpEndpoint, grpcEndpoint)))
+`, projectID, datasetID, tableID, emu.httpEndpoint, emu.grpcEndpoint)))
 
 	stream, err := sb.Build()
 	require.NoError(t, err)
@@ -131,7 +169,7 @@ gcp_bigquery_write_api:
 	// Then all 3 rows land in the BigQuery table.
 	t.Log("Then all 3 rows are present in the BigQuery table")
 	assert.Eventually(t, func() bool {
-		it := bqClient.Dataset(datasetID).Table(tableID).Read(t.Context())
+		it := emu.bqClient.Dataset(datasetID).Table(tableID).Read(t.Context())
 		var count int
 		for {
 			var row map[string]bigquery.Value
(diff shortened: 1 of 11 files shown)


internal/impl/gcp/enterprise/bigquery/schema_evolution.go
```diff
@@ -0,0 +1,185 @@
...
+		meta, err := client.Dataset(datasetID).Table(tableID).Metadata(ctx)
+		if err != nil {
+			return false, fmt.Errorf("fetching table metadata for evolution: %w", err)
+		}
+
+		missing, err := diffMissingColumns(md, meta.Schema)
+		if err != nil {
+			return false, fmt.Errorf("computing missing columns: %w", err)
+		}
+		if len(missing) == 0 {
```
josephwoodward, line 176: Concurrent schema evolution permanently drops messages

`schema_evolution.go:155–161` — when `missing == 0` on the first attempt, Evolve returns `(false, nil)`. `output.go:546–550` turns that into `permanentBatchError`. Under concurrent writes, the second batch to hit `SCHEMA_MISMATCH` sees columns already added by the first and is permanently dropped rather than retried.

internal/impl/gcp/enterprise/bigquery/schema_evolution.go
```diff
@@ -0,0 +1,185 @@
...
+	existingNames := make(map[string]struct{}, len(existing))
+	for _, f := range existing {
+		existingNames[f.Name] = struct{}{}
+	}
+
+	var missing bigquery.Schema
+	fields := md.Fields()
+	for i := range fields.Len() {
+		fd := fields.Get(i)
+		if _, ok := existingNames[string(fd.Name())]; ok {
```
josephwoodward: `schema_evolution.go:107–116` — `existingNames` map uses exact string keys. BQ column names are case-insensitive and stored lowercase; a proto field named UserID against a BQ column userid is treated as missing and an add-column request is sent. BQ rejects it as a duplicate, evolution fails, batch is dropped.

internal/impl/gcp/enterprise/bigquery/output.go
```diff
@@ -196,30 +204,34 @@ func bigQueryWriteAPIConfigFromParsed(pConf *service.ParsedConfig) (conf bigQuer
 		return
 	}
 	epConf := pConf.Namespace(bqwaFieldEndpoint)
-	if conf.EndpointHTTP, err = epConf.FieldString(bqwaFieldEndpointHTTP); err != nil {
+	if conf.EndpointHTTP, err = epConf.FieldString(bqwaepFieldHTTP); err != nil {
 		return
 	}
-	if conf.EndpointGRPC, err = epConf.FieldString(bqwaFieldEndpointGRPC); err != nil {
+	if conf.EndpointGRPC, err = epConf.FieldString(bqwaepFieldGRPC); err != nil {
```
josephwoodward, line 256: question(non-blocker): Why the rename here out of interest? The config variable (`EndpointGRPC`) is still includes endpoint?

Same question applies to `bqwaFieldEndpointHTTP` above.

internal/impl/gcp/enterprise/bigquery/output.go
```diff
@@ -196,30 +204,34 @@ func bigQueryWriteAPIConfigFromParsed(pConf *service.ParsedConfig) (conf bigQuer
...
-		batchesSent:  m.NewCounter("bigquery_write_api_batches_sent_total"),
-		batchLatency: m.NewTimer("bigquery_write_api_batch_latency_ns"),
-		retries:      m.NewCounter("bigquery_write_api_retries_total"),
+		rowsSent:                m.NewCounter("bigquery_write_api_rows_sent_total"),
+		rowsFailed:              m.NewCounter("bigquery_write_api_rows_failed_total"),
+		batchesSent:             m.NewCounter("bigquery_write_api_batches_sent_total"),
+		batchLatency:            m.NewTimer("bigquery_write_api_batch_latency_ns"),
+		retries:                 m.NewCounter("bigquery_write_api_retries_total"),
+		schemaEvolutions:        m.NewCounter("bigquery_write_api_schema_evolutions_total"),
+		schemaEvolutionFailures: m.NewCounter("bigquery_write_api_schema_evolution_failures_total"),
```
josephwoodward: Would it be worth changing this metric to `bigquery_write_api_schema_evolutions_failures_total` so it's consistent with the one above (`bigquery_write_api_schema_evolutions_total`)?

internal/impl/gcp/enterprise/bigquery/output.go
```diff
@@ -269,11 +298,30 @@ type streamWithDescriptor struct {
 	lastUsed   atomic.Int64 // UnixNano timestamp, safe for concurrent access
 }
 
+// isPermanentBQError reports whether an error from the BigQuery REST or Storage
+// Write APIs should be treated as permanent (no retry). 4xx HTTP status codes
+// are permanent except for 408 (timeout) and 429 (throttling); all gRPC errors
+// classified as permanent by classifyGRPCError are also permanent.
+func isPermanentBQError(err error) bool {
```
josephwoodward: thought (non-blocking): Feels like this kind of behaviour would be good to move into a custom big query error type so we could do `err.IsRetryable()` for instance. It could then contain a lot of other functionality like `classifyGRPCError`, clearing up the output.

internal/impl/gcp/enterprise/bigquery/output.go
```diff
@@ -495,13 +601,25 @@ func (o *bigQueryWriteAPIOutput) Close(_ context.Context) error {
 		o.client = nil
 	}
 
+	if err := ctx.Err(); err != nil {
+		errs = append(errs, err)
+	}
+
+	// Drop the resolver's schema cache so a subsequent Connect doesn't reuse
+	// state tied to the now-closed client.
+	o.resolver.cache.Range(func(k, _ any) bool {
```
josephwoodward: Can we use `o.resolver.cache.Clear()` here instead?

internal/impl/gcp/enterprise/bigquery/output.go
squiidz: Old constant bqwaFieldEndpointHTTP = "http" was a bit off — the name said "EndpointHTTP" but the actual value was just "http". That's because Namespace().FieldString() takes the leaf key inside the namespace, not the full path. bqwaep is the prefix for fields nested under endpoint, so the constant name now matches its value.

The struct field stays EndpointGRPC because the config struct is flat: no nested sub-struct, so each field needs a unique name within it.

internal/impl/gcp/enterprise/bigquery/schema_resolver.go
```diff
@@ -0,0 +1,126 @@
...
+func (r *schemaResolver) Resolve(ctx context.Context, client *bq.Client, datasetID, tableID string) (*resolvedSchema, error) {
+	if cached, ok := r.cache.Load(tableID); ok {
+		return cached.(*resolvedSchema), nil
+	}
+
+	if client == nil {
+		return nil, errors.New("no schema source available: ensure the table exists in BigQuery")
+	}
+
+	v, err, _ := r.sf.Do(tableID, func() (any, error) {
```
josephwoodward, line 76: Claude picked up this one:

**Issue 1 - Singleflight/Evict Race Can Re-store Stale Schema**

schema_resolver.go:64-83 and output.go:519,571

Evict(tableID) deletes from sync.Map, but an in-flight singleflight.Do for the same key can race and re-store stale data after the eviction.

Timeline:
1. Goroutine A enters sf.Do(tableID, ...) — fetching the old schema (before evolution)
2. Schema evolution completes; resolver.Evict(tableID) deletes the cache entry
3. Goroutine A's singleflight completes → r.cache.Store(tableID, rs) writes the stale schema back

Result: the next Resolve call hits the re-stored old schema, causing another SCHEMA_MISMATCH_EXTRA_FIELDS cycle. Not a correctness failure (it converges), but it causes silent extra retry loops after every evolution.

Fix: After singleflight.Do completes successfully, verify the cache entry wasn't evicted mid-flight before storing:
if _, alreadyEvicted := r.cache.LoadOrStore(tableID, rs); alreadyEvicted {
  // another writer evicted while we were in-flight; don't re-store stale data
}
Or use sync.Map.CompareAndSwap / swap the singleflight group key on eviction (e.g. include a generation counter).

internal/impl/gcp/enterprise/bigquery/output.go
```diff
@@ -469,11 +592,19 @@ func (o *bigQueryWriteAPIOutput) Close(_ context.Context) error {
 	// so it does not access shared state after shutdown.
 	o.sweepWg.Wait()
 
+	// Wait for any in-flight async stream closes (from evictStream or
+	// sweepIdleStreams) so they don't race with client shutdown.
+	o.closeWg.Wait()
+
 	o.streamsMu.Lock()
 	streams := o.streams
 	o.streams = make(map[string]*streamWithDescriptor)
```
josephwoodward, line 617: Given this streams are an unbounded map, would it be worth considering setting a max limit on the number of streams it has with a setting such as `max_cached_streams` where we remove the oldest idle stream (I see we have a `lastUsed` property)?

internal/impl/gcp/enterprise/bigquery/integration_test.go
josephwoodward, line 54: Maybe add the following to force to run with emulation in ARM based MacOS:

```go
testcontainers.WithImagePlatform("linux/amd64"),
```

Otherwise it fails with:
```
generic container: create container: Error response from daemon: no matching manifest for linux/arm64/v8 in the manifest list entries: no match for platform in manifest: not found
```

internal/impl/gcp/enterprise/bigquery/integration_test.go
```diff
@@ -147,3 +185,128 @@ gcp_bigquery_write_api:
 		return count >= 3
 	}, 30*time.Second, 500*time.Millisecond)
 }
+
+func TestIntegrationSchemaEvolution(t *testing.T) {
```
josephwoodward, line 193: This test appears to fail when I run it locally:


```go
gotest -run "^TestIntegrationSchemaEvolution$" -v -count=1 ./internal/impl/gcp/enterprise/bigquery
=== RUN   TestIntegrationSchemaEvolution
    integration_test.go:199: Given a BigQuery emulator running with HTTP and gRPC ports
2026/05/18 12:28:16 github.com/testcontainers/testcontainers-go - Connected to docker: 
  Server Version: 29.4.1
  API Version: 1.54
  Operating System: Docker Desktop
  Total Memory: 23996 MB
  Labels:
    com.docker.desktop.address=unix:///Users/joseph.woodward/Library/Containers/com.docker.docker/Data/docker-cli.sock
  Testcontainers for Go Version: v0.42.0
  Resolved Docker Host: unix:///var/run/docker.sock
  Resolved Docker Socket Path: /var/run/docker.sock
  Test SessionID: b68ba4bc4d7c4f17152431818e1e4e1555b20a1284b17a2234ee290ebbd470c1
  Test ProcessID: 1d19e35c-e7f2-4053-a21c-6aed190612da
2026/05/18 12:28:16 🐳 Creating container for image ghcr.io/goccy/bigquery-emulator:latest
2026/05/18 12:28:16 🐳 Creating container for image testcontainers/ryuk:0.13.0
2026/05/18 12:28:16 ✅ Container created: e8e00e699a37
2026/05/18 12:28:16 🐳 Starting container: e8e00e699a37
2026/05/18 12:28:16 ✅ Container started: e8e00e699a37
2026/05/18 12:28:16 ⏳ Waiting for container id e8e00e699a37 image: testcontainers/ryuk:0.13.0. Waiting for: port 8080/tcp to be listening
2026/05/18 12:28:16 Shell not found in container
2026/05/18 12:28:16 🔔 Container is ready: e8e00e699a37
2026/05/18 12:28:16 ✅ Container created: 597676a73778
2026/05/18 12:28:16 🐳 Starting container: 597676a73778
2026/05/18 12:28:16 ✅ Container started: 597676a73778
2026/05/18 12:28:16 ⏳ Waiting for container id 597676a73778 image: ghcr.io/goccy/bigquery-emulator:latest. Waiting for: all of: [all of: [port 9050/tcp to be listening, port 9060/tcp to be listening]]
2026/05/18 12:28:16 🔔 Container is ready: 597676a73778
    integration_test.go:201: Given a table with name and age columns
    integration_test.go:218: When we call Evolve with a descriptor that has an extra email column
    integration_test.go:224: Then the table schema includes the email column
    integration_test.go:232: 
        	Error Trace:	/Users/joseph.woodward/code/connect_contrib/connect/internal/impl/gcp/enterprise/bigquery/integration_test.go:232
        	Error:      	[]string{"name", "age"} does not contain "email"
        	Test:       	TestIntegrationSchemaEvolution
    integration_test.go:236: When Evolve is called again with the same descriptor (simulating a concurrent batch arriving after another writer already evolved the table)
2026/05/18 12:28:16 🐳 Stopping container: 597676a73778
2026/05/18 12:28:17 ✅ Container stopped: 597676a73778
2026/05/18 12:28:17 🐳 Terminating container: 597676a73778
2026/05/18 12:28:17 🚫 Container terminated: 597676a73778
--- FAIL: TestIntegrationSchemaEvolution (1.16s)
FAIL
FAIL	github.com/redpanda-data/connect/v4/internal/impl/gcp/enterprise/bigquery	1.585s
FAIL
```

josephwoodward (approved):

PR #4392, opened 2026-05-04
ci: sign RPM packages before pushing to GCP Artifact Registry

PrzemekZglinicki:
## Summary

- Move RPM signing from a shell helper in \`push_pkg_to_gcp_ar.sh\` to goreleaser's native nfpm support: \`nfpms[*].rpm.signature.key_file\` in both \`.goreleaser/connect.yaml\` and \`.goreleaser/connect-fips.yaml\`
- nfpm signs the RPM at **package build time** (before upload) using Go's internal OpenPGP library — no dependency on system \`rpmsign\` or \`gpg\`
- The GPG private key is written to a temp \`.asc\` file in a new "Prepare RPM signing key" step and passed to goreleaser via \`RPM_KEY_FILE_PRIVATE\` in \`$GITHUB_ENV\`; the env var name makes clear it holds the private key material
- DEBs are unchanged — GCP AR provides repository-level signing for APT
- Add "Verify RPM signature in GCP AR" step in \`release.yml\`: after goreleaser publishes, downloads the RPM from GCP AR and verifies its GPG signature using \`rpm --dbpath <tmpdb>\` (avoids the silent-pass bug where plain \`rpm --checksig\` checks the system keyring and returns exit 0 for NOKEY)
- Fix \`SEARCH_TERM\` construction to use \`tr '-' '~'\` instead of \`\${VERSION//-/~}\` — bash expands \`~\` to \`\$HOME\` in parameter substitution replacements even inside double quotes
- \`test-push-to-gcp-ar\` job in \`test.yml\` tests **routing only** (GA → \`redpanda-yum\`, RC → \`redpanda-unstable-yum\`, DEB → \`redpanda-apt\`) against a mocked \`gcloud\` CLI; actual signing correctness is validated by the "Verify RPM signature in GCP AR" step in \`release.yml\`, which does a full nfpm build + GCP AR round-trip

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,13 +1,12 @@
 name: Release
 
 on:
-  pull_request:
-#  push:
-#    tags:
-#      - 'v*'
-#  schedule:
-#    - cron: '0 2 * * *' # run at 2 AM UTC
-#  workflow_dispatch:
+  push:
+    tags:
+      - 'v*'
+  schedule:
+    - cron: '0 2 * * *' # run at 2 AM UTC
+  workflow_dispatch:
 
 jobs:
   goreleaser:
@@ -20,11 +19,11 @@ jobs:
       fail-fast: false
       matrix:
         variant:
-#          - connect-ai
-#          - connect-cgo
-#          - connect-cloud
+          - connect-ai
+          - connect-cgo
+          - connect-cloud
           - connect-fips
-#          - connect-lambda
+          - connect-lambda
           - connect
 
     steps:
@@ -46,7 +45,6 @@ jobs:
             ,sdlc/prod/github/dockerhub
             ,sdlc/prod/github/rpm_signing_key
           parse-json-secrets: true
-
 
       - name: Configure AWS credentials for access to Amazon ECR Public
         uses: aws-actions/configure-aws-credentials@v6
@@ -77,8 +75,6 @@ jobs:
           tar -C "$RUNNER_TEMP/microsoft" -xf "$RUNNER_TEMP/msgo.tgz"
           echo "$RUNNER_TEMP/bin" >> "$GITHUB_PATH"
 
-
-
       - name: Release Notes
         run: ./resources/scripts/release_notes.sh > ./release_notes.md
 
@@ -87,15 +83,18 @@ jobs:
           python-version: '3.12'
 
       - name: Authenticate to GCP
+        if: ${{ github.event_name == 'push' }}
         uses: google-github-actions/auth@v3
         with:
           workload_identity_provider: projects/606234194099/locations/global/workloadIdentityPools/rp-cicd/providers/gh-connect
           service_account: gh-connect@devprod-cicd-infra.iam.gserviceaccount.com
 
       - name: Set up gcloud CLI
+        if: ${{ github.event_name == 'push' }}
         uses: google-github-actions/setup-gcloud@v3
 
       - name: Exclude GCP credentials from git state
+        if: ${{ github.event_name == 'push' }}
         run: echo 'gha-creds-*.json' >> .git/info/exclude
 
       - name: Install cloudsmith CLI (for publishing Linux packages)
@@ -130,14 +129,11 @@ jobs:
           printf '%s\n' "$RPM_SIGNING_KEY_PRIVATE" > "$tmpkey"
           echo "RPM_KEY_FILE_PRIVATE=$tmpkey" >> "$GITHUB_ENV"
 
-      - name: Create local tag for test build
-        run: git tag v0.0.0-rc.test${{ github.run_number }}
-
       - name: GoReleaser Release
-        if: ${{ github.event_name == 'pull_request' }}
+        if: ${{ github.event_name == 'push' }}
         uses: goreleaser/goreleaser-action@v7
         with:
-          args: release --timeout 120m --config ./.goreleaser/${{ matrix.variant }}.yaml
+          args: release --release-notes=./release_notes.md --timeout 120m --config ./.goreleaser/${{ matrix.variant }}.yaml
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           CLOUDSMITH_API_KEY: ${{ env.CLOUDSMITH_API_KEY }}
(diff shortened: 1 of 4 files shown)


PrzemekZglinicki:
@claude review once

PrzemekZglinicki:
Test purpose packages was uploaded to gcp ar:
image

image

PrzemekZglinicki:
> **Commits**
> 
>     1. Commits `a2d3e5d` ("ci: Test purpose changes. Drop before merge!") and `67ed751` ("ci: Remove test-purpose changes, keep RPM signing fixes") cancel each other out — the message of `a2d3e5d` itself says "Drop before merge!". They should be squashed/dropped before merge so the final history contains only the meaningful change in `c340f6e`.
> 
>     2. Both `a2d3e5d` and `67ed751` violate the message-format rule that the message after `system:` must start lowercase (`ci: Test...`, `ci: Remove...`).
> 
> 
> **Review** The actual change set in `c340f6e` (RPM signing in `push_pkg_to_gcp_ar.sh`, AWS Secrets Manager fetch + `Validate secrets` + post-upload signature verification in `release.yml`, `RPM_SIGNING_KEY_PRIVATE` plumbing through both `.goreleaser/*.yaml` files, and the `test-push-to-gcp-ar` job in `test.yml`) is consistent and looks correct. The subshell + EXIT trap in `sign_rpm` and the use of a temp `--dbpath` for `rpm --checksig` (called out in the PR description) are sound.
> 
> LGTM

In this repo we use the squash and merge concept, so the names of the particular commit doesn't matter, but also I can update them. What is more I think it is worth to keep test examples in the PR for future.

ivotron (changes requested):
goreleaser has support for signing rpm packages using a custom key. that should be done instead. see https://goreleaser.com/customization/sign/sign/ for more

.github/workflows/release.yml
```diff
@@ -44,6 +44,23 @@ jobs:
             ,sdlc/prod/github/cloudsmith
             ,sdlc/prod/github/dockerhub
           parse-json-secrets: true
+      - name: Get RPM signing key
+        uses: aws-actions/aws-secretsmanager-get-secrets@v3
+        with:
+          secret-ids: |
+            RPM_SIGNING_KEY_PRIVATE,sdlc/prod/github/rpm_signing_key_private
```
ivotron: this secret should be moved so that it is stored as JSON, to keep it consistent with the rest of secrets. that'd also allow for this to be moved to the previous task, with the rest of commits that are fetched as part of this workflow

ivotron:
> [CLA assistant check](https://cla-assistant.io/redpanda-data/connect?pullRequest=4392) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/redpanda-data/connect?pullRequest=4392) before we can accept your contribution.
> 
> **Przemek Zglinicki** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/redpanda-data/connect?pullRequest=4392) it.

@PrzemekZglinicki you have to sign the CLA

.goreleaser/connect.yaml
```diff
@@ -62,6 +62,9 @@ nfpms:
     formats:
       - deb
       - rpm
+    rpm:
+      signature:
+        key_file: "{{ .Env.RPM_KEY_FILE_PRIVATE }}"
```
ivotron, line 67: this is the only change that should go in IMO. this, plus loading the signing key as part of the aws secret fetch

.github/workflows/release.yml
```diff
@@ -173,6 +200,48 @@ jobs:
...
+          tmpdb=$(mktemp -d)
+          pubkey=$(mktemp)
+          trap 'rm -rf "$tmpgnupg" "$tmpdb" "$pubkey"' EXIT
+          printf '%s\n' "$RPM_SIGNING_KEY_PRIVATE" | GNUPGHOME="$tmpgnupg" gpg --batch --import
+          KEY_ID=$(GNUPGHOME="$tmpgnupg" gpg --list-secret-keys --with-colons | awk -F: '/^sec/{print $5}' | head -1)
+          GNUPGHOME="$tmpgnupg" gpg --armor --export "$KEY_ID" > "$pubkey"
+          rpm --dbpath "$tmpdb" --import "$pubkey"
+          rpm --dbpath "$tmpdb" --checksig "$RPM"
+          echo "RPM signature verification passed for: $(basename "$RPM")"
+
```
ivotron: instead of this being here, it should be in a separate bash script and referenced in the goreleaser config, as done for cloudsmith here https://github.com/redpanda-data/connect/blob/main/.goreleaser/connect.yaml#L97-L100

PrzemekZglinicki: It is verification, we can move it to the script. It is similar approach as in the smokechecks in the vtools repo, we sign the package, push it to the gcp ar and at the end verify that this package is available and correctly signed, it isn't a big effort and we are sure that package is correct.

.github/workflows/release.yml
PrzemekZglinicki: https://github.com/redpanda-data/vtools/pull/4219#discussion_r3192152576 - answered here

.goreleaser/connect.yaml
PrzemekZglinicki: I think that we should have verification step after push to gcp ar. We can remove part of the logic from .github/workflows/release.yml, step `name: Validate secrets` and isn't needed.

I think it is worth to keep test.yml to have some verification of `push_pkg_to_gcp_ar.sh` script. We can replace current approach with bats tests and call them in the CI.

.github/workflows/release.yml
PrzemekZglinicki: Whole logic of the push to gcp ar is in the script.

.github/workflows/release.yml
```diff
@@ -19,11 +20,11 @@ jobs:
...
-          - connect-ai
-          - connect-cgo
-          - connect-cloud
+#          - connect-ai
+#          - connect-cgo
+#          - connect-cloud
           - connect-fips
-          - connect-lambda
+#          - connect-lambda
           - connect
```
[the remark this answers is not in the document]
ivotron: @PrzemekZglinicki should this be resolved?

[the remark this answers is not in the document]
ivotron: @PrzemekZglinicki same here

.github/workflows/release.yml
```diff
@@ -74,6 +76,9 @@ jobs:
           tar -C "$RUNNER_TEMP/microsoft" -xf "$RUNNER_TEMP/msgo.tgz"
           echo "$RUNNER_TEMP/bin" >> "$GITHUB_PATH"
 
+      - name: Install rpm (for RPM signature verification)
+        run: sudo apt-get install -y rpm
```
ivotron: this is not needed by goreleaser, right?  it'd be better if it's moved to where it's needed, so as part of the `Verify RPM signature in GCP AR` task

.github/workflows/release.yml
```diff
@@ -173,6 +201,47 @@ jobs:
...
+          tmpgnupg=$(mktemp -d)
+          tmpdb=$(mktemp -d)
+          pubkey=$(mktemp)
+          trap 'rm -rf "$tmpgnupg" "$tmpdb" "$pubkey"' EXIT
+          printf '%s\n' "$RPM_SIGNING_KEY_PRIVATE" | GNUPGHOME="$tmpgnupg" gpg --batch --import
+          KEY_ID=$(GNUPGHOME="$tmpgnupg" gpg --list-secret-keys --with-colons | awk -F: '/^sec/{print $5}' | head -1)
+          GNUPGHOME="$tmpgnupg" gpg --armor --export "$KEY_ID" > "$pubkey"
+          rpm --dbpath "$tmpdb" --import "$pubkey"
+          rpm --dbpath "$tmpdb" --checksig "$RPM"
+          echo "RPM signature verification passed for: $(basename "$RPM")"
```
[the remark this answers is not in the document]
ivotron: @PrzemekZglinicki should this thread be resolved?

.github/workflows/release.yml
ivotron: i still think this should be moved to a script so it can be tested

.github/workflows/release.yml
```diff
@@ -173,6 +185,43 @@ jobs:
...
+            --project=production-devprod --destination=/tmp/rpms/ "$RPM_FILE"
+          RPM=$(find /tmp/rpms -name "*.rpm" | head -1)
+          [[ -n "$RPM" ]] || { echo "ERROR: no .rpm file found in /tmp/rpms"; exit 1; }
+          tmpdb=$(mktemp -d)
+          pubkey=$(mktemp)
+          trap 'rm -rf "$tmpdb" "$pubkey"' EXIT
+          printf '%s\n' "$RPM_SIGNING_KEY_PUBLIC" > "$pubkey"
+          rpm --dbpath "$tmpdb" --import "$pubkey"
+          rpm --dbpath "$tmpdb" --checksig "$RPM"
+          echo "RPM signature verification passed for: $(basename "$RPM")"
```
ivotron: to avoid this blocking releases, when this check fails for reasons other than the package not being signed as expected, this can ignore those non-blocking errors. for example ignore errors on `gcloud *` or `apt *`

ivotron (changes requested):

.github/workflows/release.yml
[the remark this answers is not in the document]
PrzemekZglinicki: It isn't commented, it was commented for test stage, I would like to keep this test purpose commit in the repo, especially that this repo allows to squash commit before merge, then it will be really simple to come back to test method if it will be required in the future.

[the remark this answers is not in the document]
PrzemekZglinicki: It isn't commented, it was commented for test stage, I would like to keep this test purpose commit in the repo, especially that this repo allows to squash commit before merge, then it will be really simple to come back to test method if it will be required in the future.

.github/workflows/release.yml
PrzemekZglinicki: Yep, we can move it there. I keep it here to have all install steps in one block.

.github/workflows/release.yml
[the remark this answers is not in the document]
PrzemekZglinicki: Test purpose change, I will resolve the comment.

.github/workflows/release.yml
PrzemekZglinicki: We can, but then we will have test for tests ;d

PrzemekZglinicki: I remove it, I feel it is just problematic.

ivotron (approved):
lgtm; assuming test commits get dropped

PR #4401, opened 2026-05-06
iceberg: add configurable Parquet string encoding for Redshift Spectrum compatibility

josephwoodward:
image

### DIFF
--- a/internal/impl/iceberg/output_iceberg.go
+++ b/internal/impl/iceberg/output_iceberg.go
@@ -17,6 +17,7 @@ import (
 	"github.com/apache/iceberg-go"
 	"github.com/apache/iceberg-go/io"
 	_ "github.com/apache/iceberg-go/io/gocloud"
+	"github.com/parquet-go/parquet-go"
 
 	"github.com/redpanda-data/benthos/v4/public/service"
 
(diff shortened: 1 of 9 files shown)


internal/impl/iceberg/output_iceberg.go
```diff
@@ -102,8 +103,20 @@ func newIcebergOutputFromConfig(conf *service.ParsedConfig, mgr *service.Resourc
...
 	}
 
+	// Parse parquet config
+	var writerOpts []parquet.WriterOption
+	if conf.Contains(ioFieldParquet) {
+		strEnc, err := conf.FieldString(ioFieldParquet, ioFieldParquetStringEncoding)
+		if err != nil {
+			return nil, fmt.Errorf("parsing %s: %w", ioFieldParquetStringEncoding, err)
+		}
+		if strEnc == "plain" {
```
twmb: I think this should be a switch,
```go
switch strEnc {
case "plain":
   // do this
case "delta thingy":
   // comment: this is a default no op
default:
   return nil, fmt.Errorf("unrecognized input feature gap in connect please raise an issue or something
}
```

squiidz (approved):

internal/impl/iceberg/output_iceberg.go
josephwoodward: Addressed in commit https://github.com/redpanda-data/connect/pull/4401/changes/77d7709cdb7c939fe12025faf8e3920eb07abe61

twmb (approved):

Issue #4399, opened 2026-05-06
Iceberg connector converts avro timestamp-millis to INT, not TIMESTAMP


WesWWagner:
Reported by James Graham of REDspace originally.

When parsing an Avro schema like

{
    "type": "long",
    "logicalType": "timestamp-millis"
}

ecsAvroFromAnyMap in `ecs_avro.go` only looks at the "type" field and maps it to schema.Int64. The "logicalType" field is ignored, so the field travels through the pipeline as a plain 64-bit integer. When the schema.Int64 is later converted to an Iceberg type in `type_resolver.go` it correctly becomes iceberg.Int64Type, i.e. BIGINT, with no information to signify it was intended to be a TIMESTAMP.

This problem is compounded for nullable/optional fields. The reported schema wraps the type in a union:

[
    "null",
    {
        "type": "long",
        "logicalType": "timestamp-millis"
    }
]

ecsAvroIsUnionJustOptional only handles the case where the second element is a plain string, e.g. "long". This being an object returns false, and the field falls through to ecsAvroHydrateLameUnion, which processes it as a full union and again ignores "logicalType".

Here is a potential fix:
1. Apply logicalType overrides after resolving the base Avro primitive type.
2. Modify ecsAvroIsUnionJustOptional to unwrap ["null", {primitive + logicalType}] as a simple optional field with the enriched type.

PR #4402, opened 2026-05-06
confluent,iceberg: preserve avro logical types end-to-end (#4399)

twmb:
## Summary

Issue [#4399](https://github.com/redpanda-data/connect/issues/4399): the Confluent Schema Registry Avro decoder was dropping every logical-type annotation other than `decimal`. Downstream sinks that consume `schema_metadata` (notably `iceberg`) inferred BIGINT / INT / STRING columns where the source schema declared timestamp / date / time / uuid. This PR makes every Avro logical type round-trip through `schema.Common` and lands at the right Iceberg column type.

The new `schema.Common` types and parameter blocks land in benthos v4.73.0 ([redpanda-data/benthos#429](https://github.com/redpanda-data/benthos/pull/429)), which is the version this PR consumes.

## Changes

- **Decoder** (`internal/impl/confluent/ecs_avro.go`): replace the decimal-only logical block with a full dispatcher covering `timestamp-{millis,micros,nanos}`, `local-timestamp-{millis,micros,nanos}`, `date`, `time-{millis,micros}`, `uuid`. Per Avro 1.10 spec, unrecognised annotations and primitive/logical-type mismatches fall back silently to the base primitive.
- **Encoder** (`internal/impl/confluent/common_to_avro.go`): symmetric encode. Legacy `Timestamp` schemas (nil `Logical`) keep emitting `timestamp-millis` via `EffectiveTimestamp()` for backwards compatibility. Avro-inexpressible shapes (e.g. `time-of-day` with `AdjustToUTC=true`, `time-of-day` with nanos) are rejected with field-naming errors rather than silently downcast.
- **Iceberg type-resolver** (`internal/impl/iceberg/type_resolver.go`): adds `Date`, `TimeOfDay`, `UUID` arms and uses `EffectiveTimestamp()` to pick `TimestampType` / `TimestampTzType` / `TimestampNsType` / `TimestampTzNsType` based on the schema's declared unit and `AdjustToUTC`. Iceberg has no time-with-tz, so `TimeOfDay` with `AdjustToUTC=true` is rejected loudly.
- **Iceberg shredder** (`internal/impl/iceberg/shredder/{shredder,temporal}.go`): plumbs a fieldID → `*schema.Common` map onto `RecordShredder` via `SetFieldSchemaMetadata`. The leaf-value converter consults it for time-typed columns and uses the declared unit to scale numeric inputs into the column's internal representation. **Closes the silent-corruption case**: a numeric millisecond value declared by the schema as `timestamp-millis` would previously land ~50,000 years in the future when the column type flipped from BIGINT to TIMESTAMPTZ. Now the schema's declared unit is the source of truth for unit interpretation.
- **Iceberg writer** (`internal/impl/iceberg/{writer,router}.go`): `NewWriter` accepts the `*typeResolver`. The writer parses `schema_metadata` from the first message of a batch and builds the field-ID lookup the shredder consults. Internal API change only.
- **`schema_evolution.require_schema_metadata`** (default `false`): opt-in strict mode. When `true` along with `schema_evolution.schema_metadata`, numeric values shredded into `timestamp` / `timestamptz` / `date` / `time` columns without registered schema metadata are rejected loudly instead of falling through to the bloblang Unix-seconds default. Use this when you cannot guarantee the upstream attaches schema metadata and prefer a hard error to silently corrupting dates by ~50,000 years. No effect on time-typed columns receiving native `time.Time` / `time.Duration` Go values.

## Breaking changes

Pipeline values flow through unchanged in both `preserve_logical_types=false` (default — values stay numeric) and `preserve_logical_types=true` (values stay rich Go time types). Bloblang behaviour and JSON-output bytes are unaffected.

The breakage is concentrated in:

1. **Existing Iceberg tables** with BIGINT / INT / STRING columns that were created from this bug. Iceberg disallows BIGINT → TIMESTAMP schema evolution, so the first write after upgrade will fail loudly. Drop and re-create, or use Iceberg-native column-rename + add-new-column tooling.
2. **Custom code reading `schema_metadata` bytes via `meta()`** and pattern-matching the historical INT64 shape: schemas now contain `TIMESTAMP` / `DATE` / `TIME_OF_DAY` / `UUID` along with new `unit` and `adjust_to_utc` fields.

## Test plan

- [x] `task lint` — 0 issues.
- [x] `go test ./internal/impl/confluent/...` — pass. Tests cover every Avro logical type, optional-union shape, decode/encode round-trip, decimal-on-wrong-primitive fallback.
- [x] `go test ./internal/impl/iceberg/...` (top-level + shredder + icebergx + catalogx + integration + e2e/{glue,polaris-aws,polaris-azure}) — pass. Tests cover Common→Iceberg mapping for each variant, schema-aware numeric conversion (including the year-50000 case), pre-epoch dates, time-of-day location semantics, container-recursion through nested struct/list/map metadata, strict-mode rejection of wrong-shape metadata, and panic-on-invalid TimeUnit defense.
- [ ] Manual smoke: schema-registry-decode → iceberg integration with a `timestamp-millis` Avro field, default mode. Verify TIMESTAMPTZ column with correct epoch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ All notable changes to this project will be documented in this file.
   - **Pipelines whose own code reads the `schema_metadata` bytes via `meta()`** and pattern-matches the historical INT64 shape: schemas now contain `TIMESTAMP` / `DATE` / `TIME_OF_DAY` / `UUID` along with new `unit` and `adjust_to_utc` fields. Update the pattern.
   - **iceberg shredder** is now schema-aware for numeric inputs: a numeric millisecond value declared by the schema as `timestamp-millis` is correctly interpreted as milliseconds rather than as Unix seconds. This closes a previously-silent corruption case where an int64 millis input into a TIMESTAMPTZ column would land ~50,000 years in the future.
 
+### Added
+
+- iceberg: `schema_evolution.require_schema_metadata` (default `false`). When enabled along with `schema_evolution.schema_metadata`, numeric values shredded into a `timestamp`, `timestamptz`, `date`, or `time` column without registered schema metadata are rejected loudly instead of falling through to the bloblang Unix-seconds default. Use this when you cannot guarantee the upstream attaches schema metadata and prefer a hard error to silently corrupting dates by ~50,000 years. No effect on time-typed columns receiving native `time.Time` / `time.Duration` Go values.
+
 ### Changed
 
 - iceberg: `NewWriter` now takes a `*typeResolver` argument so the writer can use schema metadata to interpret numeric inputs into time-typed columns at shredding time. Internal API change only.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,23 @@ Changelog
 
 All notable changes to this project will be documented in this file.
 
+## Unreleased
+
+### Fixed
+
+- **BREAKING:** schema_registry_decode (avro): Avro logical types — `timestamp-{millis,micros,nanos}`, `local-timestamp-{millis,micros,nanos}`, `date`, `time-{millis,micros}`, and `uuid` — are now preserved end-to-end in the schema metadata produced by the schema-registry decoder. Previously only `decimal` was honoured; every other logical type silently degraded to its base primitive (`long`, `int`, or `string`). Downstream sinks that consume `schema_metadata` (notably `iceberg`) now create the correct column type. ([#4399](https://github.com/redpanda-data/connect/issues/4399))
+
+  Pipeline values flow through unchanged in both `preserve_logical_types=false` (default — values stay numeric) and `preserve_logical_types=true` (values stay rich Go time types). Bloblang behaviour and JSON-output bytes are unaffected.
+
+  **What this changes for existing pipelines:**
+  - **iceberg with existing tables that have BIGINT / INT / STRING columns from this bug**: the connector now wants to create or evolve those columns to TIMESTAMP / TIMESTAMPTZ / DATE / TIME / UUID. Iceberg disallows BIGINT → TIMESTAMP schema evolution, so the first write after upgrade will fail loudly. Drop and re-create the table, or use Iceberg-native column-rename + add-new-column tooling to migrate before upgrading.
+  - **Pipelines whose own code reads the `schema_metadata` bytes via `meta()`** and pattern-matches the historical INT64 shape: schemas now contain `TIMESTAMP` / `DATE` / `TIME_OF_DAY` / `UUID` along with new `unit` and `adjust_to_utc` fields. Update the pattern.
+  - **iceberg shredder** is now schema-aware for numeric inputs: a numeric millisecond value declared by the schema as `timestamp-millis` is correctly interpreted as milliseconds rather than as Unix seconds. This closes a previously-silent corruption case where an int64 millis input into a TIMESTAMPTZ column would land ~50,000 years in the future.
+
+### Changed
+
+- iceberg: `NewWriter` now takes a `*typeResolver` argument so the writer can use schema metadata to interpret numeric inputs into time-typed columns at shredding time. Internal API change only.
+
 ## 4.90.0 - 2026-04-30
 
 ### Added
--- a/docs/modules/components/pages/outputs/iceberg.adoc
+++ b/docs/modules/components/pages/outputs/iceberg.adoc
@@ -149,6 +149,7 @@ output:
       table_location: s3://my-iceberg-bucket/ # No default (optional)
       schema_metadata: ""
       new_column_type_mapping: "" # No default (optional)
+      require_schema_metadata: false
     commit:
       manifest_merge_enabled: true
       max_snapshot_age: 24h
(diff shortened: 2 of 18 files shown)


squiidz:
From claude
  ---
  1. Avro encoder is missing the AdjustToUTC=true rejection for TimeOfDay

  File: internal/impl/confluent/common_to_avro.go
  Lines: 93–106 (the case schema.TimeOfDay: arm)

  ▎ The PR description says time-of-day with AdjustToUTC=true is rejected with a field-naming error, and the Iceberg type-resolver does reject it (type_resolver.go line 268). But this branch only
  ▎  switches on Unit — AdjustToUTC is never read. Avro time-{millis,micros} carries no zone semantics, so encoding a TimeOfDay{AdjustToUTC=true} silently drops that bit. Add an if
  ▎ c.Logical.TimeOfDay.AdjustToUTC { return nil, fmt.Errorf(...) } ahead of the unit switch, or update the PR description.

  ---
  2. convertDate truncates toward zero for pre-epoch dates

  File: internal/impl/iceberg/shredder/temporal.go
  Line: 42 (and the comment at lines 38–41 is also wrong)

  ▎ Go integer division truncates toward zero, not floors. 1969-12-31 23:59:59 UTC (Unix() == -1) yields -1 / 86400 == 0, producing day index 0 (Jan 1 1970) instead of -1 (Dec 31 1969). The
  ▎ comment "values not at UTC midnight are floored to their date" is incorrect for negative timestamps. Iceberg DATE supports negative day indices, so this is a real (if rare) corruption case.
  ▎ Use int32((v.UTC().Unix() - (86400 - 1)) / 86400) for negatives or compute via v.UTC().Sub(epoch).Hours() / 24 with explicit floor.

  ---
  3. Writer reads schema metadata only from batch[0]

  File: internal/impl/iceberg/writer.go
  Line: 201

  ▎ This assumes every message in the batch was produced from the same upstream schema. If the router can ever group messages with differing schema_metadata into the same batch (e.g. schema
  ▎ evolution mid-stream, or a topic with mixed schemas going to one table), batch[1..N] silently use the wrong unit interpretation. Either confirm the upstream batching invariant in a comment
  ▎ here, or do a per-message parseSchemaMetadata lookup with a tiny LRU.

  ---
  4. visitIcebergSchemaFields treats MapType as a leaf

  File: internal/impl/iceberg/writer.go
  Lines: 357–369 (the switch t := f.Type.(type) block)

  ▎ *iceberg.MapType falls through to default: out[f.ID] = child, registering the entire map as a leaf. A map<string, timestamp-millis> value column would never get its unit metadata. Add a case
  ▎ *iceberg.MapType: branch that recurses into the value type, or add a comment documenting that map values aren't unit-aware.

  ---
  5. ListType branch builds a synthetic Object parent — awkward

  File: internal/impl/iceberg/writer.go
  Lines: 360–365

  ▎ The &schema.Common{Type: schema.Object, Children: []schema.Common{child.Children[0]}} synthetic parent and Name: "" element rely on strings.EqualFold("", "") matching at lookup. Works today,
  ▎ but reads as a hack and will surprise the next reader. Extract a helper recordLeafIfTimeTyped(fieldID int, elemType iceberg.Type, elemCommon *schema.Common, out map[int]*schema.Common) that
  ▎ handles the leaf-vs-recurse decision directly.

  ---
  6. Silent default: in unit-scaling helpers

  File: internal/impl/iceberg/shredder/temporal.go
  Lines: 104 (numericToTimeMicros), 159 (scaleTimestampNumeric nanos branch), 172 (scaleTimestampNumeric non-nanos branch)

  ▎ All three return n unchanged on an unrecognized TimeUnit. The benthos PR's Validate() should prevent that, but defense-in-depth: an invalid TimeUnit(0) slipping through silently scales by 1 —
  ▎  i.e. interprets seconds as micros. Either log at debug, return a sentinel the caller can detect, or panic on unreachable.

  ---
  7. numericInt64 truncates float to int

  File: internal/impl/iceberg/shredder/temporal.go
  Line: 180 (function declaration; the float64 and float32 arms are inside)

  ▎ int64(float64Value) silently truncates fractional sub-second precision and overflows for very large floats. Probably acceptable but worth a doc comment noting that float inputs lose sub-unit
  ▎ precision.

  ---
  8. avroLogicalTimestampUnit returns invalid 0 sentinel

  File: internal/impl/confluent/ecs_avro.go
  Lines: 230–241

  ▎ default: return 0 returns an invalid TimeUnit(0) that would fail Unit.valid() if it ever escaped. Only called from one site today (line 213) where the cases are exhaustive, so it's
  ▎ unreachable — but if reused, this propagates a malformed Logical. Either return (TimeUnit, bool) and have the caller handle missing, or panic("unreachable") to make the contract explicit.

  ---
  9. Foot-gun: schema-metadata stripping mid-pipeline reintroduces the bug

  File: internal/impl/iceberg/shredder/temporal.go
  Lines: 131–141 (the bloblang fallback in convertTimestamp)

  ▎ Not a bug per se — the fallback is intentional for backwards compat. But after this PR, the type-resolver has already picked TIMESTAMPTZ for the column, so a downstream mapping that drops
  ▎ schema_metadata between decoder and sink reintroduces the year-50000 corruption silently. CHANGELOG mentions this once but it's the highest-blast-radius regression path. Consider adding a
  ▎ writer config option require_schema_metadata: true that errors when a time-typed column is reached without metadata.

twmb:
**Commits**
LGTM

**Review**
Reviewed the Avro logical-type preservation work and the `require_schema_metadata` strict-mode addition. Decoder/encoder symmetry, type-resolver mapping, shredder schema-aware conversion, and writer plumbing all look coherent. License headers on the new files match the iceberg package's RCL classification, the new config field is wired through with proper validation, the panic("unreachable") sites are correctly gated behind the same upstream switches that filter their input, and the `convertTimestamp` strict-mode check correctly distinguishes nil-Logical-Timestamp legacy schemas (millis default via EffectiveTimestamp) from no-metadata cases (bloblang seconds fallback).

LGTM

squiidz (approved):

PR #4404, opened 2026-05-07
iceberg: bump iceberg-go to allow v1 manifests inside v2 manifest lists

Jeffail:
## Summary

- Bumps `github.com/apache/iceberg-go` to the merge commit of [apache/iceberg-go#1030](https://github.com/apache/iceberg-go/pull/1030), which relaxes `ManifestListWriter.AddManifests` so a v2 manifest list may reference v1 manifest files (and a v3 list may reference v1/v2 manifests), per the Iceberg spec's mixed-version inheritance rules.
- Without that fix, every commit on a table that was upgraded from v1 to v2 — whether via `Transaction.UpgradeFormatVersion` in `internal/impl/iceberg/committer.go` or out-of-band by another engine such as Spark or Hive — failed permanently with `invalid argument: ManifestListWriter only supports version 2 manifest files`, because historical v1 manifests resurfaced through `existingManifests()` on every commit.
- Adds a regression test (`TestCON459_V2ManifestListAcceptsV1Manifest`) that constructs a v1 `ManifestFile` via the public `iceberg-go` API, writes it through `WriteManifestList(2, ...)`, and asserts that the decoded entry inherits `content=data` and `sequence_number=min_sequence_number=0`. If the upstream fix ever regresses, this test fails.

Resolves CON-459.

## Test plan

- [x] `task fmt`
- [x] `task lint` — 0 issues
- [x] `task test` — full unit + template suite passes, including the new regression test
- [ ] Customer (Comcast) confirms the failure no longer reproduces against an edge build that includes this bump


internal/impl/iceberg/con459_repro_test.go
```diff
@@ -0,0 +1,70 @@
+// Copyright 2025 Redpanda Data, Inc.
```
josephwoodward:
```suggestion
// Copyright 2026 Redpanda Data, Inc.
```

josephwoodward (approved):

PR #4417, opened 2026-05-10
oracledb_cdc: extend checkpoint cache to support multiple callers

josephwoodward:
This change updates the internal OracleDB checkpoint cache to support multiple connect instances.

Upon launch if it detects the original checkpoint schema then it'll migrate it to the new schema, including:
- Updating the default cache key to the configured default, whilst leaving the SCN untouched
- Updating `CACHE_KEY` length to 128

First run that detects migration:

image

Subsequent runs shows no log output as table has been migrated:

image

Change in table schema to allow custom cache keys and renames the key value from the original default `max_scn` to the value of `checkpoint_cache_key` between before and after runs:

Before:
image

After:

image

Sharing table with multiple instances, arriving at the same SCN:

image

### DIFF
--- a/docs/modules/components/pages/inputs/oracledb_cdc.adoc
+++ b/docs/modules/components/pages/inputs/oracledb_cdc.adoc
@@ -324,7 +324,7 @@ A https://www.docs.redpanda.com/redpanda-connect/components/caches/about[cache r
 
 === `checkpoint_cache_table_name`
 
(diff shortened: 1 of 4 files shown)


squiidz:
From Claude:

1. Partial migration can't recover (checkpoint_cache.go:175-194). ALTER auto-commits; if the UPDATE then fails, next startup sees charLen == 128, skips the whole block, leaves the max_scn row behind, and the input
restarts from SCN 0. Decouple the two steps so each is independently idempotent.
  
2. No validation of checkpoint_cache_key. Empty string violates the NOT NULL PK; > 128 chars hits ORA-12899. Both surface as cryptic startup-time DB errors instead of config errors.

squiidz (approved):

Issue #1347, opened 2022-07-28
Have you considered adding support for HDFS Kerberos auth？


Salud77:

Salud77:
Hi, @Jeffail ,I see you have labeled this issue.  If necessary, I can provide some code related to HDFS Kerberos auth, although the code doesn't look that good.

mihaitodor:
I wonder if it's a matter of switching to https://pkg.go.dev/github.com/colinmarc/hdfs/v2, since, right now, Benthos is still on v1. Not sure how much effort that would require, though. I don't have any experience with HDFS. Is there any way to spin up a Docker container which has HDFS in it with Kerberos auth configured?

Khalid-Nowaf:
Is there any plans to adopt kerberos for this case?

mihaitodor:
Hey @Khalid-Nowaf, unfortunately I don't think anyone looked into it so far. Please feel free to send a PR if you'd like to work on it.

twmb:
Closing as stale — no PR materialized despite interest over 2+ years. This would require upgrading to hdfs/v2 and adding Kerberos support. If someone wants to contribute this, feel free to open a new issue or PR.

Khalid-Nowaf:
@twmb a PR has been created #4419 . thanks.

PR #4419, opened 2026-05-11
WIP: hdfs: add Kerberos authentication support

Khalid-Nowaf:
Adds Kerberos authentication support for the HDFS connector by upgrading to github.com/colinmarc/hdfs/v2 and introducing shared HDFS auth configuration for input and output.

  **Changes**

  - Adds auth.kerberos config for HDFS output.
  - Adds limited Kerberos support for HDFS input.
  - Supports keytab-based Kerberos auth with:
      - krb5_conf
      - keytab
      - principal
      - service_principal
      - data_transfer_protection

**Related Issues**
#1347

### DIFF
--- a/docs/modules/components/pages/inputs/hdfs.adoc
+++ b/docs/modules/components/pages/inputs/hdfs.adoc
@@ -25,8 +25,15 @@ component_type_dropdown::[]
 
 Reads files from a HDFS directory, where each discrete file will be consumed as a single message payload.
 
+
+[tabs]
+======
+Common::
++
+--
+
 ```yml
-# Config fields, showing default values
+# Common config fields, showing default values
 input:
   label: ""
   hdfs:
@@ -34,6 +41,32 @@ input:
     user: ""
     directory: "" # No default (required)
 ```
+
+--
+Advanced::
++
+--
+
+```yml
+# All config fields, showing default values
+input:
+  label: ""
+  hdfs:
+    hosts: [] # No default (required)
+    user: ""
+    auth:
+      kerberos:
+        enabled: false
+        krb5_conf: /etc/krb5.conf
+        keytab: ""
+        principal: ""
+        service_principal: nn/_HOST
+        data_transfer_protection: ""
+    directory: "" # No default (required)
+```
+
+--
+======
 
 == Metadata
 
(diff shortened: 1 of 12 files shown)


Khalid-Nowaf:
I would like to know how I trigger the review process.

Jeffail (commented):
Thank you for the contribution, @Khalid-Nowaf — Kerberos support here has been long awaited. I spent some time tracing through the diff and the upstream libraries; the notes below are intended as collaborative feedback.

**Kerberos client lifecycle (`internal/impl/hdfs/config.go` + `input.go`/`output.go` `Close`)**

`Login()` inside `hdfsKerberosConfig.client()` spawns a background renewal goroutine via `gokrb5`'s `enableAutoSessionRenewal` (`client/session.go`). Because the client is constructed with `NewWithKeytab`, the TGT will refresh indefinitely for the lifetime of the `*krbclient.Client` — so long-running pipelines should authenticate without issue, which is great.

However, neither `hdfsReader.Close` nor `hdfsWriter.Close` currently calls `Destroy()` on the `*krbclient.Client` (which stops the renewal goroutine, drops the session, and zeroes credential material), nor closes the underlying `*hdfs.Client`. For a process that brings the component up once and runs forever this is harmless, but on component rebuild (config reload, fatal-error retry) the previous instance's renewal goroutine retains a reference to the `Client` and the in-memory keytab, preventing GC. Stashing the Kerberos client on the reader/writer struct and calling `Destroy()` (plus closing the HDFS client) in `Close` would resolve both.

**Dead branch in `validateInput` (`config.go`)**

`hdfsConfigFromParsed` already invokes `c.kerberos.validate()` before `validateInput` runs, and `validate()` rejects any `data_transfer_protection` value outside `{\"\", \"authentication\", \"integrity\", \"privacy\"}`. The `default:` arm of the `validateInput` switch is therefore unreachable and could be removed for clarity.

**\`user\` field semantics under Kerberos**

In `hdfs/v2`, when `KerberosClient` is set, `User` becomes the effective/proxy user rather than the authenticating identity (which is derived from the principal). The current field description — \"A user ID to connect as.\" — is accurate for the non-Kerberos case but potentially misleading once Kerberos is enabled. A brief note in the description would help operators understand the interaction.

**Origin of the input integrity/privacy restriction**

Would you mind clarifying the source of the \"integrity\"/\"privacy\" limitation for the HDFS input? Knowing whether this stems from `colinmarc/hdfs/v2`, a deliberate Connect-side decision, or a gap to be closed later would be useful for operators — and a short reference in the validation error message itself would aid debugging in the field.

Khalid-Nowaf:
Thanks @Jeffail for the comprehensive review.

I’d like to provide more context on **the origin of the input integrity/privacy restriction**, before jumping to refactor the code based on your review.

Before working on this, I set up a local HDFS cluster with Kerberos enabled and created local integration tests (Im using macOS, and the current integration tests does not support my OS): [Local Integration Test Setup](https://github.com/Khalid-Nowaf/connect/tree/hdfs/integration-testing/integration-testing/hdfs). [Reproduce Failure Script](https://github.com/Khalid-Nowaf/connect/blob/hdfs/integration-testing/integration-testing/hdfs/kerberos/scripts/reproduce-readfile-privacy-failure.sh)

Once the environment was ready, the following test cases failed:

* Kerberos with ```data_transfer_protection: integrity```
    * The Hadoop CLI can read and write files successfully, so the cluster itself appears healthy.
    * ReadDir works.
    * ReadFile fails with: proto: cannot parse invalid wire-format data.
* Kerberos with ```data_transfer_protection: privacy```
    * The Hadoop CLI can read and write files successfully.
    * ReadDir works.
    * ReadFile fails with the same error.

This does not appear to be a Kerberos authentication issue or a problem with the HDFS setup. Instead, it looks like an issue in the upstream Go HDFS client’s DataNode read path when the stream is wrapped with either integrity or privacy protection.

I started debugging the upstream HDFS client, but it required more time than I initially expected, so I treated this as an upstream bug for now.

For that reason, I block these two modes only for the HDFS input path to avoid exposing a configuration that I know will fail.   But still allow them for HDFS output because output operations were tested locally and worked correctly, including with privacy.

Khalid-Nowaf:
I think the issue is with my environment or local test setup. I’ll try using the current integration tests on a Linux machine and add Kerberos (krb5) integration tests as well.

I’ll get back once I’ve tested it again.

PR #4427, opened 2026-05-14
confluent,iceberg,parquet: preserve avro logical types end-to-end across schema_metadata consumers

Jeffail:
## Summary

Closes the GF iceberg-pipeline issue — `time.Time` values from `schema_registry_decode` failing the iceberg shredder with `"expected number value, got timestamp"` against `BIGINT` columns — and cascades the same fix class through every other consumer of the `schema.Common` metadata format. Adds a Claude skill to catch the same drift in future.

The underlying bug class is a **value-vs-metadata mismatch**: the value-side decoder honours an Avro annotation but the metadata-side parser doesn't, so downstream sinks (notably `iceberg`) create the wrong column type, then fail to write the typed value into it. Four distinct annotations all triggered this shape:

1. Field-level (sibling-of-type) `logicalType` — the Java/JDBC Avro idiom emitted by Oracle CDC and most JDBC tooling.
2. Debezium `connect.name` temporal annotations.
3. Avro `duration` logical type.
4. Cross-consumer gaps (parquet, JSON Schema, iceberg Map/Union) found by the post-fix audit.

## Commit narrative

11 commits, intentionally ordered as five narrative phases:

| Phase | Commits |
|---|---|
| **Foundation** (Travis's prior #4399 work) | `1a68504fb` |
| **Customer fix** (field-level lift + upstream bump) | `7ea489942`, `d84edb87a` |
| **Rolling-upgrade support** (opt-in strict mode, then the coerce bridge that needs it, then the gate that makes the whole thing opt-in via `preserve_logical_types`) | `821468858`, `008c03bf8`, `0e771a9db` |
| **Audit-found gaps in other `schema.Common` consumers** | `9879e3b79` (Debezium), `e25f69a5b` (parquet), `25fd6455d` (iceberg Map/Union), `bda5ab073` (JSON Schema) |
| **Drift prevention** | `69275f6c0` (audit skill) |

Each commit is self-contained, passes its own test surface, and reads top-to-bottom as a story.

## What this changes for users

**Default users (`preserve_logical_types: false`):** zero behaviour change.

**Users on `preserve_logical_types: true`:**

- `schema_metadata` now faithfully reports `TIMESTAMP` / `DATE` / `TIME_OF_DAY` / `UUID` for fields whose Avro schemas declare those logical types (including the Java/JDBC sibling-of-type idiom).
- Iceberg auto-creates the correct column types on fresh tables.
- Iceberg coerce-on-write writes `time.Time` values into existing `BIGINT`/`INT` columns by scaling to the wire-equivalent integer using the schema metadata's declared `Unit` — operators with affected tables can upgrade without dropping them. An `INFO`-level log fires once per divergent column for traceability.
- Setting `require_schema_metadata: true` flips the coerce into a hard error — for operators who want loud failure on schema drift rather than silent reconciliation.

**Users on `translate_kafka_connect_types: true`:** Debezium `connect.name` temporal annotations (`io.debezium.time.Date`, `Timestamp`, `Time`, `MicroTimestamp`, `MicroTime`, `NanoTimestamp`, `NanoTime`, `ZonedTimestamp`, `Year`) are now also recognised on the metadata side, matching the value-side translation.

**Parquet-encode users:** schemas declaring `Date`, `TimeOfDay`, `UUID`, or `Map` no longer fail with the generic `unsupported by this processor` error. `Union` and `Null` still error but with messages that name the constraint and point at the upstream coercion remedy. `time.Time` / `time.Duration` values now flow through to TIMESTAMP / DATE / TIME columns without requiring upstream conversion to RFC3339 strings.

**Schema Registry encode (JSON Schema):** `Date`, `TimeOfDay`, and `UUID` now produce the matching JSON Schema draft 2019-09 `format` keywords.

## Drift prevention

`.claude/skills/common-schema-audit/SKILL.md` codifies the audit workflow for every `schema.Common` consumer. Invoke as `/common-schema-audit` to enumerate variants from the live benthos source, find every consumer via `schema.ParseFromAny`, audit type-coverage and value-coercion per consumer in parallel, and produce a matrix with per-consumer GAP / PARTIAL / COVERED verdicts. The skill caught a real iceberg `Map`/`Union` gap on its first run, which is included as a fix in this PR.

## Upstream dependency bump

`github.com/twmb/avro` bumped to `v1.7.3-0.20260513193503-1e5c2a3fc070` (pseudo-version pinning the merge SHA of [twmb/avro#38](https://github.com/twmb/avro/pull/38)). The upstream PR adds the field-level logicalType lift on the value-decode side, so `time.Time` flows through natively for sibling-form schemas. A proper `v1.7.3` tag is pending from Travis; we'll bump again to the tag once it lands.

The bump also picks up [twmb/avro#39](https://github.com/twmb/avro/pull/39)'s spec-compliance pass, which changed the default-mode JSON serialisation of decimal values to the spec-blessed codepoint-mapped string form (matching Java's JsonEncoder). Test expectations updated accordingly; users on `preserve_logical_types: true` are unaffected because our `decimal` CustomType still produces `json.Number`.

## Test plan

- [x] `go test ./internal/impl/iceberg/... ./internal/impl/confluent/... ./internal/impl/parquet/...` — 12 packages, all green.
- [x] `task lint` — `0 issues`.
- [x] Integration tests under `-tags integration` against testcontainers (MinIO + iceberg-rest-fixture) — green.
- [x] Rebased on current `origin/main` (`afff20cc5`).

### DIFF
--- a/internal/impl/iceberg/output_iceberg.go
+++ b/internal/impl/iceberg/output_iceberg.go
@@ -484,6 +484,17 @@ func parseSchemaEvolutionConfig(conf *service.ParsedConfig) (SchemaEvolutionConf
 		}
 	}
 
+	// Parse require_schema_metadata
+	if conf.Contains(ioFieldSchemaEvolution, ioFieldSchemaEvolutionRequireSchemaMetadata) {
+		cfg.RequireSchemaMetadata, err = conf.FieldBool(ioFieldSchemaEvolution, ioFieldSchemaEvolutionRequireSchemaMetadata)
+		if err != nil {
+			return cfg, err
+		}
+		if cfg.RequireSchemaMetadata && cfg.SchemaMetadata == "" {
+			return cfg, fmt.Errorf("%s.%s requires %s.%s to be set", ioFieldSchemaEvolution, ioFieldSchemaEvolutionRequireSchemaMetadata, ioFieldSchemaEvolution, ioFieldSchemaEvolutionSchemaMetadata)
+		}
+	}
+
 	return cfg, nil
 }
 
--- a/internal/impl/iceberg/shredder/shredder.go
+++ b/internal/impl/iceberg/shredder/shredder.go
@@ -80,6 +80,13 @@ type RecordShredder struct {
 	// iceberg's recommended convention and with engines like Spark and Trino
 	// in their default configurations.
 	caseSensitive bool
+	// fieldCommons optionally maps an iceberg field ID to the upstream
+	// schema.Common describing the same field. When present, the leaf
+	// value-conversion step uses Logical params (timestamp unit,
+	// AdjustToUTC, time-of-day unit) to interpret numeric inputs into
+	// typed columns instead of guessing. nil entries fall back to the
+	// pre-schema-metadata behavior — see [convertLeafValue].
+	fieldCommons map[int]*schema.Common
 }
 
 // NewRecordShredder creates a new shredder for the given schema.
@@ -91,6 +98,21 @@ func NewRecordShredder(schema *iceberg.Schema, caseSensitive bool) *RecordShredd
 		schema:        schema,
 		caseSensitive: caseSensitive,
 	}
+}
+
+// SetFieldSchemaMetadata supplies a field-ID → schema.Common map that the
+// leaf value converter consults when it sees a numeric input destined for a
+// time-typed Iceberg column (TIMESTAMP, TIMESTAMPTZ, TIMESTAMP_NS,
+// TIMESTAMP_TZ_NS, DATE, TIME). Without this map, the converter accepts
+// time.Time / time.Duration / int64 directly but cannot disambiguate the
+// unit of a bare numeric millis vs. micros vs. seconds — and historically
+// defaulted to interpreting numbers as Unix seconds. Supplying this map
+// lets the connector tell the shredder "this column was sourced from an
+// Avro timestamp-millis", which the converter then honours.
+//
+// Passing nil clears any previously-set metadata.
+func (rs *RecordShredder) SetFieldSchemaMetadata(byID map[int]*schema.Common) {
+	rs.fieldCommons = byID
 }
 
 // Shred converts a nested record into a sequence of shredded values.
@@ -219,7 +241,7 @@ func (rs *RecordShredder) shredValue(
 
 	default:
 		// Leaf/primitive type.
-		pqVal, err := convertLeafValue(value, typ)
+		pqVal, err := convertLeafValue(value, typ, rs.commonForField(fieldID))
 		if err != nil {
 			return err
 		}
@@ -331,8 +353,9 @@ func (rs *RecordShredder) shredMap(
 		}
 		first = false
 
-		// Shred the key.
-		keyVal, err := convertLeafValue(k, mapType.KeyType)
+		// Shred the key. Map keys are always leaf primitives, never time
+		// types, so the schema-metadata lookup never fires for them.
+		keyVal, err := convertLeafValue(k, mapType.KeyType, nil)
 		if err != nil {
 			return fmt.Errorf("map key: %w", err)
 		}
@@ -399,9 +422,15 @@ func (rs *RecordShredder) shredNull(
 	}
 }
 
-// convertLeafValue converts a Go value to a parquet.Value based on the Iceberg type.
-// This is a stub - full implementation would handle all type conversions.
-func convertLeafValue(value any, typ iceberg.Type) (parquet.Value, error) {
+// convertLeafValue converts a Go value to a parquet.Value based on the
+// Iceberg type. common is the upstream schema.Common describing the same
+// column (or nil); when present it carries logical params consulted by the
+// time/date arms to interpret numeric inputs using the declared unit
+// instead of guessing. Without metadata, the converter accepts time.Time /
+// time.Duration directly and treats bare numerics as already-in-the-target
+// unit (microseconds for time, seconds for timestamp via bloblang
+// ValueAsTimestamp's default — preserves the pre-PR behavior).
+func convertLeafValue(value any, typ iceberg.Type, common *schema.Common) (parquet.Value, error) {
 	if value == nil {
 		return parquet.NullValue(), nil
 	}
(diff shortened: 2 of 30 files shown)


internal/impl/iceberg/output_iceberg.go
```diff
@@ -484,6 +484,17 @@ func parseSchemaEvolutionConfig(conf *service.ParsedConfig) (SchemaEvolutionConf
 		}
 	}
 
+	// Parse require_schema_metadata
+	if conf.Contains(ioFieldSchemaEvolution, ioFieldSchemaEvolutionRequireSchemaMetadata) {
+		cfg.RequireSchemaMetadata, err = conf.FieldBool(ioFieldSchemaEvolution, ioFieldSchemaEvolutionRequireSchemaMetadata)
```
josephwoodward, line 489: minor (non-blocking): could be inlined into the following:

```go
if cfg.RequireSchemaMetadata, err = conf.FieldBool(ioFieldSchemaEvolution, ioFieldSchemaEvolutionRequireSchemaMetadata); err != nil {
    return cfg, err
}

File: internal/impl/iceberg/shredder/shredder.go [hunk 2]
```diff
@@ -93,6 +107,37 @@ func NewRecordShredder(schema *iceberg.Schema, caseSensitive bool) *RecordShredd
...
+// column with no [schema.Common] in the field map is rejected with a
+// per-field error rather than guessed-as-Unix-seconds.
+//
+// Strict mode has no effect on time.Time / time.Duration values, which
+// carry their own unit unambiguously, and no effect on non-time columns.
+//
+// Defaults to off (back-compat). Operators that cannot guarantee schema
+// metadata flows end-to-end can flip this on to fail loudly instead of
+// silently corrupting dates by ~50,000 years.
+func (rs *RecordShredder) SetStrictTemporalMode(on bool) {
```
josephwoodward, line 122: minor (non-blocking): Is a setter necessary, would we just expose `StrictTemporalMode`?

internal/impl/parquet/schema_coercion.go
```diff
@@ -125,6 +110,138 @@ func (encodingCoercionVisitor) visitLeaf(value any, schemaNode parquet.Node) (an
...
+		}
+		return scale(t)
+	case int64:
+		return v, nil
+	case int32:
+		return int64(v), nil
+	case int:
+		return int64(v), nil
+	case float64:
+		return int64(v), nil
```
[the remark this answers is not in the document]
josephwoodward: @Jeffail Has this been addressed?

internal/impl/iceberg/shredder/shredder.go
```diff
@@ -416,10 +468,34 @@ func convertLeafValue(value any, typ iceberg.Type) (parquet.Value, error) {
...
+		// but the upstream schema declares a temporal logical type (Date,
+		// TimeOfDay) the upstream value will be time.Time / time.Duration.
+		// Honour the schema's unit and emit the wire-equivalent integer.
+		// Strict mode (require_schema_metadata=true) disables this bridge
+		// and refuses the type disagreement loudly instead.
+		if n, ok := coerceTemporalToNumeric(value, common); ok {
+			if strictTemporal {
+				return parquet.NullValue(), fmt.Errorf("int column received %T while schema metadata declares type %v; require_schema_metadata=true demands the existing column type match the schema metadata — recreate the table to migrate", value, common.Type)
+			}
+			return parquet.Int32Value(int32(n)), nil
```
josephwoodward, line 481: Int32 column arm casts coerceTemporalToNumeric result to int32 without overflow check

coerceTemporalToNumeric accepts time.Time with schema.Timestamp and returns v.UnixMilli() — for any modern timestamp this is ~1.7×10¹² and vastly exceeds math.MaxInt32 (2.1×10⁹). The Int32 arm casts the result with int32(n) at line 481 with no bounds check. The Int32 arm is intended for Date/TimeOfDay (whose values fit in
int32), but nothing prevents a schema.Timestamp common from entering it, producing a silently truncated year.

Failure scenario: Iceberg table has an INT32 column with schema metadata Type=Timestamp(Millis), value is time.Time("2024-01-15") → UnixMilli() = 1,705,276,800,000 → int32(...) silently wraps to a garbage value with no error.

internal/impl/parquet/schema_coercion.go
```diff
@@ -125,6 +111,157 @@ func (encodingCoercionVisitor) visitLeaf(value any, schemaNode parquet.Node) (an
...
+			if errDate != nil {
+				// Surface both attempts — a malformed bare date like
+				// "2024-13-99" would otherwise yield only the RFC3339
+				// error, which misleadingly suggests the user must add
+				// a time component.
+				return nil, fmt.Errorf("parsing DATE string %q: tried RFC3339 (%v) and YYYY-MM-DD (%v)", v, errRFC, errDate)
+			}
+			t = t2
+		}
+		return int32(t.UTC().Unix() / 86400), nil
```
josephwoodward, line 194: Pre-epoch date string path uses truncating division instead of floor division

The time.Time path at lines 175–180 applies an explicit days-- floor correction for negative remainders. The string parse path at line 194 does int32(t.UTC().Unix() / 86400) with no such correction. Go integer division truncates toward zero, so "1969-12-31T23:59:59Z" (Unix = −1) produces day 0 (1970-01-01) instead of day −1
(1969-12-31).

Failure scenario: parquet_encode receives "1969-12-31T23:59:59Z" for a DATE column → stored as 1970-01-01, silently wrong. The same floor logic already exists in internal/impl/iceberg/shredder/temporal.go but was not applied here. (This divergence is also a code-reuse problem: the floor calculation is now duplicated in three
places and has already drifted.)

josephwoodward:
Claude highlighted this one, not sure how accurate it is but with checking.

---

Map/Array children bypass applyAvroLogicalType entirely (internal/impl/confluent/ecs_avro.go ~line 616)

When a map value type or array item type is expressed as an Avro object ({"type":"long","logicalType":"timestamp-millis"}) rather than a bare string, the Map and Array branches do as["values"].(string) — a bare string type assertion — which panics/errors on the object form. Even when the value is a bare string like "long",
ecsAvroTypeToCommon is called directly with no subsequent applyAvroLogicalType call, so a nested timestamp-millis value type stays as Int64 in schema.Common. This is the same iceberg BIGINT-instead-of-TIMESTAMP column class mismatch this PR was built to prevent, left open for all map/array nested types.

Failure scenario: Avro schema {"type":"map","values":{"type":"long","logicalType":"timestamp-millis"}} → type assertion to string fails at runtime; or if expressed as a bare "long" key, no logical type is applied and the iceberg sink creates a BIGINT column instead of TIMESTAMP.

josephwoodward (approved):

PR #4429, opened 2026-05-14
avro: resolve named-type references in nullable unions

Jeffail:
## Summary

Closes [CON-468](https://redpandadata.atlassian.net/browse/CON-468). The Avro JSON schema parser at `internal/impl/confluent/ecs_avro.go` was dropping named-type references in nullable unions — i.e. the `["null", "Fee"]` idiom where `Fee` is a previously-defined record reused across more than one field — to `schema.Any`, so downstream sinks (notably `iceberg`) saw a VARCHAR column where the customer's Avro schema asked for a nested struct.

[#4380](https://github.com/redpanda-data/connect/pull/4380) closed the **inline-record** form of this shape (`["null", {"type":"record","name":"Fee",...}]`). It did not address the more common **name-reference** form, which any non-trivial Avro schema with a reused record will produce — the Avro spec requires named types to be defined once and referenced by name thereafter.

## What changed

`ecsAvroConfig` now carries a `names map[string]schema.Common` populated as the parser walks the schema. Every `record`, `enum`, and `fixed` definition is registered under both its simple name and its fully-qualified `namespace.name` form, matching Avro's lexical-scope resolution rules. String-form type references consult the map before falling back to the primitive-name lookup, so `"Fee"` resolves to the registered record's full structure instead of collapsing to `schema.Any`.

The two existing optional-union helpers (`ecsAvroIsUnionJustOptional` for primitives, `ecsAvroIsUnionJustOptionalObject` for inline objects) are replaced by a unified `ecsAvroResolveOptionalUnion` that handles all three branch forms (primitive, named reference, inline object) and accepts either ordering — `["null", X]` and `[X, null]` are now equivalent, per CON-468 acceptance criterion 2.

Both the raw-union and lame-union (default `raw_unions: false`) hydrators route through the same `ecsAvroResolveTypeRef` helper, so the fix covers every customer regardless of their `raw_unions` setting.

## Commit narrative

| Commit | Scope |
|---|---|
| `4531c5171` | Names map, unified resolver, named-type registration, raw-union path. Five tests covering CON-468 criterion 2 verbatim. |
| `10000e8a4` | Same fix routed through the lame-union path (default config). One additional test. |

## Scope notes

- **Self-referential records** (a record whose own children reference it by name before its definition completes) remain unsupported. Closing that would require pre-registering a placeholder before walking children. Avro recursive types are rare; we can address as a follow-up if a customer hits it.
- **JSON Schema and Protobuf** decode paths in the Schema Registry processor never produced `schema.Common` metadata to begin with — `store_schema_metadata` is wired only for the Avro arm. That is a pre-existing limitation, not a regression, and a feature request worth its own ticket if a customer asks for it.

## Test plan

- [x] `go test ./internal/impl/confluent/...` — all green, including the six new tests below.
- [x] `task lint` — `0 issues`.
- [x] Verified the existing `TestEcsAvroRawUnionNestedRecord` (the [#4380](https://github.com/redpanda-data/connect/pull/4380) regression) still passes — the inline-record path is unchanged.

New tests:

- `TestEcsAvroRawUnionNullableRecordByName` — `["null", "Fee"]` name reference.
- `TestEcsAvroRawUnionNullableOrderIndependence` (3 sub-tests) — `[X, "null"]` across inline records, primitives, and name references.
- `TestEcsAvroRawUnionNullableRecordNamespaced` — short and fully-qualified namespace references.
- `TestEcsAvroRawUnionNullableRecordNested` — record-containing-record where the inner is a name-reference union.
- `TestEcsAvroLameUnionNameResolution` — same fix verified in the default-config (lame-union) path.

[CON-468]: https://redpandadata.atlassian.net/browse/CON-468?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/impl/confluent/ecs_avro.go
+++ b/internal/impl/confluent/ecs_avro.go
@@ -534,19 +534,14 @@ func ecsAvroHydrateRawUnion(cfg ecsAvroConfig, c *schema.Common, types []any) er
 func ecsAvroHydrateLameUnion(cfg ecsAvroConfig, c *schema.Common, types []any) error {
 	c.Type = schema.Union
 	for i, uObj := range types {
-		var childT schema.Common
-
-		switch ut := uObj.(type) {
-		case string:
-			childT = schema.Common{
-				Name: ut,
-				Type: ecsAvroTypeToCommon(ut),
-			}
-		case map[string]any:
-			var err error
-			if childT, err = ecsAvroFromAnyMap(cfg, ut); err != nil {
-				return fmt.Errorf("union `%v` child '%v': %w", c.Name, i, err)
-			}
+		childT, ok := ecsAvroResolveTypeRef(cfg, uObj)
+		if !ok {
+			return fmt.Errorf("union `%v` child '%v': could not resolve type %T", c.Name, i, uObj)
+		}
+		if s, isStr := uObj.(string); isStr {
+			// Lame-union children keep the type-name as the Common.Name so
+			// the tagged-JSON envelope key matches the Avro wire form.
+			childT.Name = s
 		}
 
 		if childT.Type == schema.Null {
--- a/internal/impl/confluent/ecs_avro.go
+++ b/internal/impl/confluent/ecs_avro.go
@@ -75,15 +75,19 @@ type ecsAvroConfig struct {
 	// for sibling-form Avro by [applyAvroLogicalType].
 	translateKafkaConnectTypes bool
 
-	// names accumulates resolved record/enum/fixed definitions during a
-	// single parse so that string-form references (e.g. "Fee" in
-	// ["null", "Fee"]) can be expanded to their full Common shape. The
-	// map is lazily allocated by ecsAvroParseFromBytes and shared by
-	// reference through recursive ecsAvroFromAnyMap calls; mutations from
-	// sub-trees propagate to later siblings as the Avro spec requires
-	// (named types are lexically scoped from the schema root, and a name
-	// must be defined before it is referenced).
+	// names is the lexical-scope registry of resolved record/enum/fixed
+	// definitions, keyed by Avro fullname (and short name for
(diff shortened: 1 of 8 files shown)


internal/impl/confluent/ecs_avro.go
```diff
@@ -544,6 +620,87 @@ func ecsAvroHydrateLameUnion(cfg ecsAvroConfig, c *schema.Common, types []any) e
...
+	// Common once the walk completes. Mutual recursion across distinct
+	// records is still not supported — the second record's placeholder
+	// does not exist while the first is being walked.
+	typeName, _ := as["type"].(string)
+	fullname, shortName, childNamespace := ecsAvroAssignFullname(cfg.namespace, typeName, as)
+	if fullname != "" {
+		placeholder := ecsAvroPlaceholder(typeName, shortName)
+		cfg.names[fullname] = placeholder
+		if shortName != "" && shortName != fullname {
+			cfg.names[shortName] = placeholder
```
josephwoodward: Claude highlighted this one, not sure if it's a risk? Same for line 647.

ecsAvroFromAnyMap registers every named type under both its fullname (com.a.Foo) and its short name (Foo). When two records share a short name in different
namespaces — e.g. com.a.Fee and com.b.Fee — the second parse overwrites cfg.names["Fee"]. Any unqualified reference to "Fee" that falls through to the bare-name
lookup (i.e. from root scope or a namespace that has no Fee of its own) silently resolves to whichever record was defined last.

The qualified lookup in ecsAvroLookupName (cfg.namespace + "." + ref) saves you when the reference is made from within the correct namespace — but references from
root scope or a third namespace skip that path and hit the stale entry directly.

josephwoodward (approved):

PR #4430, opened 2026-05-15
oracledb_cdc: handle LOB_TRIM redo SQL separately from SELECT_LOB_LOCATOR

josephwoodward:
When updating an existing LOB field via a SQL `UPDATE` command, OracleDB will handle that update in a number of ways.

1. SecureFile (Oracle Free 23c default): emits LOB_WRITE(s) → LOB_TRIM(N). The trim finalises the value after fragments are written, so clearing fragments on LOB_TRIM was the bug. May emit no DML UPDATE — in that case a synthetic UPDATE is synthesized from the assembled accumulator.

2. BASICFILE: emits LOB_TRIM(0) → LOB_WRITE(s) (clear-then-write). The DML UPDATE is always present; fragments arrive after the trim, so LOB_TRIM(0) must be a no-op against the fragment list.

3. BASICFILE out-of-row (DISABLE STORAGE IN ROW): no SELECT_LOB_LOCATOR is emitted. The inferLOBLocator path creates the accumulator from an existing DML event rather than from the locator SQL, and the DML UPDATE is always present for PK matching.

This change ensures we support these three variants and includes a test to verify each. 

image

### DIFF
--- a/internal/impl/oracledb/integration_test.go
+++ b/internal/impl/oracledb/integration_test.go
@@ -1893,4 +1893,159 @@ oracledb_cdc:
 		}
 		require.NoError(t, stream.StopWithin(time.Second*10))
 	})
+
+	t.Run("LOB_TRIM handling for BASICFILE LOB updates", func(t *testing.T) {
+		// BASICFILE storage emits LOB_TRIM(0) → LOB_WRITE(s) on UPDATE (clear-then-write),
+		// whereas SecureFile emits LOB_WRITE(s) → LOB_TRIM(N). The LOB_TRIM(0) must not
+		// discard the fragments written after it, and the assembled value should appear in
+		// the UPDATE CDC event merged via MergeLOBsIntoDMLEvents.
+		require.NoError(t, db.CreateTableWithSupplementalLoggingIfNotExists(t.Context(), "testdb.lobtrimbasic",
+			`CREATE TABLE testdb.lobtrimbasic (
+			id      NUMBER GENERATED ALWAYS AS IDENTITY (NOCACHE) PRIMARY KEY,
+			clobcol CLOB
+		) LOB(clobcol) STORE AS BASICFILE`))
+
+		var batch oracledbtest.Batch
+
+		cfg := fmt.Sprintf(`
+oracledb_cdc:
+  connection_string: %s
+  stream_snapshot: false
+  logminer:
+    lob_enabled: true
+    scn_window_size: 20000
+    backoff_interval: 1s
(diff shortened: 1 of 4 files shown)


Jeffail (approved):
The storage-variant breakdown in the description and the integration tests across SecureFile, BASICFILE, and BASICFILE out-of-row are a meaningful improvement, and the `dbms_lob.trim(loc, N)` finalisation vs. clear-then-write distinction is now correctly drawn. Approving overall.

A few non-blocking points worth surfacing:

**1. Pass 3 fallback (`internal/impl/oracledb/logminer/sqlredo/lob.go:189-201`)** — the code unconditionally selects the most-recent matching event, while the comment describes "exactly one candidate ... otherwise most recent". Largely a documentation cleanup, but there is a latent edge case: if a transaction contains two LOB-only UPDATEs on the same table for different rows *and* Oracle strips PK columns from both WHERE clauses, both accumulators bypass Pass 1/2 and converge on the same Pass 3 target — one row's LOB silently overwrites the other's. Likely rare, but worth either tightening the predicate or adding a warn-log when multiple accumulators target the same event.

**2. `LOB_TRIM(N)` warning predicate (`internal/impl/oracledb/logminer/logminer.go:362-371`)** — the warning only fires when `N > 0 && len(Fragments) == 0`. Two adjacent cases also produce a potentially incorrect assembled value but do not warn:
- `N < M` (writes extend past N): `Assemble` produces M bytes; the real LOB is N bytes.
- `N > M` with `M > 0`: a preserved prefix exists beyond what we have.

`Assemble` is also not currently truncated to `N`. The current scope catches the most blatant case (nothing to publish at all) and is defensible given SecureFile's typical full-rewrite UPDATE pattern; worth noting in the comment that this is an explicit tradeoff rather than a complete check.

**3. Synthetic UPDATE shape (`internal/impl/oracledb/logminer/logminer.go:433-442`)** — the synthesized event publishes only `{<pk_columns>, <lob_column>}`; other row columns are unavailable from redo. Downstream consumers expecting a full-row UPDATE will see a sparse payload. Is sparse-row UPDATE the intended downstream contract? It may be worth documenting, or carrying over column values from a prior INSERT/UPDATE in the same transaction when one is present.

**4. (Minor)** Form A in `OpLobTrim` (`logminer.go:340`) overwrites `state.Accumulators[key]` unconditionally, whereas `OpSelectLobLocator` (line 298) uses `if !exists`. The comment "establish (or reset)" signals intent, but if Form A can ever fire after fragments have already accumulated for the same key, those fragments would be lost. Worth confirming Oracle does not emit that sequence.

None of the above are blockers — happy for this to land.

PR #4432, opened 2026-05-16
gcp: migrate Pub/Sub input and output to the pubsub/v2 SDK

highlyavailable:
Google's Cloud Go SDK moved Pub/Sub to a v2 package, so this migrates the `gcp_pubsub` input and output over to it.

The main issue is that v2 splits the data clients from admin operations. Topic and subscription existence checks, IAM, and CreateSubscription aren't on `Publisher`/`Subscriber` anymore, so those now go through `TopicAdminClient`/`SubscriptionAdminClient` with a `codes.NotFound` check.

One behaviour change worth calling out: v2 dropped synchronous pull, so the input's `sync` field doesn't map to anything now. I left it in and marked it deprecated rather than removing it since this is a Stable connector, and the reader logs a warning if someone sets it. Can rip it out instead if you'd prefer.

Ran the unit tests and the full `gcp_pubsub` integration suite against the emulator, all green.

Happy to take feedback and iterate on any of this.

Fixes #4041

### DIFF
--- a/docs/modules/components/pages/inputs/gcp_pubsub.adoc
+++ b/docs/modules/components/pages/inputs/gcp_pubsub.adoc
@@ -41,7 +41,6 @@ input:
     credentials_json: ""
     subscription: "" # No default (required)
     endpoint: ""
-    sync: false
     max_outstanding_messages: 1000
     max_outstanding_bytes: 1e+09
 ```
@@ -60,7 +59,6 @@ input:
     credentials_json: ""
     subscription: "" # No default (required)
     endpoint: ""
-    sync: false
     max_outstanding_messages: 1000
     max_outstanding_bytes: 1e+09
     create_subscription:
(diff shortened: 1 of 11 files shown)


josephwoodward:
Thank you for this @highlyavailable!

highlyavailable:
Quick heads up on the most recent commit (c4528b6): two of the failing golangci-lint findings were in `internal/impl/aws/s3/bench/read/redpanda-connect/main.go`, which is unrelated to the gcp_pubsub migration in this PR. They look like new modernize rules picking up pre-existing patterns (a `sync.WaitGroup` use that can use `wg.Go`, and an `if/assign` that can use `max`). I included the two-line fix here just to keep this PR's CI green, but if you'd rather see those land in their own PR on main I'm happy to split them out.

squiidz (approved):

highlyavailable:
hey @josephwoodward, anything blocking this one? It's approved and CI is green... the only red check is the claude-review bot, which errored out on a missing API key rather than finding anything

highlyavailable:
Hi @squiidz, following up to see if anything is blocking this

PR #4433, opened 2026-05-16
mysql_cdc: smooth IAM auth setup for RDS

david-yu:
## Summary

Two small papercuts surfaced while validating the [redpanda-operator Pipeline CRD against RDS with native IAM database authentication](https://github.com/redpanda-data/redpanda-operator/pull/1337#issuecomment-4464604928). Both can be fixed inside `mysql_cdc` itself.

- **`AllowCleartextPasswords` is now set automatically when `aws.enabled` is true.** The IAM token is sent to MySQL as the user's password, which the Go MySQL driver refuses unless this flag is set. Users no longer need to remember `&allowCleartextPasswords=1` in the DSN. TLS on the DSN is what protects the token in transit — the field docstring calls this out.
- **`aws.endpoint` is more forgiving.** The AWS SDK's RDS token signer requires `host:port`, but the field was documented as a bare hostname. Connect now derives the port from the DSN when the user omits it (falling back to `3306`), and falls back to the DSN address entirely when `aws.endpoint` is empty. Field is now optional with a `""` default.

Refactored DSN parsing and IAM driver-flag toggling into small testable helpers (`parseMySQLConfig` / `applyIAMDriverDefaults`) and added unit tests for `normalizeIAMEndpoint`.

The N9 friction point from the operator PR comment ("`REQUIRE SSL;` as a standalone statement") is purely user-side SQL — no Connect change applies.

### 1. `allowCleartextPasswords=1` in the DSN

**Before** — without the flag, the driver refuses to send the IAM token and the pipeline crashes:

```yaml
input:
  mysql_cdc:
    dsn: "${MYSQL_USER}@tcp(${MYSQL_HOST}:3306)/shop?tls=skip-verify"
    aws:
      enabled: true
      region: us-east-2
      endpoint: "${MYSQL_HOST}:3306"
    tables: [orders]
```
```
ERROR  mysql.Connect: this user requires clear text authentication.
       If you still want to use it, please add 'allowCleartextPasswords=1' to your DSN
```

**Workaround users had to apply:**
```diff
- dsn: "${MYSQL_USER}@tcp(${MYSQL_HOST}:3306)/shop?tls=skip-verify"
+ dsn: "${MYSQL_USER}@tcp(${MYSQL_HOST}:3306)/shop?tls=skip-verify&allowCleartextPasswords=1"
```

**After** — Connect sets the flag itself whenever `aws.enabled` is `true`, so the original config just works. TLS on the DSN keeps the token off the wire in cleartext.

### 2. `aws.endpoint` accepts a bare hostname

**Before** — passing a hostname without a port produces a confusing error from inside the AWS SDK:

```yaml
input:
  mysql_cdc:
    dsn: "user@tcp(mydb.abc123.us-east-1.rds.amazonaws.com:3306)/shop"
    aws:
      enabled: true
      region: us-east-2
      endpoint: mydb.abc123.us-east-1.rds.amazonaws.com   # missing :3306
```
```
ERROR  building IAM auth token: the provided endpoint is missing a port,
       or the provided port is invalid
```

**After** — the port from the DSN is spliced in (or 3306 as a final fallback), so the same config works. Better still, `aws.endpoint` can now be omitted entirely:

```yaml
input:
  mysql_cdc:
    dsn: "user@tcp(mydb.abc123.us-east-1.rds.amazonaws.com:3306)/shop"
    aws:
      enabled: true
      region: us-east-2
      # endpoint inherited from the DSN
```

## Backwards compatibility

**Fully backwards compatible — no breaking changes.** Every previously valid config continues to behave identically.

| Existing config | Before | After |
| --- | --- | --- |
| `aws.enabled: true` + `allowCleartextPasswords=1` in DSN | works | works (the flag is idempotent — setting it twice does nothing extra) |
| `aws.enabled: false` (or unset) | unchanged | unchanged — `AllowCleartextPasswords` is only touched when IAM is enabled |
| `aws.endpoint: "host:3306"` | works | works — values with an explicit port are passed through untouched |
| `aws.endpoint: "host"` (bare hostname) | errored at startup | now works (port derived from DSN, fallback `3306`) |
| `aws.endpoint` omitted | rejected at config parse (field was required) | accepted — falls back to the DSN's address |

The `aws.endpoint` schema change (`required` → `optional` with `""` default) is a relaxation: existing CRDs / config files that set the field continue to parse and validate. Nothing that previously parsed will now fail.

### Implications worth flagging for reviewers

- **Security model is unchanged.** `AllowCleartextPasswords` only lets the driver respond to a server's cleartext-auth challenge — it doesn't downgrade TLS or weaken anything on the connection. RDS IAM auth always pairs with TLS in production; the `dsn` field's updated docstring spells out the recommended setup (`tls=true` + RDS CA bundle in prod, `tls=skip-verify` for local tests).
- **No change for non-IAM users.** If `aws.enabled` is false or unset, the new code paths (cleartext toggle, endpoint normalization) are skipped entirely.
- **No DSN rewrite is visible to the user.** The cleartext flag is set on the parsed `mysql.Config` before the driver opens the connection; nothing is written back into the user's DSN string.
- **No new dependencies.** Only standard-library additions (`net`, `errors`, `strings`).

## Test plan

- [x] \`go test ./internal/impl/mysql/...\`
- [x] \`golangci-lint run ./internal/impl/mysql/...\`
- [ ] Integration test against RDS with \`aws.enabled: true\` and a bare-hostname \`aws.endpoint\`

### DIFF
--- a/docs/modules/components/pages/inputs/mysql_cdc.adoc
+++ b/docs/modules/components/pages/inputs/mysql_cdc.adoc
@@ -422,7 +422,6 @@ The MySQL endpoint as `host:port` (for example `mydb.abc123.us-east-1.rds.amazon
 
 *Default*: `""`
 
-
 === `aws.id`
 
 The ID of credentials to use.
--- a/docs/modules/components/pages/inputs/mysql_cdc.adoc
+++ b/docs/modules/components/pages/inputs/mysql_cdc.adoc
@@ -86,7 +86,7 @@ input:
     aws:
       enabled: false
       region: "" # No default (optional)
-      endpoint: "" # No default (required)
+      endpoint: ""
       id: "" # No default (optional)
       secret: "" # No default (optional)
       token: "" # No default (optional)
@@ -137,7 +137,7 @@ The type of MySQL database to connect to.
 
 === `dsn`
 
-The DSN of the MySQL database to connect to.
+The DSN of the MySQL database to connect to. When `aws.enabled` is `true`, the Go MySQL driver's `allowCleartextPasswords=1` flag is set automatically so the IAM auth token can be transmitted; pair it with TLS (`tls=true` plus the RDS CA bundle, or `tls=skip-verify` for non-production) to protect the token on the wire.
 
 
 *Type*: `string`
@@ -147,6 +147,8 @@ The DSN of the MySQL database to connect to.
 # Examples
 
 dsn: user:password@tcp(localhost:3306)/database
+
+dsn: user@tcp(mydb.abc123.us-east-1.rds.amazonaws.com:3306)/database?tls=true
 ```
 
 === `tables`
@@ -413,10 +415,12 @@ The AWS region where the MySQL instance is located. If no region is specified th
 
 === `aws.endpoint`
 
-The MySQL endpoint hostname (e.g., mydb.abc123.us-east-1.rds.amazonaws.com).
+The MySQL endpoint as `host:port` (for example `mydb.abc123.us-east-1.rds.amazonaws.com:3306`). The AWS IAM token signer requires the port; if it is omitted the port from the DSN is used, falling back to `3306`. When this field is empty the address from the DSN is used.
 
 
 *Type*: `string`
+
+*Default*: `""`
 
 
 === `aws.id`
--- a/internal/impl/mysql/aws/aws.go
+++ b/internal/impl/mysql/aws/aws.go
@@ -16,7 +16,10 @@ package aws
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net"
+	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	awsconfig "github.com/aws/aws-sdk-go-v2/config"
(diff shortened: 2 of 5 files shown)


david-yu:
Heads-up on the red checks on the first run:

- **`test → Docs`**: my hand-patched `.adoc` had one extra blank line vs what `task docs` produces. Pushed `2597bc5` to align with `docs_gen` output; CI is re-running.
- **`golangci-lint`**: failures are in `internal/impl/aws/s3/bench/read/redpanda-connect/main.go` (`modernize: waitgroup` / `minmax`) and are pre-existing on `main` — the latest [Test run on `main`](https://github.com/redpanda-data/connect/actions/runs/25932040554) since `bed70d9e7` (`bench/docs: add S3 benchmark (#4369)`) fails on the exact same two lines. Not introduced by this PR.
- **`claude-review`**: fails with `Could not fetch an OIDC token` — the action can't get an `id-token: write` from a forked-PR runner. Infra-level, not code.

Let me know if you'd like me to tack the two-line `modernize` fix onto this PR to unblock the lint job here, or keep this PR scoped to `mysql_cdc` and address the bench file in a separate chore.

david-yu:
@claude review

PR #4443, opened 2026-05-20
BigQuery: pending streams partitioned

squiidz:


internal/impl/gcp/enterprise/bigquery/table_creator.go
```diff
@@ -0,0 +1,160 @@
...
+
+// tableCreator builds bigquery.TableMetadata from the parsed config and
+// creates BigQuery tables idempotently. It is invoked lazily by schemaResolver
+// when Metadata returns 404 and auto_create_table is enabled, so it handles
+// both static and interpolated table names.
+type tableCreator struct {
+	schema          bigquery.Schema
+	partitioning    *bigquery.TimePartitioning
+	clustering      *bigquery.Clustering
+	requirePartFltr bool
```
josephwoodward: Is it worth just calling this `requirePartitionFilter` similar to how it's defined in bigquery.TableMetadata?

squiidz: agree, I've change it

internal/impl/gcp/enterprise/bigquery/write_mode.go
```diff
@@ -0,0 +1,161 @@
...
+			return fmt.Errorf("AppendRows ack at offset %d: %w", offset, err)
+		}
+		offset += int64(len(chunk))
+	}
+
+	if _, err := ms.Finalize(ctx); err != nil {
+		return fmt.Errorf("FinalizeWriteStream %s: %w", stream.GetName(), err)
+	}
+
+	if _, err := p.storage.BatchCommitWriteStreams(ctx, &storagepb.BatchCommitWriteStreamsRequest{
```
josephwoodward: Do we need to handle `StreamErrors` on the response here?

internal/impl/gcp/enterprise/bigquery/write_mode.go
```diff
@@ -0,0 +1,161 @@
...
+		Parent:       parent,
+		WriteStreams: []string{stream.GetName()},
+	}); err != nil {
+		return fmt.Errorf("BatchCommitWriteStreams %s: %w", stream.GetName(), err)
```
josephwoodward: Given this is a customer facing error, would it be worth adding a little more context to it so it's not just the calling function?

internal/impl/gcp/enterprise/bigquery/output.go
```diff
@@ -278,6 +550,8 @@ func newBQWAMetrics(m *service.Metrics) *bqwaMetrics {
 		retries:                 m.NewCounter("bigquery_write_api_retries_total"),
 		schemaEvolutions:        m.NewCounter("bigquery_write_api_schema_evolutions_total"),
 		schemaEvolutionFailures: m.NewCounter("bigquery_write_api_schema_evolutions_failures_total"),
+		cachedStreams:           m.NewGauge("bigquery_write_api_cached_streams"),
```
josephwoodward: Perhaps `bigquery_write_api_streams_cached` for discoverability and alignment with existing`bigquery_write_api_streams_*` metrics?

internal/impl/gcp/enterprise/bigquery/output.go
```diff
@@ -515,6 +818,24 @@ func (o *bigQueryWriteAPIOutput) WriteBatch(ctx context.Context, batch service.M
...
+	// Create/Append/Finalize/Commit lifecycle per batch. Schema-mismatch and
+	// permanent errors flow through handleWriteError just like the default
+	// path so evolution + DLQ semantics stay consistent.
+	if o.conf.WriteMode == "pending_stream" {
+		if pending == nil {
+			return service.ErrNotConnected
+		}
+		parent := o.tableCacheKey(projectID, tableID)
+		if err := pending.Write(ctx, parent, swd.descriptorProto, rows); err != nil {
+			return o.handleWriteError(ctx, client, err, batch, tableID, swd.descriptor, "pending stream write")
```
josephwoodward, line 836: Should we be calling `o.evictStream(cacheKey)` before `handleWriteError` here to evict stale descriptorProto?

internal/impl/gcp/enterprise/bigquery/writer.go
```diff
@@ -0,0 +1,161 @@
+// Copyright 2026 Redpanda Data, Inc.
+//
+// Licensed as a Redpanda Enterprise file under the Redpanda Community
+// License (the "License"); you may not use this file except in compliance with
+// the License. You may obtain a copy of the License at
+//
+// https://github.com/redpanda-data/connect/blob/main/licenses/rcl.md
+
+package bigquery
```
josephwoodward, line 9: Instead of `write_mode.go`, what about just `writer.go`?

internal/impl/gcp/enterprise/bigquery/write_mode.go
```diff
@@ -0,0 +1,161 @@
...
+
+// chunkRowsByBytes splits a row slice into smaller slices such that each
+// chunk's total payload byte size stays under maxBytes. A single row that
+// already exceeds maxBytes is emitted in its own chunk; the API will reject
+// it server-side and the caller surfaces that as a permanent error.
+func chunkRowsByBytes(rows [][]byte, maxBytes int) [][][]byte {
+	if len(rows) == 0 {
+		return nil
+	}
+	var chunks [][][]byte
```
josephwoodward: These could be declared in a chunk, ie:

```go
var (
...
)

File: internal/impl/gcp/enterprise/bigquery/table_creator.go [hunk 8]
```diff
@@ -0,0 +1,160 @@
...
+	if err == nil {
+		tc.log.Infof("Auto-created BigQuery table %s.%s", datasetID, tableID)
+		return nil
+	}
+	var apiErr *googleapi.Error
+	if errors.As(err, &apiErr) && apiErr.Code == http.StatusConflict {
+		// Another writer created the table first — that's the contract.
+		return nil
+	}
+	return fmt.Errorf("creating table %s.%s: %w", datasetID, tableID, err)
```
josephwoodward: Could this be restructured to the following, handling the error case first? (as is generally most common in Go):

```go
if err := client.Dataset(datasetID).Table(tableID).Create(ctx, meta); err != nil {
	var apiErr *googleapi.Error
	if errors.As(err, &apiErr) && apiErr.Code == http.StatusConflict {
		// Another writer created the table first — that's the contract.
		return nil
	}
	return fmt.Errorf("creating table %s.%s: %w", datasetID, tableID, err)
}

tc.log.Infof("Auto-created BigQuery table %s.%s", datasetID, tableID)
return nil
```

josephwoodward: Thinking about this further, it might be better to split the error handling up. First to see if it can convert to a `googleapi.Error` type, then checking the status code. If it's an API error then it'll have a lot more context of the failure which would be good to include in the error message (such as status code, and Google specific diagnostics).

josephwoodward (approved):

PR #4453, opened 2026-05-25
BigQuery: cdc upsert

squiidz:
Implements CON-395 on top of the pending-streams branch. Two new
opt-in write modes layered onto gcp_bigquery_write_api: upsert and
upsert_delete. Both inject BigQuery's _CHANGE_TYPE pseudo-column
per row and optionally _CHANGE_SEQUENCE_NUMBER for out-of-order
resolution.

Write modes
- write_mode: upsert — UPSERT-only rows.
- write_mode: upsert_delete — UPSERT and DELETE rows.
- Both use the default stream (the only stream type BigQuery's CDC
  contract supports). pending_stream + CDC is rejected at parse time.

Pseudo-column injection
- change_type: Bloblang field resolving per row to UPSERT or DELETE
  (case-insensitive; INSERT is rejected — BigQuery doesn't allow
  mixing INSERT and UPSERT/DELETE in the same write).
- change_sequence_number: optional Bloblang field. Validated against
  BigQuery's 1-to-4-sections-of-1-to-16-hex-chars format.
- Injection happens at the JSON byte level via map[string]json.RawMessage
  so user values (int-as-string, base64 bytes, RFC3339 timestamps)
  pass through byte-exact. No round-trip through map[string]any that
  would corrupt int64 strings above 2^53.
- Per-row validation failures route to DLQ via BatchError.Failed
  so a single bad row never poisons the batch. rowsFailed metric
  counts every reject path.

Primary keys
- New primary_keys config (≤16 columns). Required when auto_create_table
  is true; otherwise the connector reads the table's declared PKs via
  TableConstraints.PrimaryKey and uses those.
- When both config and table-declared PKs are present, they must match
  (same columns, same order). Mismatches fail loudly at createStream.
- CDC mode without any PK source fails at createStream with a message
  pointing users to ALTER TABLE … ADD PRIMARY KEY.

Descriptor wrapping
- _CHANGE_TYPE and _CHANGE_SEQUENCE_NUMBER (when configured) are
  appended as proto fields at max(existing)+1, +2 in the user's
  descriptor. Wrapped descriptors live on the per-stream
  streamWithDescriptor so concurrent createStream calls cannot race.
- Stream cache keying includes the write_mode suffix so CDC and
  non-CDC pipelines pointed at the same table cannot share a stream.

message_format
- CDC modes require message_format: json. The protobuf path is
  rejected at parse time (and defensively at runtime) because the
  injection works on JSON bytes.

Docs and tests
- New migration guide at docs/modules/components/pages/outputs/
  bigquery_cdc_migration.adoc covering config translation, schema
  requirements, snapshot vs streaming tradeoffs, and the DML
  restriction on CDC tables. Bidirectional xref with the output page.
- Unit tests cover: validateChangeType, validateChangeSequenceNumber,
  validateCDCPrimaryKeys, wrapDescriptorForCDC, injectCDCJSON (incl.
  byte-fidelity for >2^53 integers), cdcInjector.validateAndResolveCDC,
  extractPrimaryKeysFromMetadata, tableCreator.buildMetadata with PKs,
  and config parsing for the five new fields plus cross-field
  validation.
- Integration tests stubbed (t.Skip) — goccy emulator does not
  implement CDC; real-BQ verification deferred.

### DIFF
--- a/internal/impl/gcp/enterprise/bigquery/cdc.go
+++ b/internal/impl/gcp/enterprise/bigquery/cdc.go
@@ -1,0 +1,192 @@
+// Copyright 2026 Redpanda Data, Inc.
+//
+// Licensed as a Redpanda Enterprise file under the Redpanda Community
+// License (the "License"); you may not use this file except in compliance with
+// the License. You may obtain a copy of the License at
+//
+// https://github.com/redpanda-data/connect/blob/main/licenses/rcl.md
+
+package bigquery
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"regexp"
+	"strings"
+
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/descriptorpb"
+
+	"github.com/redpanda-data/benthos/v4/public/service"
+)
+
+// validateChangeType normalises and validates a Bloblang-resolved _CHANGE_TYPE
+// value against BigQuery's CDC contract. BigQuery only accepts UPSERT and
+// DELETE; INSERT is expressed by omitting the pseudo-column entirely, and
+// mixing INSERT with UPSERT/DELETE in the same write is unsupported.
+// allowDelete is true for write_mode=upsert_delete.
+func validateChangeType(raw string, allowDelete bool) (string, error) {
+	v := strings.ToUpper(strings.TrimSpace(raw))
+	switch v {
+	case "UPSERT":
+		return v, nil
+	case "DELETE":
+		if !allowDelete {
+			return "", errors.New("change_type DELETE is only valid when write_mode is upsert_delete")
+		}
+		return v, nil
+	case "":
+		return "", errors.New("change_type resolved to an empty value")
+	default:
+		return "", fmt.Errorf("change_type %q is not UPSERT or DELETE", raw)
+	}
+}
+
+// changeSequenceNumberPattern matches BigQuery's _CHANGE_SEQUENCE_NUMBER format:
+// 1 to 4 sections of 1 to 16 hexadecimal characters each, separated by '/'.
+var changeSequenceNumberPattern = regexp.MustCompile(`^[0-9A-Fa-f]{1,16}(/[0-9A-Fa-f]{1,16}){0,3}$`)
+
+// validateChangeSequenceNumber checks that raw conforms to BigQuery's CDC
+// sequence-number format. BigQuery rejects values that do not match this
+// format with INVALID_ARGUMENT at AppendRows time, so we validate per-message
+// before the network round trip.
+func validateChangeSequenceNumber(raw string) error {
+	if !changeSequenceNumberPattern.MatchString(raw) {
+		return fmt.Errorf("change_sequence_number %q does not match BigQuery format (1-4 sections of 1-16 hex chars separated by /)", raw)
+	}
+	return nil
+}
+
+// wrapDescriptorForCDC returns a copy of base with `_CHANGE_TYPE` and
+// optionally `_CHANGE_SEQUENCE_NUMBER` appended as STRING fields at the next
+// free field numbers (max existing + 1, +2). The base descriptor is left
+// untouched. Injection is name-based (see injectCDCJSON), so the field numbers
+// don't need to be returned — they only need to be valid and unused.
+func wrapDescriptorForCDC(base *descriptorpb.DescriptorProto, withSequenceNumber bool) (*descriptorpb.DescriptorProto, error) {
+	if base == nil {
+		return nil, errors.New("wrapDescriptorForCDC: nil base descriptor")
+	}
+	var maxNum int32
+	for _, f := range base.Field {
+		if n := f.GetNumber(); n > maxNum {
+			maxNum = n
+		}
+	}
+	changeTypeFieldNumber := maxNum + 1
+	wrapped := proto.Clone(base).(*descriptorpb.DescriptorProto)
+	wrapped.Field = append(wrapped.Field, &descriptorpb.FieldDescriptorProto{
+		Name:   new("_CHANGE_TYPE"),
+		Number: new(changeTypeFieldNumber),
+		Type:   descriptorpb.FieldDescriptorProto_TYPE_STRING.Enum(),
+		Label:  descriptorpb.FieldDescriptorProto_LABEL_OPTIONAL.Enum(),
+	})
+	if withSequenceNumber {
+		changeSeqFieldNumber := maxNum + 2
+		wrapped.Field = append(wrapped.Field, &descriptorpb.FieldDescriptorProto{
+			Name:   new("_CHANGE_SEQUENCE_NUMBER"),
+			Number: new(changeSeqFieldNumber),
+			Type:   descriptorpb.FieldDescriptorProto_TYPE_STRING.Enum(),
+			Label:  descriptorpb.FieldDescriptorProto_LABEL_OPTIONAL.Enum(),
+		})
+	}
+	return wrapped, nil
+}
+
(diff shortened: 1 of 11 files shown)


internal/impl/gcp/enterprise/bigquery/cdc.go
```diff
@@ -0,0 +1,192 @@
...
+		})
+	}
+	return wrapped, nil
+}
+
+// cdcInjector holds the immutable per-output CDC configuration: validated
+// Bloblang fields and the allowDelete flag. Safe to share across concurrent
+// WriteBatch goroutines because all fields are read-only after construction.
+type cdcInjector struct {
+	log         *service.Logger
```
josephwoodward: I see this isn't used anywhere, is it worth removing it until we need it?

internal/impl/gcp/enterprise/bigquery/cdc.go
```diff
@@ -0,0 +1,192 @@
...
+		if _, exists := raw["_CHANGE_SEQUENCE_NUMBER"]; exists {
+			return nil, errors.New("payload already contains _CHANGE_SEQUENCE_NUMBER; remove it before CDC injection")
+		}
+		encodedSeq, err := json.Marshal(changeSeq)
+		if err != nil {
+			return nil, fmt.Errorf("encoding change_sequence_number: %w", err)
+		}
+		raw["_CHANGE_SEQUENCE_NUMBER"] = encodedSeq
+	}
+	return json.Marshal(raw)
```
josephwoodward, line 141: Given this is a `map[string]json.RawMessage` does it matter that'll be in a non-deterministic byte order when unmarshalled?

internal/impl/gcp/enterprise/bigquery/cdc.go
```diff
@@ -0,0 +1,192 @@
...
+// non-empty) into the JSON object bytes without round-tripping through a Go
+// value tree. Using json.RawMessage preserves the exact byte representation of
+// every user field — int64 strings stay strings, NUMERIC strings keep their
+// precision, base64 BYTES stay base64. Returns an error if jsonBytes is not a
+// JSON object, or if the payload already carries a key the injector would
+// otherwise overwrite (forces explicit user remediation rather than silent
+// data loss).
+func injectCDCJSON(jsonBytes []byte, changeType, changeSeq string) ([]byte, error) {
+	dec := json.NewDecoder(bytes.NewReader(jsonBytes))
+	var raw map[string]json.RawMessage
```
josephwoodward: nit (non-blocker) Perhaps the following for clarity?

```go
	var (
		reader   = bytes.NewReader(jsonBytes)
		raw map[string]json.RawMessage
	)
	if err := json.NewDecoder(reader).Decode(&raw); err != nil {
		return nil, fmt.Errorf("parsing JSON: %w", err)
	}

```

internal/impl/gcp/enterprise/bigquery/integration_test.go
```diff
@@ -498,3 +498,29 @@ gcp_bigquery_write_api:
...
 	}, 30*time.Second, 500*time.Millisecond)
 }
+
+// CDC integration tests require a real BigQuery instance. The goccy emulator
+// does not implement the _CHANGE_TYPE / _CHANGE_SEQUENCE_NUMBER pseudo-columns,
+// PRIMARY KEY constraints, or the UPSERT/DELETE semantics that CDC mode
+// depends on. These tests are gated with t.Skip so the package compiles
+// against the real BQ API and so a future real-BQ test job can flip the gate.
+
+func TestIntegrationCDCUpsert(t *testing.T) {
```
josephwoodward: question (non-blocker): Would it make sense to move these into one logical "change" test with individual subtest? And are they covered elsewhere?

squiidz: Good call, collapsed them into a single TestIntegrationCDC with upsert / upsert_delete / out_of_order / mixed_batch subtests.

internal/impl/gcp/enterprise/bigquery/cdc.go
```diff
@@ -0,0 +1,189 @@
...
+	if len(configPKs) > 0 && len(tablePKs) > 0 {
+		if !equalStringSlices(configPKs, tablePKs) {
+			return fmt.Errorf("primary_keys config %v does not match table %q PRIMARY KEY %v",
+				configPKs, tableID, tablePKs)
+		}
+	}
+	return nil
+}
+
+func equalStringSlices(a, b []string) bool {
```
josephwoodward: This can be replaced with `slices.Equal`, it's [exactly the same implementation](https://cs.opensource.google/go/go/+/refs/tags/go1.26.4:src/slices/slices.go;l=20).

Do they need to be sorted before comparison? (`slices.Equal` compares order)

internal/impl/gcp/enterprise/bigquery/output.go
```diff
@@ -890,6 +1014,137 @@ func permanentBatchError(batch service.MessageBatch, err error) error {
...
+	result, err := swd.stream.AppendRows(ctx, rows)
+	if err != nil {
+		o.evictStream(cacheKey)
+		return o.handleWriteError(ctx, client, err, batch, tableID, swd.descriptor, "appending CDC rows")
+	}
+	resp, err := result.FullResponse(ctx)
+	if err != nil {
+		o.evictStream(cacheKey)
+		return o.handleWriteError(ctx, client, err, batch, tableID, swd.descriptor, "waiting for CDC append result")
+	}
```
josephwoodward, line 1131: Claude has reported the following that sounds like it'd be worth checking out:

Bug 1 - Schema evolution in CDC mode tries to ALTER TABLE to add _CHANGE_TYPE (medium severity)

output.go:1113–1122

When AppendRows or FullResponse returns SCHEMA_MISMATCH_EXTRA_FIELDS in CDC mode, writeBatchCDC calls:

return o.handleWriteError(ctx, client, err, batch, tableID, swd.descriptor, "appending CDC rows")

swd.descriptor is the wrapped descriptor — it contains the user's fields plus _CHANGE_TYPE (and _CHANGE_SEQUENCE_NUMBER when configured). handleWriteError passes this to
Evolve, which calls diffMissingColumns(md, meta.Schema). Since _CHANGE_TYPE is not a real column in the BigQuery table schema, it appears as "missing" and Evolve issues an
ALTER TABLE to add it.

That call will either fail (BigQuery rejects the reserved CDC name) or silently add a regular _CHANGE_TYPE STRING column, which likely breaks the CDC contract. Either way,
handleWriteError returns a permanent BatchError and every row in the batch goes to the DLQ. Any future batch that needs genuine schema evolution in CDC mode will hit the
same failure.

Fix: pass the resolver's unwrapped descriptor (rs.messageDescriptor) to handleWriteError from the CDC path, or strip fields whose names start with _CHANGE_ before calling
Evolve. The simplest approach is to store the unwrapped descriptor alongside the wrapped one in streamWithDescriptor (or derive it by looking up rs from the resolver cache)
and pass that to handleWriteError:

// output.go — writeBatchCDC error paths
return o.handleWriteError(ctx, client, err, batch, tableID, swd.descriptor, "appending CDC rows")
//                                                            ^^^^^^^^^^^^^
// should be the unwrapped descriptor so Evolve doesn't attempt to add _CHANGE_TYPE

internal/impl/gcp/enterprise/bigquery/output.go
josephwoodward: minor (nit): should this be `bqWriteAPIConfig` to keep it consistent with the names above?

internal/impl/gcp/enterprise/bigquery/output.go
squiidz: fixed

josephwoodward (approved):

PR #4462, opened 2026-05-28
 Feature/sap hana

ness-david-dedu:


ness-david-dedu (commented):
**Non-inline issues**

**`internal/impl/saphana/bench/ngdbc.jar`** — A proprietary SAP HANA JDBC driver binary is committed to the repository. The README already documents how to obtain it from the SAP Software Downloads Center; ship the instructions, not the file. Remove it and add it to `.gitignore`.

**`docs/benchmark-results/SUMMARY.md`** — The SAP HANA connector adds a results file (`docs/benchmark-results/sap-hana.md`) but `SUMMARY.md` is not updated. Per [`docs/benchmarking.md`](https://github.com/redpanda-data/connect/blob/442127ecbd66374b4d3f622032663c920dfed57d/docs/benchmarking.md#L322-L326): "When adding a new benchmark suite — update the table in this document, and update `docs/benchmark-results/SUMMARY.md`."

**`docs/benchmarking.md`** — The "Existing Benchmarks" table in this file does not include the SAP HANA entry. The same maintenance rule applies: add a row for the new suite.

internal/impl/saphana/input_sap_hana.go
```diff
@@ -0,0 +1,543 @@
...
+	s.db = db
+	s.schemas = newSchemaCache(db, s.log)
+	s.log.Debug("Connected to SAP HANA.")
+	return nil
+}
+
+// tableRef returns a quoted table reference: "SCHEMA"."TABLE" or just "TABLE".
+func (s *sapHANAInput) tableRef() string {
+	if s.schemaName != "" {
+		return `"` + s.schemaName + `"."` + s.tableName + `"`
```
ness-david-dedu: **SQL injection — unescaped schema/table identifiers**

`schemaName` and `tableName` are wrapped in double-quotes but internal `"` characters are not escaped. Standard SQL identifier escaping requires doubling them (`"` → `""`). A user-controlled value containing a literal double-quote can terminate the identifier and inject arbitrary SQL.

Fix: apply `strings.ReplaceAll(s.schemaName, "\"", "\"\"")` (and the same for `tableName`) before concatenating into the identifier.

internal/impl/saphana/input_sap_hana.go
```diff
@@ -0,0 +1,543 @@
...
+func (s *sapHANAInput) openRows(ctx context.Context) (*sql.Rows, error) {
+	switch s.mode {
+	case shModeBulk:
+		q := `SELECT * FROM ` + s.tableRef()
+		return s.db.QueryContext(ctx, q)
+
+	case shModeIncrementing:
+		if s.hwm == "" {
+			q := `SELECT * FROM ` + s.tableRef() +
+				` ORDER BY "` + s.incrementingCol + `"`
```
ness-david-dedu: **SQL injection — unescaped column name identifiers**

`incrementingCol` (here and at lines 303, 325, 329–331) and `timestampCol` (lines 313, 317, 323–331) are interpolated directly into the SQL string wrapped in double-quotes without escaping internal `"` characters. A value containing `"` can break out of the identifier and inject SQL.

Fix: escape each column name with `strings.ReplaceAll(col, "\"", "\"\"")` before building the query string.

internal/impl/saphana/output_sap_hana.go
```diff
@@ -0,0 +1,205 @@
...
+		return nil, err
+	}
+	if len(columns) == 0 {
+		return nil, fmt.Errorf("field %q must not be empty", shoFieldColumns)
+	}
+	o.numCols = len(columns)
+
+	quotedCols := make([]string, len(columns))
+	for i, c := range columns {
+		quotedCols[i] = `"` + c + `"`
```
ness-david-dedu: **SQL injection — column names not identifier-escaped**

Column names are wrapped in double-quotes but internal `"` characters are not escaped (cf. `strings.ReplaceAll`). A column name containing `"` can terminate the identifier and inject SQL.

Fix: `quotedCols[i] = "\"" + strings.ReplaceAll(c, "\"", "\"\"") + "\""`

internal/impl/saphana/output_sap_hana.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	placeholders := strings.Repeat("?,", len(columns))
+	placeholders = placeholders[:len(placeholders)-1]
+	o.insertSQL = fmt.Sprintf(`INSERT INTO %s (%s) VALUES (%s)`,
```
ness-david-dedu: **SQL injection — `table` field is completely unquoted in INSERT SQL**

The `table` value is interpolated raw into the INSERT statement via `%s` with no quoting or escaping. A value containing SQL metacharacters (`;`, `--`, `'`, etc.) injects arbitrary SQL. The field should be split into schema and table components and each part should be quoted with escaped identifiers (same fix as `tableRef()` in the input).

go.mod
```diff
@@ -230,6 +230,7 @@ require (
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric v0.55.0 // indirect
 	github.com/ProtonMail/go-crypto v1.4.1 // indirect
 	github.com/RoaringBitmap/roaring/v2 v2.15.0 // indirect
+	github.com/SAP/go-hdb v1.16.8 // indirect
```
ness-david-dedu: **Wrong dependency classification — should be direct, not indirect**

`github.com/SAP/go-hdb` is classified as `// indirect` but the connector directly imports `gohdb "github.com/SAP/go-hdb/driver"` in both `input_sap_hana.go` and `output_sap_hana.go`. It should appear in the main `require` block without the `// indirect` annotation. Run `go mod tidy` to fix.

ness-david-dedu:
**Commits**

1. `2ef9e7e2` — `fix(sap_hana): ...`: `fix` is not a valid system name in this repo's commit format. The pattern is `system: message` or `system(subsystem): message` where *system* is a known area name. Should be `sap_hana: fix read ...` (or the subsystem form `sap_hana(input): ...`).
2. `24eb7ef3` — `working bulk read`: vague WIP message. Use imperative form describing the change (e.g. `sap_hana: add bulk read mode`).
3. `9e36dff8` — `imrpove hana rows write speed`: typo ("imrpove") and missing system prefix. Should be e.g. `sap_hana(output): improve bulk write throughput`.
4. `cf122214` — `add bencmark results`: typo ("bencmark") and missing system prefix. Should be e.g. `sap_hana: add benchmark results`.
5. `4673e059` — ` add incrementing-mode benchmark...`: leading whitespace in the message.

**Review**

Five issues found across security and dependency classification.

1. **SQL injection — `tableRef()`** ([input_sap_hana.go#L283–L287](https://github.com/redpanda-data/connect/blob/442127ecbd66374b4d3f622032663c920dfed57d/internal/impl/saphana/input_sap_hana.go#L283-L287)): `schemaName` and `tableName` are concatenated into double-quoted SQL identifiers without escaping internal `"` characters. Fix: `strings.ReplaceAll(name, `"`, `""`)` before concatenation.
2. **SQL injection — column names in `openRows()`** ([input_sap_hana.go#L296–L332](https://github.com/redpanda-data/connect/blob/442127ecbd66374b4d3f622032663c920dfed57d/internal/impl/saphana/input_sap_hana.go#L296-L332)): `incrementingCol` and `timestampCol` are interpolated into SQL identifiers without escaping internal `"`. Same fix.
3. **SQL injection — output column names** ([output_sap_hana.go#L119–L121](https://github.com/redpanda-data/connect/blob/442127ecbd66374b4d3f622032663c920dfed57d/internal/impl/saphana/output_sap_hana.go#L119-L121)): same pattern, same fix.
4. **SQL injection — `table` field completely unquoted** ([output_sap_hana.go#L125](https://github.com/redpanda-data/connect/blob/442127ecbd66374b4d3f622032663c920dfed57d/internal/impl/saphana/output_sap_hana.go#L125)): `table` is interpolated raw into the INSERT SQL via `%s` with no quoting. Needs schema/table splitting and identifier escaping.
5. **`go.mod` misclassified as indirect** ([go.mod#L233](https://github.com/redpanda-data/connect/blob/442127ecbd66374b4d3f622032663c920dfed57d/go.mod#L233)): `github.com/SAP/go-hdb` is `// indirect` but is directly imported. Run `go mod tidy`.
6. **Proprietary binary committed** (`internal/impl/saphana/bench/ngdbc.jar`): SAP HANA JDBC driver JAR should not be in the repo. The README documents how to obtain it — remove the file and add it to `.gitignore`.
7. **`docs/benchmark-results/SUMMARY.md`** not updated with SAP HANA entry (required per [docs/benchmarking.md#L322–L326](https://github.com/redpanda-data/connect/blob/442127ecbd66374b4d3f622032663c920dfed57d/docs/benchmarking.md#L322-L326)).
8. **`docs/benchmarking.md` Existing Benchmarks table** not updated with SAP HANA suite (same rule).

internal/impl/saphana/input_sap_hana.go
```diff
@@ -0,0 +1,541 @@
...
+
+Every message produced by this input carries the following metadata fields:
+
+- ` + "`sap_hana_schema`" + `: The HANA schema name.
+- ` + "`sap_hana_table`" + `: The HANA table name.
+- ` + "`schema`" + `: Avro-compatible schema derived from ` + "`SYS.TABLE_COLUMNS`" + `, suitable for use with ` + "`schema_registry_encode`" + `. Column additions are detected automatically without a pipeline restart. Only present when ` + "`schema_name`" + ` is configured.
+`).
+	Field(service.NewStringField(shFieldDSN).
+		Description("SAP HANA connection DSN.").
+		Example("hdb://user:password@host:39017"),
```
ness-david-dedu: **Input DSN not marked `.Secret()`**

The `dsn` field embeds credentials (`hdb://user:password@host:port`) but is not marked `.Secret()`. Without it the value appears in DEBUG logs and the config API. The output connector's DSN is correctly marked Secret — apply the same here.

Fix: add `.Secret()` after `.Example("hdb://user:password@host:39017")`.

internal/impl/saphana/input_sap_hana.go
```diff
@@ -0,0 +1,541 @@
...
+				}
+				s.dbMut.Lock()
+				if !timerFired {
+					return nil, nil, service.ErrEndOfInput
+				}
+				// Pre-warm schema cache before opening rows (same reason as bulk mode).
+				if s.schemaName != "" {
+					_, _ = s.schemas.schemaForEvent(ctx, s.schemaName, s.tableName, nil)
+				}
+				rows, err := s.openRows(ctx)
```
ness-david-dedu: **Nil pointer dereference: `s.db` not guarded after re-acquiring lock**

`dbMut` is released at line 377 while waiting for the poll timer. `Close()` can acquire the lock in that window, set `s.db = nil`, and release the lock. When `ReadBatch` re-acquires the lock at line 389, `s.db` is nil. The `timerFired` guard at line 390 is dead code (the `ctx.Done` and `stopChan` branches both `return` before reaching line 389, so `timerFired` is always `true` here) and does not protect against this.

Fix: after line 389 add:
```
if s.db == nil {
    return nil, nil, service.ErrNotConnected
}
```

internal/impl/saphana/bench/saphana-read/bulk/benchmark_config.yaml
```diff
@@ -0,0 +1,23 @@
+input:
```
ness-david-dedu, line 1: **Missing `benchmark:` processor and wrong output sink (applies to all 4 benchmark configs)**

Per [`docs/benchmarking.md §3`](https://github.com/redpanda-data/connect/blob/f301c6aa6d99a77730b70c9c51894dc8186d81d0/docs/benchmarking.md#L117-L150):
- "The key element is the **`benchmark` processor** which logs rolling throughput statistics."
- "**`drop: {}`** — Eliminates output overhead so you measure only input throughput."

All four configs (`saphana-read/bulk`, `saphana-read/incrementing`, `saphana-read/query`, `saphana-write`) are missing the `benchmark:` processor entirely and route to `kafka_franz` instead of `drop: {}`. This means Connect's built-in `msg/sec` and `MB/sec` stats are not produced, and the numbers in `sap-hana.md` include Kafka write overhead rather than isolating connector throughput.

Fix: add a `benchmark:` processor to each output section and change the output to `drop: {}` (or document the Kafka overhead explicitly if end-to-end measurement is the intent, and add a separate pass with `drop: {}` for connector isolation).

ness-david-dedu (commented):
**`docs/benchmark-results/SUMMARY.md`** — SAP HANA row is added to the "At a Glance" table but there is no corresponding paragraph under the "What These Numbers Mean" section. Every other connector has one. Add a short paragraph explaining what the numbers mean and what the benchmark measured (e.g. fetch_size sensitivity, Kafka overhead, recommended config).

ness-david-dedu:
**Commits**
LGTM

**Review**

Three code issues and one docs gap found in the updated PR.

1. **Input DSN not ** ([input_sap_hana.go#L69–L72](https://github.com/redpanda-data/connect/blob/f301c6aa6d99a77730b70c9c51894dc8186d81d0/internal/impl/saphana/input_sap_hana.go#L69-L72)): The input `dsn` field is missing `.Secret()`, exposing embedded credentials in DEBUG logs and the config API. The output connector marks its DSN Secret correctly — same fix needed here.
2. **Nil pointer dereference in poll loop** ([input_sap_hana.go#L377–L397](https://github.com/redpanda-data/connect/blob/f301c6aa6d99a77730b70c9c51894dc8186d81d0/internal/impl/saphana/input_sap_hana.go#L377-L397)): `dbMut` is released during the poll sleep (line 377); `Close()` can acquire the lock, set `s.db = nil`, and release it. When `ReadBatch` re-acquires at line 389, `s.db` is nil and `openRows()` panics. The `timerFired` guard at line 390 is dead code (both other select branches return before reaching line 389). Fix: add `if s.db == nil { return nil, nil, service.ErrNotConnected }` after line 389.
3. **Benchmark configs missing `benchmark:` processor and using wrong output sink** ([saphana-read/bulk/benchmark_config.yaml](https://github.com/redpanda-data/connect/blob/f301c6aa6d99a77730b70c9c51894dc8186d81d0/internal/impl/saphana/bench/saphana-read/bulk/benchmark_config.yaml)): All four configs route to `kafka_franz` instead of `drop: {}` and have no `benchmark:` processor. Per [docs/benchmarking.md §3](https://github.com/redpanda-data/connect/blob/f301c6aa6d99a77730b70c9c51894dc8186d81d0/docs/benchmarking.md#L117-L150), the benchmark processor is required for `msg/sec`/`MB/sec` stats and `drop: {}` is required to isolate connector throughput from output overhead.
4. **`docs/benchmark-results/SUMMARY.md` missing SAP HANA paragraph** ([SUMMARY.md#L16–L24](https://github.com/redpanda-data/connect/blob/f301c6aa6d99a77730b70c9c51894dc8186d81d0/docs/benchmark-results/SUMMARY.md#L16-L24)): SAP HANA row added to the table but no paragraph under "What These Numbers Mean". Every other connector has one.

ness-david-dedu:
**Commits**

1. `ea5af3be` — message starts with two leading spaces (`  sap_hana: fix DSN secret lea…`). Message must not start with whitespace per commit policy.

**Review**

Two previous rounds of findings have been resolved (DSN Secret field, poll-loop nil guard, benchmark config gaps, SQL injection, output identifier quoting). Three remaining issues:

1. **`quoteIdentifier` has no unit test** — this is the load-bearing SQL injection prevention function. It correctly doubles embedded `"` characters per SQL standard, but there is no test exercising the escape path. A test with an input containing a `"` character would confirm the doubling behaviour under regression. [schema.go#L24-L26](https://github.com/redpanda-data/connect/blob/ea5af3be8c6f58f151f67783b83163b364ac9af3/internal/impl/saphana/schema.go#L24-L26)

2. **Integration tests are missing Given-When-Then `t.Log()` structure** — the project convention (followed by e.g. `internal/impl/gcp/enterprise/bigquery/integration_test.go`) uses `t.Log("Given ...")` / `t.Log("When ...")` / `t.Log("Then ...")` annotations to make CI log output triageable. None of the seven `TestIntegration*` functions in [integration_test.go](https://github.com/redpanda-data/connect/blob/ea5af3be8c6f58f151f67783b83163b364ac9af3/internal/impl/saphana/integration_test.go) follow this pattern.

3. **`timerFired` guard is unreachable dead code** — at [input_sap_hana.go#L391-L393](https://github.com/redpanda-data/connect/blob/ea5af3be8c6f58f151f67783b83163b364ac9af3/internal/impl/saphana/input_sap_hana.go#L391-L393) both the `ctx.Done()` and `stopChan` cases `return` before reaching the `if !timerFired` check, so `timerFired` is always `true` when that line executes. The guard can be removed without changing behaviour.

squiidz:
1. timestamp+incrementing mode silently breaks after the first empty poll — input_sap_hana.go advances s.timestampHWM unconditionally at end of ReadBatch, but s.hwm only
  advances when a row is scanned. After one empty poll the next query binds inc > ? with an empty string against a numeric column. Fix: don't advance timestampHWM past the first
  observed row, or fall back to pure timestamp comparison while s.hwm == "".
  
  2. HWM round-trips through fmt.Sprintf("%v", v) — input_sap_hana.go:451-455. For a TIMESTAMP incrementing_column (which the docs explicitly suggest), %v on time.Time produces
  2024-01-01 00:00:00 +0000 UTC, which HANA cannot parse. Even for BIGINT/DECIMAL the string→column-type implicit cast is fragile. Keep the raw any and pass it directly to
  QueryContext.
  
  3. Output's "execMany bulk path" claim is not what the code does — output_sap_hana.go:170-189 calls db.ExecContext with a flat arg slice. That triggers go-hdb's argument-count
  detection but not a true prepared/MtInsert flow. Either prepare the statement once in Connect and cache *sql.Stmt, or remove the "single MtInsert RPC" claim from
  docs/benchmark-results/sap-hana.md.
  
  4. go.mod has a stray blank line in the // indirect block — go mod tidy will revert it and any tidiness check on CI will fail.
  
  5. public/components/all/package.go import order — saphana was inserted before salesforce. The block is alphabetised everywhere else; this will fail any goimports/lint
  enforcement.

twmb:
The approach here is the right one and matches the scope we defined for SAP HANA: a dedicated polling `sap_hana` input (bulk / incrementing / query) plus `hana` driver support in `sql_insert` / `sql_raw` for writes. The schema-metadata path (from `SYS.TABLE_COLUMNS`, addition-only drift detection, precision-aware DECIMAL mapping) is the most important piece and it's in good shape, and the SQL is parameterized with identifier quoting throughout.

There are also a few good ideas in #4490 worth pulling in here — it grows the scope of this PR slightly, but keeps everything in one place.

A few things to address before merge:

**Config surface**
- [ ] Decide on the connection model: separate `connection_url` / `connection_user` / `connection_password` vs. the single `dsn` here, plus `table_name` / `incrementing_column_name` naming. The spec's own `sql_insert`/`sql_raw` examples use `dsn: hdb://…`, so a single `dsn` may be the more consistent choice — let's just pick one deliberately.
- [ ] `poll_interval` defaults to 5s; spec calls for 60s.
- [ ] Missing knobs: `numeric_mapping` (`none` / `best_fit` / `best_fit_eager_double`), `partition_count`, `max_retries`.

**Schema metadata (ideas from #4490)**
- [ ] Add primary-key discovery — #4490's `SYS.INDEX_COLUMNS` ⋈ `SYS.INDEXES WHERE CONSTRAINT = 'PRIMARY KEY'` query is a clean way to do it; there's no PK info in the emitted schema today.
- [ ] Use real per-column nullability (`IS_NULLABLE`) instead of marking every field optional.
- [ ] Emit DECIMAL as a canonical decimal string rather than `float64`. This matches every other CDC input here (oracle / mysql / mongodb / postgres all emit canonical decimal strings via `sqlutil.CanonicaliseDecimal`), and bare floats get rejected by downstream Avro string/decimal encoding — the current path is both lossy and inconsistent with the rest of the repo.

**Behavior**
- [ ] The high-water mark is in-memory only, so it doesn't survive a restart — consider persisting it (a `cache` resource) so incrementing mode resumes without re-reading from the initial value.
- [ ] The dedicated `sap_hana` output is outside the defined scope (writes go through `sql_insert`/`sql_raw` with `driver: hana`, which you've already wired up). Either drop it, or let's confirm we want it as a deliberate extra.
- [ ] Metadata keys: `schema` is correct; `table_name` / `database_schema` were the agreed names (currently `sap_hana_table` / `sap_hana_schema`).

**Tests / CI**
- [ ] #4490 has a useful type/encoding edge-case suite (Unicode, max int64, DECIMAL precision, date boundaries, sub-second timestamps, LOBs) that exercises the go-hdb round-trip — worth porting.
- [ ] Run `task docs` so the generated component page + `info.csv` land in the PR, and add license headers to the `bench/**/main.go` helpers.

meticulo3366:
i would love to merge this one in soon

squiidz:
1- checkpoint persisted on read with a no-op ack (data loss on crash). This is the most important one and it's not a duplicate: the checkpoint cache was added precisely in response to twmb's "persist the HWM" request, but the implementation commits the HWM at read time instead of in the AckFunc. The feature reviewers asked for has a correctness gap in how it was delivered. Nobody has flagged this.

2- incrementing strict > + per-row HWM advance loses rows when a tie spans a fetch_size boundary. Novel. Sharpened by the fact that the field doc still suggests a timestamp column (non-unique) as a valid incrementing column.

3- incrementing_initial_value is bound as a string on the first poll. Related to squiidz second issue but distinct: squiidz's runtime HWM concern was fixed, but the initial config value is still always a string until the first row re-types it.

squiidz (approved):

squiidz:
**1. Transient scan error can permanently lose rows in `incrementing` mode**
`scanRow` advances `s.hwm` as rows are scanned, but if the batch is then discarded by an error path (`scanning columns` / `iterating rows`), those rows are now below the in-memory HWM and the next poll's `WHERE inc > ?` skips them. That next poll then looks empty, and the empty-poll path persists `s.hwm` (not `hwmSafe`) immediately — so a single transient mid-cursor error makes the loss permanent across restarts. Suggest rewinding `s.hwm = s.hwmSafe` on the error-discard paths, and persisting `hwmSafe` in the empty-poll `saveCheckpoint`.

**2. Checkpoint persistence isn't ack-order aware**
Each batch's `AckFunc` persists its own HWM snapshot unconditionally. With multiple batches in flight, batch N+1 can ack (persisting a higher HWM) while batch N is still unacked — a crash at that point loses batch N's rows. Other inputs in the repo use `checkpoint.Capped` for ordered ack tracking; the same pattern fits here. (The "nothing in flight, safe to persist immediately" comment on the empty-poll path has the same problem — prior batches may still be unacked.)

**3. `hana` driver in the generic `sql_*` components breaks non-enterprise builds**
`conn_fields.go` adds `hana` to the driver enum and aliases it to `hdb`, but the go-hdb driver is only registered via the enterprise-only saphana package import. In the community binary (which has `sql_*` but not saphana), `driver: hana` fails at runtime with `sql: unknown driver "hdb"`. Either add `_ "github.com/SAP/go-hdb/driver"` to `public/components/sql/package.go` (go-hdb is Apache-2.0, so that's fine) or drop the `sql_*` driver support from this PR.

**4. Missing generated docs page**
There's no `docs/modules/components/pages/inputs/sap_hana.adoc` — please run `task docs` and commit the output, otherwise CI will fail.

PR #4465, opened 2026-05-29
oracledb: emit CDC decimal values as canonical strings in both modes

Jeffail:
## What

The Oracle CDC snapshot (`sql.Scan`) and streaming (LogMiner redo) paths derived NUMBER column types and coerced their values through separate code, which could disagree. Most importantly, an integer-valued decimal arriving from the redo log as a bare numeric literal was converted to an `int64` and emitted as a **bare JSON number** rather than a canonical decimal string. Downstream Avro encoding into a string-typed field rejects a JSON number, failing with `cannot use json.Number with Avro type string`.

## How

- Introduce a single shared coercion, `sqlutil.CoerceToCommon`, mapping any numeric Go type (`string`, `int64`, `float64`, `json.Number`) to the canonical form for a common-schema column: `Int64`→`int64`, `Float`→`float64`, and `Decimal`/`BigDecimal`→canonical decimal **strings** (never a bare number).
- Unify type derivation behind `replication.NumberToCommon` so both modes classify a NUMBER identically.
- Streaming value coercion now routes every numeric input through the shared helper.
- The snapshot mapper classifies columns with `NumberToCommon`, keeping the native `int64` scan for integer-width columns and routing `Decimal`/`BigDecimal` columns through the shared string coercion. This also resolves a latent drift where negative-scale NUMBER columns emitted a value type inconsistent with their schema.

## Tests

- Unit tests for the shared coercion and the streaming path (including the integer-valued-decimal regression).
- An integration test asserting that snapshot and streaming agree on both the yielded schema and the value type for every column, and that decimals are always strings.

## Note

`Int64`-classified columns (`NUMBER(p,0)` with p ≤ 18) and `BINARY_FLOAT`/`BINARY_DOUBLE` are still emitted as JSON numbers by design, consistent with their common schema; this change scopes to `Decimal`/`BigDecimal`.

### DIFF
--- a/internal/impl/oracledb/schema.go
+++ b/internal/impl/oracledb/schema.go
@@ -11,10 +11,7 @@ package oracledb
 import (
 	"context"
 	"database/sql"
-	"encoding/json"
 	"fmt"
-	"math"
-	"strconv"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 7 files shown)


squiidz (approved):

internal/sqlutil/coerce.go
```diff
@@ -0,0 +1,160 @@
...
+//   - Float32 / Float64   → float64
+//   - Decimal             → canonical decimal string (never a bare number)
+//   - BigDecimal          → canonical big-decimal string (never a bare number)
+//   - anything else       → returned unchanged
+//
+// A nil value is returned as-is. On a recoverable conversion failure (e.g. a
+// non-numeric string for a numeric column) the original value is returned
+// alongside a non-nil error so callers can log and preserve the input rather
+// than emitting a wrong type.
+func CoerceToCommon(c schema.Common, v any) (any, error) {
```
josephwoodward, line 42: Given this is redo log specific (judging by the comment), would this be better in the `impl/oracledb` package?

internal/impl/oracledb/schema.go
```diff
@@ -46,37 +43,11 @@ func oracleTypeToCommonType(dataType string) schema.CommonType {
...
-//     (38, 255) for bare NUMBER): BigDecimal.
-//   - scale == 0 && 0 < precision <= 18: Int64 (existing optimisation —
-//     scale-0 NUMBER fits losslessly in int64).
-//   - scale < 0: rounded to Decimal(precision, 0) since Avro/Parquet/Iceberg
-//     can't represent negative scale.
-//   - otherwise: Decimal(precision, scale).
+// oracleNumberToCommon builds a schema.Common entry for a NUMBER column. It
+// delegates to replication.NumberToCommon so the streaming schema cache and the
+// snapshot value mapper derive identical types from the same precision/scale.
 func oracleNumberToCommon(name string, precision, scale int64, hasDecimalInfo bool) schema.Common {
```
josephwoodward: Is there much value in having this function versus saving the indirection by calling `replication.NumberToCommon` directly?

josephwoodward (approved):

josephwoodward:
All integration tests (including the new one) are passing:

image

internal/sqlutil/coerce.go
Jeffail: Good catch on the wording — the doc comment over-emphasised redo logs and made it read as Oracle-specific. The helper is actually generic to the SQL-CDC sources (string / int64 / float64 / json.Number → canonical form), so I've reworded the comment to be source-neutral and kept it in `sqlutil` alongside its `CanonicaliseDecimal`/`CanonicaliseBigDecimal` siblings. It can't move into `impl/oracledb` regardless: `replication/snapshot.go` also calls it, and since `oracledb` imports `replication`, placing it there would create an import cycle. Addressed in the latest commit.

internal/impl/oracledb/schema.go
Jeffail: Agreed — there's no value in it now that it's pure delegation. I've removed the wrapper and call `replication.NumberToCommon` directly at both call sites (and in the test). Addressed in the latest commit.

josephwoodward (approved):

PR #4473, opened 2026-06-01
aws_bedrock_embeddings: support Cohere input_type and v4 response

squiidz:
Cohere embedding models on Bedrock require an `input_type` field
  (`search_document`, `search_query`, etc.) to produce correct asymmetric
  embeddings, and Cohere Embed v4 changed the response schema from
  `{"embeddings": [[...]]}` to `{"embeddings": {"float": [[...]]}}`.

  Dispatch the request body shape and response parsing on the model ID,
  using a substring match on "cohere" so regional inference profiles
  like `us.cohere.embed-v4:0` are detected. Both v3 and v4 response
  shapes are now accepted. Titan model code paths are unchanged


Jeffail (approved):
LGTM, my claude is a nerd and has a minor nit to raise but feel free to ignore

To add a small note beneath the above: the change looks correct and is well-tested. I verified against the AWS Bedrock documentation that `input_type` is indeed marked **required** for the Cohere Embed v4 model (as it is for v3), so failing fast at configuration-parse time for Cohere models is the right call rather than over-restrictive. The dual-shape response parser is also appropriately defensive — AWS's own v4 documentation is internally inconsistent about when the `embeddings` field is returned as an array versus an object keyed by embedding type, so handling both shapes is prudent.

The only minor nit, entirely non-blocking: `TestNewProcessor_CohereRequiresInputType` uses an underscore, whereas the project's test convention favours camelCase names without underscores. Happy to see this merged as-is.

PR #1335, opened 2022-07-20
Remove some heavy dependencies from public/service

mihaitodor:
Related to #1332.

While I don't expect these values to change much in the future, it might be nicer to separate configs into pure and non-pure packages.

Given a hello world `main.go` which only imports `"github.com/benthosdev/benthos/v4/public/service"` and `"github.com/benthosdev/benthos/v4/public/components/pure"`, I noticed that there are a bunch of heavy Google dependencies that are being pulled in. Here's the diff in dependencies after this change:

```diff
diff --git a/go.mod b/go.mod
index 7571134..4eac6f8 100644
--- a/go.mod
+++ b/go.mod
@@ -5,10 +5,6 @@ go 1.18
 require github.com/benthosdev/benthos/v4 v4.4.1

 require (
-	cloud.google.com/go v0.100.2 // indirect
-	cloud.google.com/go/compute v1.5.0 // indirect
-	cloud.google.com/go/iam v0.1.0 // indirect
-	cloud.google.com/go/pubsub v1.17.1 // indirect
 	cuelang.org/go v0.4.2 // indirect
 	github.com/Azure/go-amqp v0.17.0 // indirect
 	github.com/Jeffail/gabs/v2 v2.6.1 // indirect
@@ -28,11 +24,8 @@ require (
 	github.com/gocql/gocql v0.0.0-20211222173705-d73e6b1002a7 // indirect
 	github.com/gofrs/uuid v4.2.0+incompatible // indirect
 	github.com/golang-jwt/jwt v3.2.2+incompatible // indirect
-	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
 	github.com/golang/protobuf v1.5.2 // indirect
 	github.com/golang/snappy v0.0.4 // indirect
-	github.com/google/go-cmp v0.5.7 // indirect
-	github.com/googleapis/gax-go/v2 v2.2.0 // indirect
 	github.com/gorilla/css v1.0.0 // indirect
 	github.com/gorilla/handlers v1.5.1 // indirect
 	github.com/gorilla/mux v1.8.0 // indirect
@@ -76,7 +69,6 @@ require (
 	github.com/youmark/pkcs8 v0.0.0-20201027041543-1326539a0a0a // indirect
 	go.mongodb.org/mongo-driver v1.8.2 // indirect
 	go.nanomsg.org/mangos/v3 v3.3.0 // indirect
-	go.opencensus.io v0.23.0 // indirect
[...]

PR #4476, opened 2026-06-02
kafka: add opt-in rebootstrap field for KIP-899/KIP-1102

david-yu:
## What

Adds an opt-in `rebootstrap` boolean to the shared franz-go connection fields, surfaced on the `redpanda`/`kafka` inputs and outputs. When enabled, the client wires up franz-go's `kgo.OnRebootstrapRequired` (KIP-899, refined by KIP-1102): on a broker `REBOOTSTRAP_REQUIRED` (error code 129) the client re-resolves its configured `seed_brokers` and reconnects.

## Why

franz-go v1.20.7 supports client rebootstrap, but **Connect never registered the callback**, so the broker's `REBOOTSTRAP_REQUIRED` signal was silently ignored (franz-go acts on it only when `OnRebootstrapRequired` is set). This PR lets users opt in.

The target topology is a **single stable DNS endpoint shared by a shadow and primary cluster** (e.g. `rp-cluster.example.com`) whose A-records repoint to the surviving cluster on failover. With `rebootstrap: true`, a `REBOOTSTRAP_REQUIRED` signal causes the client to drop its discovered brokers and re-resolve the seed endpoint (re-running DNS), reconnecting to the cluster the alias now points at.

### Scope note (tested)

A `seed_brokers` list that **statically enumerates both clusters' brokers does _not_ need this flag** — stock franz-go already fails over to a surviving seed via its normal metadata refresh. I verified this with two live Redpanda clusters: killing the active cluster, a client produced to the other in ~300ms **with `rebootstrap` off and the callback never invoked**. So this option is aimed squarely at the DNS-endpoint topology and at honoring broker-initiated `REBOOTSTRAP_REQUIRED`, not at static multi-cluster seed lists.

### `redpanda` input — shared DNS endpoint + rebootstrap

```yaml
input:
  redpanda:
    seed_brokers:
      - rp-cluster.example.com:9092   # stable alias; DNS repoints shadow<->primary on failover
    topics:
      - orders
    consumer_group: connect-orders
    rebootstrap: true                 # honor REBOOTSTRAP_REQUIRED and re-resolve the endpoint
```

### `redpanda` output — shared DNS endpoint + rebootstrap

```yaml
output:
  redpanda:
    seed_brokers:
      - rp-cluster.example.com:9092
    topic: orders
    rebootstrap: true
```

`seed_brokers` remains a list (separate items or comma-separated items are both expanded), so you can still provide multiple bootstrap addresses behind/alongside the alias if desired — but note the scope note above re: static multi-cluster lists.

## Notes

- Defaults to `false`; existing configs are unaffected.
- Implemented once in `internal/impl/kafka/franz_client.go` (shared `FranzConnectionFields`), so it applies to both input and output — and to the `kafka_franz` input/output, which share the same connection fields.
- Docs for the `redpanda` and `kafka_franz` input/output pages were updated manually because the full `task docs` generator can't link locally (unrelated `tigerbeetle-go` CGO link error). Other connectors sharing these fields (e.g. the `redpanda` cache and `redpanda_migrator`) will pick up the field when docs are regenerated in CI.

## Follow-up: timeout-based rebootstrap (KIP-1102)

KIP-1102 defines **two** rebootstrap triggers; this PR exposes only the first:

- [x] **Error-code trigger** (`REBOOTSTRAP_REQUIRED`) — implemented here via `rebootstrap: true` (franz-go's `OnRebootstrapRequired`).
- [ ] **Timeout trigger** (`metadata.recovery.rebootstrap.trigger.ms`) — **not** available in franz-go yet (verified through franz-go `master`). Proposed upstream in **twmb/franz-go#1335** (adds `kgo.RebootstrapTrigger`, with a runtime-verified demo on the PR). Once that merges and is released, this connector can bump franz-go and add a companion field (e.g. `rebootstrap_timeout`) to force a re-resolve after a bounded window even when no `REBOOTSTRAP_REQUIRED` is emitted.

This PR is intentionally scoped to the error-code trigger so it can land against the current franz-go release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ Changelog
 
 All notable changes to this project will be documented in this file.
 
+## Unreleased
+
+### Added
+
+- `redpanda` and `kafka` inputs and outputs now support an opt-in `rebootstrap` field that enables KIP-899 / KIP-1102 client rebootstrap, allowing seamless failover between Redpanda shadow and primary clusters when `seed_brokers` is a stable DNS endpoint. ([@david-yu](https://github.com/david-yu))
+
 ## 4.94.1 - 2026-05-29
 
 ### Fixed
--- a/docs/modules/components/pages/inputs/kafka_franz.adoc
+++ b/docs/modules/components/pages/inputs/kafka_franz.adoc
@@ -70,6 +70,7 @@ input:
     metadata_max_age: 5m
     request_timeout_overhead: 10s
     conn_idle_timeout: 20s
+    rebootstrap: false
     topics: [] # No default (required)
     regexp_topics: false
     rack_id: ""
@@ -518,6 +519,16 @@ The rough amount of time to allow connections to idle before they are closed.
 *Type*: `string`
 
 *Default*: `"20s"`
+
+=== `rebootstrap`
+
+When enabled, the client re-resolves its `seed_brokers` and rebootstraps when a broker signals that its cached metadata is stale (the Kafka `REBOOTSTRAP_REQUIRED` error code 129, introduced by KIP-899 and KIP-1102). This is useful for seamless failover when `seed_brokers` is a stable DNS endpoint that can resolve to a different (for example, shadow versus primary) Redpanda cluster.
+
+
+*Type*: `bool`
+
+*Default*: `false`
+Requires version 4.95.0 or newer
 
 === `topics`
 
--- a/docs/modules/components/pages/inputs/redpanda.adoc
+++ b/docs/modules/components/pages/inputs/redpanda.adoc
@@ -69,6 +69,7 @@ input:
     metadata_max_age: 1m
     request_timeout_overhead: 10s
     conn_idle_timeout: 20s
+    rebootstrap: false
     tcp:
       connect_timeout: 0s
       keep_alive:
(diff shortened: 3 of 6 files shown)


david-yu:
### Failover testing update — franz-go already covers these topologies

I ran end-to-end failover tests against two live single-node Redpanda clusters (a "primary" and a "shadow") to check how much `rebootstrap` actually contributes. Recovery was measured with a real **produce** to the surviving cluster (not just a Ping).

**Topology 1 — both clusters statically listed in `seed_brokers`:**

| Config | Failed over? | `OnRebootstrapRequired` calls |
|---|---|---|
| error-code only (`rebootstrap: true`, this PR) | ✅ ~300ms | 0 |
| timeout trigger (franz-go#1335) | ✅ ~0s | 0 |
| baseline (`rebootstrap: false`) | ✅ ~0s | 0 |

**Topology 2 — single stable DNS endpoint (hostname seed, repointed to shadow on failover):**

| Config | Failed over? | callback calls | seed re-dialed → shadow |
|---|---|---|---|
| error-code only (this PR) | ✅ ~200ms | 0 | 1 |
| timeout trigger (#1335) | ✅ ~800ms | 0 | 1 |
| baseline (`rebootstrap: false`) | ✅ ~800ms | 0 | 1 |

**Takeaway:** in **both** topologies, stock franz-go fails over on its own — its periodic seed re-query re-dials the seed list / re-resolves the DNS hostname and picks up the surviving cluster in under a second. The `rebootstrap` callback was **never invoked** in any run (including this PR's error-code config), and the timeout trigger never fired because built-in recovery completed first.

**So `rebootstrap` does not contribute to the shadow/primary failover scenarios this PR was originally framed around.** Its genuine, non-redundant value narrows to honoring a broker-initiated `REBOOTSTRAP_REQUIRED` (Kafka 4.0+), which franz-go ignores unless `OnRebootstrapRequired` is registered — I have not yet been able to make a broker emit that to test it directly.

Flagging this for reviewer visibility so we can decide whether this PR is still worth landing as-is, re-scoped strictly to `REBOOTSTRAP_REQUIRED`, or held. Test harness was throwaway; happy to share it.

david-yu:
Will go ahead and closed as it was discovered failover is already builtin to franz-go.

PR #4481, opened 2026-06-02
oracledb_cdc: Recreate stored procedure for updating checkpoint cache, removing the need to delete previous one first

josephwoodward:
This change ensures the stored procedure responsible for updating the OracleDB backed checkpoint cache is aligned with the table it's updating. This is especially useful now that the checkpoint cache table name can be configured.

This change also supports idempotent updating of the stored procedure, removing needing to verify of the checkpoint cache exists beforehand.


squiidz (approved):

josephwoodward:
Closing due to potential backwards compatibility challenges.

PR #4484, opened 2026-06-03
jira: add input component for streaming issues, comments, and changelog

squiidz:
Adds a `jira` input that streams Jira REST API events into Connect pipelines via JQL with cursor-based incremental polling. The cursor (max issue.updated timestamp) is persisted to a cache resource so progress survives restarts.

Available in both the Cloud and Self-Hosted distributions. Authenticates via API token (email + token). Supports resource={issues,comments,changelog}. The existing `jira` processor (enrichment lookups) is unchanged.

Shared HTTP/auth setup is extracted into a new internal/impl/jira/jiraauth sub-package consumed by both the processor and the input.

Limitations (v1): no OAuth, no worklogs resource, per-issue comment and changelog pagination is limited to the first response page (truncation logged and counted via the jira_input_child_truncated_total metric).


Jeffail (changes requested):
Quite impressed with my claude, looks like it caught a certified doozy, theres also two minor nits that should probably be looked at

Thank you for this — it is a genuinely impressive piece of work, and the concurrency model and test coverage in particular are excellent. One correctness concern that I believe warrants a change before merge, plus two minor points:

**1. (Blocking) The incremental cursor predicate is rendered in UTC, but JQL interprets date literals in the requesting account's timezone.**

In `buildJQL`, the threshold is formatted as a UTC wall-clock string:

```go
threshold := cur.Updated.Add(-r.cfg.cursorOverlap)
parts = append(parts, fmt.Sprintf(`updated >= "%s"`, threshold.UTC().Format("2006-01-02 15:04")))
```

Jira evaluates JQL date/time literals relative to the **profile timezone of the account making the request**, not UTC, and JQL has no syntax to attach an explicit offset to a date literal. The effective threshold instant is therefore shifted by the account's timezone offset:

- For an account **behind UTC** (e.g. any of the Americas), the effective threshold lands *later* than intended. Issues whose `updated` falls within that gap drop below the predicate on the subsequent poll and are never re-matched — a silent data-loss window that the default 60s `overlap` cannot absorb.
- For an account **ahead of UTC**, the threshold lands earlier, producing a wider window that the seen-set dedup absorbs harmlessly (at the cost of some redundant fetching).

Conveniently, the fix is already within reach: `Connect` calls `/rest/api/3/myself` for auth validation and currently discards the body. That response includes a `timeZone` field, so the threshold could be formatted in the account's location (`threshold.In(accountLoc).Format(...)`) to match Jira's interpretation exactly. If reading the account timezone is undesirable for any reason, the minimum bar would be to document prominently that the token's account must be configured to UTC — but deriving it from the `/myself` response already in hand seems both more robust and only marginally more code.

References for the JQL timezone behaviour:
- https://community.atlassian.com/forums/App-Central-articles/Timezones-in-Jira-reporting-why-dates-look-off/ba-p/3205133
- https://community.developer.atlassian.com/t/jql-on-search-with-updated-date-time-does-not-work-correctly/58754
- https://jira.atlassian.com/browse/JRASERVER-68522 (confirms JQL has no per-literal offset syntax)

**2. (Minor) `base_url` with a trailing slash produces a double slash.**

Every request URL is constructed as `BaseURL + "/rest/api/3/..."`. A `base_url` configured with a trailing `/` yields `//rest/...`. Jira most likely tolerates this, but trimming a trailing slash at parse time would be a tidy, defensive guard.

**3. (Minor) The empty-page-with-continuation-token path has no backoff.**

In `Read`, an empty page that still carries a `nextPageToken` falls through to `continue` with no pause. It is network-bound rather than a CPU spin (each iteration issues an HTTP request), so it is not dangerous, but a server that persistently returns empty pages alongside a continuation token would be polled in a tight request loop. A small guard there would be prudent.

Everything else looks great — happy to re-review once the timezone handling is addressed.

Jeffail (approved):
LGTM, the timezone bit needs a little bit extra work as it turns out the images we ship are completely broken for loading timezones. We'll need to update our build system before this ships (only a minor change), my emotionless AI can elaborate

To elaborate on the above: the three points from the previous review are all confirmed resolved in `6392dee` — the cursor predicate is now rendered in the account timezone, `base_url` trailing slashes are trimmed, and the empty-page path backs off only on a genuinely empty server page. The targeted unit tests pass locally. The implementation itself is correct, and the turnaround is appreciated.

The remaining issue is not in this PR's code but in how we package our images, which the timezone fix has usefully surfaced:

- The fix depends on `time.LoadLocation`, which requires the IANA timezone database to be present at runtime.
- Our release binary does not import `time/tzdata`, and the runtime image is `busybox`-based with only CA certificates and `/etc/passwd` copied in — there is no `/usr/share/zoneinfo`.
- As a result, in the shipped container `LoadLocation` fails for any non-UTC zone and falls back to `time.UTC`, silently reinstating the very data-loss window this change set out to close for any account whose profile timezone is behind UTC. It passes today only because development machines carry system tzdata.

Suggested fix, at the binary level rather than within this component (it affects every component that loads a location, not just this input):

1. Embed the database via `import _ "time/tzdata"` in the `cmd/*` entrypoints, or equivalently build the release artifacts with `-tags timetzdata`. The cost is ~403 KiB on go1.26.4, and the embedded copy is used only as a fallback when no system database is found — hosts that already ship tzdata are unaffected.
2. Alternatively, copy `/usr/share/zoneinfo` into the runtime image, though embedding travels with the binary and is less prone to regressing in a future Dockerfile change.

A regression test that does not rely on the host's tzdata would also be valuable, so CI catches a future stripped image. Given this is a packaging change rather than a code change here, landing it as a small separate change to the build system and gating the release on it seems cleanest — this PR can otherwise proceed.

squiidz:
Timezone embedding PR: https://github.com/redpanda-data/connect/pull/4565

PR #4489, opened 2026-06-05
dynamodb_cdc: paginate DescribeStream to discover all shards

squiidz:
DescribeStream returns at most 100 shards per response and uses
  LastEvaluatedShardId for continuation. The three call sites in
  input_cdc.go read only the first page, so on streams with more than
  ~100 lifetime shards the connector never discovers the currently-open
  shards: readers drain, the input silently stalls, no errors logged.

  Introduce describeStreamAllShards, a helper that loops DescribeStream
  chaining ExclusiveStartShardId = LastEvaluatedShardId until the token
  is nil, and route isCDCCheckpointStale, refreshShards, and
  refreshTableShards through it. Unit tests cover empty, single-page,
  multi-page accumulation, token chaining, and error paths.

### DIFF
--- a/internal/impl/aws/dynamodb/input_cdc.go
+++ b/internal/impl/aws/dynamodb/input_cdc.go
@@ -1299,24 +1299,56 @@ func (d *dynamoDBCDCInput) connectWithSnapshot(ctx context.Context, tableName st
 	return nil
 }
 
+// describeStreamPager is the minimum surface of *dynamodbstreams.Client
+// needed by describeStreamAllShards. Defined so tests can supply a fake
+// without depending on the concrete AWS SDK client.
+type describeStreamPager interface {
+	DescribeStream(ctx context.Context, in *dynamodbstreams.DescribeStreamInput, opts ...func(*dynamodbstreams.Options)) (*dynamodbstreams.DescribeStreamOutput, error)
+}
+
+// describeStreamAllShards calls DescribeStream repeatedly, chaining
+// ExclusiveStartShardId = LastEvaluatedShardId from the previous response,
+// and accumulates shards across all pages. DescribeStream returns at most
+// 100 shards per call; on high-write streams the lifetime shard count
+// exceeds the page size within hours, and a single un-paginated call hides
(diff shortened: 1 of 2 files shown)


internal/impl/aws/dynamodb/input_cdc.go
```diff
@@ -1299,24 +1299,56 @@ func (d *dynamoDBCDCInput) connectWithSnapshot(ctx context.Context, tableName st
...
+	DescribeStream(ctx context.Context, in *dynamodbstreams.DescribeStreamInput, opts ...func(*dynamodbstreams.Options)) (*dynamodbstreams.DescribeStreamOutput, error)
+}
+
+// describeStreamAllShards calls DescribeStream repeatedly, chaining
+// ExclusiveStartShardId = LastEvaluatedShardId from the previous response,
+// and accumulates shards across all pages. DescribeStream returns at most
+// 100 shards per call; on high-write streams the lifetime shard count
+// exceeds the page size within hours, and a single un-paginated call hides
+// the currently-open shards behind the LastEvaluatedShardId boundary.
+func describeStreamAllShards(ctx context.Context, c describeStreamPager, streamArn *string) ([]types.Shard, error) {
```
josephwoodward:
```suggestion
func describeStreamAllShards(ctx context.Context, client describeStreamPager, streamArn *string) ([]types.Shard, error) {
```

In isolation it wasn't immediately clear this was a network call, so perhaps calling this `client` may make it clearer (the function then does `client.DescribeStream`).

internal/impl/aws/dynamodb/input_cdc.go
```diff
@@ -1299,24 +1299,56 @@ func (d *dynamoDBCDCInput) connectWithSnapshot(ctx context.Context, tableName st
...
+	var (
+		all     []types.Shard
+		startID *string
+	)
+	for {
```
josephwoodward: Given this is an infinite loop are we happy that it would exit? If we know a maximum number of shards then maybe we set an upper limit on the number of iterations it could make then fail if it exceeds that?

josephwoodward (approved):

PR #4492, opened 2026-06-09
feat: Add option for using elasticsearch API keys

### LINKED ISSUE #4495: elasticsearch: ability to use api keys rather than basic auth
Elasticsearch supports API key authentication, and in the new serverless option in Elastic Cloud this is the only option that is supported.  It would be nice to be able to use the API keys in the plugins.

dbason:
Modern Elasticsearch (particularly in elastic cloud) supports API key authentication.  In particular this is the only mechanism (other than mTLS) that Elastic Cloud Serverless supports.  This adds the option to specify an api key for both versions of the Elasticsearch connector.

I wanted to avoid changing the current basic_auth schema so I have added this under a new section, but not sure whether this is the pattern you want to follow.

Fixes #4495

### DIFF
--- a/internal/impl/elasticsearch/v9/output.go
+++ b/internal/impl/elasticsearch/v9/output.go
@@ -53,6 +53,7 @@ const (
 	esFieldAuthEnabled     = "enabled"
 	esFieldAuthUsername    = "username"
 	esFieldAuthPassword    = "password"
+	esFieldAPIKey          = "api_key"
 	esFieldBatching        = "batching"
 )
 
@@ -128,6 +129,9 @@ func esConfigFromParsed(pConf *service.ParsedConfig) (*esConfig, error) {
 	if conf.retryOnConflict, err = pConf.FieldInt(esFieldRetryOnConflict); err != nil {
 		return nil, err
 	}
+	if conf.clientOpts.APIKey, err = pConf.FieldString(esFieldAPIKey); err != nil {
+		return nil, err
+	}
 
 	return conf, nil
 }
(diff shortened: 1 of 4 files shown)


internal/impl/elasticsearch/v9/output.go
```diff
@@ -179,6 +189,16 @@ Both the `+"`id` and `index`"+` fields can be dynamically set using function int
 			).Description("Allows you to specify basic authentication.").
 				Advanced().
 				Optional(),
+			service.NewObjectField(esFieldCloudAuth,
+				service.NewBoolField(esFieldCloudAuthEnabled).
+					Description("Whether to use Elastic Cloud API key authentication in requests.").
+					Default(false),
```
josephwoodward: Is there a specific reason for this boolean flag as opposed to assuming if the Cloud Auth API key is set then the user intends to use it? That way we could remove the need for nested object field and move the key to the root of the config (eg `cloud_auth_api_key`).

josephwoodward:
Hi @dbason, thank you for your contribution! Just the one question from me otherwise LGTM.

internal/impl/elasticsearch/v9/output.go
dbason: There isn't, I was just matching the basic auth pattern.  I can update that.

dbason:
@josephwoodward OK I've made that change now, and added an example and made the docs.

josephwoodward (approved):

internal/impl/elasticsearch/v9/output.go
josephwoodward: Thank you!

josephwoodward:
Thank you for your contribution @dbason

PR #4494, opened 2026-06-10
telemetry: adopt shared common-go/telemetry client

RafalKorepta:
## What

Replaces the bespoke telemetry transport (resty + go-jose signing + hand-rolled export loop) in `internal/telemetry` with the shared client from [`github.com/redpanda-data/common-go/telemetry`](https://github.com/redpanda-data/common-go/pull/172).

**Behavior-preserving.** Everything observable stays the same:
- `ActivateExporter` signature and the `internal/cli/enterprise.go` call site are unchanged.
- Early return when the embedded `key.pem` is empty (dev builds send nothing).
- `ExportHost`/`ExportDelay`/`ExportPeriod` overrides and their defaults (`https://m.rp.vectorized.io`, 5m delay, 24h period) preserved.
- Same wire contract: POST `/connect/telemetry`, `User-Agent: RedpandaConnect/<version>`, RS256 JWT signed with the embedded key, `key_generation: 1` protected header, identical payload (ID, uptime, components, host info, deployment type, tenant ID) with uptime recomputed each export.
- Payload extraction (`payload.go`) untouched; drop-on-error semantics retained via the shared `Reporter`.

Deleted: `telemetryExporter`, `export`, `exporterLoop`, local RSA parsing, and the resty log adapter (~65 lines of plumbing). Added: a 10-line benthos→`telemetry.Logger` shim.

### Two intentional, benign deltas
- The JWT protected header now also carries the conventional `"typ":"JWT"` (added by the shared signer). The ingestion service's JWT parsing ignores unknown header fields.
- The export cadence uses a ticker instead of `sleep`, so the interval no longer drifts by the send duration.

## Dependencies / merge order
**DEPENDS ON** redpanda-data/common-go#172. `go.mod` pins `github.com/redpanda-data/common-go/telemetry v0.0.0-20260610110850-da889a76852d` (the PR branch head) — CI stays red until that PR merges to `main`; re-pin to the merge commit if it lands with a different SHA. `resty`/`go-jose` moved to indirect requires as expected.

### DIFF
--- a/internal/telemetry/telemetry.go
+++ b/internal/telemetry/telemetry.go
@@ -96,7 +96,7 @@ func ActivateExporter(identifier, version, deploymentType, tenantID string, logg
 		JWTHeaders:    map[string]any{"key_generation": 1},
 	})
 	if err != nil {
-		logger.With("error", err).Debug("failed to build telemetry client")
+		logger.With("error", err).Debug("Failed to build telemetry client")
 		return
 	}
 
--- a/internal/telemetry/telemetry.go
+++ b/internal/telemetry/telemetry.go
@@ -15,15 +15,10 @@
 package telemetry
 
 import (
-	"crypto/rsa"
-	"crypto/x509"
-	"encoding/pem"
-	"errors"
+	"context"
 	"time"
 
-	"github.com/go-jose/go-jose/v4"
-	josejwt "github.com/go-jose/go-jose/v4/jwt"
-	"github.com/go-resty/resty/v2"
+	commontelemetry "github.com/redpanda-data/common-go/telemetry"
 
 	"github.com/redpanda-data/benthos/v4/public/service"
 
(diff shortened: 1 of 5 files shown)


internal/telemetry/telemetry.go
```diff
@@ -125,52 +82,35 @@ func ActivateExporter(identifier, version, deploymentType, tenantID string, logg
...
+	client, err := commontelemetry.New(commontelemetry.Config{
+		Endpoint:      exportHost,
+		Path:          "/connect/telemetry",
+		UserAgent:     "RedpandaConnect/" + version,
+		SigningKeyPEM: []byte(privateKey),
+		JWTHeaders:    map[string]any{"key_generation": 1},
+	})
 	if err != nil {
-		t.logger.With("error", err).Debug("Failed to get token string")
+		logger.With("error", err).Debug("failed to build telemetry client")
```
josephwoodward:
```suggestion
		logger.With("error", err).Debug("Failed to build telemetry client")
```

josephwoodward (approved):

PR #4504, opened 2026-06-11
bigquery: fix CDC row-error indexing and metric consistency

squiidz:
Report the original batch index in CDC row-error messages, count batchesSent on partial row failures in the non-CDC path to match the CDC path, and stop re-counting CDC validation failures into rowsFailed on every transient retry of the same batch.

### DIFF
--- a/internal/impl/gcp/enterprise/bigquery/output.go
+++ b/internal/impl/gcp/enterprise/bigquery/output.go
@@ -619,8 +619,10 @@ func bqWriteAPIConfigFromParsed(pConf *service.ParsedConfig) (conf bqWriteAPICon
 }
 
 type bqwaMetrics struct {
-	rowsSent                *service.MetricCounter
-	rowsFailed              *service.MetricCounter
+	rowsSent   *service.MetricCounter
+	rowsFailed *service.MetricCounter
+	// batchesSent counts batches whose append RPC succeeded, including batches
+	// with per-row failures — row-level detail lives in rowsSent/rowsFailed.
 	batchesSent             *service.MetricCounter
 	batchLatency            *service.MetricTimer
 	retries                 *service.MetricCounter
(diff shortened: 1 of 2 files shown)


Jeffail (dismissed):
LGTM but we could do with some new unit test coverage (not the metrics part as thats more awkward)

To add some specifics beneath the above: the three changes all look correct on inspection, and the metric-counting assertions are reasonably left out — counter values aren't readily observable through the public `service.MetricCounter` API, so pinning them in a unit test is genuinely awkward.

The one behaviour that is worth covering, and is testable without touching metrics, is the **row-error index fix** in `writeBatchCDC` — the change from `idx` to `rowIndex[idx]` in the failure message:

```go
batchErr = appendRowFailure(batchErr, batch, "row errors from BigQuery", rowIndex[idx],
    fmt.Errorf("row %d: code %d: %s", rowIndex[idx], re.GetCode(), re.GetMessage()))
```

This is exactly the kind of off-by-mapping bug that silently regresses, since the message text and the attached index used to disagree. The difficulty is that `writeBatchCDC` calls `swd.stream.AppendRows` directly, so there's no seam to inject a response carrying `GetRowErrors()` at chosen positions, and the existing `WriteBatch` unit tests only cover the not-connected and empty-batch paths.

The cheapest way to make it testable is to extract the row-errors-to-`batchErr` mapping loop into a small pure helper, e.g.:

```go
func mapRowErrorsToBatch(batchErr ..., batch service.MessageBatch, rowIndex []int, rowErrs []*storagepb.RowError) ...
```

A table test can then assert that, for a batch where (say) rows 0 and 2 failed validation — so `rowIndex` maps sent-position → original-position non-trivially — a BigQuery row error at sent-position `k` surfaces a failure pinned to `rowIndex[k]` **and** a message that reads `row <rowIndex[k]>`, not `row <k>`. That pins both halves of the fix that previously disagreed.

Happy to re-review once there's coverage for that path. Everything else looks good.

internal/impl/gcp/enterprise/bigquery/output.go
josephwoodward: subjective nit (non-blocker): maybe declare these in a batch? It often signifies "these all belong to the same loop below."

```go
var (
	batchErr           *service.BatchError
	rows               = make([][]byte, 0, len(batch))
	rowIndex           = make([]int, 0, len(batch))
	validationFailures = 0
)
```

Totally subjective though so do feel free to ignore if you don't agree.

josephwoodward (approved):

josephwoodward (approved):

PR #4510, opened 2026-06-12
aws_dynamodb_cdc: only checkpoint contiguously acked positions

squiidz:
Shard checkpoints previously persisted the highest acked sequence regardless of gaps, so out-of-order acks or a nacked batch could move the checkpoint past undelivered records and silently lose them after a crash. Track batches per shard with an ordered checkpoint tracker (as the kinesis input does) so only the highest contiguous acked position is ever persisted; a nacked batch pins the frontier and is redelivered on restart.

Also wait for multi-table shard readers to exit before coordinators return, fixing a send-on-closed-channel panic during shutdown.


internal/impl/aws/dynamodb/batcher.go
```diff
@@ -101,69 +168,65 @@ type checkpointer interface {
...
-			}
-			shardMessageCounts[cp.shardID]++
-			delete(b.messageTracker, msg)
-		}
+	tb, ok := b.batches[batch[0]]
+	if !ok {
+		// Already removed (nacked or double-acked); nothing to do.
+		return nil
 	}
+	b.dropBatchLocked(tb)
```
josephwoodward: dropBatchLocked before the shard-limit error

If the shard limit is hit, the batch is already dropped from b.batches but its checkpoint.Uncapped slot is never resolved. That shard's checkpoint frontier is permanently pinned for the rest of the process lifetime. On a busy pipeline where the limit is hit transiently, this could silently halt checkpointing for an entire shard until restart - which looks like stalled progress with no clear error in metrics.

josephwoodward (approved):

PR #4523, opened 2026-06-16
doris_stream_load: update version the connector was released in

### LINKED ISSUE #4519: doris_stream_load output declares .Version("4.86.0") but was added in 4.96.0
## Summary

The `doris_stream_load` output declares `.Version("4.86.0")` in its config spec, but the connector was actually introduced in **4.96.0**. This makes the generated reference docs show "Introduced in version 4.86.0", which contradicts the 4.96.0 release notes and the auto-generated "What's new" entry that lists it as a new component in 4.96.0.

## Evidence

- Source: [`internal/impl/doris/output_stream_load.go`](https://github.com/redpanda-data/connect/blob/main/internal/impl/doris/output_stream_load.go) calls `.Version("4.86.0")`.
- CHANGELOG: 4.96.0 lists "Added new stream load output". The output does not appear in any release prior to 4.96.0.

## Impact

Documentation generated from connector metadata (rp-connect-docs) reports the wrong introduction version for this component, confusing readers about when it became available.

## Suggested fix

Update the version string on the `doris_stream_load` output config spec from `4.86.0` to `4.96.0`.

## Workaround (docs side)

In the meantime, rp-connect-docs has corrected the rendered page and added a `version` override in `docs-data/overrides.json`. Correcting the source value here will keep future auto-generated docs accurate without a manual override.

josephwoodward:
Closes: https://github.com/redpanda-data/connect/issues/4519

This change also skips the `doris_stream_load` integration tests as they fail in CI due to JVM/Doris memory requirements

### DIFF
--- a/internal/impl/doris/output_stream_load.go
+++ b/internal/impl/doris/output_stream_load.go
@@ -177,7 +177,7 @@ func dorisStreamLoadSpec() *service.ConfigSpec {
 	return service.NewConfigSpec().
 		Beta().
 		Categories("Services").
-		Version("4.86.0").
+		Version("4.96.0").
 		Summary("Writes batches of messages into Apache Doris using Stream Load.").
 		Description(dorisStreamLoadDescription()).
 		Field(service.NewStringField(dsFieldURL).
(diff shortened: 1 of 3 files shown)


internal/impl/doris/output_stream_load.go
```diff
@@ -177,7 +177,7 @@ func dorisStreamLoadSpec() *service.ConfigSpec {
 	return service.NewConfigSpec().
 		Beta().
 		Categories("Services").
-		Version("4.86.0").
+		Version("4.96.0").
```
josephwoodward, line 180: https://github.com/redpanda-data/connect/releases/tag/v4.96.0

Jeffail (approved):

PR #4527, opened 2026-06-17
protobuf: disable hyperpb PGO to stop profile-retention memory leak

squiidz:
hyperpb v0.1.3 leaks profile data across profile-guided recompiles: MessageType.Recompile clones the prior type's compile options and appends another WithProfile, and CompileMessageDescriptor persists that growing slice in the new type's Library.Metadata. The live type therefore pins every prior generation's profile recorder (a 32 KB ring buffer per field), so heap grows unbounded under sustained decode throughput until the pod is OOMKilled. Only the final profile is used during compilation, so the retained generations serve no purpose.

Disable recompilation (RecompileInterval: 0) in the default profiling options used by the protobuf and schema_registry_decode processors, and set Rate to 0 since the never-recompiled profile is never consumed. Re-enable once the upstream hyperpb leak is fixed.

### DIFF
--- a/internal/impl/protobuf/common/decode_common.go
+++ b/internal/impl/protobuf/common/decode_common.go
@@ -35,7 +35,18 @@ type ProfilingOptions struct {
 
 // DefaultProfilingOptions are the standard profiling settings used across all
 // hyperpb decoder call sites.
+//
+// Profile-guided recompilation is currently disabled (RecompileInterval: 0)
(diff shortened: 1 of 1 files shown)


twmb (approved):

josephwoodward (approved):

josephwoodward:
Do we not have any integration tests over this package?

PR #4529, opened 2026-06-17
aws_dynamodb_cdc: support DynamoDB Global Tables for the checkpoint t…

squiidz:
aws_dynamodb_cdc: support DynamoDB Global Tables for the checkpoint table

    Add `global_table` / `global_table_replicas` options so the auto-created
    checkpoint table is provisioned as a DynamoDB Global Table (v2), replicating
    shard checkpoints across regions for low-RPO multi-region failover.

    Checkpoints are keyed by a region-portable identifier (the source table name)
    and store the stream ARN and record timestamp. Resume is layered: same-region
    restarts resume exactly via AfterSequenceNumber (unchanged behavior), while a
    failed-over region resumes from the trim horizon, skipping records at or before
    the low-water-mark timestamp of the prior region's checkpoints. Table creation
    enables DynamoDB Streams and provisions/reconciles replica regions; all new
    behavior is gated behind `global_table` and is a no-op when disabled.

### DIFF
--- a/internal/impl/aws/dynamodb/input_cdc.go
+++ b/internal/impl/aws/dynamodb/input_cdc.go
@@ -1431,22 +1431,34 @@ func (d *dynamoDBCDCInput) isCDCCheckpointStale(ctx context.Context) (bool, erro
 	for _, shard := range shards {
 		shardID := *shard.ShardId
 
-		// Check if we have a checkpoint for this shard
-		checkpoint, err := d.checkpointer.Get(ctx, shardID)
-		if err != nil || checkpoint == "" {
-			if err != nil {
-				d.log.Warnf("Failed to get checkpoint for shard %s: %v", shardID, err)
-			}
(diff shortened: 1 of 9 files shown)


internal/impl/aws/dynamodb/input_cdc.go
```diff
@@ -742,6 +775,9 @@ func newDynamoDBCDCInputFromConfig(pConf *service.ParsedConfig, mgr *service.Res
 	if err != nil {
 		return nil, err
 	}
+	if conf.globalTable && awsConf.Region == "" {
+		return nil, errors.New("global_table requires an AWS region to be configured (set `region` or the AWS_REGION environment)")
```
josephwoodward:
```suggestion
		return nil, errors.New("global_table requires an AWS region to be configured (set `region` or the AWS_REGION environmental variable)")
```

josephwoodward (approved):

PR #4545, opened 2026-06-23
aws_dynamodb_cdc: recover from expired shard iterators

squiidz:
DynamoDB Streams shard iterators are valid for only 15 minutes. When an iterator expired (prolonged output backpressure, a long idle period, or a stale iterator obtained just before a restart), GetRecords returned an ExpiredIteratorException that fell through to the generic error handler: log, wait poll_interval, retry with the same dead iterator. The affected shard looped forever and only recovered on a pod restart, accumulating a data gap in the meantime.

Both shard reader paths now detect ExpiredIteratorException and obtain a fresh iterator, resuming from the most precise position available: the last sequence number actually read from the shard, then the persisted checkpoint, and finally the configured start position. Resuming after the last read sequence reproduces exactly where a healthy iterator would be, so recovery introduces no data gap and no duplicates beyond the pipeline's normal at-least-once guarantee.

Fixes CON-508.

### DIFF
--- a/internal/impl/aws/dynamodb/input_cdc.go
+++ b/internal/impl/aws/dynamodb/input_cdc.go
@@ -413,6 +413,11 @@ type dynamoDBShardReader struct {
 	shardID   string
 	iterator  *string
 	exhausted bool
+	// lastSequenceNumber is the sequence number of the most recent record read
+	// from the shard. It is used to obtain a fresh iterator that resumes exactly
+	// where reading left off if the current iterator expires (see
+	// refreshExpiredIterator).
+	lastSequenceNumber string
 }
 
 // snapshotState encapsulates all state related to snapshot scanning.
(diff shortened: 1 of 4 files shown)


internal/impl/aws/dynamodb/input_cdc.go
```diff
@@ -1884,6 +1955,33 @@ func (d *dynamoDBCDCInput) startTableShardReader(ctx context.Context, tableName
...
+				d.log.Warnf("Shard %s (table %s) iterator expired, refreshing from last position", shardID, tableName)
+				ts.mu.RLock()
+				var lastSeq string
+				if reader, ok := ts.shardReaders[shardID]; ok {
+					lastSeq = reader.lastSequenceNumber
+				}
+				ts.mu.RUnlock()
+
+				newIter, refreshErr := d.refreshExpiredIterator(ctx, ts.checkpointer, ts.streamArn, shardID, lastSeq)
+				if refreshErr != nil {
```
josephwoodward, line 1967: What does the experience look like if we're unable to refresh the expired iterator? Does it keep trying?

internal/impl/aws/dynamodb/input_cdc.go
```diff
@@ -1799,6 +1804,72 @@ func (d *dynamoDBCDCInput) getShardIterator(shardID string) *string {
...
+		seq = checkpoint
+	}
+	if seq != "" {
+		return types.ShardIteratorTypeAfterSequenceNumber, &seq
+	}
+	if startFrom == "latest" {
+		return types.ShardIteratorTypeLatest, nil
+	}
+	return types.ShardIteratorTypeTrimHorizon, nil
+}
```
josephwoodward, line 1840: subjective nit (non-blocker):

Perhaps the following would be a little clearer:

```go
func resolveResumeIterator(lastSeq, checkpoint, startFrom string) (types.ShardIteratorType, *string) {
	switch {
	case lastSeq != "":
		return types.ShardIteratorTypeAfterSequenceNumber, &lastSeq
	case checkpoint != "":
		return types.ShardIteratorTypeAfterSequenceNumber, &checkpoint
	case startFrom == "latest":
		return types.ShardIteratorTypeLatest, nil
	default:
		return types.ShardIteratorTypeTrimHorizon, nil
	}
}
```

josephwoodward (approved):

internal/impl/aws/dynamodb/input_cdc.go
squiidz: done

internal/impl/aws/dynamodb/input_cdc.go
squiidz: Yes, On a refresh failure it logs, waits poll_interval, and continues

PR #4551, opened 2026-06-24
iceberg: add row-level operations (insert/upsert/delete) to the output

Jeffail:
## What

Extends the `iceberg` output beyond append-only with per-message row-level operations: `insert`, `upsert`, and `delete`.

The operation is selected per message by a new interpolated `row_operation` field, and the row identity for upserts/deletes is configured via `identifier_fields` (the equality-delete key, matching Iceberg's `identifier-field-ids`). The mapping is source-agnostic — a change-data-capture `op` field can be mapped to these operations upstream, for example — so no CDC-specific format is assumed.

## Behaviour

- **`insert`** (the default) keeps the existing append-only fast path entirely unchanged, so existing configurations are unaffected.
- **`upsert`** writes an equality delete on `identifier_fields` plus the new row, committed atomically as a merge-on-read `RowDelta` (overwrite) snapshot — the latest row wins.
- **`delete`** writes an equality delete only.
- Partitioned tables include the partition source columns in delete records so deletes route to the correct partition.

### Safety

The feature is opt-in and fails loudly rather than silently mishandling data:

- a static `upsert`/`delete` without `identifier_fields`, or an unknown static `row_operation`, is rejected at startup;
- a dynamic `row_operation` that resolves to an unknown value, or to upsert/delete without `identifier_fields`, errors per message — there is no silent fallback to insert;
- identifier and partition key column types that cannot yet be carried losslessly are rejected with a clear error.

## Testing

- Unit tests: config validation, per-message batch split, and real equality-delete file production plus commit semantics (append/delete/overwrite snapshots) over a local-filesystem catalog.
- Integration test: an insert → upsert → delete round-trip asserting the final table state via the REST catalog.

## Known limitations / follow-ups

- Equality and partition key columns are currently limited to JSON-safe primitive types; temporal, decimal, binary, and uuid keys are rejected with a clear error pending verification against a live catalog.
- Positional deletes and copy-on-write modes are out of scope here.

### DIFF
--- a/docs/modules/components/pages/outputs/iceberg.adoc
+++ b/docs/modules/components/pages/outputs/iceberg.adoc
@@ -120,6 +120,8 @@ output:
     namespace: analytics.events # No default (required)
     table: user_events # No default (required)
     case_sensitive_columns: true
+    row_operation: insert
+    identifier_fields: []
     storage:
       aws_s3:
         bucket: my-iceberg-data # No default (required)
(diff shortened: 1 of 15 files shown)


Jeffail:
Closing in favour of a fresh pull request: the review history here had accumulated a lot of stale, noisy threads that made it hard to follow. The branch and content are unchanged — now a single clean commit, with CI green. A replacement PR follows; apologies to reviewers for the churn.

Jeffail:
Continued in #4567.

PR #4553, opened 2026-06-24
fix(salesforce): include createable fields in writable set

ness-david-dedu:


Jeffail (changes requested):
Looks like theres a bit of extra work needed

Thank you for this — the underlying fix is sound. Platform Event custom fields (and other createable-but-not-updateable fields) are createable-only, so the previous updateable-only filter silently dropped them; including createable fields is the right call for `insert` and for the default `upsert` operation.

The concern is that the writable set is now operation-agnostic, while Salesforce field writability is not. `DescribeWritableFields` returns `createable ∪ updateable` for every operation, and `filterRecord` applies it uniformly across both the realtime and bulk paths:

- For `operation: update`, the union now admits createable-but-not-updateable fields, which Salesforce rejects on update with `INVALID_FIELD_FOR_INSERT_UPDATE`. In the realtime path this triggers a reject/block/retry cycle; the bulk path has no equivalent handling, so those rows simply fail.
- More importantly, `writableFields` and `blockedFields` are keyed by SObject only — not by operation — and the cache is shared across all topic mappings. When an `update` mapping causes a createable-only field to be blocked, that field is permanently removed from the shared set, so a subsequent `insert` or `upsert` mapping for the same SObject silently drops it again, reintroducing exactly the data loss this change set out to fix. The behaviour is order-dependent and persists for the lifetime of the process.

Suggested direction: make the writable set operation-aware — `insert` → createable, `update` → updateable, `upsert` → the union (the one operation where a record may insert or update, so the union plus the existing blocklist is the pragmatic choice) — and key the `writableFields`/`blockedFields` caches by `(SObject, operation)` so the two operations cannot poison one another. A test exercising a mixed insert/update configuration against the same SObject would guard the regression.

Happy to re-review once the writable set is operation-aware.

Jeffail (approved):
The operation-aware writable set and per-`(SObject, operation)` cache keying address the earlier concern in full: `insert` now uses createable fields, `update` uses updateable, and `upsert` retains the union for its per-record insert/update ambiguity, while the blocklist can no longer leak across operations on a shared SObject. The rewritten `TestDescribeWritableFields` covers the per-operation selection and passes locally.

Approving — thank you for the thorough turnaround.

josephwoodward (approved):

PR #4560, opened 2026-06-25
CDC tracing: emit snapshot/stream/checkpoint spans across all CDC inputs

twmb:
## What

Adds OpenTelemetry tracing spans around the **snapshot**, **stream**, and **checkpoint-commit** paths of every CDC input, per CONTRIBUTING.md §5.4.4.

Before this change, an audit confirmed **none** of the nine CDC inputs emitted tracing spans (verified in code — including the SQL CDC connectors, not assumed). Each connector now acquires a tracer via the established repo pattern `mgr.OtelTracer().Tracer("<connector>")` (the same API used by `internal/agent` and `internal/mcp`) and wraps each phase in a span. Errors are recorded on the span with an error status.

## Connectors

| Connector | snapshot | stream | checkpoint-commit |
|---|---|---|---|
| `aws_dynamodb_cdc` | ✅ | ✅ | ✅ |
| `oracledb_cdc` | ✅ | ✅ | ✅ |
| `mysql_cdc` | ✅ | ✅ | ✅ |
| `microsoft_sql_server_cdc` | ✅ | ✅ | ✅ |
| `postgres_cdc` | ✅ | ✅ | ✅ |
| `mongodb_cdc` | ✅ | ✅ | ✅ |
| `salesforce_cdc` | ✅ | ✅ | ✅ |
| `cockroachdb_changefeed` | n/a¹ | ✅ | ✅ |
| `gcp_spanner_cdc` | n/a¹ | ✅ | ✅ |

¹ Changefeed / change-stream connectors have no distinct snapshot phase — the initial backfill arrives as part of the change stream. This is called out in code comments and in the per-connector commit.

Span names are connector-scoped, e.g. `postgres_cdc.snapshot`, `postgres_cdc.stream`, `postgres_cdc.checkpoint_commit`.

## Notes

- One commit per connector.
- For `postgres_cdc`, the actual phases live in the `pglogicalstream` subpackage, so instrumentation is applied there (tracer threaded through the stream `Config`, defaulting to a no-op tracer when unset).
- For `salesforce_cdc`, the OTel `codes` package is aliased to `otelcodes` to avoid colliding with the existing `google.golang.org/grpc/codes` import.
- License headers match each connector's distribution (RCL, except the Apache-2.0 `cockroachdb_changefeed`).

## Tests

Each connector has a `TestCDCTracingSpans` unit test that drives the connector's tracing seam with an in-memory recording tracer (`tracetest.NewInMemoryExporter` + `sdktrace`) and asserts that correctly-named spans are emitted for the snapshot and stream phases, and that phase errors are recorded on the span with an error status.

### DIFF
--- a/internal/impl/aws/dynamodb/batcher.go
+++ b/internal/impl/aws/dynamodb/batcher.go
@@ -14,6 +14,9 @@ import (
 	"sync"
 
 	"github.com/Jeffail/checkpoint"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/noop"
 
 	"github.com/redpanda-data/benthos/v4/public/service"
 )
@@ -40,6 +43,10 @@ type RecordBatcher struct {
 	maxTrackedShards   int
 	maxTrackedMessages int
 	log                *service.Logger
+	// tracer wraps the live checkpoint-commit path (CONTRIBUTING.md §5.4.4).
+	// Defaults to a no-op tracer; the input installs the real tracer after
+	// construction.
+	tracer trace.Tracer
 
 	mu sync.Mutex
 
@@ -101,6 +108,7 @@ func NewRecordBatcher(maxTrackedShards, checkpointLimit int, log *service.Logger
 		maxTrackedShards:   maxTrackedShards,
 		maxTrackedMessages: maxTrackedMessages,
 		log:                log,
+		tracer:             noop.NewTracerProvider().Tracer("aws_dynamodb_cdc"),
 		batches:            make(map[*service.Message]*trackedBatch),
 		shards:             make(map[string]*shardAckTracker),
 	}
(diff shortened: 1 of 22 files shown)


twmb:
Addressed the checkpoint-commit finding: the `aws_dynamodb_cdc.checkpoint_commit` span was only emitted from `flushCheckpoint` (Close path), so nothing was traced during normal streaming. The live commit in `RecordBatcher.AckMessages` (`cp.Set`) is now wrapped in the same span via a small `commitCheckpoint` helper, so dynamodb matches every other connector's live commit path. The batcher receives the tracer from the input (no-op default so existing batcher tests are unaffected), and a new unit test acks past the checkpoint limit and asserts the live commit emits the span.

Also rebased onto latest `main` (resolved conflicts in `oracledb` and `dynamodb/batcher.go` — the latter picking up main's new `Set` signature) and force-pushed. Full build + all nine scoped `TestCDCTracing` suites pass.

PR #4565, opened 2026-06-26
cmd: embed the IANA tz database in distributed binaries

squiidz:
Components that resolve named locations via time.LoadLocation (e.g. the jira input, which renders JQL date predicates in the account's timezone) depend on a tz database being available at runtime. The release container images are busybox-based and ship no /usr/share/zoneinfo, so LoadLocation fails for any non-UTC zone and callers silently fall back to UTC, reintroducing the data-loss windows those features close. It passes in development only because dev machines carry system tzdata.

Add an internal/tzdata package that blank-imports time/tzdata, and import it from every distributed entrypoint (the four distributions plus the Lambda serverless build) so the embedded database (~450 KiB, used only as a fallback when no system database is found) travels with the binary rather than depending on the runtime image.

A host-independent regression test asserts each shipping entrypoint pulls in time/tzdata, so CI catches a future change that drops the embed before a stripped image silently regresses to UTC.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ Changelog
 
 All notable changes to this project will be documented in this file.
 
+## Unreleased
+
+### Fixed
+
+- The IANA time zone database is now embedded in all distributed binaries, so components that resolve named time zones (such as the `jira` input rendering JQL date predicates in the account timezone) work correctly in the stripped release container images, which carry no system zoneinfo.
+
 ## 4.98.0 - 2026-06-26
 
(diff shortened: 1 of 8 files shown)


mihaitodor:
I was expecting this to be handled by https://github.com/redpanda-data/connect/blob/d33d3abd864cbb896cf7c8e2d49da388adfeadc8/.goreleaser/connect.yaml#L26. Might be annoying for people who import the bundles and build custom binaries if the code embeds the tz data directly and they can’t opt out.

PR #4567, opened 2026-06-29
iceberg: add row-level operations (insert/upsert/delete) to the output

Jeffail:
## What

Extends the `iceberg` output beyond append-only with per-message row-level operations — `insert`, `upsert`, and `delete` — selected by a new interpolated `row_operation` field. The row identity for upserts/deletes is configured via `identifier_fields` (the equality-delete key, matching Iceberg's `identifier-field-ids`). The mapping is source-agnostic: a change-data-capture `op` field can be mapped to these operations upstream, so no CDC-specific format is assumed.

> Supersedes the previous pull request, which had accumulated stale review noise. The branch, the (now single-commit) history, and the content are otherwise current, and CI is green.

## Behaviour

- **`insert`** (the default) keeps the existing append-only fast path unchanged — existing configurations are unaffected.
- **`upsert`** writes an equality delete on `identifier_fields` plus the new row, committed atomically as a merge-on-read `RowDelta` snapshot (latest row wins).
- **`delete`** writes an equality delete only.
- Within a batch, the last operation per identifier key wins; each delete-bearing batch commits as its own snapshot, so concurrent changes to the same key cannot leave duplicates.
- Any primitive identifier type is supported (integers, decimal, uuid, temporal, …), each encoded so the equality-delete key matches what the insert path writes. Floating-point and bare-numeric temporal keys are rejected as unreliable equality keys.
- Partitioned tables include the partition source columns in delete records and require every partition source to be an identifier field (equality deletes are partition-scoped).
- On auto-created tables, `identifier_fields` are registered as the schema's identifier-field-ids and marked required, so downstream engines see the primary key.
- Version-1 tables are upgraded to version 2 (required for equality deletes) on the first upsert/delete; the upgrade is irreversible and warned once.

### Ordering

Keyed writes rely on per-key ordering, so concurrent in-flight batches could otherwise commit out of order. Both a config lint rule and a startup warning flag a mutating `row_operation` combined with `max_in_flight > 1`; set `max_in_flight: 1` for keyed / change-data-capture workloads.

### Safety

Opt-in, and fails loudly rather than silently mishandling data:

- a static `upsert`/`delete` without `identifier_fields`, or an unknown static `row_operation`, is rejected at startup;
- a dynamic `row_operation` that resolves to an unknown value, or to upsert/delete without `identifier_fields`, errors per message — there is no silent fallback to insert;
- unsupported key types and missing/null identifier values produce clear errors.

### Observability

Adds `iceberg_rows_inserted`, `iceberg_rows_upserted`, `iceberg_rows_deleted`, and `iceberg_commit_failures` metrics, plus best-effort cleanup of written-but-uncommitted files when a commit fails.

## Testing

- **Unit:** config and lint-rule validation, per-message batch split and per-key collapse, equality-delete file production, and commit / format-upgrade / retry / cleanup semantics over a local-filesystem catalog.
- **Integration (real REST catalog + MinIO + DuckDB):** insert → upsert → delete round-trips across key types, composite keys, partitioned tables, and batch collapse — all passing.

## Follow-ups (post-merge)

- Add a CI lane for the iceberg integration tests (currently opt-in and skipped by default).
- Document table-maintenance guidance (snapshot expiry and `rewrite_data_files`) for merge-on-read delete-file accumulation.
- Positional deletes and copy-on-write modes are out of scope here.

### DIFF
--- a/internal/impl/iceberg/committer.go
+++ b/internal/impl/iceberg/committer.go
@@ -1,4 +1,4 @@
-// Copyright 2025 Redpanda Data, Inc.
+// Copyright 2026 Redpanda Data, Inc.
 //
 // Licensed as a Redpanda Enterprise file under the Redpanda Community
 // License (the "License"); you may not use this file except in compliance with
@@ -13,6 +13,7 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/apache/iceberg-go"
@@ -29,9 +30,16 @@ import (
 const CurrentIcebergVersion = 2
 
 // CommitInput holds data files and the schema ID they were written with.
+//
+// Files are inserted (appended) data files. DeleteFiles are equality-delete
+// files produced for upsert/delete operations (ContentType EntryContentEqDeletes
+// with EqualityFieldIDs set). When DeleteFiles is empty the commit is a pure
+// append and takes the optimised AddDataFiles fast path; otherwise the changes
+// are applied atomically through a RowDelta.
 type CommitInput struct {
-	Files    []iceberg.DataFile
-	SchemaID int
+	Files       []iceberg.DataFile
+	DeleteFiles []iceberg.DataFile
+	SchemaID    int
 }
 
 // CommitConfig holds configuration for the committer.
@@ -60,7 +68,12 @@ type committer struct {
 	cfg         CommitConfig
 	reloadTable func(ctx context.Context) (*table.Table, error)
 	batcher     *asyncroutine.Batcher[CommitInput, struct{}]
-	logger      *service.Logger
+	// commitMu serializes all commits and guards c.table. The batcher's
+	// doCommit and the direct commitRowDelta path both take it.
+	commitMu        sync.Mutex
+	upgradeWarnOnce sync.Once
+	metrics         *opMetrics
+	logger          *service.Logger
 }
 
 // NewCommitter creates a new committer for a specific table.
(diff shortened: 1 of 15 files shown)


internal/impl/iceberg/output_iceberg.go
```diff
@@ -67,6 +67,25 @@ type icebergOutput struct {
...
+	deleted        *service.MetricCounter
+	commitFailures *service.MetricCounter
+}
+
+func newOpMetrics(m *service.Metrics) *opMetrics {
+	return &opMetrics{
+		inserted:       m.NewCounter("iceberg_rows_inserted"),
+		upserted:       m.NewCounter("iceberg_rows_upserted"),
+		deleted:        m.NewCounter("iceberg_rows_deleted"),
+		commitFailures: m.NewCounter("iceberg_commit_failures"),
```
josephwoodward: We should follow the [Prometheus naming convention](https://prometheus.io/docs/practices/naming/) here of suffixing metrics names with their unit (in this case `_total`).

Also, wouldn't it be better to collapse these into a single metric where the verb is a label? This is what the HTTP client does and enables better easier aggregation.

For instance:

```
sum(rate(iceberg_row_operations_total[5m]))
```

versus 

```
sum(rate(iceberg_rows_inserted[5m])) + sum(rate(iceberg_rows_upserted[5m])) + ...
````

File: internal/impl/iceberg/output_iceberg.go [hunk 2]
```diff
@@ -67,6 +67,25 @@ type icebergOutput struct {
 	logger *service.Logger
 }
 
+// opMetrics holds counters for row-level operations and commit health. Fields
+// are nil in tests that construct writers/committers directly; callers must
+// nil-check the holder before incrementing.
+type opMetrics struct {
```
josephwoodward, line 76: Seeing as these metrics are encapsulated in their own type, would it make sense to use nil-safe methods to make them no-op calls and avoid needing to check for nils?

ie:

```go
func (m *opMetrics) incrInserted(n int64)      { if m != nil { m.inserted.Incr(n) } }
func (m *opMetrics) incrUpserted(n int64)      { if m != nil { m.upserted.Incr(n) } }
func (m *opMetrics) incrDeleted(n int64)        { if m != nil { m.deleted.Incr(n) } }
func (m *opMetrics) incrCommitFailures(n int64) { if m != nil { m.commitFailures.Incr(n) } }
```

internal/impl/iceberg/output_iceberg.go
```diff
@@ -497,6 +535,39 @@ func parseSchemaEvolutionConfig(conf *service.ParsedConfig) (SchemaEvolutionConf
 	return cfg, nil
 }
 
+// parseRowOpConfig parses the row-level operation configuration and validates
+// the static case at startup. When row_operation is a static literal it must be
+// a known operation, and a static upsert/delete requires identifier_fields.
+// Dynamic (interpolated) operations cannot be checked here, so they are
+// validated per message at write time (a hard error on an unknown value or a
+// missing identifier_fields key, never a silent insert).
+func parseRowOpConfig(conf *service.ParsedConfig) (RowOpConfig, error) {
```
josephwoodward, line 567: Would it make sense to move this to the config.go file?

internal/impl/iceberg/output_iceberg.go
```diff
@@ -497,6 +535,39 @@ func parseSchemaEvolutionConfig(conf *service.ParsedConfig) (SchemaEvolutionConf
...
+	cfg := RowOpConfig{}
+
+	op, err := conf.FieldInterpolatedString(ioFieldRowOperation)
+	if err != nil {
+		return cfg, err
```
josephwoodward, line 572: If there's an error, is there a reason for returning an incomplete config here (and below) instead of nil? I couldn't see an instance where we use the config even when we get an error.

internal/impl/iceberg/writer.go
```diff
@@ -75,39 +134,102 @@ func (w *writer) Write(ctx context.Context, batch service.MessageBatch) error {
...
+		}
+		return nil
+	}
+
+	// Row-level mutations: split the batch by operation, write inserted rows as
+	// data files and deleted/upserted keys as equality-delete files, then commit
+	// them together so a single snapshot reflects the whole batch.
+	inserts, deletes, err := w.splitByOperation(batch)
+	if err != nil {
+		return err
```
josephwoodward: Would it be helpful to the user to add context to the errors returned from this and the errors below?

internal/impl/iceberg/writer.go
```diff
@@ -75,39 +134,102 @@ func (w *writer) Write(ctx context.Context, batch service.MessageBatch) error {
...
+}
+
+// cleanupFiles best-effort removes written-but-uncommitted parquet files after a
+// failed commit, to limit orphaned objects in storage. Failures are logged at
+// debug and otherwise ignored (table maintenance / remove_orphan_files remains
+// the backstop).
+func (w *writer) cleanupFiles(ctx context.Context, groups ...[]iceberg.DataFile) {
+	fs, err := w.table.FS(ctx)
+	if err != nil {
+		return
```
josephwoodward, line 196: Should we be logging this error too?

internal/impl/iceberg/writer.go
```diff
@@ -163,12 +285,331 @@ func (w *writer) Write(ctx context.Context, batch service.MessageBatch) error {
...
+		if ko.op == rowOpUpsert {
+			inserts = append(inserts, ko.msg)
+		}
+	}
+
+	if w.metrics != nil {
+		w.metrics.inserted.Incr(nInsert)
+		w.metrics.upserted.Incr(nUpsert)
+		w.metrics.deleted.Incr(nDelete)
+	}
```
josephwoodward: Shouldn't these be moved to after the commit so they more accurately reflect the metrics of actual insert, updates and deletes? Also, nUpsert counts every upsert in a batch but then deduplictes them so would show over counting.

internal/impl/iceberg/writer.go
```diff
@@ -163,12 +285,331 @@ func (w *writer) Write(ctx context.Context, batch service.MessageBatch) error {
...
+// equality-delete record — the identifier fields (the equality key) plus, for
+// partitioned tables, the partition source columns needed to route each delete
+// to its partition — together with the identifier field IDs (the equality key).
+// The field list is ordered and de-duplicated (an identifier field that is also
+// a partition source appears once). Every contributing column must be a
+// primitive type (a fundamental Iceberg constraint on identifier fields).
+func (w *writer) deleteRecordFields(tableSchema *iceberg.Schema, spec *iceberg.PartitionSpec) ([]iceberg.NestedField, []int, error) {
+	seen := map[int]struct{}{}
+	var fields []iceberg.NestedField
+	eqFieldIDs := make([]int, 0, len(w.rowOpCfg.IdentifierFields))
```
josephwoodward: nit (subjective so feel free to ignore):

```go
	var (
		fields     []iceberg.NestedField
		seen       = map[int]struct{}{}
		eqFieldIDs = make([]int, 0, len(w.rowOpCfg.IdentifierFields))
	)
```

internal/impl/iceberg/committer.go
```diff
@@ -81,28 +94,83 @@ func NewCommitter(tbl *table.Table, cfg CommitConfig, reloadTable func(ctx conte
...
+}
+
+// commitRowDelta applies a single input's inserts and equality deletes as one
+// atomic snapshot, outside the batcher so it is never coalesced with another
+// keyed commit.
+func (c *committer) commitRowDelta(ctx context.Context, input CommitInput) error {
+	c.commitMu.Lock()
+	defer c.commitMu.Unlock()
+
+	if input.SchemaID != c.currentSchemaID() {
```
josephwoodward: Would it be worth calling `c.currentSchemaID()` once and assigning it to a local variable that can be used here and below? Theoretically the ID could change if the underlying `c.table` is modified.

internal/impl/iceberg/writer.go
```diff
@@ -163,12 +285,331 @@ func (w *writer) Write(ctx context.Context, batch service.MessageBatch) error {
...
+		case json.Number:
+			return n.String(), nil
+		case int:
+			return strconv.FormatInt(int64(n), 10), nil
+		case int32:
+			return strconv.FormatInt(int64(n), 10), nil
+		case int64:
+			return strconv.FormatInt(n, 10), nil
+		case float64:
+			if n != math.Trunc(n) {
```
josephwoodward, line 432: Claude found this one:

1. Float64 → int64 overflow in deleteKeyJSONValue (writer.go:421)

For integer identifier columns where the incoming value is a float64 >= 2^63, int64(n) overflows — Go wraps it to a garbage value. The math.Trunc check only rejects non-integers, not out-of-range values. This silently writes a corrupted equality-delete key, meaning the delete targets the wrong row (or no row at all) and the upsert leaves a duplicate. Data corruption with no error surfaced.

A secondary tier of the same bug: values in the range (2^53, 2^63) pass the check but float64 can't represent them exactly, so the key can silently mismatch the stored row for the same reason.

josephwoodward (approved):

PR #4576, opened 2026-07-02
postgres_cdc: add signalling support

josephwoodward:
This change introduces signalling support into the PostgreSQL CDC connector allowing you to perform operations such as signal a new snapshot on an adhoc basis. It does this via a custom created signal table which is also being tracked by PostgreSQL's replication functionality, with control signals propagating from the WAL. When a signal is detected the signaller we determine the signal type then handle it accordingly. Currently only snapshots are supported but other signal types can added over time (such as logging).

Upon receiving a signal event, the process will:

1. Received snapshot signal, setting the state to re-snapshot on restart then pauses streaming and performs snapshotting of the table(s) specified in the signal payload (it can be all tables, or just new tables).
2. Once snapshotting is completed the connector resume streaming from its original cached checkpoint.

### Demonstration

Given the following config:

```yaml
input:
  postgres_cdc:
    dsn: ${PG_DSN:postgres://postgres:postgres@localhost:5432/testdb?sslmode=disable}
    stream_snapshot: true
    schema: public
    tables:
      - users
      - products
      - cart
    slot_name: bench_slot
    signal_table_name: rpcn_signal_table
    batching:
      period: 1s

output:
  file:
    path: "./benchmark_results.json"

```

In the screenshot below the process goes:

1. Launch the component and stream an initial 100 records into `benchmark_results.json`
2. Run `task psql:signal:snapshot` to publish a snapshot signal to the configured signalling table which the connector receives, pauses streaming then restarts to snapshot (now 201 records, 100x2 plus 1 signal event)
4. Connector returns to streaming (301 records)

image

### DIFF
--- a/docs/modules/components/pages/inputs/postgres_cdc.adoc
+++ b/docs/modules/components/pages/inputs/postgres_cdc.adoc
@@ -99,6 +99,7 @@ input:
       role: "" # No default (optional)
       role_external_id: "" # No default (optional)
       roles: [] # No default (optional)
+    signal_table_name: ""
     auto_replay_nacks: true
     batching:
       count: 0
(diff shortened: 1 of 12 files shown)


josephwoodward:
Closing to clean up PR history

PR #4584, opened 2026-07-06
postgres_cdc: wait for snapshot acks before promoting the replication…

squiidz:
During the initial snapshot the connector emitted snapshot rows (which carry no LSN and are therefore never tracked for postgres acknowledgement) and then immediately promoted the temporary replication slot to a permanent one and began streaming from startLSN, without waiting for those rows to be acknowledged downstream. A crash in that window left the permanent slot in place, so on restart the snapshot phase was skipped and the un-acked rows were silently lost.

Add an acknowledgement barrier across the stream/input boundary, mirroring the mysql input's snapshot_complete sentinel:

- The stream emits a SnapshotCompleteOpType sentinel after the last snapshot row, then blocks (until acked or shutdown) before CopyReplicationSlot.
- The input tracks in-flight snapshot batches with a WaitGroup and, on the sentinel, flushes and waits for every snapshot batch to be acknowledged before calling Stream.MarkSnapshotAcknowledged to release the barrier.

A crash before the snapshot rows are acked now leaves the temporary slot to expire and the permanent slot uncreated, so the snapshot re-runs on restart with no row loss.

Add an integration test that reproduces the crash-during-handoff: it receives the snapshot rows without acking, cancels the run, asserts the slot was not promoted, then restarts and asserts the snapshot re-runs. Verified to fail without the barrier fix.

### DIFF
--- a/internal/impl/postgresql/input_pg_stream.go
+++ b/internal/impl/postgresql/input_pg_stream.go
@@ -482,10 +482,15 @@ func (p *pgStreamInput) processStream(pgStream *pglogicalstream.Stream, batcher 
 				flushedBatch, err := batcher.Flush(ctx)
 				if err != nil {
 					p.logger.Debugf("error flushing snapshot completion batch: %s", err)
+					// The sentinel is a one-shot signal; if we bail here without
+					// acking, the barrier's snapshot goroutine blocks on
+					// snapshotAcked forever. Trigger a restart instead of stalling.
+					p.stopSig.TriggerSoftStop()
 					break
 				}
 				if err := p.flushBatch(ctx, pgStream, cp, flushedBatch); err != nil {
 					p.logger.Debugf("failed to flush snapshot completion batch: %s", err)
+					p.stopSig.TriggerSoftStop()
 					break
 				}
 				drained := make(chan struct{})
--- a/internal/impl/postgresql/input_pg_stream.go
+++ b/internal/impl/postgresql/input_pg_stream.go
@@ -15,6 +15,7 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/Jeffail/checkpoint"
@@ -396,6 +397,12 @@ type pgStreamInput struct {
 	replicationLag  *service.MetricGauge
 	stopSig         *shutdown.Signaller
 
+	// snapshotAckWG tracks in-flight snapshot batches: incremented when a
+	// snapshot batch (nil LSN) is enqueued and decremented when it is
+	// acknowledged. The snapshot->stream handoff blocks until it drains so the
+	// replication slot is not promoted before snapshot rows are durable.
+	snapshotAckWG sync.WaitGroup
+
 	// IAM authentication fields
 	iamAuthEnabled bool
 }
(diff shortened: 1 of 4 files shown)


Jeffail:
Hello squiidz — one robustness concern about the recovery path this change makes the norm, plus two smaller notes. The barrier logic itself (WaitGroup sequencing, streaming only starting past `CopyReplicationSlot`, abort paths all biasing toward *not* promoting the slot) looks correct.

**1. A lingering temporary slot can make restart crash-loop** (`internal/impl/postgresql/pglogicalstream/logical_stream.go:234`)

The barrier deliberately leaves the connector in the temporary-slot state — `<slot>_tmp` created, permanent slot not promoted — across a crash during the handoff. On restart, startup only checks for the *permanent* slot (`:187`, `:196`); finding none, it takes the snapshot path and unconditionally issues `CREATE_REPLICATION_SLOT <slot>_tmp`. `CreateReplicationSlot` is a raw command with no `IF NOT EXISTS` and no drop-first (`pglogrepl.go:221`), so if the crashed process's `<slot>_tmp` has not yet been reaped, this errors, `Connect` fails (`input_pg_stream.go:418-420`), and the input retries — a crash-loop until the slot is released.

A temporary slot is released only when its owning backend goes away. For a local crash with a prompt socket close that is quick; for an ungraceful failure with no FIN (node loss, partition) Postgres holds it until `wal_sender_timeout` (60s default), and a fast automated restart can also race the reap. Confirmed against Postgres 16 — while the owning session is still alive:

```
SELECT pg_create_logical_replication_slot('handoff_tmp','pgoutput', true);
--> ERROR: replication slot "handoff_tmp" already exists
```

No data is lost, so this is a conditional robustness gap rather than a correctness bug, but it can turn a clean re-run into a period of failed restarts under exactly the crash types the fix targets. Suggested guard: before creating `<slot>_tmp` on the snapshot path, drop any pre-existing one (tolerating "does not exist"). It only ever belongs to a dead session, so reaping it on the way in is safe.

**2. The integration test does not exercise this path** (`integration_test.go`)

The test calls `pg_terminate_backend` on the process still holding `<slot>_tmp` and waits for it to disappear before restarting, which side-steps the lingering-slot behaviour above. With the guard in place the terminate could be dropped so the test asserts the real restart path directly; without it, an added case that restarts *without* terminating would cover the scenario.

**3. Minor** — in the sentinel handler, a `batcher.Flush`/`flushBatch` error `break`s without releasing the barrier or triggering a soft stop, leaving the snapshot goroutine blocked on `snapshotAcked`. A `TriggerSoftStop()` on those branches would turn a silent stall into a clean restart.

squiidz:
Thanks for the detailed review — pushed 9edeafe addressing all three points:

**1. Lingering temp slot / crash-loop:** added a drop-before-create guard in `Connect()` — before issuing `CREATE_REPLICATION_SLOT <slot>_tmp`, it now drops any pre-existing one first, tolerating "does not exist". Since we only reach that path when no permanent slot exists, a leftover `_tmp` is necessarily stale.

**2. Integration test coverage:** I tried dropping the `pg_terminate_backend` step as suggested, and verified empirically it doesn't work in this harness — Postgres kept reporting the old `_tmp` slot as `active for PID ...` for 30+ seconds after the in-process context cancel, since Go's shutdown path doesn't force-close the replication connection promptly (unlike a real process crash, where the OS closes the socket immediately). The guard correctly refuses to drop a slot that's still genuinely active — that's not a bug, just not something this harness can trigger without also killing the connection. Kept the terminate-and-wait step; it now exercises the guard's "does not exist" path once the old slot is actually gone. Happy to add a second, longer-timeout test case specifically for the still-active branch if you think it's worth the CI time — didn't add one unprompted.

**3. Sentinel flush errors:** added `TriggerSoftStop()` to both `batcher.Flush`/`flushBatch` error branches in the snapshot-completion handling, so a flush failure there triggers a clean restart instead of leaving the barrier's snapshot goroutine blocked on `snapshotAcked` forever (it's a one-shot sentinel, so the old silent `break` had no way to recover).

Full postgresql integration suite + lint are green locally.

Jeffail (approved):
Thank you, @squiidz — the follow-up looks good and addresses the earlier feedback. The snapshot-completion flush paths now trigger a soft stop rather than silently breaking out of the select, which cleanly releases the acknowledgement barrier and restarts instead of stalling; I traced the release chain through to the stream's soft-stop cancellation and it holds. The core data-loss barrier remains sound. Happy to approve.

PR #4585, opened 2026-07-07
oracledb_cdc: handle row id based predicates

josephwoodward:
In Oracle, a ROWID is Oracle's internal physical address for a row - it's not a column in the table; it's a pseudo-column Oracle maintains internally and under some circumstances it's possible that OracleDB will fall back to using it to locate rows (tables without primary keys, LOB based operations).

So this change handles such scenario, publishing the change as an event and logging the SQL via the `DEBUG` flag (it's possible it could contain sensitive information so should be an opt-in behaviour).


Jeffail (approved):
Ace, thanks @josephwoodward!

PR #4589, opened 2026-07-07
postgres_cdc: Add multi-schema support

ness-david-dedu:
Multi-schema replication — schema field now accepts a glob pattern (e.g. tenant_*) in addition to an exact name. All
  matching schemas are replicated through a single replication slot and publication, removing the need for one slot per
  tenant.

  pg_schema metadata — schema name of the originating table, useful for per-tenant routing downstream.

  Bug fix — validateSchemaPattern accepted "" (empty quoted identifier, length 2) due to an off-by-one; corrected to
  require at least one character inside the quotes.


squiidz (commented):
Re-reviewed after the rebase — the critical issue from the earlier pass is resolved: `stream_message.go` and `replication_message_decoders.go` are no longer touched, so the duplicate `CommitTs`/`Before`/tuple-decoding reimplementation (which had drifted from what `main` already shipped in #4554/#4555) is gone. The diff is now correctly scoped to the new schema-glob + `pg_schema` + publication-batching work layered on top of `main`'s existing `commit_ts_ms`/`before` support. 👍

A few remaining items from the original pass, still present:

1. **gofmt**: the struct literal in `input_pg_stream.go` isn't gofmt-clean —
   ```go
   DBSchemaPattern: schema,
   DBTables:        tables,
   ```
   `gofmt -w` wants `DBSchemaPattern:  schema,` / `DBTables:         tables,` to stay column-aligned with `RefreshAuthToken:` in the same literal. This will fail a `gofmt -l`/CI format check as-is.

2. `validateSchemaPattern`'s quoted-identifier check (`input_pg_stream.go`) reimplements a shallow prefix/suffix/length check instead of delegating to `sanitize.UnquotePostgresIdentifier`, which this PR already uses in `schema_resolver.go`. The existing helper correctly handles escaped double-quotes (`"foo""bar"`); the new check doesn't.

3. No `CHANGELOG.md` entry — recent `postgres_cdc` PRs (#4554, #4555, ...) all added one under "Unreleased".

4. `tests/current/` (Taskfile/docker-compose/setup.sql/test_config.yaml) reads like a personal manual-test harness — worth confirming this is meant to be committed long-term vs. a scratch dir that should be dropped or renamed to something more descriptive.

5. Worth a doc callout: schema-pattern resolution happens once at stream construction (`resolveSchemas` runs at startup only). For the stated multi-tenant use case, a new `tenant_c` schema created after the pipeline starts won't be picked up until restart. Not a bug, but non-obvious given the motivating use case.

6. Test coverage gap: no test for the "zero schemas matched" startup-error path (`no schemas found matching pattern %q`).

None of these block correctness — (1) is the only one that will mechanically fail CI as-is.

squiidz (approved):

Jeffail (changes requested):
nice feature but theres a failure mode that i think takes down the whole input:

with a glob `schema` + explicit `tables`, `logical_stream.go` builds the cartesian product `schemas × tables` and every combo has to exist. `tables` is required so we always hit the `FOR TABLE ...` path in `CreatePublication` (runs at startup before snapshot). if any matched schema is missing a listed table:

- fresh run → `CREATE PUBLICATION ... FOR TABLE "tenant_x"."missing"` errors `relation does not exist`
- restart → same thing via the batched `ALTER PUBLICATION ... ADD TABLE`

so one drifted/half-provisioned schema that merely *matches* the glob halts replication for **every** tenant, not just the bad one. and this is exactly the documented flow — new schemas need a restart to get picked up, so a freshly created `tenant_*` schema thats not fully migrated yet will kill the whole thing on the next restart.

what id like:
- resolve tables per-schema — skip (with a warn log) any listed table thats missing from a given matched schema, so drift degrades gracefully instead of taking everything down
- if thats too much for v1, at minimum document that every matched schema must contain every listed table + that restarting mid-provision will stop the input

nit: the PR description lists `commit_ts_ms` and `before` as features but those are already on main, this pr is just the multi-schema glob (changelog is correct though 👍)

rest looks solid — the O(n²)→O(n) publication reconcile + batched ALTER are a nice touch, and `pg_schema` sourcing from the WAL relation is right

internal/impl/postgresql/pglogicalstream/config.go
```diff
@@ -24,8 +24,10 @@ type Config struct {
 	DBConfig  *pgconn.Config
 	DBRawDSN  string
 	TLSConfig *tls.Config
-	DBSchema  string
-	DBTables  []string
+	// DBSchemaPattern is the schema to replicate from. Accepts an exact schema
+	// name or a glob pattern using '*' as a wildcard (e.g. "tenant_*", "*").
+	DBSchemaPattern string
```
josephwoodward: Can we use the `confix.RegexpFilter` type here instead of plain strings? See the [Oracle connector's use here](https://github.com/redpanda-data/connect/blob/main/internal/impl/oracledb/input_oracledb_cdc.go?plain=1#L233) for an example of how to use it.

internal/impl/postgresql/pglogicalstream/schema_resolver.go
```diff
@@ -0,0 +1,168 @@
+// Copyright 2024 Redpanda Data, Inc.
```
josephwoodward:
```suggestion
// Copyright 2026 Redpanda Data, Inc.
```

internal/impl/postgresql/pglogicalstream/config.go
josephwoodward: Ignore this for the time being.

internal/impl/postgresql/pglogicalstream/schema_resolver.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+func resolveSchemas(ctx context.Context, conn *pgconn.PgConn, pattern string) ([]string, error) {
+	likePattern, err := schemaPatternToLike(pattern)
+	if err != nil {
+		return nil, err
+	}
+
+	q, err := sanitize.SQLQuery(
+		"SELECT schema_name FROM information_schema.schemata WHERE schema_name LIKE $1 ESCAPE '!' AND schema_name NOT LIKE 'pg!_%' ESCAPE '!' AND schema_name != 'information_schema'",
```
josephwoodward: `information_schema.schemata ` only shows schemas the role has access to, which is fine but may be misleading when a user wants to include a schema the role currently doesn't have access to then they're silently ignored. This could be resolved with `pg_catalog.pg_namespace` as it's not privilege-filtered the way `information_schema.schemata` is, so we'd be able to match on schemas the role doesn't have privilege to and make the user aware via an error or warning.

internal/impl/protobuf/processor_protobuf_test.go
```diff
@@ -51,7 +51,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/net/http2"
-	"golang.org/x/net/http2/h2c" //nolint:staticcheck
```
josephwoodward, line 54: What are these changes for? They don't appear related to the PostgreSQL changes?

internal/impl/otlp/mock_policy_server_test.go
```diff
@@ -18,7 +18,7 @@ import (
 	policymaterializerv1 "buf.build/gen/go/redpandadata/common/protocolbuffers/go/redpanda/policymaterializer/v1"
 	"connectrpc.com/connect"
 	"golang.org/x/net/http2"
-	"golang.org/x/net/http2/h2c" //nolint:staticcheck
+	"golang.org/x/net/http2/h2c"
```
josephwoodward: These changes don't seem related either?

internal/gateway/authz_endpoint_test.go
```diff
@@ -20,7 +20,7 @@ import (
 	policymaterializerv1 "buf.build/gen/go/redpandadata/common/protocolbuffers/go/redpanda/policymaterializer/v1"
 	"connectrpc.com/connect"
 	"golang.org/x/net/http2"
-	"golang.org/x/net/http2/h2c" //nolint:staticcheck
+	"golang.org/x/net/http2/h2c"
```
josephwoodward: These changes don't appear related, is there a failure on the main branch? If so it'd be great if we could move these to a separate PR.

internal/impl/postgresql/pglogicalstream/logical_stream.go
```diff
@@ -98,18 +98,43 @@ func NewPgStream(ctx context.Context, config *Config) (*Stream, error) {
...
+	schemas, inaccessibleSchemas, err := resolveSchemas(ctx, dbConn, config.DBSchemaPattern)
 	if err != nil {
-		return nil, fmt.Errorf("invalid schema name %q: %w", config.DBSchema, err)
+		return nil, fmt.Errorf("resolving schema pattern %q: %w", config.DBSchemaPattern, err)
 	}
+	if len(inaccessibleSchemas) > 0 {
+		config.Logger.Warnf("schema pattern %q matches schema(s) %v that the configured role cannot see (missing USAGE privilege); they will be skipped", config.DBSchemaPattern, inaccessibleSchemas)
+	}
+	if len(schemas) == 0 {
+		return nil, fmt.Errorf("no schemas found matching pattern %q", config.DBSchemaPattern)
```
josephwoodward: **Behaviour regression for the documented FOR ALL TABLES mode**

This hard-fails startup whenever the pattern resolves to zero schemas, unconditionally. But when tables is left empty the connector documents schema as being ignored:

> If left empty, the underlying PostgreSQL publication is created FOR ALL TABLES, which replicates every table in every schema of the database, ignoring schema.

(see [input_pg_stream.go L128-L134](https://github.com/redpanda-data/connect/blob/0e662de9d78a8903fcb2a77d0a709127442ae83f/internal/impl/postgresql/input_pg_stream.go#L128-L134))

schema is a required field, so a user running in FOR ALL TABLES mode must set it to something. Previously any syntactically valid value worked; now a value that happens to match no existing schema aborts the pipeline with no schemas found matching pattern, even though the field has no effect in that mode. The code right below already special-cases len(normalizedTables) > 0, so the same guard is needed here — skip schema resolution (or downgrade to a warning) when config.DBTables is empty.

internal/impl/postgresql/input_pg_stream.go
```diff
@@ -381,6 +401,37 @@ func newPgStreamInput(conf *service.ParsedConfig, mgr *service.Resources) (s ser
...
+	for i, ch := range s {
+		if (ch >= 'a' && ch <= 'z') || (ch >= 'A' && ch <= 'Z') || (ch >= '0' && ch <= '9') || ch == '_' || ch == '*' {
+			continue
+		}
+		return fmt.Errorf("invalid character %q at position %d in schema pattern %q", ch, i, s)
+	}
+	first := rune(s[0])
+	if first != '_' && first != '*' && (first < 'a' || first > 'z') && (first < 'A' || first > 'Z') {
+		return fmt.Errorf("schema pattern %q must start with a letter, underscore, or '*'", s)
+	}
```
josephwoodward: **Regression: unquoted schema names containing non-ASCII letters are now rejected at startup.**

This validator only accepts `[a-zA-Z0-9_*]`, but the code path it replaces — `sanitize.NormalizePostgresIdentifier`, previously called on `config.DBSchema` in `NewPgStream` — accepts any `unicode.IsLetter`/`unicode.IsDigit` rune (plus `.`): see [sanitize.go#L443-L457](https://github.com/redpanda-data/connect/blob/620c9da5c1509ada217260b9c255cbf4af373ed2/internal/impl/postgresql/pglogicalstream/sanitize/sanitize.go#L443-L457).

Failure scenario: an existing pipeline with `schema: münchen` (a legal unquoted PostgreSQL identifier that previously normalised to `"münchen"` and worked) now fails config construction with `invalid schema: invalid character 'ü' at position 1 in schema pattern "münchen"`. The user has no way to know quoting is the workaround.

Suggested fix: mirror `NormalizePostgresIdentifier`'s character classes here (`unicode.IsLetter`/`unicode.IsDigit`, plus `_` and `*`) so the pattern validator is a superset of what was previously accepted, rather than a stricter ASCII-only rule.

internal/impl/postgresql/pglogicalstream/logical_stream.go
```diff
@@ -98,18 +98,43 @@ func NewPgStream(ctx context.Context, config *Config) (*Stream, error) {
...
+		for _, table := range normalizedTables {
+			if _, ok := existingTables[table]; !ok {
+				config.Logger.Warnf("table %s.%s not found, skipping (schema %s matched pattern %q but does not contain this table)", schema, table, schema, config.DBSchemaPattern)
+				continue
+			}
+			tables = append(tables, TableFQN{Schema: schema, Table: table})
+		}
+	}
+	if len(tables) == 0 && len(normalizedTables) > 0 {
+		return nil, fmt.Errorf("none of the configured tables %v were found in any schema matching pattern %q", config.DBTables, config.DBSchemaPattern)
```
josephwoodward: **Missing tables are now silently skipped even when `schema` is an exact name (behaviour regression + doc mismatch).**

Before this change, a configured table that didn't exist reached `CreatePublication`'s `FOR TABLE` clause and failed the connect with a hard error. Now any table not found in a schema is only warned about and dropped from `tables`, and the error at [logical_stream.go#L136-L138](https://github.com/redpanda-data/connect/blob/620c9da5c1509ada217260b9c255cbf4af373ed2/internal/impl/postgresql/pglogicalstream/logical_stream.go#L136-L138) only fires when *every* table is missing.

Failure scenario: `schema: public`, `tables: [orders, ordres]` (typo). Previously the pipeline failed loudly with the offending relation named; now it starts, replicates `orders`, and the typo'd entry is only visible as a `WARN` line — silent partial data loss for a config error.

This also contradicts the documentation added in this PR, which scopes the leniency to glob patterns: "When `schema` is a glob pattern, this list is resolved against each matched schema independently: a table missing from a given schema is skipped (with a warning logged)" ([input_pg_stream.go#L133-L135](https://github.com/redpanda-data/connect/blob/620c9da5c1509ada217260b9c255cbf4af373ed2/internal/impl/postgresql/input_pg_stream.go#L133-L135)).

Suggested fix: only tolerate a missing table when the pattern actually resolved to more than one schema (or when the pattern contains a wildcard), and keep the hard error for the single/exact-schema case. Note the same skip path also masks a privilege problem: `information_schema.tables` is privilege-filtered, so a table the role can't see is reported as "not found".

Ref: [CONTRIBUTING.md §1.2.4](https://github.com/redpanda-data/connect/blob/620c9da5c1509ada217260b9c255cbf4af373ed2/CONTRIBUTING.md) — "Strongly lints and validates user-provided configuration, clearly telling users of any problems."

internal/impl/postgresql/pglogicalstream/logical_stream.go
josephwoodward: Addressed in [6803eca](https://github.com/redpanda-data/connect/pull/4589/commits/6803eca6c66378366969e56d025ddb0b02df2951)

internal/impl/postgresql/pglogicalstream/logical_stream.go
josephwoodward: Addressed in [045cde2](https://github.com/redpanda-data/connect/pull/4589/commits/045cde2e534650affb661c1a6fd1c7acbe04e7ce)

internal/impl/postgresql/pglogicalstream/logical_stream.go
```diff
@@ -98,18 +98,107 @@ func NewPgStream(ctx context.Context, config *Config) (*Stream, error) {
 		return nil, err
 	}
 
-	schema, err := sanitize.NormalizePostgresIdentifier(config.DBSchema)
-	if err != nil {
-		return nil, fmt.Errorf("invalid schema name %q: %w", config.DBSchema, err)
-	}
+	var tables []TableFQN
+	var schema string
```
josephwoodward:
```go
var (
    tables []TableFQN
	var schema string
)
```

internal/impl/postgresql/input_pg_stream.go
```diff
@@ -310,6 +348,44 @@ func newPgStreamInput(conf *service.ParsedConfig, mgr *service.Resources) (s ser
...
+			schemaPattern = strings.ToLower(schemaPattern)
+		}
+	}
+
+	if excludeSchemas, err = conf.FieldStringList(fieldExcludeSchemas); err != nil {
+		return nil, err
+	}
+	if len(excludeSchemas) > 0 {
+		if schemaPattern == "" {
+			return nil, errors.New("exclude_schemas requires schema_pattern to be set")
```
josephwoodward: Claude Review highlighted the following:

**`exclude_schemas` is silently ignored when `tables` is empty**

This is the only guard on `exclude_schemas`, and it only rejects the case where `schema_pattern` is unset. But when `tables` is left empty, the resolved-and-filtered schema list is never used: `NewPgStream` ends up with an empty `tables` slice, and `CreatePublication` falls back to `FOR ALL TABLES` ([pglogrepl.go#L296-L307](https://github.com/redpanda-data/connect/blob/52dbd2bb8984ec4aff9a3a769cdc05026ec7e9c7/internal/impl/postgresql/pglogicalstream/pglogrepl.go#L296-L307)), which replicates every schema in the database.

So this config starts cleanly and then streams `tenant_c` anyway:

```yaml
postgres_cdc:
  schema_pattern: tenant_*
  exclude_schemas: [tenant_c]
  # tables omitted -> FOR ALL TABLES
```

The `schema_pattern` field docs do call out the empty-`tables` caveat, but the `exclude_schemas` docs only say "Only valid when `schema_pattern` is set" ([input_pg_stream.go#L145-L150](https://github.com/redpanda-data/connect/blob/52dbd2bb8984ec4aff9a3a769cdc05026ec7e9c7/internal/impl/postgresql/input_pg_stream.go#L145-L150)) — so a user excluding a tenant schema for data-isolation reasons gets no signal that the exclusion did nothing.

Suggested fix: extend this check to also fail (or at minimum log a warning) when `exclude_schemas` is set and `tables` is empty, and document the interaction on the field. Per [CONTRIBUTING.md §1.2.4](https://github.com/redpanda-data/connect/blob/52dbd2bb8984ec4aff9a3a769cdc05026ec7e9c7/CONTRIBUTING.md), certified connectors must "strongly lint and validate user-provided configuration, clearly telling users of any problems".

josephwoodward: In this case we don't want the customer to have to specify every table in the `tables` configuration so I don't think the suggested fix would work.

What we want is when `schema_pattern` is set and tables is empty, to auto-discover every table in each matched schema and build the publication as FOR TABLE <scoped list> instead of falling back to global FOR ALL TABLES. This makes exclude_schemas actually work as intended and lets customers use schema_pattern without listing tables at all - which is the whole feature they're asking for.

internal/impl/postgresql/input_pg_stream.go
```diff
@@ -37,6 +40,8 @@ const (
 	fieldSnapshotMemSafetyFactor   = "snapshot_memory_safety_factor"
 	fieldSnapshotBatchSize         = "snapshot_batch_size"
 	fieldSchema                    = "schema"
+	fieldSchemaPattern             = "schema_pattern"
+	fieldExcludeSchemas            = "exclude_schemas"
```
josephwoodward: CDC config-naming conformance ([CONTRIBUTING §5.3.1](https://github.com/redpanda-data/connect/blob/55cf17dda5c835ff935ddeb77ee4f5c13c059f7f/CONTRIBUTING.md), §5.3.3, §5.5.7): the fleet's established mechanism for selecting relations across multiple schemas is the canonical `include`/`exclude` pattern-list pair, not per-purpose bespoke fields — `oracledb_cdc` ([input_oracledb_cdc.go#L205-L214](https://github.com/redpanda-data/connect/blob/55cf17dda5c835ff935ddeb77ee4f5c13c059f7f/internal/impl/oracledb/input_oracledb_cdc.go#L205-L214)) takes `include`/`exclude` as regexes over `SCHEMA.TABLE`, and `microsoft_sql_server_cdc` uses the same pair ([input_mssqlserver_cdc.go#L36-L37](https://github.com/redpanda-data/connect/blob/55cf17dda5c835ff935ddeb77ee4f5c13c059f7f/internal/impl/mssqlserver/input_mssqlserver_cdc.go#L35-L38)). §5.3.1 lists `tables` **or** `include`/`exclude` as canonical, §5.3.3 says not to invent bespoke names where a canonical one exists, and §5.5.7 anchors include/exclude matching to `oracledb_cdc`.

`schema_pattern` + `exclude_schemas` introduces a third selection shape with glob rather than regex semantics, which also means a user moving between CDC connectors has to learn two different multi-schema selection idioms (§3.1.5). Worth confirming with the CDC-standard owners whether this should instead land as `include`/`exclude` over `schema.table` patterns (postgres_cdc's existing `schema` + `tables` is grandfathered, but this is new surface, which §5 requires to conform from the start). Note the gate at `internal/plugins/cdctest` does not currently assert include/exclude naming, so it won't catch this either way.

josephwoodward: Perhaps `schema_include` and `schema_exclude`?

internal/impl/postgresql/pglogicalstream/logical_stream.go
ness-david-dedu: done

internal/impl/postgresql/input_pg_stream.go
ness-david-dedu: done

internal/impl/postgresql/input_pg_stream.go
ness-david-dedu: done

PR #4600, opened 2026-07-13
feat(mongodb): prevent binary fields from being incorrectly saved as base64 strings

disaykin:
Fixed an issue where binary data (`[]byte`) was always serialized as a `base64` string in MongoDB. This occurred because the output plugin was converting the structure to a JSON string and then re-parsing it using the `extJSON` parser, which loses the native binary type.

Replaced `extJSONFromMap` with `structuredFromMap` leveraging `msg.AsStructured()`. This prevents immediate string serialization and preserves native Go slices, enabling the MongoDB driver to properly handle binary payloads.

#### Backward Compatibility
To maintain full backward compatibility and prevent breaking downstream client applications, the old serialization behavior (saving binary fields as base64 strings) is kept by default.

A new configuration option, `preserve_binary` (defaulting to `false`), has been introduced for both MongoDB output and processor components. Users can set this flag to `true` to opt-in to the correct native BSON `BinData` representation.

Some integration and unit tests have been added to verify both the default and new execution paths.

### DIFF
--- a/docs/modules/components/pages/outputs/mongodb.adoc
+++ b/docs/modules/components/pages/outputs/mongodb.adoc
@@ -53,6 +53,7 @@ output:
     filter_map: ""
     hint_map: ""
     upsert: false
+    preserve_binary: false
     max_in_flight: 64
     batching:
       count: 0
@@ -86,6 +87,7 @@ output:
     filter_map: ""
     hint_map: ""
     upsert: false
+    preserve_binary: false
     max_in_flight: 64
     batching:
       count: 0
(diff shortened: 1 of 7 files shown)


disaykin:
Hi team! I've added a dedicated unit and integration tests (output_test.go) inside the internal mongodb package to verify this change using testcontainers. The test successfully passes and confirms that Go-native []byte fields are now correctly preserved and stored as native bson.Binary (BinData) inside MongoDB instead of being forced into base64 strings. Regarding backward compatibility: since this alters the data type stored in the database for raw bytes, please let me know if we should safely gate this behind a new configuration flag (like preserve_binary: false by default), or if you prefer to accept this as the correct default behavior for the connector. Ready for your review!

disaykin:
Update: I went ahead and implemented the `preserve_binary` option (defaulting to false) for both the output and processor components to guarantee absolute backward compatibility out of the box. I've also added integration and unit tests covering both default and new behaviors.

disaykin:
Hi @mmatczuk, could you please take a look when you have a moment?

PR #4602, opened 2026-07-15
aws_dynamodb_cdc: add checkpoint_namespace for checkpoint table sharing

squiidz:
## Summary

Adds an optional `checkpoint_namespace` field to the `aws_dynamodb_cdc` input so multiple independent pipelines (e.g. one per developer or environment) can share a single checkpoint table without overwriting each other's checkpoints.

Checkpoints were keyed only by `(StreamArn, ShardID)` — or `(TableId, ShardID)` in `global_table` mode — so concurrent consumers of the same stream pointing at the same `checkpoint_table` raced on the same rows, causing skipped/duplicated events. The workaround (a checkpoint table per consumer) means IAM and infrastructure sprawl.

## Behavior

- When `checkpoint_namespace` is set, the namespace is prefixed to the hash key **value** with a `#` delimiter: `StreamArn = "dev-alice#arn:aws:dynamodb:..."` (default mode), `TableId = "dev-alice#my-table"` (`global_table` mode). Each namespace occupies its own partition; shard checkpoints, snapshot progress rows, the snapshot completion sentinel, and the global-mode resume query are all scoped by it.
- The default (empty) leaves checkpoint keys byte-identical to previous behavior — no schema change, no migration, existing checkpoint tables are reused as-is.
- Namespaces are mutually invisible: changing (or removing) the namespace makes the pipeline restart from `start_from`. Namespaces do not coordinate consumers — two pipelines sharing the *same* namespace still race, exactly like today.
- A namespace containing `#` is rejected at config validation time. Since DynamoDB table names and stream ARNs cannot contain `#`, the encoding is unambiguous.

## Testing

- Key construction across all four mode × namespace combinations, including that the range key is never namespaced.
- Cross-namespace isolation round-trips through a shared in-memory fake (two namespaces plus an un-namespaced reader; no crosstalk in either direction).
- Snapshot progress reads/writes and the global-mode resume partition query use the namespaced hash value.
- Config parsing (set + default-empty) and validation (`#` rejected).
- All pre-existing checkpointer tests pass unmodified, which pins the empty-namespace backward-compatibility guarantee.

### DIFF
--- a/internal/impl/aws/dynamodb/input_cdc.go
+++ b/internal/impl/aws/dynamodb/input_cdc.go
@@ -66,6 +66,7 @@ const (
 	dciFieldTableTagFilter         = "table_tag_filter"
 	dciFieldTableDiscoveryInterval = "table_discovery_interval"
 	dciFieldCheckpointTable        = "checkpoint_table"
+	dciFieldCheckpointNamespace    = "checkpoint_namespace"
 	dciFieldGlobalTable            = "global_table"
 	dciFieldGlobalTableReplicas    = "global_table_replicas"
 	dciFieldBatchSize              = "batch_size"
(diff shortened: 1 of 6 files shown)


internal/impl/aws/dynamodb/input_cdc.go
```diff
@@ -668,6 +675,10 @@ func validateDynamoDBCDCConfig(conf dynamoDBCDCConfig) error {
 		return errors.New("global_table requires at least one replica region in global_table_replicas")
 	}
 
+	if strings.Contains(conf.checkpointNamespace, "#") {
```
josephwoodward, line 678: thought (non-blocking): would it be worth moving this into a lint rule for the connector?

josephwoodward (approved):

PR #3940, opened 2026-01-29
snowflake: use new parquet-go parallelized row group construction

rockwotj:
Based on https://github.com/parquet-go/parquet-go/pull/339

| Test                | Old build_µs | New build_µs | Speedup | Old rows/sec | New rows/sec | Speedup |
  |---------------------|--------------|--------------|---------|--------------|--------------|---------|
  | 1K_rows_1_worker    | 7,109        | 2,418        | 2.9x    | 142K         | 500K         | 3.5x    |
  | 1K_rows_2_workers   | 2,826        | 1,448        | 2.0x    | 500K         | 1,000K       | 2.0x    |
  | 1K_rows_4_workers   | 3,134        | 1,134        | 2.8x    | 333K         | 1,000K       | 3.0x    |
  | 10K_rows_1_worker   | 31,581       | 20,969       | 1.5x    | 323K         | 500K         | 1.5x    |
  | 10K_rows_2_workers  | 18,549       | 14,387       | 1.3x    | 556K         | 714K         | 1.3x    |
  | 10K_rows_4_workers  | 15,873       | 8,178        | 1.9x    | 667K         | 1,250K       | 1.9x    |
  | 10K_rows_8_workers  | 23,504       | 8,325        | 2.8x    | 435K         | 1,250K       | 2.9x    |
  | 50K_rows_1_worker   | 103,469      | 103,873      | ~1.0x   | 485K         | 485K         | ~1.0x   |
  | 50K_rows_2_workers  | 71,121       | 62,706       | 1.1x    | 704K         | 806K         | 1.1x    |
  | 50K_rows_4_workers  | 57,291       | 38,316       | 1.5x    | 877K         | 1,316K       | 1.5x    |
  | 50K_rows_8_workers  | 51,514       | 16,406       | 3.1x    | 980K         | 3,125K       | 3.2x    |
  | 100K_rows_1_worker  | 208,778      | 203,142      | ~1.0x   | 481K         | 493K         | ~1.0x   |
  | 100K_rows_4_workers | 101,990      | 75,440       | 1.4x    | 990K         | 1,333K       | 1.3x    |
  | 100K_rows_8_workers | 76,764       | 45,790       | 1.7x    | 1,316K       | 2,222K       | 1.7x    |

PR #4605, opened 2026-07-16
snowflake_streaming: verify parquet row counts before upload

squiidz:
## Summary

Snowflake identified a row-count metadata discrepancy in BDEC parquet files uploaded by the `snowflake_streaming` output — the same class of bug as Snowflake's internal SNOW-1465503, which the Java SDK fixed by adding a `verifyRowCounts` guard before upload ([snowflake-ingest-java#784](https://github.com/snowflakedb/snowflake-ingest-java/pull/784)).

The Go path gained parallel row-group construction in [#3940](https://github.com/redpanda-data/connect/pull/3940) (via parquet-go's `ConcurrentRowGroupWriter`) but has no equivalent guard. Worse, the EP registration metadata (`EPS.Rows`) is derived from the same parquet footer, so an internally inconsistent file self-consistently registers with Snowflake and the corruption is only detected server-side.

## Fix

After closing the parquet writer and before encryption/upload, `constructBdecPart` now verifies the footer against the rows actually serialized:

- footer `NumRows` == total rows written
- number and size of row groups == the chunks we wrote
- every column chunk's `NumValues` == its row group's `NumRows` (the BDEC schema is strictly flat — one leaf value per row per column, nulls included — so this is a hard invariant)

Any mismatch fails the batch with a detailed `refusing to upload corrupted file` error naming the offending row group or column, surfacing through normal output retry semantics instead of corrupting the table.

## Testing

`TestVerifyRowCounts` builds a real two-row-group parquet file through the same write path (`writeRowGroupFromObject` → `Commit` → `Close`) and asserts the happy path plus six mismatch cases: wrong expected totals, wrong row-group count, corrupted footer `NumRows`, corrupted row-group `NumRows`, and a short column chunk. All `./internal/impl/snowflake/...` tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 
 - general: The CGO-enabled distribution binary now embeds the IANA time zone database via the `timetzdata` build tag, matching the other distributions, so `time.LoadLocation` works in minimal runtimes without system tzdata instead of silently falling back to UTC (which shifts JQL date predicates in the `jira` input). ([@squiidz](https://github.com/squiidz), [#4583](https://github.com/redpanda-data/connect/pull/4583))
+- snowflake_streaming: BDEC parquet files are now verified for internally consistent row counts (footer, row groups, and column chunks) before upload, so an inconsistent file fails the batch instead of being registered with Snowflake. ([@squiidz](https://github.com/squiidz))
 
 ## 4.99.0 - 2026-07-02
 
--- a/internal/impl/snowflake/streaming/parquet.go
+++ b/internal/impl/snowflake/streaming/parquet.go
@@ -12,6 +12,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/parquet-go/parquet-go"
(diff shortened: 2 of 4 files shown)


Jeffail (approved):
Approving — thank you, @squiidz. The guard is well-placed: it runs after `Close()` but before encryption, upload, and EP registration, so an inconsistent file fails the batch rather than self-consistently registering with Snowflake. Anchoring `expectedRows` to the input message count (rather than footer-internal consistency alone) is a nice touch — since both write paths emit exactly one row per message and abort on any conversion error, it ties the footer back to what was actually fed in, a stronger invariant than a purely self-referential check.

Two non-blocking observations for follow-up, neither of which need hold up this change:

1. This detects rather than cures — the root cause in the concurrent row group writer remains. For a batch that corrupts deterministically, the failed-batch-and-retry path could behave as a poison pill rather than recovering. It would be worth confirming the corruption is understood to be non-deterministic, or tracking the underlying writer fix separately.

2. The check validates every column chunk present but does not assert the number of column chunks per row group, so a footer that dropped a column entirely would pass. That is a different class from the reported discrepancy and likely out of scope here, but noting it for completeness.

PR #4624, opened 2026-07-23
postgres_cdc: add signalling support (without the need to restart)

josephwoodward:
Adds support for signalling a re-snapshot, but fundamentally changes the way the PostgreSQL connector works as the snapshot/streaming modes are quite tightly coupled.


josephwoodward:
This was an experimental branch to understand how a "no restart" version would look and it carries too much complexity versus restarting the connector on a signal.

PR #4636, opened 2026-07-28
ci: dispatch rpk connect plugin docs regeneration on releases

JakeSCahill:
## Summary

The `rpk connect` plugin releases weekly-ish and publishes straight to the rpk plugin manifest, but its command docs in redpanda-data/docs only refresh when a *Redpanda* release triggers full rpk doc regeneration — so they drift for weeks at a time.

This adds one dispatch to the existing **Trigger docs updates** step in the `update-downstream` job: `update-rpk-plugin-docs` with `{plugin: connect, version}`. The docs workflow ([docs#1834](https://github.com/redpanda-data/docs/pull/1834)) installs that connect version against an rpk binary matching its committed command snapshot, refreshes only the connect subtree, and opens a docs PR against each eligible docs branch (currently main and beta).

## Details

- Uses the `ACTIONS_BOT_TOKEN` already present in the step for the other docs dispatches.
- Inherits the job's existing guard (`push` + `!contains(ref_name, '-')`), so RCs and prereleases never dispatch.
- Rapid successive releases update the same open docs PR (version-less branch on the docs side), so there's no PR pile-up.

Generator side is validated against the docs repo's real snapshots (details in [docs-extensions-and-macros#225](https://github.com/redpanda-data/docs-extensions-and-macros/pull/225)).


JakeSCahill:
@mmatczuk could you take a look when you have a moment? One-step addition to the existing docs dispatches in `update-downstream` — it keeps the `rpk connect` reference docs current on each release. The receiving side (redpanda-data/docs#1834) is reviewed and ready.

mmatczuk:
cc: @josephwoodward

JakeSCahill:
Pre-emptive hardening in 798262a, prompted by a review finding on the equivalent cloudv2 sender (redpanda-data/cloudv2#28552): the `Trigger docs updates` step runs inside `update-downstream`, which `notify-slack` depends on, so any dispatch failure (token, API, network) would have failed the job and silenced the release Slack notification. The step now carries `continue-on-error: true` — which also protects the pre-existing docs-ui and rp-connect-docs dispatches that had the same property. Docs failures can never affect the release or its notifications.

PR #4638, opened 2026-07-28
Add short descriptions to component config fields

prakhargarg105:
## What

Adds `short_description` to component config fields so the pipeline configuration UI can render concise inline help instead of full documentation prose.

Field `description` values are Asciidoc and frequently run to hundreds of characters — `postgres_cdc`'s `slot_name` is 460 characters and `heartbeat_interval` is 715 — which makes them unusable as form help text. Rather than shorten the existing descriptions (they are good docs, and the docs site needs them), this adds a separate short form alongside them.

Connect-side follow-up to redpanda-data/benthos#463, which added `FieldSpec.ShortDescription` and the `ConfigField.ShortDescription(...)` builder.

## Scope

- 457 unique short descriptions across 511 call sites, 144 files, 68 directories
- 908 field nodes in the emitted schema now carry `short_description`
- **No existing `description` is modified anywhere**

Fields still uncovered are predominantly defined in benthos itself — shared specs such as TLS, retry/backoff, and auth blocks. Those belong in a follow-up benthos PR, where a single string covers many call sites (`.tls.root_cas_file` alone appears 78 times in the schema).

## Style rules applied

- Plain text only: no Asciidoc, no `xref:`, no backticks, no URLs, no newlines
- Maximum 140 characters, target around 80
- Keep constraints that change what a user types — units, required-together fields, allowed values
- Drop rationale, background, and cross-references
- Never reword, truncate, or otherwise touch the existing `description`

## Dependency note

Pins benthos to the pseudo-version `v4.76.1-0.20260728144106-174e176163e7`, because benthos#463 is merged but not yet tagged. `git log v4.76.0..174e176163e7` contains exactly that one commit, so this is v4.76.0 plus the docs field and nothing else. Happy to rebase onto a real tag before merge if that is preferred — `go.mod` already carries 17 direct-dependency pseudo-versions, so there is precedent either way.

## Verification

- `go build ./...` — clean
- `task test:unit` — 141 packages pass, 0 failures
- `task docs` — zero diff
- **Schema equivalence:** dumped `list --format json-full` from this branch and from `main`, then stripped `short_description` from this branch's output. The two are byte-identical apart from the build `version`/`date` stamp — confirming zero descriptions altered, zero fields lost, and no reordering.
- 0 authored strings exceed 140 characters; 0 contain markup, URLs, or newlines

## Follow-ups (not in this PR)

- benthos-side short descriptions for shared/framework field specs
- The Cloud API and console each carry their own hand-maintained mirror of the field spec, so they need a corresponding update before this surfaces in the UI. Worth noting that `encoding/json` drops unknown fields without error, so until then the value is discarded silently rather than failing loudly.


prakhargarg105:
### CI status note

Both red checks are unrelated to the content of this change — flagging so they aren't misread during review.

**`integration-test-filter` — it timed out, it did not fail.** 110 tests passed, 0 failed, 16 skipped, and 26 of 49 packages completed before the 30-minute cap was reached partway through `mssqlserver`. There is no `FAIL` line anywhere in the log.

The cause is the breadth of the diff rather than anything in it. The detect step derives the package filter from changed paths:

```
git diff --name-only … | grep '^internal/impl/' | cut -d/ -f3 | sort -u
```

Since this PR adds short descriptions to fields in nearly every component, that resolves to **54 categories** (`amqp09` through `zeromq`), so the "filtered" run becomes an effectively full integration sweep. The filtered job is capped at `timeout-minutes: 30`, whereas the full nightly job is given `timeout-minutes: 120` — so this check cannot pass on a PR this broad regardless of correctness.

Happy to run a targeted subset via `/integration <packages>` if reviewers want signal on specific components. Worth noting for CI upkeep more generally: any broad PR — dependency bumps, lint or doc sweeps — will hit this same wall, so capping the detected filter or scaling the timeout with the number of detected categories may be worth considering.

**`claude-review` — missing secret on a fork PR.** The job log shows `ANTHROPIC_API_KEY` as empty. `claude-code-review.yml` triggers on `pull_request` and reads `secrets.ANTHROPIC_API_KEY`, but GitHub does not expose secrets to `pull_request` runs originating from a fork, and this PR is cross-repo. `auto-label-integration.yml` uses `pull_request_target` and does receive secrets, so that is likely the pattern needed here. This affects every fork PR, not just this one.

**On the correctness of this change specifically,** the strongest evidence is schema equivalence: dumping `list --format json-full` from this branch and from `main`, then stripping `short_description` from this branch's output, yields byte-identical JSON apart from the build `version`/`date` stamp. That rules out altered descriptions, lost fields, and reordering. `task test:unit` (141 packages, 0 failures) and `task docs` (zero diff) also pass, and `test` plus `golangci-lint` are green here in CI.

josephwoodward (approved):

josephwoodward:
Kafka tests are passing since pinning the wazero version to main that includes a fix in the dependency.

image

PR #4656, opened 2026-08-02
slack: only enforce token prefixes on literal values

### LINKED ISSUE #4511: Minor issue with slack_post component in Redpanda Connect.
Very low priority, no real impact on me at the moment, just something I noticed that might count as a bug.

When using the "slack_post" output component in a Redpanda Connect pipeline, my intention was to use this YAML in the output stanza:

```yaml
slack_post:
    bot_token: "${secrets.SLACK_APP_BOT_TOKEN}"
```
Unfortunately, it looks like there is some kind of validation testing going on, because trying to create the pipeline returns an error stating that the "bot_token" value must start with "xoxb-". Although that's technically true (all Slack bot tokens that I've seen are in fact prefixed with "xoxb-") it does mean that I can't include that part of the token in the secret -- I have to drop the prefix from the secret value and then code the pipeline YAML as:

```yaml
slack_post:
    bot_token: "xoxb-${secrets.SLACK_APP_BOT_TOKEN}"
```

Not a big deal, and I can leave comments/documentation explaining the issue, but it's the kind of thing that will be confusing to the next person after me who has to maintain the code and/or the app in production, so it might be better to adjust the component validation rules to allow the prior syntax above.

As a quick guess, the fix would be implemented on or about this line of code:

https://github.com/redpanda-data/connect/blob/9f8f9943dc0a5d19384251ac54252a6e1e093f90/internal/impl/slack/output_post.go#L43

hsdfat:
Fixes #4511.

`LintRule(has_prefix("xoxb-"))` runs against the raw config value, before interpolation — so a valid `bot_token: "${secrets.SLACK_APP_BOT_TOKEN}"` was rejected at lint time.

Adds a `tokenLintRule(prefix)` helper that skips the prefix check when the value is empty or starts with `${`, and keeps enforcing it for literals so a malformed literal token is still caught.

The issue reports `output_post.go`, but the same rule is duplicated in `input.go` (xapp- and xoxb-), `input_users.go`, `output_reaction.go` and `processor_thread.go` — all five now go through the helper.

A missing required `bot_token` still produces the separate "field is required" lint, so this doesn't hide a misconfiguration.

Adds `lint_test.go` covering all five components: unresolved `${secrets...}`, an env lookup resolving to empty, valid literal, invalid literal, and a secret whose value starts with the prefix.

### DIFF
--- a/internal/impl/slack/input.go
+++ b/internal/impl/slack/input.go
@@ -33,12 +33,8 @@ func inputSpec() *service.ConfigSpec {
 	return service.NewConfigSpec().
 		Description(`Connects to Slack using https://api.slack.com/apis/socket-mode[^Socket Mode]. This allows for receiving events, interactions and slash commands. Each message emitted from this input has a @type metadata of the event type "events_api", "interactions" or "slash_commands".`).
 		Fields(
-			service.NewStringField(iFieldAppToken).Description("The Slack App token to use.").LintRule(`
-        root = if !this.has_prefix("xapp-") { [ "field must start with xapp-" ] }
-      `),
-			service.NewStringField(iFieldBotToken).Description("The Slack Bot User OAuth token to use.").LintRule(`
-        root = if !this.has_prefix("xoxb-") { [ "field must start with xoxb-" ] }
-      `),
+			service.NewStringField(iFieldAppToken).Description("The Slack App token to use.").LintRule(tokenLintRule("xapp-")),
+			service.NewStringField(iFieldBotToken).Description("The Slack Bot User OAuth token to use.").LintRule(tokenLintRule("xoxb-")),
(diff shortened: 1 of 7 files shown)


hsdfat:
Gentle ping on this one — it's been open since 2 August, and the CLA is signed.

The branch is still green and conflict-free. It fixes #4511: `LintRule(has_prefix("xoxb-"))` runs against the raw config value before interpolation, so a valid `bot_token: "${secrets.SLACK_APP_BOT_TOKEN}"` fails linting. The change adds one helper that skips the prefix check for an empty value or one starting with `${`, and still enforces it for literals, applied to all five places the rule is duplicated.

Happy to adjust the approach if you'd rather handle it elsewhere in the config layer.

PR #4674, opened 2026-08-06
oracledb_cdc: Propagate `V$LOGMNR_CONTENTS.USERNAME` as metadata

josephwoodward:
This change propagates the `username` column from [`$VLOGMNR_CONTENTS`](https://docs.oracle.com/en/database/oracle/oracle-database/21/refrn/V-LOGMNR_CONTENTS.html) as metadata. 

For snapshots (which don't query LogMiner) the metadata value is omitted, it is also omitted if it's `NULL` or empty (`""`) when reading from LogMiner.

### Proof of Work

Running snapshot and streaming shows that:
- During snapshot the `username` is `NULL`, so the configured bloblang converts it to `""`
- When streaming the `username` value is set to `TESTDB`

image


bfbarkhouse-redpanda:
@josephwoodward ETA on this one?

Jeffail (approved):

Leward (approved):

PR #4685, opened 2026-08-10
postgres_cdc: fail the snapshot barrier on nack (CON-504)

squiidz:
Part of CON-504 (CDC at-least-once / ack-gated progress).

Follow-up to #4584: the snapshot ack barrier treated a nack exactly like an ack. The ackFn ignored its error argument, so with `auto_replay_nacks: false` a rejected snapshot batch released the barrier, the replication slot was promoted, and the rejected rows were skipped forever on restart. A nacked batch also resolved its checkpoint slot, letting later acks acknowledge LSNs past undelivered streaming rows.

This applies the pattern hardened through the review rounds on #4675 and #4677:

- A nack never resolves the checkpoint: nothing can be acknowledged past the rejected batch's rows, and an error log identifies the pinned position and the stall-once-`checkpoint_limit`-fills consequence so operators can attribute it.
- A nacked snapshot batch fails the promotion barrier: the input restarts without promoting the slot and the snapshot re-runs.
- The gate error resets per connection attempt, so one nack fails only the attempt it belongs to instead of livelocking reconnects.

### Proof of Work

- `TestIntegrationPostgresSnapshotNackFailsBarrier`: the consumer nacks the first snapshot batch (`auto_replay_nacks: false`), asserting the slot is not promoted while the snapshot is rejected; the input then self-heals — re-runs the snapshot, delivers every row, and only then promotes. Verified it **fails** against the pre-fix code: the rejected rows are never redelivered (`the snapshot should have re-run and re-delivered every row after the nack`, 5m timeout).
- The full postgresql integration suite (13 tests, including the #4584 crash barrier test) passes.
- Unit tests green under -race; lint and docs clean.

### DIFF
--- a/internal/impl/postgresql/input_pg_stream.go
+++ b/internal/impl/postgresql/input_pg_stream.go
@@ -418,9 +418,41 @@ type pgStreamInput struct {
 	// acknowledged. The snapshot->stream handoff blocks until it drains so the
 	// replication slot is not promoted before snapshot rows are durable.
 	snapshotAckWG sync.WaitGroup
+	// snapshotNackErr records the first snapshot batch nack. auto_replay_nacks
+	// is user-toggleable, so a nack can be terminal: the barrier must fail
+	// rather than promote the replication slot over undelivered rows. Cleared
+	// per connection attempt (the input outlives reconnects).
+	snapshotNackMu  sync.Mutex
+	snapshotNackErr error
 
 	// IAM authentication fields
 	iamAuthEnabled bool
+}
+
+func (p *pgStreamInput) recordSnapshotNack(err error) {
+	p.snapshotNackMu.Lock()
+	defer p.snapshotNackMu.Unlock()
+	if p.snapshotNackErr == nil {
+		p.snapshotNackErr = err
+	}
+}
+
+func (p *pgStreamInput) snapshotNackError() error {
+	p.snapshotNackMu.Lock()
+	defer p.snapshotNackMu.Unlock()
+	return p.snapshotNackErr
+}
+
+// resetSnapshotGate clears any nack recorded by a previous connection attempt
+// so the promotion barrier judges only the current run: the input outlives
+// reconnects, and a stale error would fail every retry even after a clean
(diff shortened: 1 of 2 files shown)


internal/impl/postgresql/input_pg_stream.go
```diff
@@ -431,6 +463,10 @@ func (p *pgStreamInput) Connect(ctx context.Context) error {
 		}
 	}
 
+	// The input outlives reconnects: clear any nack recorded by a previous
+	// snapshot attempt so the promotion barrier judges only this run.
```
josephwoodward, line 467: Feels somewhat of a redundant comment.

internal/impl/postgresql/input_pg_stream.go
```diff
@@ -418,11 +418,43 @@ type pgStreamInput struct {
...
+	// rather than promote the replication slot over undelivered rows. Cleared
+	// per connection attempt (the input outlives reconnects).
+	snapshotNackMu  sync.Mutex
+	snapshotNackErr error
 
 	// IAM authentication fields
 	iamAuthEnabled bool
 }
 
+func (p *pgStreamInput) recordSnapshotNack(err error) {
```
josephwoodward, line 432: Is it worth moving `recordSnapshotNack`, `snapshotNackError` and `resetSnapshotGate` to the bottom of the file? Keeping the important, relevant or public ones (`Connect`, `ReadBatch` and `Close`) at the top?

josephwoodward (approved):

squiidz:
Closing: this PR's premise misread the framework contract. `auto_replay_nacks: false` is a documented opt-in to DROP rejected messages ("If set to false these messages will instead be deleted" — the field's own description), typically because failures are routed to a DLQ which acks. Failing the promotion barrier and pinning the checkpoint on nack contradicts that contract and produces permanent backpressure / re-snapshot loops for data the user explicitly chose to discard. The original #4584 behavior (a nack releases the barrier and the checkpoint advances) is correct. The nack-pinning changes made on the other CON-504 PRs during review are being unwound for the same reason.

PR #4693, opened 2026-08-12
tests: add goleak leak detection and advisory race-detector CI (CON-179)

prakhargarg105:
Part of CON-179 (test hardening). This PR adds two complementary checks: goroutine-leak detection inside the test suites of selected connector packages, and a race-detector CI job that starts out advisory and gains teeth package by package.

## 1. Goroutine-leak detection (goleak)

Adds `go.uber.org/goleak` and wires it into a package-level `TestMain` for four packages: `internal/impl/mysql`, `internal/impl/postgresql`, `internal/impl/protobuf`, and `internal/impl/sql`. After the last test in the package finishes, goleak fails the run if any goroutine started during the tests is still alive.

**How it runs:** `TestMain` is Go's standard per-package entry point, so this needs no CI wiring. It fires on every `go test` of these packages, which means every PR (via the existing `test.yml` unit suite) and every nightly integration run. A leak is a hard test failure.

**Ignore policy:** each `TestMain` carries a small set of narrow, individually commented ignores for goroutines we cannot stop from test code. Examples: prototransform's `SchemaWatcher` poll loop (its owner exposes no stop hook and the processor's `Close` is a no-op), and the license service's hourly expiry-metric loop started by `InjectTestService`. Broad ignores are deliberately avoided; each entry names the library and the reason.

**Real leak fixed along the way:** `TestIntegrationPostgresSnapshotAckBarrier` blocked a consumer goroutine on a context that benthos never cancels (consumer funcs receive `context.Background()`), which leaked the goroutine and the unacked stream behind it past the end of the test. The test now uses its own cancellable context for the simulated crash and explicitly stops the crashed stream with `StopWithin`. The protobuf mock BSR server also gets a proper `Shutdown` on cleanup.

All four packages were verified leak-clean on repeated local runs before the check was enabled, so this cannot fail a PR on a pre-existing condition.

## 2. Advisory race-detector CI job

New workflow `.github/workflows/race_test.yml`, plus a promotion list at `.github/race-blocking-packages.txt`.

**How it runs:** on every PR the job diffs against the merge-base and collects the `internal/impl/<pkg>` directories the PR touched (same auto-scoping as `integration_test.yml`, which keeps the 2-10x race-detector runtime cost bounded). If no connector package changed, the job skips entirely. Otherwise it runs `go test -count=1 -race -shuffle=on -timeout 10m` per touched package and writes a pass/fail table to the job summary. A `workflow_dispatch` trigger allows manual runs against any package by name.

**Advisory-first design:** a race failure produces a warning annotation, and the check stays green, unless the failing package is listed in `race-blocking-packages.txt`. That file starts empty, so on day one nothing can go red from this job. A package is promoted to blocking only after it holds a green `-race` baseline (repeated clean scoped runs plus a green advisory run in CI). The invariant this preserves: a blocking gate must never be able to fail a PR on a pre-existing race the PR did not introduce.

## Supporting changes

- `CONTRIBUTING.md`: contributors are asked to run `task test:unit-race` on packages they touch, with a pointer to the blocking-list semantics.
- `.claude/agents/tester.md`: recipe for wiring goleak into a new package, including the ignore policy and the requirement to verify against a clean baseline before merging.
- `go.mod`: adds `go.uber.org/goleak v1.3.0` (test-only dependency).

### DIFF
--- a/.claude/agents/tester.md
+++ b/.claude/agents/tester.md
@@ -592,6 +592,29 @@ t.Cleanup(func() {
 - Do not use `tc := tc` in loop bodies. Go 1.22+ fixed loop variable scoping.
 - Use `t.Context()` for test contexts. Exception: in `t.Cleanup()` functions, use `context.Background()` because `t.Context()` is already canceled during cleanup.
 
+# Goroutine Leak Detection (goleak)
+
+Connector packages opt in to goroutine-leak detection by wiring `go.uber.org/goleak` into a package-level `TestMain` (one `TestMain` per package — check for an existing one first). See `internal/impl/protobuf` and `internal/impl/sql` for reference.
+
+```go
+// main_test.go (with the package's usual license header)
+func TestMain(m *testing.M) {
+    goleak.VerifyTestMain(m,
+        goleak.IgnoreCurrent(),
+        // Narrow, commented ignores for known-benign goroutines only, e.g.:
+        // database/sql keeps a connection-pool opener alive per *sql.DB.
+        goleak.IgnoreTopFunction("database/sql.(*DB).connectionOpener"),
+    )
+}
+```
+
+Adding it to a new package:
+
+1. Create `main_test.go` with the pattern above and the package's license header.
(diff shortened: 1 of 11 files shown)


prakhargarg105:
Superseded by #4721 — remade from a branch directly on this repo instead of a fork.

PR #4710, opened 2026-08-18
postgres_cdc: refresh IAM auth token for each new connection

### LINKED ISSUE #4668: postgres_cdc: heartbeat connection never refreshes IAM auth token, fails every tick after ~15min
## Summary

The `postgres_cdc` input's heartbeat mechanism opens its own connection with a one-time snapshot of the IAM auth token, and never refreshes it. Once the token expires (15 min, per AWS), every subsequent heartbeat write fails with a PAM authentication error — indefinitely, on every tick, for the life of the pipeline.

## Root cause

`newHeartbeat()` in [`internal/impl/postgresql/pglogicalstream/heartbeat.go`](https://github.com/redpanda-data/connect/blob/main/internal/impl/postgresql/pglogicalstream/heartbeat.go) calls `openPgConnectionFromConfig(config)`:

```go
func openPgConnectionFromConfig(cfg *Config) (*sql.DB, error) {
	parsedCfg, err := pgxpool.ParseConfig(cfg.DBRawDSN)
	if err != nil {
		return nil, err
	}
	parsedCfg.ConnConfig.Password = cfg.DBConfig.Password
	parsedCfg.ConnConfig.TLSConfig = cfg.TLSConfig
	return stdlib.OpenDB(*parsedCfg.ConnConfig), nil
}
```

`parsedCfg.ConnConfig.Password = cfg.DBConfig.Password` copies whatever token is in `cfg.DBConfig.Password` **at the moment `newHeartbeat()` is called** (pipeline startup) into a brand-new, independent `pgconn.Config`. This is a value copy, not a live reference to the shared config. The resulting `*sql.DB` backs every future periodic heartbeat `ExecContext` call ([`heartbeat.go`](https://github.com/redpanda-data/connect/blob/main/internal/impl/postgresql/pglogicalstream/heartbeat.go)'s `run()` method), for the entire lifetime of the pipeline, using that frozen token.

Compare this to the main replication connection's `tokenBuilder` in [`internal/impl/postgresql/aws/aws.go`](https://github.com/redpanda-data/connect/blob/main/internal/impl/postgresql/aws/aws.go):

```go
// tokenBuilder will be called upon component connection to refresh token/password and reconnect.
// Tokens last ~15 minutes and will only need refreshing after a connection is lost.
tokenBuilder := func(ctx context.Context) error {
	...
	password, err := auth.BuildAuthToken(ctx, endpoint, cfg.Region, dbConf.User, cfg.Credentials)
	...
	dbConf.Password = password
	...
}
```

This is explicitly re-invoked on every reconnect of the main connection, generating a fresh token each time. The heartbeat's connection has no equivalent refresh path — it never calls `tokenBuilder`, and its `*sql.DB` has no mechanism to regenerate or re-fetch a token when its underlying physical connection needs to reopen.

## Impact

For any `postgres_cdc` pipeline using AWS IAM authentication (`aws.enabled: true`) with a `heartbeat_interval` longer than the IAM token lifetime (15 min — so effectively any non-trivial interval), every heartbeat write after the first ~15 minutes fails:

```
level=warning msg="unable to write heartbeat message: failed to connect to `user=... database=...`: ...: server error: FATAL: PAM authentication failed for user \"...\" (SQLSTATE 28000)"
```

This recurs on every tick, indefinitely, even while the main replication stream stays healthy. We reproduced this reliably: heartbeat failure at pipeline startup + 1h, again at +2h, etc., on a fresh pipeline with a 1h `heartbeat_interval`, with the main connection fully healthy throughout.

We don't yet know if this is purely cosmetic (heartbeats are best-effort keep-alives for low-traffic tables per the docs) or could eventually affect replication slot liveness / `standby_timeout` handling if it prevents the slot from being acknowledged forward — but it's a clear, 100%-reproducible bug independent of that question.

## Suggested fix direction

`newHeartbeat`'s connection needs to participate in the same token-refresh mechanism as the main connection — either by sharing/calling the same `tokenBuilder` before each `ExecContext`, or before opening a new physical connection when the pool needs one, rather than baking in a value snapshot once at construction time.

## Environment

- Redpanda Connect (benthos) `4.98.0`
- `postgres_cdc` input, `aws.enabled: true`, IAM auth via assumed cross-account role
- Amazon RDS for PostgreSQL 12.22
- `heartbeat_interval` set to a value well beyond the 15-minute token lifetime

ReguiguiMohamed:
Fixes #4668.

`openPgConnectionFromConfig` copies `cfg.DBConfig.Password` into a fresh `pgconn.Config` and hands that to `stdlib.OpenDB`. The value is a snapshot, so every physical connection the returned `*sql.DB` opens later still authenticates with the token that existed when the pool was built. With `aws.enabled: true` the IAM token lasts around 15 minutes, after which the heartbeat write fails on every tick for the rest of the pipeline's life, while the replication stream stays healthy.

`Config.RefreshAuthToken` already exists for this, and is what `Connect` calls for the main connection. This passes it to pgx's `BeforeConnect` hook, which runs before each new physical connection and hands over the config that is about to be used, so the password gets rebuilt at the point a connection actually needs one. The heartbeat, monitor, snapshotter and server-version check all open through this helper, so this covers all four.

The refresh writes to the shared `Config.DBConfig` and the hook reads it straight back, and pools open connections from separate goroutines, so a mutex guards that pair.

### Testing

`connection_test.go` exercises the hook without a database. Nothing listens on the port in the DSN, and `BeforeConnect` runs before the dial, so a refused connection still runs it. The tests check that the token is rebuilt on every connection attempt, that a failing refresh surfaces to the caller, and that a config without `RefreshAuthToken` is left alone.

Against the current code the first two fail, reporting 0 refreshes and the startup password still in place.

I left `CHANGELOG.md` alone since release notes look to be generated from PR titles at release time. Glad to add an entry if you'd prefer one.

### DIFF
--- a/internal/impl/postgresql/pglogicalstream/connection.go
+++ b/internal/impl/postgresql/pglogicalstream/connection.go
@@ -9,16 +9,23 @@
 package pglogicalstream
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 	"regexp"
 	"strconv"
+	"sync"
 
+	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgxpool"
 	"github.com/jackc/pgx/v5/stdlib"
 )
 
 var re = regexp.MustCompile(`^(\d+)`)
+
+// Every pool refreshes the token through the same Config.DBConfig, so the
+// refresh and the read that follows it have to be serialised.
+var refreshAuthTokenMu sync.Mutex
 
 func openPgConnectionFromConfig(cfg *Config) (*sql.DB, error) {
 	parsedCfg, err := pgxpool.ParseConfig(cfg.DBRawDSN)
@@ -27,7 +34,23 @@ func openPgConnectionFromConfig(cfg *Config) (*sql.DB, error) {
 	}
 	parsedCfg.ConnConfig.Password = cfg.DBConfig.Password
 	parsedCfg.ConnConfig.TLSConfig = cfg.TLSConfig
-	return stdlib.OpenDB(*parsedCfg.ConnConfig), nil
+	return stdlib.OpenDB(*parsedCfg.ConnConfig, stdlib.OptionBeforeConnect(
+		func(ctx context.Context, connCfg *pgx.ConnConfig) error {
+			if cfg.RefreshAuthToken == nil {
+				return nil
+			}
+			// IAM tokens expire long before the pipeline does, so rebuild the
+			// password for every new connection rather than reusing the one
+			// captured when the pool was opened.
+			refreshAuthTokenMu.Lock()
+			defer refreshAuthTokenMu.Unlock()
+			if err := cfg.RefreshAuthToken(ctx); err != nil {
+				return err
+			}
+			connCfg.Password = cfg.DBConfig.Password
+			return nil
+		},
+	)), nil
 }
 
 func getPostgresVersion(cfg *Config) (int, error) {
--- a/internal/impl/postgresql/pglogicalstream/connection_test.go
+++ b/internal/impl/postgresql/pglogicalstream/connection_test.go
@@ -1,0 +1,83 @@
+// Copyright 2025 Redpanda Data, Inc.
+//
+// Licensed as a Redpanda Enterprise file under the Redpanda Community
+// License (the "License"); you may not use this file except in compliance with
+// the License. You may obtain a copy of the License at
+//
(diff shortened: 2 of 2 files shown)


josephwoodward:
Thanks for looking into this, I'll take a look shortly

ReguiguiMohamed:
> Thanks for looking into this, I'll take a look shortly

hey @josephwoodward  i'm truly honoured to contribute, will be here awaiting any fixes or updates to my PR in realtime or  monitor and add anything

Issue #4722, opened 2026-08-25
aws_dynamodb_cdc: shard reader retries forever when a shard is deleted while its iterator is expired (ResourceNotFoundException)


zytek:
## Summary

In `aws_dynamodb_cdc`, a shard reader whose iterator has expired enters an **infinite retry loop** if the shard itself is deleted (aged past the stream's 24h retention) before the refresh succeeds. `GetShardIterator` then permanently returns `ResourceNotFoundException: Shard does not exist`, but the expired-iterator recovery path treats every refresh failure as transient, so the reader retries at `poll_interval` forever — logging one ERROR line per second per wedged shard until the pipeline restarts. The sibling `TrimmedDataAccessException` path already handles the same terminal condition correctly.

All line references are pinned to `04875dd` (current `main` at time of filing), `internal/impl/aws/dynamodb/input_cdc.go`.

## Observed behavior

Redpanda Cloud managed pipeline, `snapshot_mode: snapshot_and_cdc`, against a ~680M-item table. During the multi-hour snapshot, CDC shard readers spend long stretches parked in the `ShouldThrottle()` backpressure loop, so their iterators (15-minute TTL) routinely expire. Backlog shards meanwhile age past the 24h stream retention and DynamoDB deletes them. Result, repeated every second per affected shard, continuing for many hours after the snapshot completed:

```
Failed to refresh expired iterator for shard shardId-00000001787496244464-c0942b45, retrying in 1s: getting refreshed iterator for shard shardId-00000001787496244464-c0942b45: operation error DynamoDB Streams: GetShardIterator, https response error StatusCode: 400, ... ResourceNotFoundException: Requested resource not found: Shard does not exist
```

(That shard's embedded timestamp is 2026-08-23T14:44Z; the errors began once it aged out of retention — the loop can never succeed.)

## Root cause

1. `GetRecords` fails with `ExpiredIteratorException` → the handler calls `refreshExpiredIterator` (L2231/L2653 for the multi-table/single-table variants; helper at L2100).
2. `refreshExpiredIterator` resumes via `GetShardIterator` with `AfterSequenceNumber`. For a **deleted** shard this returns `ResourceNotFoundException` — a permanent error.
[...]

PR #4727, opened 2026-08-26
fix aws_dynamodb_cdc: stop retrying expired-iterator refresh when th…

squiidz:
When a shard iterator expired and the shard had since been deleted, the refresh GetShardIterator call failed with ResourceNotFoundException and was treated as a transient error, retrying at poll_interval forever. Classify the Streams API ResourceNotFoundException as permanent at both reader call sites: mark the shard exhausted, signal the coordinator, and return - matching the existing trimmed-data refresh-failure handling.

CON-546, GitHub issue #4722

### DIFF
--- a/internal/impl/aws/dynamodb/snapshot.go
+++ b/internal/impl/aws/dynamodb/snapshot.go
@@ -289,6 +289,20 @@ func isExpiredIteratorError(err error) bool {
 	return ok
 }
 
+// isStreamsResourceNotFoundError returns true when a DynamoDB Streams call
+// fails because the requested stream or shard no longer exists. Unlike a
+// throttle or a network blip this is permanent: a deleted shard can never be
+// read again, so callers should mark it exhausted and move on rather than
+// retry. Deliberately matches only the Streams API type, not the DynamoDB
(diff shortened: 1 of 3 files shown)


internal/impl/aws/dynamodb/snapshot.go
```diff
@@ -289,6 +289,20 @@ func isExpiredIteratorError(err error) bool {
 	return ok
 }
 
+// isStreamsResourceNotFoundError returns true when a DynamoDB Streams call
+// fails because the requested stream or shard no longer exists. Unlike a
+// throttle or a network blip this is permanent: a deleted shard can never be
+// read again, so callers should mark it exhausted and move on rather than
+// retry. Deliberately matches only the Streams API type, not the DynamoDB
+// table API's ResourceNotFoundException (e.g. a missing checkpoint table).
+func isStreamsResourceNotFoundError(err error) bool {
```
josephwoodward, line 298: It feels like this comment is documenting a behaviour the caller should be concerned about, not the function?

Also, is this behaviour specific to snapshotting (would, for instance,  `internal/impl/aws/dynamodb/input_cdc.go` be a better place for it)?

josephwoodward (approved):

PR #4734, opened 2026-08-27
redpanda/migrator: avoid O(N²) schema registry fan-out in translate_ids sync

prakhargarg105:
> Recreates #4731 (originally authored by @alextreichler) as a native branch so the Claude review GitHub Action runs. Commit is identical.

## Problem

With `schema_registry.translate_ids: true`, the `redpanda_migrator` output registers each schema at the destination via franz-go's `CreateSchema`, which resolves the returned ID through `SchemaUsagesByID` — a fetch of **every subject-version sharing that ID**, spawned as one unbounded goroutine per usage.

Identical schema bodies deduplicate to a single destination ID, so syncing N such subjects costs ~N(N+1)/2 destination requests instead of N, and none of those requests are bounded by `max_parallel_http_requests`. Against a registry with heavily shared schema bodies (e.g. per-environment copies of the same schemas) this produces sustained request bursts that overload single-node registries into `connection reset by peer`, which fails the sync, which fails the output connect, which restarts the pipeline and replays the whole quadratic sweep — observed in a production migration as 169 consecutive connect failures with zero topics migrated.

A side effect of the fan-out is actively misleading errors: the failing GET names an unrelated subject that merely shares the schema ID (`sync subject schema <A> ...: unable to GET ".../subjects/<B>/versions/1"`), making healthy subjects look broken.

## Fix

In the `translate_ids` branch of `syncSubjectSchema`, replace `CreateSchema` with:

- `RegisterSchema` — one POST; idempotent (returns the existing ID for an already-registered identical schema), preserving the previous create-or-reuse semantics;
- one `LookupSchema` — to resolve the destination version for `schemaInfo` and logging.

Two sequential requests per schema inside the bounded worker pool: O(N) total, concurrency capped by `max_parallel_http_requests`. franz-go's own `RegisterSchema` docs recommend exactly this trade. The `translate_ids: false` path is unchanged.

## Measurements

From the included regression test (40 subjects sharing one schema body, `max_parallel_http_requests: 2`, counting reverse proxy in front of the destination registry):

| metric | before | after |
|---|---|---|
| usage fan-out reads (`GET /subjects/S/versions/V`) | 826–838 (≈ N(N+1)/2 = 820) | 0 |
| peak concurrent in-flight requests | 42–79 | 2 (= configured limit) |
| schema registrations | 40 | 40 |

## Validation

- New `TestIntegrationSchemaRegistryMigratorSyncSharedSchemaFanout` fails on the previous code and passes with the fix.
- All `TestIntegrationSchemaRegistryMigrator*` integration tests pass except `SyncWithReferences`, which fails identically on unmodified main (pre-existing, unrelated).
- Unit tests, `go vet`, `gofumpt` clean.

## Follow-ups (not in this PR)

The same fan-out pattern exists in:
- the migrator's `translate_ids: false` branch (`CreateSchemaWithIDAndVersion` fans out internally) — fixable with `RegisterSchema(ctx, subject, sch, ss.ID, ss.Version)`, no lookup needed;
- the `internal/impl/confluent/sr` wrapper (`CreateSchema`/`CreateSchemaWithIDAndVersion` both discard everything but the ID), which the standalone `schema_registry` output and `schema_registry_encode` processor inherit;
- separately, the fatal-connect behavior (one failing subject blocks all topic/data migration via the output's Connect) is tracked as its own defect.

### DIFF
--- a/internal/impl/redpanda/migrator/migrator_schema_registry.go
+++ b/internal/impl/redpanda/migrator/migrator_schema_registry.go
@@ -814,16 +814,12 @@ func (m *schemaRegistryMigrator) syncSubjectSchema(ctx context.Context, ss sr.Su
 			return schemaInfo{}, fmt.Errorf("create schema: %w", err)
 		}
 
-		// Single lookup to resolve the destination version for this subject.
-		dss, err := m.dst.LookupSchema(ctx, dstSubject, sch)
-		if err != nil {
-			m.metrics.IncSchemaCreateErrors()
-			return schemaInfo{}, fmt.Errorf("lookup created schema with id %d: %w", id, err)
-		}
-
-		info = schemaInfoFromSubjectSchema(dss)
-		m.log.Infof("Schema migration: schema created with translated id: subject=%s version=%d id=%d => subject=%s version=%d id=%d",
-			ss.Subject, ss.Version, ss.ID, info.Subject, info.Version, info.ID)
+		// The destination version is left unset: the registration response
+		// carries only the ID, which is also the only field of schemaInfo
+		// with a functional consumer.
+		info = schemaInfo{Subject: dstSubject, ID: id}
+		m.log.Infof("Schema migration: schema created with translated id: subject=%s version=%d id=%d => subject=%s id=%d",
+			ss.Subject, ss.Version, ss.ID, info.Subject, info.ID)
 	} else {
 		dss, err := m.dst.CreateSchemaWithIDAndVersion(ctx, dstSubject, sch, ss.ID, ss.Version)
 		if err != nil {
--- a/internal/impl/redpanda/migrator/migrator_schema_registry.go
+++ b/internal/impl/redpanda/migrator/migrator_schema_registry.go
@@ -367,14 +367,6 @@ type schemaInfo struct {
 	ID      int
 }
 
-func schemaInfoFromSubjectSchema(ss sr.SubjectSchema) schemaInfo {
-	return schemaInfo{
-		Subject: ss.Subject,
-		Version: ss.Version,
-		ID:      ss.ID,
-	}
-}
-
 // schemaRegistryMigrator coordinates migration between a source and destination
 // Schema Registry.
 //
@@ -821,7 +813,11 @@ func (m *schemaRegistryMigrator) syncSubjectSchema(ctx context.Context, ss sr.Su
 		m.log.Infof("Schema migration: schema created with translated id: subject=%s version=%d id=%d => subject=%s id=%d",
 			ss.Subject, ss.Version, ss.ID, info.Subject, info.ID)
 	} else {
-		dss, err := m.dst.CreateSchemaWithIDAndVersion(ctx, dstSubject, sch, ss.ID, ss.Version)
+		// RegisterSchema instead of CreateSchemaWithIDAndVersion for the same
+		// reason as above: the latter resolves the registered ID through the
+		// unbounded SchemaUsagesByID fan-out, and the ID and version are
+		// already known here.
+		id, err := m.dst.RegisterSchema(ctx, dstSubject, sch, ss.ID, ss.Version)
 		if err != nil {
 			const conflictPattern = `Schema already registered with id \d+ instead of input id \d+`
 			if ok, _ := regexp.MatchString(conflictPattern, err.Error()); ok {
@@ -844,11 +840,10 @@ func (m *schemaRegistryMigrator) syncSubjectSchema(ctx context.Context, ss sr.Su
 			m.log.Warnf("Schema migration: schema subject=%s version=%d id=%d could not be created (server error: %s) - using existing schema with the same ID, if this is not the desired behavior, try enabling translate-ids",
 				ss.Subject, ss.Version, ss.ID, err.Error())
 
-			dss = ss
-			dss.Subject = dstSubject
+			id = ss.ID
 		}
 
-		info = schemaInfoFromSubjectSchema(dss)
+		info = schemaInfo{Subject: dstSubject, Version: ss.Version, ID: id}
 		m.log.Infof("Schema migration: schema created with fixed id: subject=%s version=%d id=%d",
 			info.Subject, info.Version, info.ID)
 	}
--- a/internal/impl/redpanda/migrator/migrator_schema_registry.go
+++ b/internal/impl/redpanda/migrator/migrator_schema_registry.go
@@ -797,12 +797,28 @@ func (m *schemaRegistryMigrator) syncSubjectSchema(ctx context.Context, ss sr.Su
 	var info schemaInfo
 	t0 := time.Now()
 	if m.conf.TranslateIDs {
-		// If the schema already exists (and is identical), this returns
-		// the existing schema
-		dss, err := m.dst.CreateSchema(ctx, dstSubject, sch)
+		// Register with a registry-assigned ID. If the schema is already
+		// registered (and identical), this returns the existing ID without
+		// creating a new version.
+		//
+		// RegisterSchema is used instead of CreateSchema because CreateSchema
+		// additionally resolves the returned ID via SchemaUsagesByID, which
+		// fetches every subject-version sharing that ID using one unbounded
+		// goroutine per usage. Identical schema bodies deduplicate to a single
+		// ID, so syncing N such subjects costs O(N^2) destination requests,
+		// none of them bounded by MaxParallelHTTPRequests.
+		const autoAssign = -1
+		id, err := m.dst.RegisterSchema(ctx, dstSubject, sch, autoAssign, autoAssign)
 		if err != nil {
 			m.metrics.IncSchemaCreateErrors()
 			return schemaInfo{}, fmt.Errorf("create schema: %w", err)
+		}
+
+		// Single lookup to resolve the destination version for this subject.
+		dss, err := m.dst.LookupSchema(ctx, dstSubject, sch)
+		if err != nil {
+			m.metrics.IncSchemaCreateErrors()
+			return schemaInfo{}, fmt.Errorf("lookup created schema with id %d: %w", id, err)
 		}
 
 		info = schemaInfoFromSubjectSchema(dss)
--- a/internal/impl/redpanda/migrator/migrator_schema_registry_fanout_integration_test.go
+++ b/internal/impl/redpanda/migrator/migrator_schema_registry_fanout_integration_test.go
@@ -1,4 +1,4 @@
-// Copyright 2025 Redpanda Data, Inc.
+// Copyright 2026 Redpanda Data, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -24,7 +24,6 @@ import (
 	"regexp"
 	"sync/atomic"
 	"testing"
-	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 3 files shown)


internal/impl/redpanda/migrator/migrator_schema_registry.go
```diff
@@ -797,14 +797,30 @@ func (m *schemaRegistryMigrator) syncSubjectSchema(ctx context.Context, ss sr.Su
 	var info schemaInfo
 	t0 := time.Now()
 	if m.conf.TranslateIDs {
```
squiidz, line 791: The `else` branch below (`translate_ids: false`, the **default**) still calls `CreateSchemaWithIDAndVersion`, which performs the exact same `SchemaUsagesByID` fan-out this PR removes — so the O(N²) behaviour remains for the common configuration.

The fix is even simpler there than here: `id` and `version` are inputs (`ss.ID`, `ss.Version`), so `RegisterSchema` alone suffices and `info` can be built directly as `{dstSubject, ss.Version, ss.ID}` with no lookup at all. The new fan-out test only covers `translate_ids: true`, so this path is also untested.

internal/impl/redpanda/migrator/migrator_schema_registry.go
```diff
@@ -797,14 +797,30 @@ func (m *schemaRegistryMigrator) syncSubjectSchema(ctx context.Context, ss sr.Su
...
+		// none of them bounded by MaxParallelHTTPRequests.
+		const autoAssign = -1
+		id, err := m.dst.RegisterSchema(ctx, dstSubject, sch, autoAssign, autoAssign)
 		if err != nil {
 			m.metrics.IncSchemaCreateErrors()
 			return schemaInfo{}, fmt.Errorf("create schema: %w", err)
 		}
 
+		// Single lookup to resolve the destination version for this subject.
+		dss, err := m.dst.LookupSchema(ctx, dstSubject, sch)
```
squiidz: This follow-up `LookupSchema` looks unnecessary: everything downstream (`checkSchemaIDConflict`, `DestinationSchemaID`) consumes only `info.ID`, which `RegisterSchema` already returned — `info.Subject`/`info.Version` only feed the two `Infof` lines. So it doubles destination requests per subject-version (10k subject-versions ⇒ 10k extra sequential POSTs, ~2× sync latency) to fetch values nothing needs.

It's also a new failure surface the old ID-keyed resolution didn't have: it resolves by posted schema body, and with `normalize: true` both calls carry `?normalize=true`. A registry honouring normalize on register but not on the subject-lookup endpoint (the kind of server divergence this file already works around elsewhere) turns this into a 40403 that fails the whole subject sync where the old code succeeded.

Suggest dropping the lookup and building `info = schemaInfo{Subject: dstSubject, ID: id}` (log the version at Debug or omit it). If the lookup stays, one more thing: the old `CreateSchema` path guaranteed the returned ID matched the register-assigned one; here `id` is only used in the error string. A one-line `if dss.ID != id` warning would restore that integrity check for free.

internal/impl/redpanda/migrator/migrator_schema_registry_fanout_integration_test.go
```diff
@@ -0,0 +1,164 @@
+// Copyright 2025 Redpanda Data, Inc.
```
squiidz: New file — header should be `Copyright 2026` (this one was copied from the older sibling migrator files).

internal/impl/redpanda/migrator/migrator_schema_registry_fanout_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/redpanda-data/connect/v4/internal/impl/redpanda/migrator"
+)
+
+// countingProxy is a reverse proxy in front of the destination Schema
+// Registry that records request counts and peak concurrent in-flight
+// requests.
+type countingProxy struct {
+	server *httptest.Server
+
+	total       atomic.Int64 // all requests
```
squiidz, line 42: `total` is incremented on every request but never read by any assertion or log — either drop it, or better, use it for the missing traffic-happened assertion (see the comment on the assertions below).

internal/impl/redpanda/migrator/migrator_schema_registry_fanout_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+	t.Log("And: a counting reverse proxy in front of the destination Schema Registry")
+	proxy := newCountingProxy(t, dstCluster.SchemaRegistryURL)
+	dstSR, err := sr.NewClient(sr.URLs(proxy.server.URL))
+	require.NoError(t, err)
+
+	t.Logf("And: %d source subjects sharing one identical schema body", numSubjects)
+	const sharedSchema = `{"type":"record","name":"Shared","fields":[{"name":"a","type":"int"}]}`
+	ctx := t.Context()
+	for i := range numSubjects {
+		_, err := srcSR.CreateSchema(ctx, fmt.Sprintf("shared-%03d-value", i), sr.Schema{Schema: sharedSchema})
```
squiidz: This setup loop seeds the subjects via `srcSR.CreateSchema`, which itself performs the O(N²) `SchemaUsagesByID` fan-out this test guards against — just pointed at the *source* registry (~820 extra requests, each call spawning up to 40 unbounded parallel version GETs; potential flake source on slow CI). The returned `SubjectSchema` is discarded, so `srcSR.RegisterSchema` (one POST per subject) is a drop-in.

internal/impl/redpanda/migrator/migrator_schema_registry_fanout_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+	t.Log("When: the schema migrator syncs with translate_ids enabled")
+	conf := migrator.SchemaRegistryMigratorConfig{
+		Enabled:      true,
+		Versions:     migrator.VersionsAll,
+		TranslateIDs: true,
+	}
+	// NB: the testing constructor sets MaxParallelHTTPRequests to 2.
+	m := migrator.NewSchemaRegistryMigratorForTesting(t, conf, srcSR, dstSR)
+
+	syncCtx, cancel := context.WithTimeout(ctx, 3*time.Minute)
```
squiidz: Nit: every other `Sync` call in this package bounds with `context.WithTimeout(t.Context(), redpandaTestWaitTimeout)` — reusing that avoids a second timeout knob (`go test -timeout` already covers the outer bound).

internal/impl/redpanda/migrator/migrator_schema_registry_fanout_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+	t.Logf("Destination registry traffic for %d identical-body subjects:", numSubjects)
+	t.Logf("  schema registrations (POST): %d", registers)
+	t.Logf("  usage listings (GET /schemas/ids/N/versions): %d", idGets)
+	t.Logf("  usage fan-out (GET /subjects/S/versions/V): %d (O(N) expectation: <=%d, O(N^2) worst case: %d)",
+		versionGets, numSubjects, numSubjects*(numSubjects+1)/2)
+	t.Logf("  peak concurrent in-flight requests: %d (max_parallel_http_requests: 2)", maxInFlight)
+
+	// Syncing N subjects must cost O(N) destination requests. 2*N leaves room
+	// for one extra lookup per subject.
+	assert.LessOrEqual(t, versionGets, int64(2*numSubjects),
```
squiidz: Two problems with the guard as written:

1. **It can pass vacuously.** The only assertions are `versionGets <= 2*N` and `maxInFlight <= 2`; nothing asserts any destination traffic happened. A future change that makes `Sync` sync nothing while returning nil (empty roots, filter change, `enabled()` gating) passes with every counter at 0. Add e.g. `assert.GreaterOrEqual(t, registers, int64(numSubjects))`.

2. **The bound rests on a wrong premise.** The comment says the `2*N` ceiling "leaves room for one extra lookup per subject", but `LookupSchema` is `POST /subjects/{s}` — it never matches `reVersionGet`. The honest post-fix expectation is `versionGets == 0`; the 80-request slack would let a partial ~N-scale fan-out regression pass silently.

Also worth noting: since this is Docker-gated via `integration.CheckSkip`, a revert to `CreateSchema` sails through normal CI. The counting proxy would work just as well against a stub `httptest` SR as an always-run unit test.

squiidz (commented):
The `RegisterSchema` swap itself is sound — registration semantics are identical to the old path and this removes the `SchemaUsagesByID` fan-out for the `translate_ids: true` arm. A few things though, the first one I'd consider blocking since the PR title doesn't hold for the default config (inline comments below).

Two points that fall outside the diff:

**The root cause lives in our own wrapper, and two other components still hit it.** `sr.Client.CreateSchema` and `CreateSchemaWithIDAndVersion` (`internal/impl/confluent/sr/client.go:167,181`) pay the full `SchemaUsagesByID` fan-out only to return `ss.ID`. The `schema_registry` output (`internal/impl/kafka/output_schema_registry.go:478,503` — which has its own `translate_ids` mode, hit per source subject-version) and the SR encode processor (`internal/impl/confluent/processor_schema_registry_encode.go:653`, auto-registration on cache miss) call them with the exact same O(N²) behaviour. Switching the two wrapper methods to `RegisterSchema` is a two-line change that fixes every caller at once — worth doing here or as an immediate follow-up.

**`max_parallel_http_requests` is only enforced as a Sync worker count**, not at the HTTP layer, so any other franz-go internal fan-out (e.g. `SchemaReferences`, the batch endpoints) — or a future library upgrade — silently re-breaks the invariant exactly as `CreateSchema` did. A semaphore `RoundTripper` wrapping the client transport with `MaxParallelHTTPRequests` tokens would make the configured limit unconditionally true, and demote this PR's swap to a request-count optimization rather than the sole enforcement.

internal/impl/redpanda/migrator/migrator_schema_registry.go
```diff
@@ -797,19 +789,35 @@ func (m *schemaRegistryMigrator) syncSubjectSchema(ctx context.Context, ss sr.Su
...
+		// creating a new version.
+		//
+		// RegisterSchema is used instead of CreateSchema because CreateSchema
+		// additionally resolves the returned ID via SchemaUsagesByID, which
+		// fetches every subject-version sharing that ID using one unbounded
+		// goroutine per usage. Identical schema bodies deduplicate to a single
+		// ID, so syncing N such subjects costs O(N^2) destination requests,
+		// none of them bounded by MaxParallelHTTPRequests.
+		const autoAssign = -1
+		id, err := m.dst.RegisterSchema(ctx, dstSubject, sch, autoAssign, autoAssign)
```
Leward, line 803: yup, [the doc](https://pkg.go.dev/github.com/twmb/franz-go/pkg/sr#Client.RegisterSchema) on `RegisterSchema` indicates:

> If you want to register a schema and receive more details back, you can use [CreateSchema] or [CreateSchemaWithIDAndVersion] at the expense of more HTTP requests to the registry. 

As I understand the downside here is that we are only return with `id int` instead of [`dss SubjsctSchema`](https://pkg.go.dev/github.com/twmb/franz-go/pkg/sr#SubjectSchema).

Leward (approved):

internal/impl/redpanda/migrator/migrator_schema_registry.go
prakhargarg105: Addressed in 575e738: the `translate_ids: false` path now calls `RegisterSchema(ctx, dstSubject, sch, ss.ID, ss.Version)` with no lookup, building `info` directly from the known ID and version. The fanout test runs both modes; locally each subtest shows 40 registrations, 0 usage-endpoint requests, and peak concurrency 2.

internal/impl/redpanda/migrator/migrator_schema_registry.go
prakhargarg105: Right, and nothing needs more than the ID here: `checkSchemaIDConflict` and `DestinationSchemaID` read only `info.ID`, and `info.Subject`/`info.Version` only fed log lines. In the fixed-ID path the version is an input, so `info` still carries it. The translate path drops it from the log.

prakhargarg105:
Following up on @squiidz's review points outside the diff:

- **The sr wrapper has the same fan-out** (`internal/impl/confluent/sr/client.go`, affecting the `schema_registry` output and `schema_registry_encode` processor): tracked in [CON-581](https://redpandadata.atlassian.net/browse/CON-581). It changes two other components that need their own review and tests, so it is kept out of this PR. That ticket also notes that the only guard test is Docker-gated, with no always-run coverage.
- **`max_parallel_http_requests` is enforced as a worker count, not at the HTTP layer**: tracked in [CON-582](https://redpandadata.atlassian.net/browse/CON-582). With this PR applied, no current migrator path goes through a fanning-out franz-go method, so the gap is latent rather than active.

On the red `integration-test-filter` check: `TestIntegrationSchemaRegistryMigratorSyncWithReferences` and `TestIntegrationMigratorMultiPartitionSchemaAwareWithConsumerGroups` fail identically on this branch and its merge base with `main` (`expected: "{\"type\":\"int\"}" actual: "\"int\""`), so they predate this change. `TestIntegrationTopicMigratorSyncConfig` timed out in CI but passes locally on both.

[CON-581]: https://redpandadata.atlassian.net/browse/CON-581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CON-582]: https://redpandadata.atlassian.net/browse/CON-582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

squiidz (approved):

PR #4756, opened 2026-09-02
CI: Support using Claude to review OSS contributions

josephwoodward:
Currently Claude is only invoked for PRs created by users with write access to avoid abuse. It would be good to extend this, whilst retaining the same guard rails to reviewing contributions by those outside of Redpanda, greatly reducing the time it takes to review OSS contributions.

This change adds support for invoking Claude reviews for users outside of Redpanda by using a `Claude Review` label that once assigned will automatically be removed after the review. This way it can be reapplied once users have addressed any feedback whilst also avoiding abuse as it still verifies write access as part of the check.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -2,42 +2,72 @@ name: Claude Code Review
 
 on:
   pull_request:
-    types: [opened, synchronize, ready_for_review, reopened]
+    types: [opened, synchronize, ready_for_review, reopened, labeled]
 
 concurrency:
   group: claude-review-${{ github.event.pull_request.number }}
   cancel-in-progress: true
 
 jobs:
-  # Gate runs as its OWN job so that, for non-write authors, the review job below is
-  # skipped at the JOB level. A job-level skip reports the `claude-review` check as
-  # "skipped" (neutral/grey), not "success" (green) — a skipped review must never read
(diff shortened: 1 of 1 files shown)


squiidz (approved):

.github/workflows/claude-code-review.yml
```diff
@@ -2,42 +2,72 @@ name: Claude Code Review
...
         env:
           GH_TOKEN: ${{ github.token }}
+          # No checkout in this job, so gh needs GH_REPO to resolve the repo.
+          GH_REPO: ${{ github.repository }}
         run: |
+          if [[ "${{ github.event.action }}" == "labeled" ]]; then
+            SENDER="${{ github.event.sender.login }}"
+            PERMISSION=$(gh api repos/${{ github.repository }}/collaborators/$SENDER/permission --jq '.permission')
+            if [[ "$PERMISSION" != "admin" && "$PERMISSION" != "write" ]]; then
+              echo "::error::@$SENDER does not have write access ($PERMISSION); only maintainers can invoke a review via the 'Claude Review' label."
```
Leward, line 42: Can anyone pick labels when creating PRs?

Hence this check?

Leward (approved):

.github/workflows/claude-code-review.yml
josephwoodward: Users with the `triage` permission can select labels, so this is to align the permissions to existing checks.

PR #4783, opened 2026-09-08
redis_script: fix the dead LUA scripts link in the summary

JakeSCahill:
## Problem

`redisScriptProcConfig`'s summary links to `https://redis.io/docs/manual/programmability/eval-intro/`, which returns HTTP 404. Redis moved the page under `docs/latest/develop`.

The summary is published verbatim as the first line of the `redis_script` reference page, so the dead link ships in the docs. It was picked up by the weekly Redpanda docs link check.

## Fix

Point at the current page: `https://redis.io/docs/latest/develop/programmability/eval-intro/` (HTTP 200).

No behavior change: the string is documentation metadata only.

### DIFF
--- a/docs/modules/components/pages/processors/redis_script.adoc
+++ b/docs/modules/components/pages/processors/redis_script.adoc
@@ -23,7 +23,7 @@
 component_type_dropdown::[]
 
 
-Performs actions against Redis using https://redis.io/docs/manual/programmability/eval-intro/[LUA scripts^].
+Performs actions against Redis using https://redis.io/docs/latest/develop/programmability/eval-intro/[LUA scripts^].
 
 Introduced in version 4.11.0.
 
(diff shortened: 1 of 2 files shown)


JakeSCahill:
`integration-test-filter` fails on `TestIntegrationRedisFailoverCache`, twice (original run and a re-run of just the failed job). It is not from this change: the diff is one URL inside a `Summary()` string, and the failing test builds a Redis master plus sentinel from `bitnami/redis:latest` and `bitnami/redis-sentinel:latest` in containers.

The same job failed on #4638, the last PR to touch `internal/impl/redis`, and that was merged. The redis package is only selected by the filter when a PR touches it, so there is no green baseline on `main` to compare against, and both `:latest` image tags have moved since April, when the sentinel setup was last worked on (#4197).

Worth a separate look at that test. Every other check here is green.

josephwoodward:
@JakeSCahill Fixed the test here https://github.com/redpanda-data/connect/pull/4793

josephwoodward (approved):

PR #4784, opened 2026-09-09
iceberg: cut two per-record map allocations from the shredder

Jeffail:
One slice of #4712, split out per review — this is just the shredder allocation cut (Leward's "A"), with no measurement tooling attached.

## The change

Shredding a record built two maps per struct — an index of input keys by their match-key, and a set of keys that matched a schema field. Both exist for case-insensitive matching, where several input keys can fold onto one schema field and that ambiguity needs reporting rather than silently resolving.

When matching is case-sensitive though, the match-key is the identity function, so neither map earns its keep: a field's value is just `value[field.Name]`, and case-collisions can't happen because Go map keys are themselves case-sensitive. That's the default (`case_sensitive_columns: true`), so from what I can tell it's the path almost all traffic takes.

So this splits the two apart — `shredStructExact` for case-sensitive matching with direct lookups, and the existing body moves to `shredStructFolded` unchanged. Unknown-field detection now counts how many input keys a schema field claimed and skips the scan entirely when they're all accounted for (the steady state once a schema settles), building a lookup set only on the rare path where something genuinely is unknown. Duplicate field names within a struct disable that shortcut for the schema, since the shortcut assumes names are unique and a genuinely unknown key could otherwise hide behind a duplicate that matched.

`BenchmarkShredWide` at `GOMAXPROCS=1`, benchstat, n=8:

```
                sec/op                      B/op                        allocs/op
before          4.369µ                      4.312Ki                     71
after           1.472µ  -66.3% (p=0.000)    1.609Ki  -62.7% (p=0.000)    41  -42.3% (p=0.000)
```

Worth being clear that's the shredder in isolation. Earlier profiling put shredding at roughly 27% of the sink's CPU, so my read is the end-to-end saving should be appreciable but a good deal smaller than 66% — I haven't measured that yet, so please treat the sink-level number as unquantified rather than implied.

Since it's purely an optimisation it shouldn't change observable behaviour at all, so `TestShredStructPathsAgree` runs the same schema and record through *both* implementations and requires identical emitted values (in order), identical new-field notifications (compared as sets, since map iteration order is random) and identical errors — full matches, unknown keys (top-level, nested, and behind a duplicate name), missing optionals, explicit nulls, both required-field error paths, and the empty-record / empty-schema edges.

## Not in this PR

The measurement/profiling tooling that motivated this, the commit-regime harness, the write-path throughput harness, and the parquet compression field — those are following as their own PRs per Leward's split on #4712.

### DIFF
--- a/internal/impl/iceberg/shredder/shredder.go
+++ b/internal/impl/iceberg/shredder/shredder.go
@@ -87,6 +87,14 @@ type RecordShredder struct {
 	// iceberg's recommended convention and with engines like Spark and Trino
 	// in their default configurations.
 	caseSensitive bool
+	// duplicateFieldNames records whether any struct in the schema repeats a
+	// field name. When it does, shredStructExact cannot use its matched-count
+	// shortcut: the count is per field, so two fields sharing a name inflate it
+	// and an unaccounted-for input key can be mistaken for a full match, which
+	// would silently drop an unknown-field notification and with it schema
+	// evolution of that column. Pinned by
+	// TestShredStructPathsAgreeWithDuplicateFieldNames.
+	duplicateFieldNames bool
 	// fieldCommons optionally maps an iceberg field ID to the upstream
 	// schema.Common describing the same field. When present, the leaf
 	// value-conversion step uses Logical params (timestamp unit,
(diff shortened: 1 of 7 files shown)


internal/impl/iceberg/shredder/shredder.go
```diff
@@ -227,6 +291,89 @@ func (rs *RecordShredder) shredStruct(
...
+			// Field is null or missing - emit null for all leaf descendants.
+			if err := rs.shredNull(field.Type, field.ID, repLevel, defLevel, sink); err != nil {
+				return err
+			}
+			continue
+		}
+
+		if err := rs.shredValue(field.Type, field.ID, fieldValue, fieldPath, repLevel, fieldDefLevel, maxRepLevel, sink); err != nil {
+			return fmt.Errorf("field %q: %w", field.Name, err)
+		}
```
Leward, line 344: suggeestion: extract repeated logic into a helper function

Looks like this block of code is repeated in both `shredStruct` and `shredStructExact`. Maybe it could be extracted into its own `shredField` helper function?

Leward (approved):

internal/impl/iceberg/shredder/shredder.go
Jeffail: Good instinct, and I tried it — pulled the shared block into a `shredField(field, fieldValue, exists, path, repLevel, defLevel, maxRepLevel, sink)` helper called from both. Benchmarked before/after with benchstat (n=10, GOMAXPROCS=1):

```
ShredWide/declared_schema=false   1.328µ -> 1.382µ   +4.07% (p=0.000)
ShredWide/declared_schema=true    1.495µ -> 1.561µ   +4.42% (p=0.000)
```

Same allocs/op and B/op either way, so it's pure call/dispatch overhead, not a GC story — but it's the same class of regression the file already documents for the sibling case (the comment above `shredStruct`'s case-insensitive body: "~5% slower on BenchmarkShredCaseInsensitive for the extra call"). Given the whole point of this PR is shaving CPU off this exact path, I don't think a ~4% giveback to deduplicate ~20 lines is a good trade, so I've left the duplication in place rather than apply the suggestion. Happy to discuss further if you feel differently, but wanted to bring data rather than just push back on vibes.

PR #4785, opened 2026-09-09
iceberg: add an optional parquet compression codec

Jeffail:
Another slice of #4712, split out per review — this is Leward's "B": just the `parquet.compression` field, no throughput measurement attached (that's following separately as its own PR, since it needs this one to read the codec back out of written files).

## The problem

Data files were always written uncompressed — the only writer option this output ever passed was a string encoding, so `parquet-go` fell back to its uncompressed default, and nothing consulted the table's own `write.parquet.compression-codec` property either.

## The change

Adds a `parquet.compression` field accepting `uncompressed`, `snappy`, `gzip` and `zstd`. It's deliberately optional rather than defaulted, so "unset" is distinguishable from an explicit "uncompressed" and can mean "defer to the table". Resolved per table, once, when its writer is built:

1. `parquet.compression`, when set
2. otherwise the table's `write.parquet.compression-codec` property
3. otherwise uncompressed, preserving existing behaviour

Only codecs every targeted engine reads are offered — the original `lz4` codec is ambiguously specified and readers disagree on what it means, `brotli` and `lzo` have patchy support. A table property naming one of those is reported in the log and treated as unset rather than failing the write, since that property isn't this output's configuration to validate.

Worth calling out: setting the table property is usually the better lever than the field, because copy-on-write rewrites whole data files from inside the Iceberg library, out of reach of a writer option we pass — but that code reads the same property (and defaults it to `zstd`). So an otherwise unconfigured table can end up holding a mix of codecs already, which is legal and transparent (parquet records its codec per column chunk) but worth knowing about. The field stays as the escape hatch for catalogs that reject client-set table properties, like Databricks Unity Catalog.

Compression codec names are folded before lookup on both sides — the table property obviously needs it, and less obviously, so does the configured value (the config framework's enum linter lower-cases before validating, so `compression: ZSTD` passes and arrives verbatim).

Tests cover the resolution order, both spellings of "no compression", the declined codecs and both `LZ4_RAW` spellings, casing variants, whitespace-only property values, and that the warning names the table and is de-duplicated per table (a router is inherently multi-table). A second test reads the codec back out of the written parquet footer rather than trusting the resolver's return value.

## Not in this PR

The write-path throughput harness that measures these codecs (depends on this one), the shredder change, the commit-regime harness, the profiling configs, and the Databricks e2e bench — those are their own PRs off #4712.


docs/modules/components/pages/outputs/iceberg.adoc
```diff
@@ -299,6 +300,26 @@ To guarantee an existing table never ends up with a mix of the two annotations,
...
 
+== Data file compression
```
Leward, line 303: thought: too verbose and repeated by `parquet.compression` field

docs/modules/components/pages/outputs/iceberg.adoc
```diff
@@ -299,6 +300,26 @@ To guarantee an existing table never ends up with a mix of the two annotations,
...
+
+*Codec support.* `snappy`, `gzip` and `zstd` are read by every engine this output targets. Parquet permits others that are not offered here: the original `lz4` codec was ambiguously specified and readers disagree on what it means, `lz4_raw` is its unambiguous replacement but has younger and less universal reader support, and `brotli` and `lzo` are patchily supported. If the table property names one of those it is reported in the log and the files this output writes are uncompressed instead. That governs only those files — the Iceberg library does map those codecs, so if the intent is that nothing writes one, change the table property rather than relying on this.
```
Leward: This could be part of the field documentation itself. 

These notes will show up in the the main section of the doc, but they are specific to how to set the `parquet.compression` config.

internal/impl/iceberg/config.go
```diff
@@ -99,8 +99,35 @@ const (
...
+	"2. otherwise the table's own `write.parquet.compression-codec` property, if the table has one.\n" +
+	"3. otherwise uncompressed.\n" +
+	"\n" +
+	"`parquet.compression` is optional rather than defaulted precisely so that step 2 is reachable: an unset field means \"whatever the table says\", which is not the same as explicitly choosing `uncompressed`.\n" +
+	"\n" +
+	"*Prefer the table property where you can.* Several kinds of file are written by the Iceberg library rather than by this output's own writer, and `parquet.compression` cannot reach those: the whole-file rewrites performed by `merge_strategy: copy-on-write`, and the equality-delete files written by `merge-on-read`. The table property does reach them, because the library reads the same property (and defaults it to `zstd` when absent). So the property is the lever that gets every file in the table on the same codec, whereas the field governs only the data files this output writes itself — appends and merge-on-read data files. Reach for the field when the property is not available to you, notably on catalogs that reject client-set table properties, such as the Databricks Unity Catalog.\n" +
```
Leward: Do you think this would be clearer/easier to read? 

```suggestion
	"*Prefer the table property for compression settings whenever possible.* `parquet.compression` only applies to files this connector writes directly. The Iceberg library can perform additional writes, such as copy-on-write rewrites and equality-delete files, and these do not honour this field. They follow the table's `write.parquet.compression-codec` property instead, which defaults to `zstd`. Setting the table property is the only way to get one codec for every file in the table. Use this field when you cannot set table properties, for example on Databricks Unity Catalog.\n" +
```

internal/impl/iceberg/config.go
```diff
@@ -99,8 +99,35 @@ const (
...
+	"\n" +
+	"*Cost.* Less than you might expect. Measuring the write path against local object storage found every codec within a few percent of uncompressed, in both directions, at one core as well as four — so there is no throughput penalty worth planning around. What varies enormously is the size benefit, and that depends on the data: on a repetitive record shape `zstd` was ~15x smaller than uncompressed, while on genuinely random content it saved ~2%, because parquet's dictionary and byte-array encodings have already compacted what they can before any codec runs. See the benchmark results for the numbers and their caveats.\n"
```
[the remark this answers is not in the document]
Leward: All good as this is handled in a follow-up PR: https://github.com/redpanda-data/connect/pull/4786

internal/impl/iceberg/config.go
```diff
@@ -99,8 +99,35 @@ const (
...
+	"\n" +
+	"*Prefer the table property where you can.* Several kinds of file are written by the Iceberg library rather than by this output's own writer, and `parquet.compression` cannot reach those: the whole-file rewrites performed by `merge_strategy: copy-on-write`, and the equality-delete files written by `merge-on-read`. The table property does reach them, because the library reads the same property (and defaults it to `zstd` when absent). So the property is the lever that gets every file in the table on the same codec, whereas the field governs only the data files this output writes itself — appends and merge-on-read data files. Reach for the field when the property is not available to you, notably on catalogs that reject client-set table properties, such as the Databricks Unity Catalog.\n" +
+	"\n" +
+	"NOTE: Use a lower-case codec name in the property. This output accepts any casing, but the Iceberg library's own lookup is lower-case only and silently falls back to uncompressed for anything else — so a property of `ZSTD` would give you compressed appends and uncompressed rewrites.\n" +
```
Leward: I would move this to the field description as it is quite specific to it

internal/impl/iceberg/config.go
Leward: I am not sure if we need to expand this much in the outward facing doc about this. IMO we could simply say that setting compression has a negligible impact on throughput while it can drive significant storage saving depending on the same of the data.

internal/impl/iceberg/config.go
Jeffail: Reworded per your suggestion (kept the substance, tightened the wording). `c34233296`.

internal/impl/iceberg/config.go
Jeffail: Moved into the field description, next to the table-property paragraph it's actually about. `c34233296`.

internal/impl/iceberg/config.go
Jeffail: Cut to exactly that — negligible throughput, size savings depend on the data, no numbers inline. `c34233296`.

docs/modules/components/pages/outputs/iceberg.adoc
Jeffail: Agreed — trimmed the long-form section down (reworded the table-property paragraph, moved the casing note into the field description, cut the Cost paragraph to two sentences with no numbers). Should read a lot less like a rehash of the field description now. `c34233296`.

Leward (approved):

PR #4811, opened 2026-09-11
postgres_cdc: initial support for incremental snapshotting via signalling

josephwoodward:
This change adds support for incremental snapshotting in the PostgreSQL connector, including a more general purpose package that other connectors can use to support such a feature.

### How it Works

I'm going to document how the implementation works (which in this case is more specific to PostgreSQL due to some of its powerful primitives), but from a high level it works by recording the database's transaction visibility state (a snapshot window open state), reads a chunk of (snapshot) rows bounded by primary key, then records transaction visibility state again (snapshot window closed). Those two readings bracket the read: any transaction that could have modified those rows has an id between them. The chunk is held in a buffer while replication continues, and each streamed change evicts its own row from that buffer, meaning the live, current value always wins. Once a commit arrives with an id past both readings, every racing transaction has been seen, and whatever survives in the buffer is released.

Below is a diagram that hopefully makes things a little clearer, however in this case we don't need to use a signal table in Postgres for anything beyond launching the snapshot:

image

### What Does This Give Us?

A few things:

- Without needing to perform a blocking snapshot, users can continue to stream changes whilst also performing a snapshot at the same time.
- We can easily snapshot/re-snapshot existing or new tables on an ad hoc basis.

This does come with a few trade-offs (see the `Incremental snapshot (whilst streaming)` header below) in the semantics, but still delivers the guarantee that the complete dataset is replicated after snapshot finished.

---

#### Snapshotting with no concurrent streaming

Below triggers snapshotting of two tables (`cart` followed by `products`), once `cart` has finished and `products` begins, the connector is restarted and continues to snapshot:

image

Metrics of the run (you can see where we stopped the connector in the metrics below):

image

---

#### Incremental snapshot (whilst streaming):

This run signals a snapshot of 10M rows, whilst simultaneously streaming updates to the rows being read, resulting in collisions. When this happens the record in the transaction log always wins, so in this case the read is discarded. Hence the `read` vs `update` difference in the terminal output below.

This changes the semantics of the traditional blocking snapshot in that consumers are required to use upset behaviour. What is guaranteed though is the complete dataset after snapshot is completed.

image

You can also see an attempt to queue another snapshot gets rejected as it's already been snapshotted.

image


internal/impl/postgresql/input_pg_stream.go
```diff
@@ -52,6 +57,13 @@ const (
 	// FieldAWSIAMAuthEnabled enabled field.
 	FieldAWSIAMAuthEnabled = "enabled"
 	shutdownTimeout        = 5 * time.Second
+
+	fieldIncSnapshot                   = "incremental_snapshot"
+	fieldIncSnapshotEnabled            = "enabled"
+	fieldIncrementalSnapshotChunkSize  = "chunk_size"
```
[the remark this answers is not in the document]
josephwoodward: Perhaps `batch_size` would be better as it aligns with the blocking snapshot terminology.

squiidz:
There's 2 things that might be worth looking at before merging:

1. Unchanged TOAST columns. Under REPLICA IDENTITY DEFAULT, an update to a row while its chunk is buffered arrives with the unchanged_toast_value placeholder for untouched TOAST columns. deduplicateStreamedRow evicts the buffered snapshot row, which was the only real copy, so the sink keeps the placeholder forever. Fix: take the buffered row rather than delete it, and fill the placeholders from it before emitting the update.

2. Stale snapshot rows can be emitted after newer streamed changes in three cases:
   - timestamptz primary keys: the chunk decoder gives time.Local, the pgoutput text decoder a fixed zone, so the %v window keys differ whenever host and server offsets differ. canonicalizePKValue needs a time.Time -> UTC case.
   - PK-changing updates: only the new key is evicted, so the old-key row is later released as a "read". Also evict the key from BeforeData.
   - Partitioned parents: the publication is created without publish_via_partition_root, so changes stream under leaf names and never match the parent in the window. Reject relkind = 'p' in checkBackfillable.

internal/impl/postgresql/pglogicalstream/incremental_snapshot.go
```diff
@@ -0,0 +1,819 @@
...
+	table := incrementalsnapshot.TableID{Schema: message.Schema, Table: message.Table}
+	if !s.incSnapshot.coordinator.Snapshotting(table) {
+		return nil
+	}
+
+	pk, err := s.incrementalStreamedRowPK(ctx, table, message.Data)
+	if err != nil {
+		return fmt.Errorf("resolving primary key for incremental snapshot deduplication on table %s: %w", table, err)
+	}
+	if s.incSnapshot.coordinator.OnStreamedRow(table, pk) {
```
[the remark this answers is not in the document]
josephwoodward: a2f82246ad79

internal/replication/incrementalsnapshot/coordinator.go
```diff
@@ -0,0 +1,519 @@
...
+// while a replication stream runs, dropping each buffered row the stream has
+// already delivered.
+//
+// It holds only the algorithm: which table and key range to read next, when
+// a chunk is safe to emit, and what the caller may checkpoint. Side effects
+// come from Deps, window comparisons from Watermark.
+//
+// Not safe for concurrent use: call OnStreamedRow and OnCommit from one
+// goroutine, in stream order. OnCommit's chunk read blocks on I/O by design.
+type Coordinator[P any, W Watermark[P]] struct {
```
Leward, line 37: (optional) Could be a good idea to document P and W types in the struct godoc, just for each of reading.

internal/replication/incrementalsnapshot/coordinator.go
```diff
@@ -0,0 +1,519 @@
...
+	cfg CoordinatorConfig[P, W]
+
+	// resume holds the state given to NewCoordinator; nil once Start has
+	// consumed it.
+	resume *State
+
+	remaining []TableID
```
Leward, line 44: That is a lot going on for the internal state of a single component. Maybe some logical grouping could help reducing the cognitive complexity that comes with it.

Leward: An example:

```go
// progress is the resumable position in the backfill: what State reports.
type progress struct {
	remaining  []TableID
	current    *TableID
	maxPK      PrimaryKey
	lastSentPK PrimaryKey
}


func (p progress) clone() progress { p.remaining = slices.Clone(p.remaining); return p }

type Coordinator[P any, W Watermark[P]] struct {
	cfg CoordinatorConfig[P, W]
	resume *State

	live      progress // advances as chunks are fetched
	committed progress // advances only after a flush; State reads this
	...
}
```

Leward: Most of the state here belongs to the one table being read: current, maxPK, lastSentPK, the window fields, etc. Only the queue, the flags, and the committed mirror are cross-table.

Suggestion: move the per-table fields into a `tableSnapshot` type with its own lifecycle (resolve keys, fetch, close window, flush, done). The coordinator then only manages the queue, checkpoints, and picks the next table.

I think this would make both the code and the tests easier to follow, but it can be a follow-up. Do you agree with the split?

internal/impl/postgresql/input_pg_stream.go
```diff
@@ -254,10 +267,75 @@ a JSON object with a ` + "`message`" + ` key, whose value is written to the conn
 
 ` + "```sql" + `
 INSERT INTO <schema>.<signal_table_name> (type, data) VALUES ('log', '{"message": "Signal message"}');
-` + "```").
+` + "```" + `
+
+**` + "`snapshot`" + `** — backfills the named tables incrementally, alongside streaming. Requires
```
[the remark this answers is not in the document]
josephwoodward: f0269c5430ce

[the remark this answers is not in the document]
josephwoodward: f0269c5430ce

internal/replication/incrementalsnapshot/coordinator.go
josephwoodward: Good idea 👍  They were originally documented on `NewCoordinator` but moved them to here.

josephwoodward: Done in 9dcbda6172fc

internal/impl/postgresql/inc_snapshot_integration_test.go
```diff
@@ -0,0 +1,1342 @@
...
+	stream, err := builder.Build()
+	require.NoError(t, err)
+	license.InjectTestService(stream.Resources())
+
+	stopped := make(chan struct{})
+	go func() {
+		defer close(stopped)
+		// Cancellation is how the harness ends the stream, not a failure.
+		if err := stream.Run(t.Context()); err != nil && !errors.Is(err, context.Canceled) {
+			t.Logf("stream error: %v", err)
```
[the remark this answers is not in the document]
josephwoodward: 7c373b38d375

internal/replication/incrementalsnapshot/coordinator.go
josephwoodward: Sounds like a good improvement. As this is in the package that will be shared across all the connectors it'd be great to invest some time in cleaning it up and maturing it.

internal/replication/incrementalsnapshot/coordinator.go
```diff
@@ -0,0 +1,534 @@
...
+// W is the watermark type, which does interpret P: it decides when a chunk's
+// window opens and closes. Refer to Watermark for the contract, including
+// that implementations must be comparable.
+//
+// Both are pinned once per connector rather than threaded through call sites.
+// Postgres does this with aliases in
+// internal/impl/postgresql/incrementalsnapshot, binding P to uint32 and W to
+// its own Watermark, so the rest of that connector names an unparameterised
+// Coordinator.
+//
```
Leward: Not sure if that block in the godoc comment is that useful.

internal/replication/incrementalsnapshot/state.go
```diff
@@ -0,0 +1,128 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+)
+
+// CurrentStateVersion contains the version of the state.
+const CurrentStateVersion = 3
```
Leward, line 19: Versions 1 and 2 only existed on this branch, so the 1..3 range is compatibility with checkpoints nobody has.

Shall we reset it back to 1?

internal/replication/incrementalsnapshot/state.go
```diff
@@ -0,0 +1,128 @@
...
+	var raw stateJSON
+	if err := dec.Decode(&raw); err != nil {
+		return err
+	}
+
+	*s = State(raw)
+	narrowPrimaryKey(s.LastSentPK)
+	narrowPrimaryKey(s.MaxPK)
+
+	if s.Version < minSupportedStateVersion || s.Version > CurrentStateVersion {
```
Leward, line 73: Could this be a problem in the future when someone waits too long between two upgrades of Connect?

internal/impl/postgresql/pglogicalstream/incremental_snapshot.go
```diff
@@ -0,0 +1,868 @@
...
+			// necessarily the one these rows came from.
+			s.monitor.TrackSnapshotTable(ctx, tableFQN(rows[0].Table))
+			s.monitor.UpdateSnapshotProgressForTable(tableFQN(rows[0].Table), len(rows))
+		} else {
+			s.logger.Debugf("Incremental snapshot: checkpoint advanced with no rows to flush (fully deduplicated)")
+		}
+		checkpoint := s.incSnapshot.coordinator.State()
+		s.reportTableTransition(ctx, checkpoint.CurrentTable)
+
+		state, err := json.Marshal(checkpoint)
```
Leward, line 634: Poked around with Claude Code

-----

A float8 (or float4) primary key holding Infinity or NaN breaks the backfill permanently. Postgres sorts Infinity above every finite value, so ResolveMaxKey returns it as MaxPK. The chunk fetches fine, but json.Marshal rejects non-finite floats when the checkpoint is built:

    advancing incremental snapshot: serializing incremental snapshot state: json: unsupported value: +Inf

The stream restarts, replays the signal, and fails again. I ran it against PG16 for 30s: 60 restarts, zero rows delivered, slot never advanced, other tables blocked.

Repro:

    CREATE TABLE float_keyed (id float8 PRIMARY KEY, name text);
    INSERT INTO float_keyed VALUES (1.5, 'a'), (2.5, 'b'), ('Infinity', 'c');
    -- then signal a snapshot of float_keyed

Two ways out: encode non-finite floats as the strings "Infinity" / "-Infinity" / "NaN" in the checkpoint (Postgres accepts those literals when bound, same as numeric today), or reject float key columns in checkKeyTypesBindable like bytea. Either is fine, but the failure should drop the table rather than loop the stream.

Leward: **How likely is it in real life?**

Rare. Two things have to be true at once, and each is uncommon on its own.

 - **A float primary key**. Floats are a poor key because equality is imprecise, and almost every schema tool and ORM steers people to integers, UUIDs, or text. You see them mostly in scientific data, legacy imports, and tables where someone keyed on a measurement.
 - **A non-finite value in that key.** `Infinity` in a float column is usually an application bug or a sentinel. Somebody would have to insert it deliberately or through unchecked arithmetic.

Leward: A unit test that can be added to `state_test.go` that don't require any live DB:

```go
func TestStateRoundTripNonFiniteFloatKey(t *testing.T) {
	for _, v := range []float64{math.Inf(1), math.Inf(-1), math.NaN()} {
		s := &State{Version: CurrentStateVersion, MaxPK: PrimaryKey{v}}
		_, err := json.Marshal(s)
		require.NoError(t, err, "a float8 key of %v must be checkpointable", v)
	}
}
```

internal/replication/incrementalsnapshot/deps.go
```diff
@@ -0,0 +1,97 @@
...
+
+	// ResolveWatermark reads a fresh watermark.
+	ResolveWatermark(ctx context.Context) (W, error)
+
+	// ForceFreshTransaction assigns the connection a real transaction id, so
+	// the stream has a known position for the first watermark.
+	//
+	// Start calls it once. Calling it per watermark would make every chunk's
+	// two watermarks differ, which disables the drain entirely.
+	ForceFreshTransaction(ctx context.Context) error
```
Leward: This method is only needed because Postgres assigns transaction ids lazily. Other engines expose their watermark as a global counter (Oracle SCN, MySQL GTID set, SQL Server LSN), so for them this would be a no-op.

This package is meant to be engine agnostic, so a Postgres concept in the interface, and a comment that explains Postgres behaviour, feel out of place here.

Since Start calls it once, before the first watermark, could the Postgres glue run txid_current() itself before coordinator.Start? Then the method can go, and the xid reasoning lives next to the Postgres code that needs it.

Leward: On second thought, removing the method has a downside: Start would depend on the caller running `txid_current()` first, and nothing in the code would enforce that.

Alternative: keep the hook but give it a generic name and doc, e.g. Prepare(ctx), "called once by Start before the first watermark read, return nil if nothing is needed". The xid explanation then moves to the Postgres implementation.

internal/replication/incrementalsnapshot/coordinator.go
```diff
@@ -0,0 +1,534 @@
...
+	}
+
+	// State never captures an unflushed chunk, so resuming always means
+	// planning the next one. Watermarks are always re-derived.
+	return c.planNextChunk(ctx)
+}
+
+// commitLiveState snapshots the live fields into the committed ones. Call
+// only once everything fetched so far has been flushed.
+func (c *Coordinator[P, W]) commitLiveState() {
```
Leward: "Commit" already means a DB commit (OnCommit) and a persisted checkpoint (commitCheckpoint in the input). Here it is an in-memory copy, which is easy to misread as a save.

What about emitted* for the fields and markEmitted for this function? It says what happened (the rows went to emit) without suggesting they were persisted or acked. I'd avoid "flushed" since the input already has flushBatch.

internal/replication/incrementalsnapshot/coordinator.go
```diff
@@ -0,0 +1,534 @@
...
+//
+// It never emits: it buffers the first chunk and returns, leaving OnCommit to
+// release everything. Callers may therefore start the coordinator before
+// their downstream is consuming. With nothing queued it goes idle and waits
+// for AddTables.
+func (c *Coordinator[P, W]) Start(ctx context.Context) error {
+	resume := c.resume
+	c.resume = nil
+
+	if resume != nil {
```
Leward: Wouldn't it be easier to add this initialization logic directly in `NewCoordinator`?

Today, `NewCoordinator` stores the resume state and Start applies it. `Start` also re-adds the tables that `AddTables` queued before `Start`, so they end up behind the resumed queue.

`AddTables` already does this on its own: it appends to the end of the queue and skips tables in `knownTables`. So if `NewCoordinator` applied the resume state, a later `AddTables` call would give the same queue, before or after Start.

That would remove the resume field and the seeded block, and Start would only begin reading.

Side note: the Postgres input only calls `AddTables` from the signal handler, after Start. So today the seeded path only runs in tests.

Leward (commented):
Given the complexity in `incrementalsnapsho` package. What about adding a package level doc from a [`doc.go` file](https://go.dev/blog/godoc#:~:text=That%20package%20uses%20another%20convention%20for%20packages%20that%20need%20large%20amounts%20of%20introductory%20documentation%3A%20the%20package%20comment%20is%20placed%20in%20its%20own%20file%2C%20doc.go%2C%20which%20contains%20only%20those%20comments%20and%20a%20package%20clause.)? 

```go
//
// Licensed as a Redpanda Enterprise file under the Redpanda Community
// License (the "License"); you may not use this file except in compliance with
// the License. You may obtain a copy of the License at
//
// https://github.com/redpanda-data/connect/blob/main/licenses/rcl.md

// Package incrementalsnapshot backfills database tables in key-ordered,
// key-bounded chunks while a replication stream runs, dropping each buffered
// row the stream has already delivered.
//
// Coordinator holds the algorithm. Everything database-specific is injected:
// side effects through Deps, window comparisons through Watermark. The
// package therefore needs no database driver.
//
// A component for one database supplies a Deps and a Watermark, and usually
// aliases Coordinator to its own position and watermark types to keep the
// type arguments out of its call sites.
//
// # Why a chunk waits before it is emitted
//
// The coordinator reads a chunk while the stream runs. A row in the chunk
// can be older than a change that the stream has not delivered yet.
//
// Example. Transaction T1 sets row 7 to "b" (it was "a") and commits while
// the chunk read runs:
//
//	low                              high
//	 |---------- chunk read ----------|
//	          T1 commits (row 7 = "b")
//
// The read returns row 7, but with which value? That depends on whether the
// read got to row 7 before or after T1 committed. The coordinator cannot
// know. Then T1 streams, with row 7 = "b". This value is certainly current,
// so the coordinator drops row 7 from the chunk.
//
// Any commit between low and high can do this. So the coordinator holds the
// chunk until the stream passes high, and then emits it. Commits after high
// stream after the chunk, so downstream they overwrite it in the correct
// order.
//
// # Watermarks
//
// low and high are watermarks: positions in the stream that the coordinator
// gets just before and just after the chunk read. It checks each streamed
// commit against them to find when the stream passes high.
//
// A watermark is not an exact point. Transactions that are still running
// when it is taken have no position in the stream yet. So each watermark
// covers an uncertain range. Watermark.OpensAt and Watermark.ClosesAt give
// the start and the end of that range.
//
// If no transaction was running at low or at high, and nothing started
// between them, the chunk cannot be stale. The coordinator then emits it
// immediately, without a wait for the next commit. Refer to
// Watermark.Quiesced.
package incrementalsnapshot
``` 

Edit: I realize there is already a package level godoc in `types.go`.

internal/replication/incrementalsnapshot/coordinator.go
Leward: Idea for later: With a per-table type, we could read one chunk from each of N tables inside the same watermark pair.

internal/replication/incrementalsnapshot/deps.go
josephwoodward: Oh yeah, `Prepare` is a good idea.

Leward:
Follow-up from offline discussion: we may want to restrict incremental snapshot to simple PK layout first

internal/replication/incrementalsnapshot/coordinator.go
josephwoodward: This has been done.

internal/replication/incrementalsnapshot/coordinator.go
josephwoodward: I've renamed it to `captureLiveState` as `commit` is already uses elsewhere and could lead to confusion.

internal/replication/incrementalsnapshot/state.go
josephwoodward: Reset back to 1, good catch 👍

PR #4820, opened 2026-09-14
ci: download Go modules before integration tests

Leward:
The integration runner lets `go test` fetch modules lazily. When the Go proxy drops a connection, the first package that needs the module fails with `[setup failed]` and the run is red with no visible cause.

Examples:
- https://github.com/redpanda-data/connect/actions/runs/34874961896 
- https://github.com/redpanda-data/connect/actions/runs/34375846077 
- https://github.com/redpanda-data/connect/actions/runs/34375888625

Run `go mod download` in a dedicated step first. Modules missing from the restored cache are fetched before any test starts, and a proxy error fails in a named step instead of inside a random test package.

Ref: CON-126


Leward:
Because this change the pipeline file, this runs the full suite which times-out in a PR context (30 minutes max).

josephwoodward (approved):

PR #4822, opened 2026-09-15
snowflake: upgrade gosnowflake driver to v2.2.0 (CVE-2026-85525)

squiidz:
Jira: [CON-562](https://redpandadata.atlassian.net/browse/CON-562)

## Why

`github.com/snowflakedb/gosnowflake v1.19.0` is affected by [CVE-2026-85525](https://nvd.nist.gov/vuln/detail/CVE-2026-85525) ([GHSA-h6vc-8q8q-7wvw](https://github.com/advisories/GHSA-h6vc-8q8q-7wvw), CWE-295, CVSS 3.1 7.4 high). The driver accepted an OCSP response without binding it to the certificate being validated and treated definitive verification failures as transient, so a revoked certificate could be accepted. The CVE record lists every Go driver v1 release up to 1.19.1 as affected, with the fix only in v2.2.0, so this requires the module path move to `gosnowflake/v2`.

Reachability is real: `snowflake_put` and the `sql_*` components (via the `snowflake` driver) open TLS connections to Snowflake, so the OCSP path runs in normal operation.

## What changed

- `go.mod`: `gosnowflake v1.19.0` replaced by `gosnowflake/v2 v2.2.0`. The only other movement is the `aws-sdk-go-v2/feature/s3/manager` indirect dropping out (v2 moved to `transfermanager`, already in the tree).
- Import path updated in `output_snowflake_put.go`, `integration_test.go` and `public/components/sql/snowflake.go`.
- `snowflake_put` PUT call uses `WithFilePutStream` (renamed from `WithFileStream`). The `WithFileTransferOptions{RaisePutGetError: true}` wrapper is removed: v2 always raises PUT/GET errors, so behaviour is unchanged.
- DSN documentation link now points at the v2 package page; `task docs` regenerated the eight affected `sql_*` pages.
- New unit test `internal/impl/sql/snowflake_dsn_test.go` pins the key-pair DSN format documented on those pages (`...?authenticator=snowflake_jwt&privateKey=<base64url PKCS8>`) against `gosnowflake.ParseDSN` in v2.
- `licenses/third_party.md` and CHANGELOG updated.

## v2 breaking-change review

Checked against the v1.19.0 and v2.2.0 sources:

- Every `Config` field we set (`Account`, `User`, `Password`, `Authenticator`, `Role`, `Database`, `Warehouse`, `Schema`, `PrivateKey`, `Params`) still exists. `Config` is now an alias for `internal/config.Config`.
- DSN `privateKey` parsing is byte-for-byte identical (base64 URL encoding, PKCS8), so the documented DSN format keeps working. No translation layer needed.
- `file_compression_type.go` extension table is identical, so the `snowflake_put` compression mapping needs no change.
- The `client_session_keep_alive` session parameter is still recognised.
- `InsecureMode`, `SnowflakeTransport`, `KeepSessionAlive`, Arrow batches and the logrus logger were never used by this repo.
- The `TMPDIR` troubleshooting paragraph stays: v2.2.0 removes the import-time `gosnowflake-cgo` directory (gosnowflake#1807) but PUT still needs a writable temp dir.

## Verification

- `go build ./...` for all distributions, `task lint` (0 issues), `task fmt`, `task docs`.
- Unit tests for `internal/impl/snowflake/...` and `internal/impl/sql/...` pass.
- The macOS `ld: warning: ignoring duplicate libraries: '-ldl'` predates this change (reproduces on main).

## Not covered here

- `public/bundle/free/go.mod` and `public/bundle/enterprise/go.mod` still list v1 as `// indirect`. They pin the published `connect/v4 v4.108.0` and only move when the release chore runs `update_bundles.sh`, so they cannot pick up v2 in this PR.
- The Snowflake integration tests need real credentials (RSA key plus `SNOWFLAKE_*` env). They cover key-pair auth through `sql_select`; password auth and a real `snowflake_put` stage upload need a manual run before merge.

[CON-562]: https://redpandadata.atlassian.net/browse/CON-562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ Changelog
 
 All notable changes to this project will be documented in this file.
 
+## Unreleased
+
+### Changed
+
+- snowflake_put, sql_*: Upgraded the Snowflake `gosnowflake` driver from v1.19.0 to v2.2.0. This fixes CVE-2026-85525 (OCSP responses were accepted without checking that they matched the certificate being validated, so a revoked certificate could be accepted) and stops the driver creating a temporary directory at start-up when the `snowflake` driver is never used. ([@squiidz](https://github.com/squiidz), [#4822](https://github.com/redpanda-data/connect/pull/4822))
```
Leward, line 10: The changelog is generated at release time, this will get ignored.

Leward (approved):

PR #4830, opened 2026-09-17
gcp_spanner_cdc: derive metadata index names deterministically

bfbarkhouse-redpanda:
## Summary
- `TableNamesFromExistingTable` generated a fresh `uuid.New()` for the watermark and created-at index names on every call. Since `Subscriber.Setup()` re-issues `CREATE TABLE/INDEX IF NOT EXISTS` DDL on **every startup** (not just first run), the always-different index name means `IF NOT EXISTS` never matches a prior run's index.
- Practical impact: every connector restart silently adds two new indexes to the metadata table instead of reusing the existing ones. On a connector that restarts often (crash loops, redeploys, autoscaling), this eventually hits Spanner's 128-index-per-table limit and `Setup()` starts failing with `FailedPrecondition: too many indices (limit 128 per table)`.
- Fix: derive the index name suffix from a SHA-256 hash of `databaseID + table` instead of a random UUID, so repeated calls for the same metadata table always produce the same index names and the DDL is actually idempotent. `RandomTableNames` (used only for testing, per its own log message) is unchanged in behavior.

## Test plan
- [x] `go build ./internal/impl/gcp/enterprise/...` passes
- [x] `go test ./internal/impl/gcp/enterprise/changestreams/metadata/...` passes, including updated `TestTableNamesFromExistingTable` which now asserts index names are stable across repeated calls for the same table, and distinct across different tables in the same database
- [ ] Would appreciate a maintainer sanity-check against the integration test suite (`metadata_integration_test.go`), which needs a Spanner emulator/instance I didn't have available


josephwoodward (approved):

internal/impl/gcp/enterprise/changestreams/metadata/name.go
```diff
@@ -45,7 +63,7 @@ type TableNames struct {
 // The watermark index will be in the form of "WatermarkIdx_<databaseId>_<uuid>".
 // The createdAt / start timestamp index will be in the form of "CreatedAtIdx_<databaseId>_<uuid>".
 func RandomTableNames(databaseID string) TableNames {
-	id := uuid.New()
+	id := uuid.New().String()
 	return TableNames{
 		TableName:          genName(tableNameFormat, databaseID, id),
```
josephwoodward, line 68: What will happen to customers using the older index, especially those that are close to hitting this? Will they need to manually clean up the index so they can migrate to this newer deterministic version?

bfbarkhouse-redpanda: On the first restart after upgrading, one new index will be created. Every index accumulated from prior stay orphaned unless someone drops them manually. If they're close but not at the ceiling, this fix stops further growth, but they still have all the old cruft sitting there. If they're already at (or one-away-from) 128, they will still hit FailedPrecondition on the very next restart after upgrading. Either can document the manual cleanup steps or ship a one-time migration helper. Thoughts?

PR #4839, opened 2026-09-18
fix(sftp): close SFTP client on every write failure to prevent SSH channel leak

### LINKED ISSUE #3584: sftp output does not retry after a failed transfer operation. (Connection failed)
Version 4.61.0:

 It seems that when there is an issue with the sftp, say a permissions issue on overwriting a file, the sftp output will never reconnect to the sftp server. First there is a permission denied issue overwriting a file on the remote:
```
time="2025-08-12T17:45:56Z" level=error msg="Failed to send message to sftp: failed to open remote file: permission denied" @service=redpanda-connect label="" path=root.output.broker.outputs.0 stream=sftp
```
Then after the failure I start seeing these messages in the log:
```
time="2025-08-12T17:45:57Z" level=error msg="Failed to send message to sftp: failed to create SFTP client: ssh: rejected: connect failed (open failed)" @service=redpanda-connect label="" path=root.output.broker.outputs.0 stream=sftp
```
And on and on forever until this morning:
```
time="2025-08-13T16:21:14Z" level=error msg="Failed to send message to sftp: failed to create SFTP client: ssh: rejected: connect failed (open failed)" @service=redpanda-connect label="" path=root.output.broker.outputs.0 stream=sftp
```

If I restart my redpanda-connect pod, transfers resume normally. I believe this is a bug with the reuse of the ssh connection in the sftp output.

brokenjacobs:
## Summary
- The `sftp` output leaked an SSH channel every time a write failed after `sftp.NewClient` succeeded (e.g. `MkdirAll`/`OpenFile` failing with a permission error). The old client/channel was only closed if a file handle had also been opened, so a run of failed writes (e.g. a permission-denied path) accumulated leaked channels until the SSH server's per-connection channel limit was hit, permanently breaking the output (`ssh: rejected: connect failed (open failed)` forever, requiring a restart).
- Every failure path in `sftpWriter.Write` (client creation, `MkdirAll`, `OpenFile`, the append-mode `Lstat`/`Seek`, and the final write) now closes and clears the SFTP client/handle before returning.
- Also fixed a related bug where a failed `Lstat`/`Seek` in append mode left `s.handle`/`s.handlePath` set on the writer, which would have caused the *next* write to that path to skip the seek and write at the wrong offset.

## Test plan
- [x] `go test ./internal/impl/sftp/...` (unit tests)
- [x] Added `TestIntegrationSFTPOutputRecoversFromRepeatedWriteFailures`, which reproduces the issue against a real SFTP server (20 failed writes to a path blocked by an existing file) and asserts no stale client/handle remains and the connection is still usable afterward. Verified it fails on the pre-fix code and passes with the fix.
- [x] `go vet ./internal/impl/sftp/...`

Fixes #3584


internal/impl/sftp/output.go
```diff
@@ -180,21 +180,22 @@ func (s *sftpWriter) Write(_ context.Context, msg *service.Message) (wErr error)
...
-		if path == s.handlePath {
-			return s.writeTo(s.handle, msg)
-		}
+	if s.handle != nil && path == s.handlePath {
+		return s.writeTo(s.handle, msg)
+	}
 
-		// If the path changes, we reset the handle and open the new file.
+	// The path changed, so close the previous handle and SFTP client before
+	// opening the new file.
```
Leward, line 188:
```suggestion
	// No usable handle for this path. If a handle for a different path is
	//  still open, close it and its SFTP client before opening the new file.
```

internal/impl/sftp/output.go
```diff
@@ -210,42 +211,62 @@ func (s *sftpWriter) Write(_ context.Context, msg *service.Message) (wErr error)
 	}
 
 	if err := s.sftpClient.MkdirAll(filepath.Dir(path)); err != nil {
+		if err := s.sftpClient.Close(); err != nil {
```
Leward, line 214: (optional) I realize we have the closing logic repeated at multiple place. 

We could replace it with the following helper. It also has the added benefit that the main `Write` function will be streamlined since the error logging will be handled in the helpers. 

```go
func (s *sftpWriter) closeClient() {
      if s.sftpClient == nil {
              return
      }
      if err := s.sftpClient.Close(); err != nil {
              s.log.With("error", err).Error("Failed to close SFTP client")
      }
      s.sftpClient = nil
}

func (s *sftpWriter) closeHandle(h io.WriteCloser) {
      if err := h.Close(); err != nil {
              s.log.With("error", err).Error("Failed to close written file")
      }
      s.handle = nil
      s.handlePath = ""
}
```

internal/impl/sftp/output.go
```diff
@@ -210,42 +211,62 @@ func (s *sftpWriter) Write(_ context.Context, msg *service.Message) (wErr error)
...
 	}
-	s.handle = handle
-	s.handlePath = path
 
 	if s.appendMode {
 		// Need to seek to the end when appending to an existing file.
 		// Details here: https://github.com/pkg/sftp/issues/295
 		fi, err := s.sftpClient.Lstat(path)
 		if err != nil {
+			s.closeHandleAndClient(handle)
```
Leward, line 235: If going with my previous suggestion of adding `closeClient()` and `closeHandle()`, I would replace that line and `closeHandleAndClient` with:

```go
s.closeHandle()
s.closeClient()
```

internal/impl/sftp/integration_test.go
```diff
@@ -206,6 +206,60 @@ cache_resources:
...
+
+	writer, err := newWriterFromParsed(parsed, service.MockResources())
+	require.NoError(t, err)
+	require.NoError(t, writer.Connect(t.Context()))
+	t.Cleanup(func() { require.NoError(t, writer.Close(context.Background())) })
+
+	// More attempts than a typical SSH server's default per-connection
+	// channel limit (OpenSSH defaults to 10), so a channel leak on failure
+	// would eventually surface as a connection-level error here.
+	for range 20 {
```
Leward, line 249: The comment above does not hold for this emulator.

`sftpgo` has no per-connection channel cap, so on main all twenty writes leak a channel and the final write to `/upload/ok.txt` still succeeds. The only thing catching the leak today is the two `assert.Nil` checks.

`sftpgo` tracks each SFTP channel as a connection and lists them at `GET /api/v2/connections` with the admin token. I checked against this image: opening five `sftp.Client` on one SSH connection gives five entries, and the count drops as each one is closed. Asserting the count is at most one after the loop would test the leak against the server instead of against private fields.

To make that work:

- `runEmulator` needs to return the HTTP base address and the admin token. It currently drops both after creating the user.
- Wrap the count assertion in `require.Eventually`. In my check the count dropped right after `Close`, but do not rely on it being synchronous.
- With that assertion in place, the twenty iterations make sense and the comment can stay. Without it, two or three iterations are enough and the OpenSSH rationale should go.

Leward (commented):
Thanks for the fix. The diagnosis is right and the cleanup on each failure path is correct. Approving the design as is.

One thing to note for a follow-up PR, by you or by us.

The fix keeps one `sftp.Client` per file and per attempt. That has two costs:

- Each path change and each retry opens a new SSH channel and redoes the SFTP handshake. With a timestamped path or frequent failures, that is one extra round trip per file.
- The leak is prevented by discipline, not by design. Every early return between `sftp.NewClient` and `s.handle = handle` must close the client. Forget it once and #3584 is back.

One `sftp.Client` can open many files. If we create it once in `Connect` next to `ssh.Dial`, keep it as long as the SSH connection, and close it in `Close`, a failed write never opens a channel. There is nothing left to leak. A path change then only closes and reopens the file handle.

Two things from this PR carry over:

1. The handle still needs closing when `Lstat`, `Seek`, or the write fails after `OpenFile` succeeded.
2. On `ErrSSHFxConnectionLost`, drop the SFTP client and the cached handle too, so `Connect` recreates both. Today a static path reuses a handle from the dead session after reconnect. That is a separate, pre-existing bug.

```
Before this PR (client per file attempt, never closed on failure)

  ssh.Client ──┬── sftp.Client #1 ── OpenFile ✗ permission denied   (channel stays open)
               ├── sftp.Client #2 ── OpenFile ✗ permission denied   (channel stays open)
               ├── sftp.Client #3 ── OpenFile ✗ permission denied   (channel stays open)
               ├── ...
               └── sftp.NewClient ✗ "connect failed (open failed)"  (channel limit reached)
```

```
Follow-up (client per connection)

  ssh.Client ──── sftp.Client ──┬── OpenFile ✗ permission denied
                                ├── OpenFile ✗ permission denied
                                ├── ...
                                └── OpenFile ✓ ── handle ── Write, Write, Write
                                                    │
                                     path changes ──┘ Close handle, OpenFile new path
```

brokenjacobs:
I'll give these a shot this weekend, thank you so much for taking a look at this. I'm not super experienced with Golang, but I know enough to apply these suggestions without running through Claude again. I'll give it a go!

brokenjacobs:
I ended up having claude implement the requested fixes. Does it look good to you?

internal/impl/sftp/integration_test.go
brokenjacobs: I think this is fixed now

Leward (approved):

Leward:
Fix looks good to me, thank you for the contribution

Leward:
Ran the integrations tests locally

```
--- PASS: TestIntegrationSFTP (1.44s)                                                                                                                                                                                                                   
  --- PASS: TestIntegrationSFTPDeleteOnFinish (1.59s)                                                                                                                                                                                                     
  --- PASS: TestIntegrationSFTPOutputRecoversFromRepeatedWriteFailures (0.49s)                                                                                                                                                                            
  --- PASS: TestIntegrationSFTPReadBatchRotatesFiles (1.11s)                                                                                                                                                                                              
  ok    github.com/redpanda-data/connect/v4/internal/impl/sftp  5.312s       
  ```

File: internal/impl/sftp/output.go [hunk 5]
```diff
@@ -180,21 +180,15 @@ func (s *sftpWriter) Write(_ context.Context, msg *service.Message) (wErr error)
...
-
-		// If the path changes, we reset the handle and open the new file.
-		if err := s.handle.Close(); err != nil {
-			s.log.With("error", err).Error("Failed to close written file")
-		}
-		if err := s.sftpClient.Close(); err != nil {
-			s.log.With("error", err).Error("Failed to close SFTP client")
-		}
+	if s.handle != nil && path == s.handlePath {
+		return s.writeTo(s.handle, msg)
```
josephwoodward: Shouldn't we also be calling the `closeHandle` and `closeClient` functions here as well?

Leward: Good catch. This path is the same as on `main`, so the PR did not introduce it. But you are right, it is the one failure path left without cleanup.

brokenjacobs: Good catch, pushed this as well. It definitely goes towards @Leward 's suggestion of refactoring this logic.

josephwoodward (approved):

Leward:
Integration test run:

```task test:integration -- sftp
task: [test:integration] go run ./cmd/tools/integration run sftp
Integration Tests (1 packages)
Output: .integration/20260922114420

[1/1] sftp
  RUN  TestIntegrationSFTPDeleteOnFinish
  PASS TestIntegrationSFTPDeleteOnFinish (1.58s)
  RUN  TestIntegrationSFTPReadBatchRotatesFiles
  PASS TestIntegrationSFTPReadBatchRotatesFiles (1.36s)
  RUN  TestIntegrationSFTPOutputRecoversFromRepeatedWriteFailures
  PASS TestIntegrationSFTPOutputRecoversFromRepeatedWriteFailures (0.59s)
  RUN  TestIntegrationSFTP
  PASS TestIntegrationSFTP (1.02s)
✓ sftp (5.4s)

========================================
Results: 1 passed, 0 failed, 0 skipped
========================================
``` 

Looks good, let's merge it 🎉

Issue #4648, opened 2026-07-31
redpanda (kafka (eventhubs)) input stalling for exactly 15 minutes


JoranDox:
Hi, we're seeing occasional input stalls in our redpanda connect pods consuming eventhubs data. Crucially: logstash (which is consuming the same data in parallel while we're trying to replace it with redpanda connect) is not seeing this, so it can't be purely an eventhubs problem. 

Relevant parts of our setup:
- right before the data enters eventhubs, we timestamp it
- at the moment we start processing it in logstash/connect we timestamp it again and compute the diff with the first timestamp to monitor processinglag
- we also tag the data with the eventhubs partition
- both output to the same elasticsearch in different indices, and we monitor processinglag, and connect is seeing peaks of 900s on specific partitions

enabling debug logging (see collapsible below for detailed log) shows about this:
- happily processing, printing something with every batch sent
- something happens like a partition rebalance, takes at most some seconds to finish
- 15 minutes of nothing
- happy processing again

what we've tried:
- we used to use the kafka input at version 1.0.0 (probably an old default we copied since forever), so we switched this to the redpanda input instead, same thing happens
- changed settings from defaults (some of these variables play/interfere with each other so have changed multiple times) but no improvement observed
  - session_timeout: default 1m → 10s
  - conn_idle_timeout: raised to 20s
  - request_timeout_overhead: raised to 40s
  - rebalance_timeout: pinned explicitly at its default 45s (not actually changed, just made explicit)
  - tcp.tcp_user_timeout: added (default 0/disabled) →  8s
  - tcp.keep_alive.idle: raised default 15s → 30s
  - unordered_processing on & off again
  - max_yield_batch_bytes: 10000000
  - fetch_max_partition_bytes: 10000000
  - partition_buffer_bytes: 12000000

again, crucially the logstash is not experiencing these stalls so it's almost certain it's connect-specific, but it's hard to debug especially since the debuglogs are showing nothing at all in that 15 minute window

Any suggestions in helping debugging or potentially relevant 15-minute timeouts (or 5-minute timeouts with 2 retries etc etc) are very welcome!

<details>
<summary>debug logs</summary>

```log
time="2026-07-22T01:50:51Z" level=debug msg="Successfully dispatched [3] documents in 45ms (66.666667 docs/sec)" @service=benthos label="" path=root.output.broker.outputs.0
[...]

PR #4844, opened 2026-09-22
elasticsearch: bound bulk requests with a timeout

dkheyman:
## Why

A bulk request that never returns wedges `elasticsearch_v8` / `elasticsearch_v9` permanently. The output has no deadline of its own, so the write sits in `bulk.Do` until the process is restarted — the pipeline stops draining its input and the only symptom is silence.

Two independent causes.

**There is no timeout field.** The config spec exposes `urls, index, action, id, pipeline, routing, retry_on_conflict, tls, max_in_flight, api_key, basic_auth, batching`, none of which bound a request, and `elasticsearch.Config` has no timeout of its own — it only accepts an `http.RoundTripper`. `http_client` has had `timeout` for comparison, and Elastic's own Java REST client defaults to a 1s connect / 30s socket timeout ([docs](https://www.elastic.co/docs/reference/elasticsearch/clients/java/_timeouts)).

**Enabling TLS discards Go's default transport timeouts**, so a TLS user is worse off than a plaintext one:

https://github.com/redpanda-data/connect/blob/main/internal/impl/elasticsearch/v9/output.go#L107-L111

```go
if tlsEnabled {
    conf.clientOpts.Transport = &http.Transport{
        TLSClientConfig: tlsConf,
    }
}
```

That is a zero-value `http.Transport`. Supplying one also bypasses the fallback in `elastic-transport-go`, which would otherwise `http.DefaultTransport.Clone()`:

| | `http.DefaultTransport` | zero value |
|---|---|---|
| dial | `30s` | none — nil `DialContext` falls back to `var zeroDialer net.Dialer` |
| TLS handshake | `10s` | none — "Zero means no timeout" |
| idle connection | `90s` | none — "Zero means no limit" |

Go [`net/http` field docs](https://github.com/golang/go/blob/go1.25.1/src/net/http/transport.go#L183-L230); `zeroDialer` at [L1274](https://github.com/golang/go/blob/go1.25.1/src/net/http/transport.go#L1274).

**Neither `retry` nor `fallback` can rescue this**, because both act only on a *returned error* and a hang never returns one. The `retry` output blocks on the child's result channel and reaches `MaxElapsedTime` only once an error arrives:

https://github.com/redpanda-data/benthos/blob/v4.77.0/internal/impl/pure/output_retry.go#L187-L209

```go
select {
case res = <-resChan:
case <-r.shutSig.HardStopChan():
    return
}

if res != nil {
    ...
    nextBackoff := backOff.NextBackOff()
    if nextBackoff == backoff.Stop {
```

If no result is ever sent, `max_elapsed_time` never fires. There is no configuration that bounds this today.

## What changed

Two commits, one per cause, plus the regenerated docs.

- **`retain default transport timeouts when TLS is enabled`** — clone `http.DefaultTransport` and set `TLSClientConfig` on the clone, which is what `elastictransport` would have done on our behalf. No config change, and it applies whether or not `timeout` is set.
- **`add timeout field`** — a `timeout` duration applied via `context.WithTimeout` around `bulk.Do` and `Info().Do`. The typed API reads the response body inside `Do`, so this bounds the whole request — connect, send, response headers and body — the same coverage `http.Client.Timeout` gives `http_client`. It defaults to `0s`, which waits indefinitely, so the change is additive and no existing pipeline behaves differently unless it opts in.

`v8` and `v9` are maintained as parallel copies and the file header asks for changes to be mirrored, so both get the same patch.

## Tests

New `output_test.go` in each package. No Docker, so these run in normal CI.

- `TestConfigTLSRetainsDefaultTransportTimeouts` — asserts the TLS transport keeps `DefaultTransport`'s handshake, idle and expect-continue timeouts and a non-nil `DialContext`. Fails against the previous code, where all three were zero.
- `TestWriteBatchTimesOutOnUnresponsiveServer` — `httptest` server that accepts a request and never answers; asserts the write is abandoned with `context.DeadlineExceeded`, returning in ~250ms against a configured 250ms.
- `TestWriteBatchZeroTimeoutDefersToCallerContext` — with `timeout` unset, the caller's context is the only remaining bound. Pins the pre-existing behaviour so the additive default stays honest.
- `TestConfigTimeout` — unset, explicit value, explicit `0s`.

- [x] `go test ./internal/impl/elasticsearch/...` — v8 `1.587s`, v9 `1.897s`
- [x] `golangci-lint run ./internal/impl/elasticsearch/...` (v2.10.1, pinned in `.versions`) — 0 issues
- [x] `go vet ./internal/impl/elasticsearch/...`
- [x] `docs_gen` — only the two elasticsearch pages changed
- [ ] Integration tests not run locally (no Docker available on this machine); unchanged by this PR


dkheyman:
Related issue: https://github.com/redpanda-data/connect/issues/4648

